PR #10, opened 2019-01-11
Streamline CMake setup

psalz:
This is an incremental improvement with the goal of streamlining the whole CMake setup a bit.

- Adds checks that the `hcc`/`nvcc` and `clang` executables were found in PATH, so that we can error during generation, not just at compile time.
- Use `add_subdirectory` instead of `ExternalProject` for the rewrite and transform executables. I understand you tried to have those compiled with the system compiler instead of `syclcc`, but the same effect can be achieved by setting `CMAKE_CXX_COMPILER` within the CMakeLists.txt specific to `libhipSYCL` (as the variable is scoped to that folder). The advantage of this approach is that any configuration problems are again reported during the initial CMake call, instead of during compilation.
- Various minor cleanups/refactorings/consistency changes.

PS: I'm impressed how well the approach you're taking with hipSYCL works - very cool!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,12 @@
+cmake_minimum_required (VERSION 3.5)
 project(hipSYCL)
 
-set(CMAKE_CXX_COMPILER ${PROJECT_SOURCE_DIR}/bin/syclcc)
+# Make sure either hcc or nvcc can be found
+find_program(VENDOR_DEVICE_COMPILER NAMES hcc nvcc)
+if(VENDOR_DEVICE_COMPILER MATCHES "-NOTFOUND")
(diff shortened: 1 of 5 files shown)


illuhad:
I'm glad to hear that hipSYCL works well for you!

Thank you once more for this excellent PR :) You are right of course that `add_subdirectory()` is a much cleaner solution than using external projects.

If you plan on doing more work on the build system, be advised that there are some build-system related  changes about to be merged from the `cpu-backend` branch (basically, one version of libhipSYCL will be built for each of the backends cuda, rocm, cpu). So don't be suprised if things look a bit different soon ;)

PS: If any questions about the code or project structure pop up, feel free to contact me any time either with an issue or per email :)

PR #11, opened 2019-01-11
Add destructor implementation for device_selector

psalz:
It appears that `device_selector` was missing a destructor implementation, which made it so subclasses couldn't be instantiated (throwing a linker error).


illuhad:
Awesome, thank you Philip!

PR #12, opened 2019-01-11
Fix diagnostic location printing in includes rewriter

psalz:
This fixes a typo in the includes rewriter that causes the diagnostic location not to be correctly set (always printing `<unknown location>`).


illuhad:
Good catch, thank you!

PR #13, opened 2019-01-13
Handle skipped includes in inclusion rewriter

psalz:
Consider the following example program:

`foo.cpp`:
```c++
#include "foo.h"
#include "foo.h"
int main() { return 0; }
```

`foo.h`:
```c++
#pragma once
struct foo {};
```

This currently causes `syclcc` (`nvcc` actually, `hcc` as well, presumably) to fail with either an error that `foo.h` cannot be found, or because of a duplicate definition of `struct foo` -- depending on whether the build is done out-of-source or not.

Of course this is a contrived example, but the same happens for transitive inclusions of the same header. The reason for this is that after the first `#include` has been processed, all further include directives for the same file are not commented out.

Instead of keeping track of includes on a per-file basis, Clang's `InclusionRewriter` tracks them based on the source location within the _including_ file. This location is naturally different for the two directives in the example above (being on a different line). Crucially, if a file is skipped due to an include guard, the location is not tracked at all. This causes the current implementation to think that a file should not be rewritten, and thus does not comment out the directive. Note that this is not a problem for the mainline Clang implementation, as it always comments out all inclusion directives.

This PR adds an additional data structure that keeps track of any skipped includes and uses that to check whether a `#include` directive should be commented out, even if it is not to be rewritten.

### DIFF
--- a/src/hipsycl_rewrite_includes/InclusionRewriter.cpp
+++ b/src/hipsycl_rewrite_includes/InclusionRewriter.cpp
@@ -16,6 +16,8 @@
 /// This file is almost entirely identical to clang's mainline InclusionRewriter.cpp.
 /// However, this version does not rewrite includes of system headers.
 
+#include <unordered_set>
+
 #include "clang/Rewrite/Frontend/Rewriters.h"
 #include "clang/Basic/SourceManager.h"
 #include "clang/Frontend/PreprocessorOutputOptions.h"
@@ -53,6 +55,10 @@ class InclusionRewriter : public PPCallbacks {
   bool UseLineDirectives; ///< Use of line directives or line markers.
   /// Tracks where inclusions that change the file are found.
   std::map<unsigned, IncludedFile> FileIncludes;
+  /// hipSYCL: Keeps track of where skipped inclusions are found.
+  /// This is required so we can decide whether we want to comment out the
+  /// corresponding #include directive or not.
+  std::unordered_set<unsigned> SkippedIncludes;
   /// Tracks where inclusions that import modules are found.
   std::map<unsigned, const Module *> ModuleIncludes;
   /// Tracks where inclusions that enter modules (in a module build) are found.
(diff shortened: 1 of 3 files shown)


src/hipsycl_rewrite_includes/InclusionRewriter.cpp
```diff
@@ -195,6 +201,9 @@ void InclusionRewriter::FileSkipped(const FileEntry &/*SkippedFile*/,
                                     SrcMgr::CharacteristicKind /*FileType*/) {
   assert(LastInclusionLocation.isValid() &&
          "A file, that wasn't found via an inclusion directive, was skipped");
+  auto P = SkippedIncludes.insert(LastInclusionLocation.getRawEncoding());
+  (void)P;
```
illuhad: What does this do? `P` is just an `std::pair`, right?

psalz: I copied this from the implementation of `FileChanged`, it's most likely just to prevent a compiler warning about `P` being unused during release builds (when `assert` is a no-op).

illuhad: Okay, then let's keep it that way. Perhaps you could add a comment?

illuhad: Okay, that's fine :)

psalz:
Yes, I usually advocate using `#pragma once` but it's interesting to finally encounter a somewhat real-world example of where it is actually inferior to traditional include guards ;-).

Sure, I will add a test case tomorrow! Also I wanted to talk about the approach to testing in general anyway, so I'll open an issue regarding that as well :-).

illuhad:
Perfect, thanks! :)

psalz:
I added the test case (also had to force push a typo-fix)!

keryell:
> Yes, I usually advocate using `#pragma once` but it's interesting to finally encounter a somewhat real-world example of where it is actually inferior to traditional include guards ;-).

Another problem is that `#pragma once` is non standard...

Issue #14, opened 2019-01-14
Forward declared template functions not correctly pruned


psalz:
Consider this simple program:

```c++
template<typename T>
void foo(T);

template <typename T>
void bar(T) {}

template<typename T>
void foo(T value) {
    bar(value);
}

int main() {
    return 0;
}                                                              
```

After the rewrite and transformation passes, `nvcc` complains with `#endif without #if` and `unterminated #if`. The temporary *.cu file reveals the reason:

```c++
# 1 "<built-in>"
# 1 "/tmp/forward_decl.cpp"
template<typename T>
void foo(T)
#endif
;

template <typename T>
void bar(T) {}

template<typename T>
void foo(T value) ;
#if 0 // -- definition stripped by hipsycl_transform_source
{
        bar(value);
}

int main() {
        return 0;
}
```

Apparently, according to the Clang AST, the beginning of the body is at the definition, while the end is at the declaration. This of course causes a complete mess and breaks compilation of larger headers (e.g. Boost). I've been told that his might be annoying to fix, so my question is: Why are you pruning these uninstantiated templates anyway? Is this just an optimization or is there a practical reason?

illuhad:
The reason for pruning uninstantiated templates is that clang and nvcc have subtle differences with regards to overload resolution in uninstantiated templates. In order to correct `__host__` and `__device__` attributes perfectly, we need to have an accurate call graph that, in particular, matches the call graph that nvcc works with. 
As it turns out, there are cases where clang cannot resolve called overloaded functions within uninstantiated templates while nvcc still resolves them. Consequently, we may have missing edges in the clang call graph, and hence also miss information to infer whether a function should be host or device. Since nvcc still sees these edges, it may complain about missing `__device__` or `__host__` attributes. At the time, I thought that just removing such functions entirely from the code would be the most robust solution to this problem. This was also the final ingredient that allowed me to compile SYCL parallel STL code. 

There surely must be a way to get the actual beginning and end of the code block in the AST? Otherwise we could also try removing only the problematic function calls, but I'm not sure if this would be simpler in general, especially since function calls can be parts of other statements or calls.

psalz:
I see. Well, I talked to some compiler guys in my office and was told that they encountered this issue some years ago and back then it was a pain to work around. Of course that might be different today! I'm not really familiar with LLVM/Clang, but I know someone who is! Ping @PeterTh

illuhad:
[...]

PR #15, opened 2019-01-14
hipsycl_transform_source: Use getSourceRange() when pruning uninstantiated templates

illuhad:
This attempts to fix issue #14 by using `getSourceRange()` instead of `getLocStart()` and `getLocEnd()`. While the generic implementation of `getSourceRange()` just creates a range from `getLocStart()` to  `getLocEnd()` as one would expect, the specialization in `FunctionDecl` seems to do something else [(see here)](https://clang.llvm.org/doxygen/Decl_8cpp_source.html#l03637).

While it's still unclear to me why `getLocStart()` and `getLocEnd()` behave the way they do for the case described in issue #14, `getSourceRange()` seems to work as expected.

### DIFF
--- a/src/hipsycl_transform_source/CompilationTargetAnnotator.cpp
+++ b/src/hipsycl_transform_source/CompilationTargetAnnotator.cpp
@@ -262,7 +262,7 @@ CompilationTargetAnnotator::addAnnotations()
         std::string prefix = ";";
 
         const clang::Stmt *body = d->getBody();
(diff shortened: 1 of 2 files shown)


psalz:
This seems to fix it yes, great! So it was pretty easy after all. Unfortunately I still get some errors related to pruning (which I circumvented yesterday by disabling pruning altogether); I think it's only tangentially related to this though, so I'll open another issue.

psalz:
(See #16)

Issue #16, opened 2019-01-15
Pruned definition of class template function causes re-declaration error


psalz:
In this example

```c++
class foo {
    template<typename T>
    void bar(T value);
};

template <typename T>
void baz(T) {}

template <typename T>
void foo::bar(T value) {
    baz(value);
}

int main() {
    return 0;
}
```

`nvcc` will throw a re-declaration error of `foo::bar`, as only the declaration is left after pruning. I think the correct approach in this case would be to comment out the entire definition outside the class.

illuhad:
We could indeed comment out the entire definition, or we replace the definition with a trivial one i.e. `{}`, potentially with dummy return value. I'll think about what the best solution will be, in any case I think this will be easy to fix :)

psalz:
I think you're right, and I think replacing pruned definitions with `{}` might be a good approach in general (not just in this case). I suppose the return value doesn't matter, as compilers probably won't do any static analysis on uninstantiated templates.

illuhad:
It would also simplify the code slightly since for constructors, we are already inserting `{}` to avoid having an initializer list in a constructor that has been turned into a declaration (which causes nvcc to crash with some internal assertion failure).
The reason why I went the route of commenting out definitions was that clang's replace function would always crash - but in hindsight it could be that this was just because of the weird begin/end source locations... I'll try that again later :)

illuhad:
Here are my current findings:
* Even when inserting `{}`, we still need to comment out the existing code block instead of just overwriting it, since removing several code ranges in the same rewrite buffer apparently can really screw up the rewrite buffer, leading to a garbled result. Perhaps it works better if we fill the replacement buffer with enough spaces so that it is at least of equal size as the original code block?
* gcc actually warns about missing return statements in uninstantiated templates, so a dummy return value is in fact necessary
* Dummy return values require several special cases, for example in pruned lambda functions clang defines the return value of type `auto`, so we cannot just do something like `return *reinterpret_cast<return_type*>(new int);` which I was working with. Perhaps it would be possible to just not prune lambda functions since it's sufficient to prune the enclosing function, but there are other problems, for example it seems clang typedefs `bool` internally as `_Bool` (causing `hipsycl_transform_source` to insert `_Bool` return values) which nvcc doesn't know

Perhaps it is simpler to just comment out the entire definition as before, with special treatment for class template functions.

psalz:
I see, in that case it might be easier to comment out the entire definition, yes.

[...]

PR #17, opened 2019-01-17
Fix class template function pruning

illuhad:
This fixes issue #16 by, instead of commenting out the entire definition, replacing each statement within a definition with an amount of whitespaces equal to the statement's length.
Apart from fixing issue #16, this has a couple of advantages:
* Simpler code
* Probably more robust against `#ifdef` etc macros inside the function definition since these are not removed (Previously, if there were `ifdef` macros interleaved with the function body, bad things could have happened...)
* Usually does not alter line numbers for diagnostics (to be tested in detail - a statement spanning multiple lines may be an issue)

The changes are tested with code from the SYCL parallel STL and seem to work well with that.

In general, removing all statements from the function body may trigger warnings about missing return statements. This is not a problem in this case because
* nvcc seems to have worse diagnostics than gcc/clang and does not seem to pick up on that
* AMD's compiler is clang-based, so it is expected that such warnings are triggered. However, since it is clang-based, it will work with the same call graph as the hipSYCL source transformation. Therefore, pruning is not even necessary and could be disabled entirely on AMD if this becomes a problem
* In any case we could also just suppress these warnings since we already get full clang-based diagnostics from the hipSYCL source transformation. So, not having missing return statement warnings from nvcc or hcc doesn't hurt.

### DIFF
--- a/src/hipsycl_transform_source/CompilationTargetAnnotator.cpp
+++ b/src/hipsycl_transform_source/CompilationTargetAnnotator.cpp
@@ -37,7 +37,7 @@ using namespace clang;
 
 namespace hipsycl {
 namespace transform {
-
+  
 CompilationTargetAnnotator::CompilationTargetAnnotator(Rewriter& r,
                                                        CallGraph& callGraph)
   : _rewriter{r},
@@ -259,26 +259,19 @@ CompilationTargetAnnotator::addAnnotations()
                          << std::endl;
 
       if (d->hasBody()) {
-        std::string prefix = ";";
 
         const clang::Stmt *body = d->getBody();
-        auto bodyStart = body->getSourceRange().getBegin();
+        auto range = body->getSourceRange();
 
-        if(clang::isa<CXXConstructorDecl>(d))
+        const clang::FunctionDecl* function = clang::cast<clang::FunctionDecl>(d);
+        
+        for(auto child = body->child_begin(); child != body->child_end(); ++child)
         {
-          // If we are a constructor, instead of turning the definition
-          // into a declaration with a ';', we insert an empty definition.
-          // This also works if an initalizer list is present, where
-          // a ';' would be a syntax error.
-          prefix = "{}";
+          auto range = (*child)->getSourceRange();
+          int size = _rewriter.getRangeSize(range);
+          if(range.isValid() && size > 0)
+            _rewriter.ReplaceText(range, std::string(size, ' '));
         }
-
-        auto bodyEnd = body->getSourceRange().getEnd();
-      
-        _rewriter.InsertTextBefore(bodyStart,
-            prefix+"\n#if 0 // -- definition stripped by hipsycl_transform_source\n");
-        _rewriter.InsertTextAfterToken(bodyEnd, "\n#endif\n");
-        // TODO: Correct the current #line in the source
       }
     }
   } 
--- a/src/hipsycl_transform_source/CompilationTargetAnnotator.cpp
+++ b/src/hipsycl_transform_source/CompilationTargetAnnotator.cpp
@@ -261,9 +261,6 @@ CompilationTargetAnnotator::addAnnotations()
       if (d->hasBody()) {
 
         const clang::Stmt *body = d->getBody();
-        auto range = body->getSourceRange();
-
-        const clang::FunctionDecl* function = clang::cast<clang::FunctionDecl>(d);
         
         for(auto child = body->child_begin(); child != body->child_end(); ++child)
         {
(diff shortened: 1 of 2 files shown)


psalz:
I just tried this and while it does reduce the number of errors, I still get some. I'm not sure if they're directly related to this particular issue though (I previously had disabled pruning altogether, which works for me, so it could be another pruning-related issue). I will investigate and come back to you shortly!

psalz:
Okay so it is definitely related to this change. Check this out:

```c++
class foo {
    template<typename T>
    void bar(T value);
    
    template<typename T>
    int func_with_return_value();
};

template <typename T>
void baz(T) {}

template <typename T>
void foo::bar(T value) {
    baz(value);
}

#define MACRO_MADNESS(x) x

bool predicate(int value) {
    return value;
}

template <typename T>
int foo::func_with_return_value() {
    baz(T{});

    int my_value = 3;
    if (predicate(my_value))
        MACRO_MADNESS("test");

    return T{};
}

int main() {
    return 0;
}
```

The pruned result looks like this

```c++
template <typename T>
int foo::func_with_return_value() {
            ;


    if (predicate(my_value))
        MACRO_MADNESS("test");

              ;
}

```
and compilation fails with undeclared identifier `my_value`.

While the macro in this case is useless, in the actual code I'm trying to compile this is a wrapper for `throw` or `assert`. Apparently there exists some disparity between the Clang AST, which is built using the expanded macro, and the source code, which contains the unexpanded macro. What's really weird though is how not only the macro, but also the preceding `if` statement remains un-pruned.

illuhad:
Oh dear, I don't really think that going down the macro rabbit hole is going to be worth the hassle - we will likely run into one edge case after the next...

Perhaps it will be more robust after all if we just remove the unresolved function calls in question instead of the entire function? We could replace them like so:

```c++
void unresolved_overload(int);
void unresolved_overload(float);

template<class T>
T unused(T a, T b)
{
  int b1 = 0;
  a + T(); //was e.g.: a+unresolved_overload(a);
  return b1 + T(); //was e.g.: b1+unresolved_overload(b);
}

int main(){}
```
The important thing here is that the function that we replace these calls with is dependent on a template parameter. Since the function is uninstantiated, this prevents the compiler from performing type checks. Of course, the most simple function depending on T is `T()`.
So, in the implementation, we should:
1. Find all function calls that are unresolved in all instantiations of a given template function
2. Find the dependent template parameter `T` in the current context
3. Replace the function call with `T()`. Since the compiler cannot know the signature of T's constructor, it would probably also be sufficient to just replace the function name with `T` and leave the arguments as they are.

Are there any other edge cases or issues that could be a problem for this approach?

PS: Thank you for finding all these problems and putting so much effort in investigating them! It is extremely helpful to also have hipSYCL tested against other codebases, as seen from the number of problems that turn up :)

psalz:
This could work yes, but can it be done reliably? I seem to recall you mentioning the Clang rewriter crashing on you, and this seems like a great way of confusing it a lot ;-).

Ultimately I think that while there certainly is merit in getting pruning as stable as possible, I think it is also not outrageous to suggest that device code should be kept simple to avoid some of the more esoteric issues we’ve found. As I mentioned in #20, all of these things happened in libraries that don't contain any device code in the first place.

So I'd wager a guess that this current implementation, with the addition of #20, would probably cover the vast majority of real-world use cases. If we drop `nvcc` entirely for Clang, as suggested in #24, I suppose we could even get rid of this hack for good (assuming that `hcc` doesn't need it).

illuhad:
I think this approach could at least be far more reliable than the current one. Clang rewriter crashes only seem to be an issue if you start removing things. As long as you only insert new text or replace a text segment with the same amount of characters, everything seems to work fine. So, we can just insert whitespaces if the replacement function call is shorter than the original one :)

Of course, based on the previous experiences, I don't want to make a bet if this may not lead to other unexpected edge cases popping up - but I think it's worth a try.

While I would like to keep nvcc support as mentioned in #22, we could absolutely only enable pruning for nvcc. Then only nvcc users would be punished by any weirdness that may be associated with it ;)
hcc is clang based, so I don't think that pruning is necessary, and for the CPU backend it's not important anyway if `__device__` and `__host__` attributes are correct since they will be `#define`d away.

illuhad:
I have written a test implementation of the approach based on only rewriting unresolved function calls, and it seems that there a couple of cases that require extra effort to treat correctly, for example nested calls of unresolved functions. So, I will close this PR for now since we are moving in the direction of using clang as default cuda compiler and selective rewrite directories. 
For the remaining use cases where nvcc (and hence template pruning) is needed, I think the current implementation plus selective rewrite directories are sufficient for now.

Issue #22, opened 2019-01-22
Proposal: Add support for compiling CUDA with Clang


psalz:
I’d love to see support for compiling CUDA using Clang (as detailed [here]( http://llvm.org/docs/CompileCudaWithLLVM.html)) added to `syclcc`. As far as I can tell this pretty much only has advantages over using `nvcc` (I could be wrong though, as I don’t have much CUDA experience).

For example, Clang will happily compile this program:

```c++
#include <CL/sycl.hpp>

int main() {
    cl::sycl::queue queue;

    queue.submit([](auto& cgh) {
        cgh.template parallel_for<class test>(cl::sycl::range<1>(10), [=](cl::sycl::item<1> item) {
            printf("Hello %d\n", item[0]);
        });
    });

    return 0;
}
```

As you know, generic lambas are among the things that `nvcc`’s (experimental!) extended lambda feature currently does not support at all. Don’t ask me why, but I actually need this.

As for feasibility: I added Clang as a third option into `syclcc` and successfully compiled both `libhipSYCL` itself (with very minor adjustments) as well as my application (which makes use of generic lambdas).

This would likely also allow us to omit the pruning step, as the call-graphs should be the same between the libTooling and Clang invocations.

Other advantages:
-	All (except for the `syclcc` specific) parameters can be simply passed to `clang++`, no juggling required.
-	You get Clang’s awesome error reporting.

In fact I think it might even be worth considering making Clang the default for compiling to CUDA, and dropping `nvcc` altogether. Anyone using hipSYCL has to have Clang installed anyway, so that is not a problem. However again, I’m not certain if there’s any downsides to this, like e.g. any features that aren’t supported. Dropping `nvcc` would allow us to use features that only Clang supports, like [overloading based on __device__ and __host__ attributes]( http://llvm.org/docs/CompileCudaWithLLVM.html#overloading-based-on-host-and-device-attributes). Although this might be moot as I’m not sure what `hcc`’s stance is on this.

illuhad:
This is an issue where I can give you an answer very quickly: I'm very much in favor of adding clang-cuda support, and, in fact, had already thought about this myself :) If you already have it working, please file a pull request :)

[...]

PR #27, opened 2019-01-25
Add support for compiling CUDA using Clang

psalz:
This adds support for compiling hipSYCL on CUDA using Clang, as discussed in #22.

This turned out to be easier than expected, as no code changes whatsoever are required for libhipSYCL (even before it was only adjusting the compiler detection macros, but since hipSYCL now uses `__CUDACC__`, which is also defined when compiling with Clang, it works right out of the box). All changes are thus contained to `syclcc`.

Clang CUDA is now the default method for compiling CUDA, the previous behavior (= using nvcc) can be restored by explicitly passing `--hipsycl-platform=nvcc`.

The version of Clang used for compilation can be configured using either the `--cuda-clang-compiler=<compiler>` CLI parameter, or the `HIPSYCL_CUDA_CLANG_COMPILER` environment variable.

In addition, this includes some minor refactorings of some parts of `syclcc` and adds a new configuration option `platform_needs_template_pruning`, which dictates whether the `--with-template-pruning` flag is passed to hipsycl_transform_source. This is only enabled for nvcc.

---

Now I don't think this PR is ready for prime-time just yet, as there are some remaining issues / open questions:
- I wasn't sure if we should compile a separate version of libhipSYCL for clang/nvcc, but my hunch is that it shouldn't be necessary.
- Clang requires a `--cuda-gpu-arch` parameter to be set, as it otherwise defaults to compute capability 2.0, which is [pretty old](https://developer.nvidia.com/cuda-gpus) and only supported by CUDA 7 and 8. I decided to go with `sm_52`, which corresponds to the GTX 9xx series. These should be forward-compatible however. Of course this can be overwritten by passing the parameter to `syclcc`.
- [It looks like](https://releases.llvm.org/7.0.0/docs/CompileCudaWithLLVM.html) Clang 6 and 7 can only target CUDA 7 and 8. I have CUDA 10 installed so I cannot test if everything works with Clang 6/7. Right now the version of Clang used by `syclcc` defaults to `clang++`, which I assume on many systems will be Clang 6 (at least that's the default on Ubuntu 18.04 LTS). Maybe this should be changed to Clang 8 once it releases in February?

### DIFF
--- a/bin/syclcc
+++ b/bin/syclcc
@@ -80,14 +80,17 @@ class hip_installation:
 
 ## Represents the different supported hipSYCL platforms
 class hipsycl_platform:
-  CUDA = "cuda"
+  CUDA_CLANG = "cuda"
+  CUDA_NVCC = "nvcc"
   ROCM = "rocm"
   HIPCPU = "hipcpu"
 
   def from_string(platform_name):
     name = platform_name.lower()
-    if name == "cuda" or name == "nvidia" or name == "nvcc":
-      return hipsycl_platform.CUDA
+    if name == "cuda" or name == "nvidia":
+      return hipsycl_platform.CUDA_CLANG
+    elif name == "nvcc":
+      return hipsycl_platform.CUDA_NVCC
     elif name == "rocm" or name == "amd" or name == "hcc" or name == "hip":
       return hipsycl_platform.ROCM
     elif name == "cpu" or name == "host" or name == "hipcpu":
@@ -165,6 +168,7 @@ class source_file_processor:
          "-Wno-unused-command-line-argument",
          "--hipsycl-transform-dir="+output_dir,
          "--hipsycl-main-output-file="+output_filename]
+         +(["--with-template-pruning"] if self._config.platform_needs_template_pruning else [])
          +self._additional_args
          +self._config.hipsycl_common_arguments)
 
@@ -224,8 +228,62 @@ class source_file_processor:
 
 cpp_files = set([".cpp",".cxx",".c++",".cc",".c"])
 
-## Compiler backend using nvcc
-class cuda_compiler:
+## Compiler backend for CUDA using Clang
+class cuda_clang_compiler:
+  def __init__(self,config):
+    self.device_compiler = config.cuda_clang_compiler
+    self.config = config
+    self.is_clang_available = is_program_executable(self.device_compiler)
+
+  def includes_linking_stage(self, args):
+    for arg in args:
+      if (arg.startswith("-E") or
+          arg.startswith("-fsyntax-only") or
+          arg.startswith("-S") or
+          arg.startswith("-c")):
+        return False
+    return True
+
+  def run(self, args):
+    source_processor = source_file_processor(self.config, ".cu", args)
+    result = None
+    try:
+      transformed_args = []
+      for arg in args:
+        if is_source_file(arg):
+          transformed_args.append(source_processor.process_file(arg))
+        else:
+          transformed_args.append(arg)
+
+      compiler_args = ["--cuda-gpu-arch=sm_52",
+                        "-std=c++14",
+                        "-pthread",
+                        "-Wno-unused-command-line-argument",
+                        "-Wno-deprecated-declarations"]
+
+      compiler_args += self.config.hipsycl_common_arguments
+      compiler_args += transformed_args
+
+      # To avoid warnings only add linker flags if this run will invoke the linker
+      # TODO: This mechanism could also be used to conditionally link libhipSYCL
+      if self.includes_linking_stage(args):
+        compiler_args += ["-lcudart_static",
+                           "-ldl",
+                           "-lrt"]
+
+      result = subprocess.call([self.device_compiler]+compiler_args)
+
+    finally:
+      source_processor.release_files()
+
+    return result
+
+  def is_available(self):
+    return self.is_clang_available
+
+
+## Compiler backend for CUDA using nvcc
+class cuda_nvcc_compiler:
   def __init__(self,config):
     self.device_compiler = "nvcc"
     self.config = config
@@ -246,11 +304,6 @@ class cuda_compiler:
   def is_no_nvcc_argument(self,arg):
     return not arg in self.nvcc_option_output
 
-  def is_compilation_step(self,args):
-    for arg in args:
-      if is_source_file(arg):
-        return True
-    return False
 
   def obtain_linker_arguments(self, linker_args):
     if not linker_args.startswith("-Wl,"):
@@ -405,6 +458,8 @@ class hipcpu_compiler:
 #  Supported arguments are:
 #  --force-alternative-compiler=<compiler>  Use the specified compiler
 #               directly without going through the hipSYCL toolchain
+#  --cuda-clang-compiler=<compiler>  Use an alternative version of Clang
+#               for compiling CUDA (default: clang++)
 #  --keep-temporary-files Keep temporary files created by the hipSYCL
 #               source-to-source transformation; this can be useful
 #               for debugging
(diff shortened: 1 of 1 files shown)


illuhad:
Excellent, good to hear that things are easier than expected for a change :)

As far as your questions are concerned:
* I'm pretty sure that a separate libhipSYCL for clang/nvcc is not needed. libhipSYCL only interfaces with the CUDA Runtime API. libhipSYCL is only compiled with nvcc because the HIP headers require that they are compiled with HCC or a CUDA compiler. I'm pretty sure that clang also just uses the nvidia runtime API (it's still a proprietary platform, so there are not so many choices..). So, let's first try without a separate libhipSYCL. If we hit problems we can revisit that question ;)
* I'm fine with `sm_52`. Some clusters still have some legacy Kepler GPUs in them, but this is not going to stay for long ;) However, I think that this behavior should be documented in the readme and potentially `syclcc --help`
* In `syclcc` there is already a list of clang executables that will be tried to see if they are available when looking for the CPU backend compiler [here](https://github.com/illuhad/hipSYCL/blob/8f080b64bed9d5210334fae90e76d5b4449c445a/bin/syclcc#L67). With a similar logic, we could add a function `get_newest_clang()` which iterates through a list `['clang++-8', 'clang++-7', 'clang++-6', 'clang++']` and returns the first working command? In that way, the newest clang is always preferred.

illuhad:
No need to hurry, I also need some additional time to test this since clang nightly seems to conflict with the regular clang package on Arch Linux, so I'll try installing it in a container as a next step..

My impression was that there is always both `clang++-<major>.<minor>` and `clang++-<major>`. Is that not the case on Ubuntu?

Okay, then I'll update the README in one go, including an updated image for the compilation workflow :)

psalz:
That's what's confusing me, I only have either `<major>.<minor>`, or `<major>`, not both. But now that I'm thinking about it, this might just be the nightly package (maybe this is not built by the Ubuntu maintainers).

In any case, I just went for checking all possible variants between 8 and 6. I only do this if no version has been set explicitly, to avoid the overhead.

PS: Your new Travis config seems to always try to pull from this repository, but the PR comes from my fork. I think you might have to construct the remote URL yourself from `TRAVIS_PULL_REQUEST_SLUG`.

bin/syclcc
```diff
@@ -222,8 +226,62 @@ class source_file_processor:
...
+    result = None
+    try:
+      transformed_args = []
+      for arg in args:
+        if is_source_file(arg):
+          transformed_args.append(source_processor.process_file(arg))
+        else:
+          transformed_args.append(arg)
+
+      compiler_args = ["--cuda-gpu-arch=sm_52",
```
illuhad: Won't this lead to clashes if the user wants to override the the arch and passes `--cuda-gpu-arch` himself? The `hip_compiler` only appends the GPU arch argument based on the content of  `HIPSYCL_GPU_ARCH`, if this environment variable is defined. Perhaps we can have a similar behavior here for the sake of consistency and only add `--cuda-gpu-arch=sm_52` if we don't have an arch defined in `HIPSYCL_GPU_ARCH`?

bin/syclcc
```diff
@@ -222,8 +226,62 @@ class source_file_processor:
...
+
+      compiler_args += self.config.hipsycl_common_arguments
+      compiler_args += transformed_args
+
+      # To avoid warnings only add linker flags if this run will invoke the linker
+      # TODO: This mechanism could also be used to conditionally link libhipSYCL
+      if self.includes_linking_stage(args):
+        compiler_args += ["-lcudart_static",
+                           "-ldl",
+                           "-lrt"]
```
illuhad: We should also conditionally link against libhipSYCL as you propose, since this is the behavior of the other compiler classes in `syclcc`.

illuhad: Sorry, forget that, this is of course handled by `config.hipsycl_common_arguments` :)

bin/syclcc
```diff
@@ -447,19 +501,33 @@ class syclcc_config:
...
+               hipSYCL platform; valid values are: cuda,nvcc,rocm,cpu
 --help Print help message """)
         sys.exit(0)
       else:
         self._forwarded_args.append(arg)
 
+    if self._cuda_clang_compiler == None:
+      self._cuda_clang_compiler = get_first_working_command([
+      "clang++-8.0","clang++-8","clang++-7.0","clang++-7",
+      "clang++-6.0","clang++-6","clang++"])
```
illuhad: Can we reuse this list in `get_host_compiler()`? Then we only have to maintain one list of clang executable names ;)

bin/syclcc
```diff
@@ -447,19 +501,33 @@ class syclcc_config:
...
+
+  def _parse_compound_argument(self, arg):
+    parsed_args = arg.split("=")
+    if len(parsed_args) != 2:
+      raise RuntimeError("Invalid argument: "+arg)
+    return parsed_args[1]
+
```
illuhad: That function is a very useful addition, I like it :)

bin/syclcc
psalz: Well the user-provided arguments are appended afterwards, so I thought they'd just overwrite the default. *But* I just realized that apparently [it is possible to pass multiple architectures at once](http://llvm.org/docs/CompileCudaWithLLVM.html#invoking-clang), and compile for all of them. So yes, this is not a good solution. Also I didn't realize you already had a `HIPSYCL_GPU_ARCH` in place, as `nvcc` doesn't use it. I will pull this out of `hip_compiler` into a general configuration option, is that okay with you?

bin/syclcc
psalz: I thought maybe we could at some point extract some of the common functionality that all compilers share into a base class which could then also handle stuff like this.

bin/syclcc
psalz: Sure, however would you still like to default to `g++`?

bin/syclcc
illuhad: Perfect, that sounds like a good solution :)

bin/syclcc
illuhad: `get_host_compiler()` is only used for the CPU backend (which does almost not depend on the source-to-source transformation at all), so there's no inherent advantage to preferring clang. And since g++ comes preinstalled basically everywhere, I think it makes sense to use it as default. In the future we could however perhaps add a configuration flag for this.

bin/syclcc
illuhad: Yes, that makes sense. The current approach was implemented when things were still very simple with only two backends around ;)

psalz:
I consolidated the lists of Clang executables between the hipCPU and CUDA-Clang compilers, and added support for `HIPSYCL_GPU_ARCH` to the latter, as discussed. I also rebased everything onto the current master.

I thought some time about how to handle the precedence of `--cuda-gpu-arch` parameters and the `HIPSYCL_GPU_ARCH` environment variable. Since it is possible to pass multiple architectures to Clang, I came to the conclusion that the best solution would be to simply merge them. Only if neither is provided the default (`sm_52`) is used.

I also briefly looked into supporting `HIPSYCL_GPU_ARCH` for the `nvcc` compiler, but the supported architecture parameters are a bit of a mess and I'm not sure how to best handle them. I added a TODO note for that.

---
On a somewhat unrelated note: While I tried the changes to `get_host_compiler()` I noticed that for me the hipCPU backend doesn't compile. I didn't open a separate issue because I don't have time to investigate this thoroughly right now; however I think that it's due to the order of the linker flags that are passed to `g++` (with `clang++` it does compile by the way). GCC requires that all libraries are linked after the object file that needs them. While not strictly required for shared libraries (such as libhipSYCL_cpu), this can still be enforced with the `--as-needed` flag. According to [this](https://stackoverflow.com/a/409470/2095767) the existence of this flag varies between distros, and apparently Arch doesn't specify it (which might explain why you didn't get an error).

illuhad:
Thanks, looks good :) I agree that merging `--cuda-gpu-arch` and `HIPSYCL_GPU_ARCH` is a good solution for now.

Travis (which now pulls correctly from your fork :) ) now complains that clang cannot find `libdevice` for `sm_52`. This is probably because clang 6 (as used in the travis build) is incompatible with the CUDA 10 in the docker image, so we should update the CUDA dockerfile to also install clang nightly.

Not sure what the best approach is to fix the CPU backend for `g++`, apart of course from just using `clang++`. I will look into that this evening.

psalz:
Yes, the Travis error is most likely due to the Clang version!

As for GCC: I think you can simply swap the order of `hipsycl_common_arguments` and `transformed_args` being added to `compiler_args`. Then it works for me (I still get a warning about `-Wno-unused-command-line-argument` being an unrecognized option though!). Also does the CPU backend really need the rewrite/transform passes? Aren't the `__host__` and `__device__` attributes defined to empty anyways?

illuhad:
The CPU backend does need the transformation passes due to one tiny issue (not related to `__device__` or `__host__`): In parallel hierarchical for invocations, all declarations that are not of type `cl::sycl::private_memory<T>` need to go into local memory (`__shared__` memory in CUDA/HIP). These `__shared__` attributes are also inserted during the transformation pass. This is also necessary on the CPU backend as it alters semantics so that this variable should be shared across all parallel OpenMP threads in a work group. This is at the moment implemented with `#define __shared__ static`.

Also it was my plan to use the transformation pass in the future to have some significant performance increase on CPUs: If we detect that no barrier is called we can have the OpenMP parallelization span across work groups instead of work items within a group (and iterate over work items within a group sequentially). Better vectorization, less false sharing and so on.

PR #28, opened 2019-01-25
prune templates by commenting function body content

illuhad:
As a final attempt to fix issue #16 without introducing potential new problems, this causes the pruning step to comment out the content of the template function bodies instead of the bodies themselves. I.e., instead of 
```cpp
template<class T>
void f() ;
#if 0 // -- definition stripped by hipsycl_transform_source
{
  ...
} 
#endif
```
this does
```cpp
template<class T>
void f()
{
#if 0 // -- definition stripped by hipsycl_transform_source
   ...
#endif
} 
```


psalz:
This seems like a surprisingly effective solution. That one library which caused both the duplicate declaration error (#16), as well as the undefined values/macro mess (#17) now goes through without a hitch!

illuhad:
Yay, we found a solution at last! :D

Issue #26, opened 2019-01-25
Transformation steps currently have hard dependency on OpenMP


psalz:
With the addition of the hipCPU backend the source transformation steps now include the hipCPU version of `hip_runtime.h` ([here](https://github.com/illuhad/hipSYCL/blob/8f080b64bed9d5210334fae90e76d5b4449c445a/include/CL/sycl/backend/backend.hpp#L33)). This unfortunately causes compilation to fail on my system, as apparently my OpenMP headers are located outside the default system include search paths. The culprit is [kernel_execution_context.hpp](https://github.com/illuhad/hipCPU/blob/3895f884e96ab49db075cc5925677a8fd6af2004/include/hipCPU/hip/detail/kernel_execution_context.hpp), which includes `<omp.h>`.

I don't know all the reasons for including hipCPU during transformations (presumably one being the definition of the `__host__` and `__device__` macros) so I don't know what the best course of action is here.

illuhad:
It's not only the `__host__` and `__device__` macros, we need to have the entire HIP/CUDA API available in order to have the runtime functions, HIP/CUDA thread/block ids and so on. Before the introduction of hipCPU, we had a special `dummy_hip` HIP header which would just contain the declarations without definitions, which I think was far more of an hack ;)

I think the best way to solve this is to let the source transformation parse the code with OpenMP enabled. Perhaps it's sufficient to just add `-fopenmp` to the arguments of `hipsycl_transform_source`?

illuhad:
yes, we could have hipCPU supporting sequential execution without OpenMP. In this case, launching any kernel with a block size > 1 will fail or do weird stuff since it is not possible to guarantee correct synchronization semantics (i.e. `__syncthreads()`) if we don't have as many OpenMP threads as HIP threads in the block.
This is why I originally decided not to support non-OpenMP builds of hipCPU; to avoid any surprises when accidentally compiling without OpenMP.

Is it possible that for clang 6, the OpenMP support must be installed separately and it fails because of that?

psalz:
Yes, you are right. I think I know what happened: I installed the Clang 8 nightly through their [pre-built package](http://apt.llvm.org/) to avoid those long compilation times. Apparently this also includes a newer version of `libomp` and `libomp-dev` which conflicts with the Clang 6 one. So it was completely my fault |-). I however still think not having a dependency on OpenMP for compiling the source transformations is a good thing!

PR #29, opened 2019-01-26
Don't require OpenMP during source transformation

illuhad:
This disables the requirement for OpenMP during the source transformation step, as suggested in issue #26. This is done using an updated version of hipCPU (make sure the hipCPU submodule is also up-to-date!) and the new macro `HIPCPU_NO_OPENMP`.


psalz:
Awesome! Works for me!

illuhad:
Thanks for the feedback! :)

PR #32, opened 2019-01-31
Add option to build hipSYCL with nvcc instead of clang

illuhad:
This adds the cmake option `USE_NVCC`, which can be used to compile hipSYCL with nvcc instead of the default CUDA compiler clang. This is useful if a version of clang that is new enough to compile against recent versions of CUDA is not installed.
Additionally, this also moves the code to find clang to the top-level CMakeLists, instead of having it duplicated in `hipsycl_transform_source` and `hipsycl_rewrite_includes`.
I've also added `clang-8` and `clang-8.0` to the list of clang executables that we look for.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,15 @@ endif()
 # Make sure either hcc or nvcc can be found
 find_program(NVCC_COMPILER NAMES nvcc)
 find_program(HCC_COMPILER NAMES hcc)
+find_program(CLANG_EXECUTABLE_PATH NAMES clang++-8 clang++-8.0 clang++-7.0 clang++-7 clang++-6.0 clang++-6 clang++ CACHE STRING)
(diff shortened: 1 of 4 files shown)


psalz:
LGTM!

At some point I think it would be cool if these configuration-time settings could also influence the defaults of `syclcc`, i.e., if somebody sets `USE_NVCC` they'll likely also want to use `nvcc` as their hipSYCL-platform. Likewise, if I configure with `WITH_CPU_BACKEND=0` the `hipcpu_compiler` still identifies as available as long as any host compiler can be found in PATH. Maybe these settings could be written into `syclcc` while copying it to the installation location (iirc CMake has facilities to do that). Just a thought!

illuhad:
Yes, I agree. I would also like the clang executable used by `syclcc` to be the one that is specified with cmake. We should consider that we need `syclcc` already to compile `libhipSYCL`, so it must be configured already at that point. Not sure if that makes it more complicated, but it's something to keep in mind.

Issue #20, opened 2019-01-22
Proposal: Explicit inclusion rewriting


psalz:
I propose adding a special flag for `syclcc` which controls whether an include path should be rewritten or not. Here’s why:

First of all I suspect that for many smaller SYCL applications, all kernel code will be situated within implementation files anyway, so the entire rewrite pass could be omitted, improving compilation speeds.
More importantly though, inlining of headers can sometimes cause compilation errors. As an example, the following program does not compile with syclcc (using nvcc, as of Boost 1.68):

```c++
#include <boost/optional.hpp>

int main() {
    boost::optional<int> foo, bar;
    foo = bar;
    return 0;
}
```

If I prevent the inclusion rewriter from inlining Boost (by adding a condition within `RewriteSelector::shouldIncludeBeRewritten`), everything works fine. I can’t tell you the exact reason, as Boost is huge and contains a metric ton of workarounds for various compilers (I really don’t envy those guys!), however, to me this looks like a case of compiler detection which causes the compilation to take different branches during the inclusion rewriting and the compilation with `nvcc`. I actually have no idea how they do it, I tried to create a small test case to replicate this and couldn’t come up with one. Nevertheless, it apparently does happen.

Moreover, all of the weird egde-cases I’ve encountered with pruning were within libraries that don’t contain any device code at all. Omitting them from the rewrite pass would also alleviate those problems.

Lastly, this could generally improve compilation performance, as we would have less text to dump back to disk.

---

So, instead of rewriting everything but hipSYCL and system libraries, I propose that headers which should be rewritten be made explicit through some flag for syclcc, e.g. `-IR` or `--rewrite-includes-path=`. I suspect that in many cases the rewrite step would only concern local includes (e.g. `-IR./include`) or could even be omitted entirely. Also maintaining a list of system library headers to exclude wouldn’t be required anymore either.

illuhad:
I agree with the general strategy that we should try to only fiddle with code that we need to, and ignore irrelevant parts.
That said, I would like to at least in principle support SYCL header-only libraries (like parallel STL) that are installed somewhere in the system without additional effort from the user. So, I would prefer if we still rewrite everything by default. But we could have additional flags, as you propose, to control this. I.e.
* By default everything is rewritten
* if specific rewrite include paths are passed, only these are rewritten.
The name of this argument should probably be something that the user can immediately understand, like `--device-code-path` or something like that, since this option would effectively control where kernel code is allowed.

[...]

PR #33, opened 2019-02-02
Implement selective rewriting of includes

illuhad:
This implements manual selection of paths in which header files are rewritten, as discussed in issue #20. By default, everything is rewritten (as is currently the case). If `-RDI=<path>` or `--restrict-device-header-path=<path>` (names of parameter are still up for debate) are passed to `syclcc`, then a whitelist is activated, `<path>` is added to the whitelist and only include files inside of `<path>` are rewritten. At the moment, this does not yet take into account filesystem links.

If `hipsycl_transform_source` wants to modify a source that is not rewritten, a warning is emitted. This is however still disabled by default, since the warning is apparently also triggered by many files of the STL and the hipSYCL runtime. Since no changes in these files should ever be necessary, this indicates that there may be some bug in `hipsycl_transform_source`, which is best investigated in a seperate issue.

### DIFF
--- a/bin/syclcc
+++ b/bin/syclcc
@@ -144,7 +144,8 @@ class source_file_processor:
          "-DHIPCPU_NO_OPENMP",
          "-Wno-unused-command-line-argument"]
          +self._additional_args
-         +self._config.hipsycl_common_arguments)
+         +self._config.hipsycl_common_arguments
+         +["--rewrite-include-paths="+p for p in self._config.device_header_paths])
 
       if returnvalue != 0:
         raise RuntimeError("Error while rewriting includes, aborting compilation")
@@ -484,6 +485,10 @@ class hipcpu_compiler:
 #               when building hipSYCL itself.
 #  --hipsycl-platform=<platform>  Build the program for the specified
 #               hipSYCL platform; valid values are: cuda,nvcc,rocm,cpu
+#  --restrict-device-header-path=<path>,
+#  -RDI=<path>  Restrict the paths in which include
+#               files are allowed to contain device code. Can be specified
+#               several times.
 #  --help       Print help message
 class syclcc_config:
   def __init__(self, argv):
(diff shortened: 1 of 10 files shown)


psalz:
Nice! Although in addition to links, the current implementation does not correctly deal with relative paths: First, as far as I can tell the `--restrict-device-header-path` parameter does not work for something like `./include`. While I don't think it must absolutely support relative paths, compilers generally do it for their include path parameters, so it could be good for consistency.

More importantly though, something like this wouldn't work:

/some/src/file.cpp:
```c++
#include "../include/kernel.h"
```
with `--restrict-device-header-path=/some/include` (as `/some/include` is not a prefix of `../include`).

I think most if not all of these issues (including links) could be resolved with Boost::filesystem's [canonical](https://www.boost.org/doc/libs/1_67_0/libs/filesystem/doc/reference.html#canonical).

illuhad:
Thanks for spotting this! I agree that it would be highly beneficial to support relative paths for the reasons you mentioned. I'll experiment with `boost::filesystem` this evening :)

illuhad:
I finally had the time to implement this with `boost::filesystem::canonical`, seems to work as far as I can tell. I'm not yet sure what happens if files are included using relative paths since `syclcc` is not necessarily using the directory in which the header file is as work directory, which could cause `canonical()` to expand the path incorrectly. Whether this then works depends on what clang returns as file path of include headers (if we're lucky these are already always absolute paths)- more testing is necessary here.
It seems that at the moment Travis CI fails because it cannot find `libboost_filesystem.so`. Likely an extra package for this library must be installed in the Dockerfile.

psalz:
I didn't have time to test this thoroughly, but I can tell you that I successfully compiled my project using this PR, without having to manually blacklist certain includes within `RewriteSelector` -- very cool! One hack less ;-).

Last time I played around with this it looked like Clang would always provide absolute paths already, but you could of course add a sanity check with `boost::filesystem::path::is_relative`!

For Travis I think you might need `libboost-all-dev` or simply `libboost-filesystem-dev`.

illuhad:
Thanks for the feedback! It should now print a warning when clang returns a relative path, as you suggested. I have also fixed the Travis builds :) I'll merge this PR tomorrow if no further issues pop up.

Issue #24, opened 2019-01-22
Unit testing


psalz:
While trying to compile more complex applications I often find areas of hipSYCL where I’d like to contribute something, be it a missing operator on some class or error handling for some particularly obscure situation (into which I tend to run into quite frequently for some reason). Also I’m getting a lot of (mostly minor) compiler warnings with hipSYCL right now, and I’d like to cut down on those as well. Anyway, I’d like to not break the existing implementation while doing this – so I think it’s time to add unit tests :).

While at some point it would be great to have test coverage for all the different components of hipSYCL (i.e. the source transformations, `syclcc`, ...), I think for now the most important thing is to add tests for the C++ library itself.

With SYCL being a specification, it certainly could be a good approach to try and add tests by systematically going through the spec. Unfortunately I don’t have the resources to embark on such a journey right now, so I’d suggest to take a more ad-hoc approach for now (i.e., adding tests as new features are added or when bugs are fixed). Once the conformance test suite goes public (#6) things should also become a lot easier in this regard.

A couple of questions of how to approach testing come to mind (I’m sure there’s more to figure out further down the line...):
-	Some functionality will have to be tested on the host, some on a device and some on both. It would be good to have a concise way of specifying this within a test case.
-	Should functions in tests include “\_\_device\_\_” attributes, or should they also exercise the source transformation step?

Do you have any preferences regarding a unit testing framework? I can highly recommend using [Catch2](https://github.com/catchorg/Catch2/), as it's modern, very easy to set up and doesn't get in your way.

Let me know your thoughts on this; I can begin to set this up if you’d like.

---
PS: Sorry for the issue-avalanche, as I mentioned I had some stuff in my backlog and I figured to just do it all today and be done with it ;-).

illuhad:
yes, we absolutely need this :) I agree that full unit testing of the spec is too much work, and likely unecessary anyway if the conformance test suite will be released. So, let's just add tests for new features/issues as you suggest.

There are several components to testing:
[...]

PR #35, opened 2019-03-20
Add unit tests

psalz:
This is an initial stab at adding some unit tests for hipSYCL. I'm using Boost.Test, as detailed in #24, and for now this basically only translates some of the existing test programs -- which I'd classify as smoke tests -- into a proper unit testing format. Note that I didn't include the [vec.cpp](https://github.com/illuhad/hipSYCL/blob/4658388425f9227a307fd50fc8d9852372ba2e72/tests/vector/vec.cpp) tests, as those don't compile for me at the moment (I haven't looked into it much more).

Some notes:
- I’ve included your license header but took the liberty to add “and contributors” to the copyright.
- At some point these should probably be split into multiple files, however I’d suggest we wait a bit to see what kinds of tests we’ll end up having and how they might best be grouped together.
- I didn't include the basic-kernel-with-offset example, as I discovered a bug which I'll address in a subsequent PR.

Tests are confirmed running with
- [x] Clang CUDA
- [x] nvcc
- [x] ComputeCpp 1.0.5
- [ ] HIP/hipCPU

Please let me know if you are happy with this general setup or not, so I can make any potential adjustments before I’ll open additional PRs with more tests -- I’ve already got a whole lot more in store :).


tests/unit_tests.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+  for(size_t i = 0; i < global_size / local_size; ++i) {
+    size_t expected = 0;
+    for(size_t j = 0; j < local_size; ++j) expected += i * local_size + j;
+    size_t computed = host_buf[i * local_size];
+    BOOST_TEST(computed == expected);
+  }
+}
+
+BOOST_AUTO_TEST_CASE(local_memory) {
```
illuhad: Perhaps we can rename this to `dynamic_local_memory`? The previous test `hierarchical_dispatch` uses (static) local memory as well. While `hierarchical_dispatch`  tests the source transformation to insert a `__shared__` attribute, this tests the local memory accessor mechanism.

tests/unit_tests.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+  {
+    auto acc = buf.get_access<mode::read>();
+    for(size_t i = 0; i < num_elements; ++i) {
+      BOOST_REQUIRE(acc[i] == 2 * i);
+    }
+  }
+}
+
+BOOST_AUTO_TEST_CASE(multiple_queues) {
```
illuhad: This test actually tests automatic synchonization and ordering of concurrent operations in the task graph, the multiple queues are more of an implementation detail to obtain concurrent operations on NVIDIA GPUs.. Queues are very simple objects with barely any logic from hipSYCL's point of view, so this test should perhaps be named `task_graph_synchronization` or something like that?

Maybe it's also a good idea to execute this test in a loop since synchronization issues tend to only pop up sometimes...

illuhad:
Thank you very much for all the work you put in this! I love what you have done here :)

Please give me a couple of days to play around with it and test it before I merge it; time is scarce for me at the moment. I would also like to have a look at `vec.cpp` - it should work and always has worked for me ;)

I'm absolutely fine with you changing the license header. In fact, especially if we now get more contributors I think we should start thinking about adding an `AUTHORS` file. Credit where credit is due :)

These test cases would now all be tested in a single program, correct? I'm a bit uncertain whether this is ideal since hipSYCL has a global state (the task graph) which means that the test cases are not fully independent. If there's a bug in the task graph, results could depend on the order in which tests are executed. We could of course also wait and see how the tests turn out as you suggest before reorganizing the structure. Perhaps we do not need full isolation between some groups of tests (?). Let me know what you think!

Overall I'm very happy with this (as far as I can say without actually testing it) and would love to see more of what you have in store!

psalz:
I just pushed your suggested test name changes!

> These test cases would now all be tested in a single program, correct?

Correct, and yes, this is a valid concern and I now am a bit worried about this as well as I’m currently writing some code and tests that are interacting with the task graph (I didn’t think too much about it when converting the existing tests to be honest...).

We could add a mechanism to `detail::application` which destroys and re-creates the runtime (including task graph and so on). This might also do something like `hipDeviceReset()` to ensure all streams are flushed etc. A more ad-hoc solution would be to simply run all tests prone to side-effects separately (from a script), e.g. by using `./unit-tests –run_test=task_graph_synchronization`.

> Maybe it's also a good idea to execute this test in a loop since synchronization issues tend to only pop up sometimes...

I don’t think executing this in a loop is a good solution, as (unit) tests should ideally be asserting deterministic behavior. If there are any potential synchronization hazards we should preferably figure them out and thoroughly test them (at some point...). In general it might be worthwhile to investigate ways of introspecting the task graph so tests can be more focused on its state instead of just observing the final result (which arguably turns them into integration tests, really). Maybe we could even go as far as adding some capabilities to manually control when the task graph processes tasks?

----

I briefly tried compiling `vec.cpp` again, and as it turns out it does work with nvcc, however with Clang I get (among other, similiar errors):

```
tests/vector/vec.cpp:104:42: error: no matching member function for call to
      'wzyx'
      cl::sycl::vec<float, 4> v3 = v2_lo.wzyx();
                                   ~~~~~~^~~~
include/CL/sycl/vec.hpp:463:34: note: candidate function not
      viable: call to __host__ function from __device__ function
  HIPSYCL_DEFINE_VECTOR_SWIZZLE4(wzyx, 3, 2, 1, 0)
```

illuhad:
> We could add a mechanism to detail::application which destroys and re-creates the runtime (including task graph and so on). This might also do something like hipDeviceReset() to ensure all streams are  flushed etc. A more ad-hoc solution would be to simply run all tests prone to side-effects separately (from a script), e.g. by using ./unit-tests –run_test=task_graph_synchronization.

I'm fine with both solutions, but to me running tests separately seems to be less error-prone (against e.g. forgetting to reset the runtime in a test), unless we find a way to automatically reset the runtime before each test.

> In general it might be worthwhile to investigate ways of introspecting the task graph so tests can be more focused on its state instead of just observing the final result (which arguably turns them into integration tests, really). Maybe we could even go as far as adding some capabilities to manually control when the task graph processes tasks?

Interesting idea. In general, the control we gain by such an approach would be limited, as the task graph just hands over tasks to the CUDA/HIP runtime as soon as all dependencies are fulfilled. When they are actually executed depends both on the CUDA/HIP driver and hardware capabilities regarding concurrent execution if they are in separate streams. I think that this makes it difficult to test for synchronization hazards in an isolated manner, since ultimately such issues may depend on the hardware. We could use hipCPU though, which guarantees to always process each stream in a separate thread and, since we can modify that code as we please, gives us full control over the hardware behavior.
In the context of my work I'm also interested in tuning the task graph which also requires batching tasks together and submitting them in one go. I think this would already make testing easier since we could just submit one "batch" (i.e. sub-graph) per test and therefore already have more control over when stuff gets submitted.
To focus more on one aspect in our testing, we could also rely on raw HIP as much as possible in our testing and use hipSYCL only for the parts that we actually want to test. E.g. perhaps we could do kernel testing without the task graph, accessors and buffers. For this, we would probably have to polish hipSYCL's HIP interoperability a bit more, which is on my todo list anyway :)
The existing test cases would then be more like higher-level integration tests.

I suspect that the `vec.cpp` problem can be solved easily: The definitions of the simple swizzle functions here https://github.com/illuhad/hipSYCL/blob/4658388425f9227a307fd50fc8d9852372ba2e72/include/CL/sycl/vec.hpp#L359-L375
indeed lack the required `__host__ __device__` attributes. The big question though is why this compiles fine with nvcc... Might have to do something with `syclcc` compiling with relaxed constexpr support when using nvcc (which allows these attributes to be omitted on const expressions), and these functions doing a lot of static compile-time stuff..

psalz:
I've pushed two more changes. The first adds the missing `__host__` and `__device__` attributes for the vector swizzle functions and converts the tests to unit tests. In the second I've added a new `cl::sycl::detail::application::reset` function which is called after every unit test through a test-suite-wide fixture. I'm calling `hipDeviceReset` after I've recreated the runtime as I didn't want to assume that `cl::sycl::device` (and e.g. `detail::set_device`) is usable if the runtime does not exist. I assume that all resources are cleaned up through destructors anyways, so this is just for good measure.

illuhad:
Thank you :)

src/libhipSYCL/application.cpp
```diff
@@ -25,54 +25,15 @@
...
-#include <iostream>
-#include <cassert>
+#include "CL/sycl/detail/application.hpp"
 
-#include <CL/sycl.hpp>
+namespace cl {
+namespace sycl {
+namespace detail {
 
+std::unique_ptr<runtime> application::rt = std::make_unique<runtime>();
```
illuhad: While debugging issue #47, a question popped up that you perhaps already thought about when you wrote this code and already have an answer to. I was wondering if it could happen that actually we end up with two `runtime` objects: One inside `libhipSYCL` and one inside the user application by means of the included `application.hpp`. In practice this doesn't happen on Linux because the runtime object is exported from `libhipSYCL`. My question is: Is this portable and also works on other platforms? Would it be beneficial to also move the other functions inside `application` to the `cpp` file to make this more robust?

psalz: You're right, this might be an issue on Windows (or which *other platforms* are we talking about?). I actually thought about moving the member's implementations to the new source file, but apparently then forgot about it :[. 

Are you planning on adding Windows support?

illuhad: Yes, I'm mostly talking about Windows. While I'm not personally interested in windows support, I also wouldn't like to intentionally break it (especially with such subtle issues) since, at least in principle hipSYCL should be able to run on Windows (the bigger problem would likely be clang-cuda and ROCm).
I think I would prefer if we could just move the member functions to the cpp to make to make this more robust against such subtle problems in case someone decides at some point to try using it on Windows :)

psalz: I just made a PR with these changes.

Also, regarding clang-cuda and Windows: While investigating the Clang bug I came across [this commit](https://github.com/llvm/llvm-project/commit/66a8b07cd9) - it looks like work is being done to support Windows in Clang 9 :)

illuhad: Nice, that's good to know :)

PR #36, opened 2019-03-27
Improve id and range spec conformance

psalz:
This updates some of the APIs of `cl::sycl::id` and `cl::sycl::range` to be in accordance with SYCL 1.2.1 revision 4. Notably this adds some missing "common interface members" for `id` (namely the `==` and `!=` operators) and fixes some of the issues surrounding `const` members returning non-const references that existed in earlier revisions of the spec.

However, the more interesting part really is the unit tests I wrote to test against these APIs. In order to be able to test 1-, 2-, and 3-dimensional variants for both `id` and `range` without having tons of duplicated code I'm making use of `BOOST_AUTO_TEST_CASE_TEMPLATE`. Some template juggling is required to make it work but I think the final solution is actually quite nice!

### DIFF
--- a/include/CL/sycl/id.hpp
+++ b/include/CL/sycl/id.hpp
@@ -79,10 +79,20 @@ struct id {
     : _data{dim0, dim1, dim2}
   {}
 
+  /* -- common interface members -- */
+
   __host__ __device__
   id(const id<dimensions>& other)
     : _data{other._data}
   {}
+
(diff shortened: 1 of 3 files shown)


psalz:
Oh and I forgot to mention: This is again confirmed working with nvcc, Clang and ComputeCpp 1.1.0.

keryell:
Nice to keep up with the latest spec!

illuhad:
Thank you! :)

PR #37, opened 2019-03-28
Fix item selection and linear id computation, improve spec conformance

psalz:
This addresses two related but distinct issues with the `item` class and also improves its API spec conformance.

The first thing this fixes is the way work items are selected for the user provided kernel functor whenever the global size is not a multiple of the default work group size (e.g. 16/16 for 2D kernels). Previously these were selected/discarded based on their linear id compared to the `range::size()` of the global size. While this works fine for 1D kernels it causes issues for 2D/3D kernels where the items provided to the user kernel can have coordinates that lie outside the requested global size. With this change the decision is now based on a per-dimension comparison, like you already have been using for kernels with an offset.

The second change affects the way `item::get_linear_id()` is computed. Previously, this computation was based on a global size that was determined on-the-fly using `hipBlockIdx_*`, `hipBlockDim_*` and `hipThreadIdx_*`. However, even after selecting the correct items (fixing any potentially issues with `hipThreadIdx_*`) the default work group size would always be used through `hipBlockDim_*`, which of course can lead to wrong results when, again, the global size is not a multiple of that. I've fixed this by storing the actual global size provided by the user within `detail::item_impl`, using that for the linear index computation instead.

I haven't looked into the other dispatch methods (i.e., `nd_item` and so on) for whether similar issues exist there as well, but since the user has to specify the work group size manually (and the global size has to be a multiple of it), I suspect there is no problem.

Lastly I've made some minor improvements to the `item` API, again bringing it more in line with SYCL 1.2.1 revision 4. Most notably the subscript operator now correctly includes the item offset (if any), which I understand wasn't previously possible due to the (now dropped) requirement of returning a reference, in combination with the zero-overhead storage of the offset.

---

I've also snuck in a small change that fixes the CMake includes variable for Boost (i.e., `Boost_INCLUDE_DIRS` - no idea how I previously came up with `Boost_HEADER_FILES`).

### DIFF
--- a/include/CL/sycl/handler.hpp
+++ b/include/CL/sycl/handler.hpp
@@ -62,12 +62,28 @@ __global__ void single_task_kernel(Function f)
   f();
 }
 
+template<int dimensions, bool with_offset>
+__device__
+bool item_is_in_range(const item<dimensions, with_offset>& item,
+                      const sycl::range<dimensions>& execution_range,
+                      const id<dimensions>& offset)
+{
+  for(int i = 0; i < dimensions; ++i)
+  {
+    if(item.get_id(i) >= offset.get(i) + execution_range.get(i))
+    {
+      return false;
+    }
+  }
+  return true;
+}
+
 template<int dimensions, class Function>
 __global__ void parallel_for_kernel(Function f,
                                     sycl::range<dimensions> execution_range)
 {
-  item<dimensions, false> this_item;
-  if(this_item.get_linear_id() < execution_range.size())
+  item<dimensions, false> this_item{item_impl<dimensions>{execution_range}};
+  if(item_is_in_range(this_item, execution_range, id<dimensions>{}))
     f(this_item);
 }
 
@@ -76,19 +92,8 @@ __global__ void parallel_for_kernel_with_offset(Function f,
                                                 sycl::range<dimensions> execution_range,
                                                 id<dimensions> offset)
 {
-  item<dimensions> this_item{offset};
-
-  bool item_is_in_range = true;
-  for(int i = 0; i < dimensions; ++i)
-  {
-    if(this_item.get_id(i) >= offset.get(i) + execution_range.get(i))
-    {
-      item_is_in_range = false;
-      break;
-    }
-  }
-
-  if(item_is_in_range)
+  item<dimensions> this_item{item_impl<dimensions>{execution_range}, offset};
+  if(item_is_in_range(this_item, execution_range, offset))
     f(this_item);
 }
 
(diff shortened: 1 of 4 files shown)


psalz:
I just realized that I apparently forgot to add a test case for `item::get_linear_id` (which is ironic), so I've just pushed one. (Also I fixed two typos that crept in during cleanup...)

include/CL/sycl/handler.hpp
```diff
@@ -66,8 +66,19 @@ template<int dimensions, class Function>
...
+
+  bool item_is_in_range = true;
+  for(int i = 0; i < dimensions; ++i)
+  {
+    if(this_item.get_id(i) >= execution_range.get(i))
+    {
+      item_is_in_range = false;
+      break;
+    }
+  }
```
illuhad: Could we perhaps add a function so that we can share code between `parallel_for_kernel` and the similar implementation in `parallel_for_kernel_with_offset`?

tests/unit_tests.cpp
```diff
@@ -523,4 +545,248 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(id_api, _dimensions, test_dimensions::type) {
   // TODO: In-place and binary operators
 }
 
+// Helper type to construct unique kernel names for all instantiations of
+// a templated test case.
+template<typename T, int dimensions>
+struct kernel_name {};
```
illuhad: Is it really necessary to have a unique kernel name? hipSYCL is more permissive than the SYCL spec because CUDA/HIP don't require this. kernel invocations in hipSYCL *should* work without specifying a kernel name.

include/CL/sycl/handler.hpp
psalz: Sure, I'll look into it!

tests/unit_tests.cpp
psalz: It's certainly not *necessary*, but I did this so that 1) the tests are written in valid SYCL and thus 2) can be compiled with ComputeCpp or any other spec conformant SYCL implementation! I thought that might prove useful at some point (e.g. to check for differences between implementations), but of course I can take it out, if you prefer.

illuhad: No, it's fine, I was only interested in whether there were problems compiling without a unique kernel name, since that should work :)

include/CL/sycl/handler.hpp
illuhad: Thanks!

psalz:
I moved the `item_is_in_range` functionality into a separate function that is used by both `parallel_for_kernel` and `parallel_for_kernel_with_offset`!

illuhad:
Awesome, thanks!

PR #38, opened 2019-04-02
Fix faulty buffer versioning edge case

psalz:
This fixes an edge-case with buffer versioning, where accesses using `discard_*` access modes would cause host/device versions to not be bumped appropriately, ultimately resulting in an invalid buffer state.

For example:

1. A buffer is accessed on the host using e.g. `discard_write`, resulting in `host@1` and `device@0`.
2. Then, the buffer is accessed on the device using `discard_write`. Since no copy is required, this line within `register_device_access` is executed: `_host_data_version += version_bump;` - resulting in `host@1` and `device@1`.
3. A subsequent host access, e.g. using `read`, will determine that the buffers are in sync and no transfer is required, even though the buffer has last been written to on the device.

I removed the branching decision for how to bump the buffer version based on whether a copy is required or not. Instead, I'm simply always increasing the buffer version by 1 over the current version on the other side. Is there any downsides to this that I am not seeing?

I've also added the optimization for `copy_required` when discarding previous contents to `register_host_access` that already existed within `register_device_access`.

### DIFF
--- a/src/libhipSYCL/buffer.cpp
+++ b/src/libhipSYCL/buffer.cpp
@@ -491,21 +491,25 @@ buffer_action buffer_state_monitor::register_host_access(access::mode m)
   }
   else
   {
-    // Make sure host is up-to-date before reading
-    bool copy_required = this->is_host_outdated();
-    bool read_only = (m == access::mode::read);
-
     HIPSYCL_DEBUG_INFO << "buffer_state_info: host access, current data versions: host-side is @"
                        << _host_data_version
                        << ", device side is @"
                        << _device_data_version
                        << std::endl;
 
+    // Make sure host is up-to-date before reading
+    bool copy_required = this->is_host_outdated();
+    bool read_only = (m == access::mode::read);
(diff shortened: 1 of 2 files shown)


illuhad:
Oh dear, these buffer synchronization rules are nasty... 
If you just set `_host_data_version = _device_data_version + version_bump` unconditionally, that would mean that the version of the `_host_data_version` remains the same if you have several host write accesses. This *should* be fine since the host still remains ahead of the device, but I'm not sure if this breaks some other assumptions..
I would (also for clarity) suggest that we do:
`_host_data_version = max(_host_data_version,_device_data_version) + version_bump`
I think this conveys more clearly what we want to achieve and should still fix your problem, right?

psalz:
Yes, I like the idea - pushed!

psalz:
Shoot, this included some white space changes my editor automatically made which I didn't want to push. Hold on...

illuhad:
Thank you! :)

Issue #21, opened 2019-01-22
Fully support syclcc in CMake


psalz:
I’ve been trying to make hipSYCL work with a more “complex” CMake setup than what is used for the tests, in particular, one that finds and links other libraries (standard CMake stuff, really). I did manage to get it working, however not without some changes.

Right now there's a couple of issues as far as I can tell:

The most pressing one is the in-source temporary files. CMake often uses `try_compile` for feature detection or to determine a library version. For example, the `FindMPI` module compiles a small program to see if a particular library it found actually works. The problem is that these files are often located somewhere in a read-only location (if it’s a system installation), so `syclcc` cannot write its temporaries and the `find_package` call simply fails (and the reason is not particularly well communicated by CMake).

Now within `syclcc` you [remarked](https://github.com/illuhad/hipSYCL/blob/191fa6c208ab3961c1e9d76e9ebdaffda39c7170/bin/syclcc#L176) that temporaries are located in-source for more robust include resolution, however so far I’ve had no problems with out-of-source builds. Maybe this could be made configurable through a CLI parameter? In any case I think writing into the source tree is a bit hairy and should best be avoided, at least by default.

Another thing is compiler detection. The very first thing CMake does is to compile a small program (however this time within the local `CMakeTmp` folder) to try and determine the compiler version. Right now this identifies `nvcc` as `gcc`, since (afaik) that is the default host compiler of `nvcc`. I’m not sure what the implications of this are, i.e. if there is any difference in parameters that will be passed to `syclcc`, e.g. something `gcc`-specific, which could then maybe interfere with the Clang preprocessing passes. As an aside, and to alleviate this, it might be worth considering setting the host compiler of `nvcc` to Clang by default, which would fit better with the rest of the compilation stages anyway, as they are all LLVM-based.

In addition to these points, if `/tmp` is a `tmpfs` mount, compilation performance could also be greatly improved by moving all temporaries there.

Finally, something that might be worth investigating is to provide a CMake function like ComputeCpp's `add_sycl_to_target`, as [provided by their SDK](https://github.com/codeplaysoftware/computecpp-sdk/blob/163fdf3b952873aa17d49492ce317f462042d026/cmake/Modules/FindComputeCpp.cmake#L380). I see several benefits to this:
-	First it would allow to only compile certain translation units with the device compiler. For larger projects this could really cut down on compilation times.
-	Secondly, this would improve compatibility with existing projects that work with ComputeCpp. 
[...]

PR #40, opened 2019-04-11
Improve CMake support

psalz:
This improves hipSYCL's CMake support by...

1) ...moving to out-of-source builds (and adding `-isystem` to the flags passed through to the source transformation passes) which facilitates the "basic" use case of setting `CMAKE_CXX_COMPILER=syclcc`, as detailed in #21.
2) ...generating a CMake package config file during installation, which can be used to add hipSYCL to specific targets in a more idiomatic CMake fashion.

For the former I've added support for the `--hipsycl-transform-dir=` and `--hipsycl-main-output-file=` CLI parameters to `hipsycl_rewrite_includes`, as already supported by `hipsycl_transform_source`. A single output path parameter would suffice here, but I opted to align it with the existing convention. To allow the device compilers to find headers that have not been rewritten I'm adding the source directory of every translation unit as an include path (as again is already being done for the transformation passes).

The package config file is a bit more involved. Ultimately, as discussed in #21, I wanted to to provide a `add_sycl_to_target` function, as that seems to be an emerging convention between different SYCL implementations. This would make it more convenient to support multiple SYCL implementations from within a single `CMakeLists.txt`. Due to hipSYCL's nature -- not having a separate device compilation pass which generates an integration header -- I first thought it might not be possible to provide such a function, and that instead we'd have to go for something like CUDA's (now deprecated) `cuda_add_executable`. However even that turned out to be problematic, as simply setting `CMAKE_CXX_COMPILER=syclcc` before calling `add_executable` does not work; CMake seems to rely on global state when deciding on the compiler being used for any given target (i.e., setting the compiler back to the previous value at the end of `add_sycl_to_target` will cause the SYCL target to be compiled using that compiler as well).

Together with @w4rh4wk I came up with an entirely different solution, that is pretty nifty I think: We've added `SYCL` as a completely separate CMake language (similar to e.g. `CXX` or `CUDA`). Using `set_source_files_properties` it is then easy to mark all source files passed into `add_sycl_to_target` as `SYCL`, thus instructing CMake to use `syclcc` to compile them.

I've opted to not take any additional parameters for `add_sycl_to_target`, as neither does ComputeCpp's implementation. Instead, hipSYCL is configured through CMake variables. Besides supporting optional parameters such as `HIPSYCL_GPU_ARCH`, the CMake module includes first-class support for `HIPSYCL_PLATFORM`. However, unlike bare-bones `syclcc`, the information about which platforms are available is not based on the compiler executables that can be found within `$PATH` at the time of compilation, but instead on which backends were actually compiled and installed. Furthermore, the check for whether multiple platforms are available (and none has been explicitly chosen) is done during CMake configuration, not during compilation. This I think greatly improves the overall developer experience.

The approach to using this package config is as follows:

1) Build and install hipSYCL.
2) Within the user application's `CMakeLists.txt`, do a `find_package(hipSYCL CONFIG)`.
3) (edit - forgot this step): Add a target as usual using `add_executable` or `add_library`, then call `add_sycl_to_target(TARGET <target> SOURCES <source files>)`.
4) Run `cmake` with `-DCMAKE_PREFIX_PATH=<hipSYCL installation location>/lib`. Potentially add additional parameters such as `-DHIPSYCL_PLATFORM`.
5) Success!

In the future, this approach could even enable use cases where only some translation units are compiled with `syclcc`, while others are simply compiled with the host compiler - which could be good for performance. However for now, unless the other translation units don't touch SYCL at all, this unfortunately doesn't work, as the hipSYCL headers include hardcoded `__host__` and `__device__`  attributes, which will trip up the host compiler.

----

I've marked this as a draft PR as there might very well be some usage patterns that are not yet fully supported by this approach. @VileLasagna please have a look whether this is in line with what you've been envisioning as well!


VileLasagna:
Ooohh. nice stuff. I'll try and have a look over the weekend. The `-isystem` change was actually something I ran into myself as my  own "hello world" project inherits one of those through SDL2.
Enabling it as a language is interesting, I was looking at the `findHIP.cmake` module instead which works more in line with the old findCUDA module. This is even fancier.

Can't wait to try it

VileLasagna:
Hey, there @psalz 

Just gave your branch a spin and it's looking like a massive chunk of the work done. Great stuff.
I've forked it and made a [pull request myself](https://github.com/psalz/hipSYCL/pull/1) but thought I'd kept the conversation in this thread rather than branching it.
In my PR I make 3 changes:

 - I corrected your solution to `-isystem` includes. They (at least for clang) come in the form `-isystem <some dir here>` so are ultimately two separate tokens that need to be forwarded. This is the original change I was planning to PR straight to the trunk, so depending on how @illuhad is feeling, I can split it off in its own thing.

 - I've implemented, in `syclcc` the change I mention in #21 , where syclcc will let you override the platform from what it reads from env. It does print out a message, but I'm considering suppresing it both platforms match

 - On your `CMakeSYCLInformation.cmake` I've added a few lines where it reads HIPSYCL_PLATFORM from the environment when trying to determine the default platform if it detects more than one.

With these tweaks, I've adjusted my own [CMakeLists](https://gitlab.com/VileLasagna/toyBrot/blob/master/SYCL/CMakeLists.txt) for my test project where I'm playing with SYCL and have to say, it's looking almost passable =P. Definitely a massive improvement already over having to specify syclcc as CMAKE_CXX_COMPILER (which to me also means I don't need to section off this build from the other 9 in that project).  

All the celebration out of the way, things I bumped into that to me are the next steps here are:

 - HIPSYCL_PLATFORM needs to be made a proper cache variable, which is easy enough, really

 - CMake still doesn't understand many of its own properties in the context of SYCL as a new language, a few examples I bumped into were: Language standards, `POSITION_INDEPENDENT_CODE`, `INSTALL_RPATH_USE_LINK_PATH`...

 - The brunt of CMake implicit properties still need to be properly defined (`INCLUDE_DIRECTORIES`, `LINK_DIRECTORIES`, `LINK_LIBRARIES`....). The annoying thing here is that we'll have 4 sets of each (cpu, cuda, nvcc and rocm). Need to think if the best way to do this is to indeed use implicit properties for SYCL or, perhaps, provide `hipSYCL::rocm` `hipSYCL::nvcc`... etc targets that provide them. It's a bit of a mindset choice I guess.

 - Something that may cross into `syclcc` as well: I actually failed to build the project with CUDA at all. Being in Arch meant to me that clang didn't seem to like CUDA 10.1 and nvcc didn't like my version of gcc, and I couldn't get -ccbin to be forwarded properly so I could do the equivalent of `CMAKE_CUDA_HOST_COMPILER`. I think something that will help clarify the process as a whole would be the splitting of options, if not for each language (you add `SYCL_CUDA_OPTIONS`, `SYCL_HCC_OPTIONS`....) at least `SYCLCC_OPTIONS` and `SYCL_BACKEND_OPTIONS` for things that need to be specifically forwarded to the actual underlying compilers (believe me, nvcc already gives me enough grief on this end). But, as I mentioned, maybe the best way of doing it is tweaking a bit how syclcc treats arguments, or perhaps forking/wrapping it specifically for this use, as I believe it ends up going in the opposite direction from what @illuhad wants to do, which is making syclcc as transparent as possible, and this would be making it explicitly an intermediate... 

 - Bit of a minor. you mention that you took the  `add_sycl_to_target` nomenclature from some other implementation... if they stick to it, I do think standardisation is pretty good. Otherwise, I feel it's a bit of a weird function name and, from how you use it, I guess `target_sycl_sources` would be a better description
 
I'm willing to help on polishing this up; I'm pretty excited by what's here already. Keen on hearing how you guys are feeling on how this PR currently is

psalz:
Thanks for your feedback and the PR!

> I corrected your solution to `-isystem` includes. They (at least for clang) come in the form `-isystem <some dir here>` so are ultimately two separate tokens that need to be forwarded.

You are absolutely right. When I added this a couple of months back I was just happy that my `find_package(MPI)` would finally compile (while uses a `try_compile` call with an `-isystem` internally) so I never even bothered to check whether this was actually handled correctly... Thanks!

> On your CMakeSYCLInformation.cmake I've added a few lines where it reads HIPSYCL_PLATFORM from the environment when trying to determine the default platform if it detects more than one.

Okay, sure! Why do you have that branching on `HIPSYCL_ENV == hip` though? `syclcc` should accept `hip` just fine (which is a bit of a mess btw, I wish there was only one definitive parameter for each backend...)

> CMake still doesn't understand many of its own properties in the context of SYCL as a new language

Yes, I suspected there are some issues. I didn't want to just randomly copy everything from CMake's CXX definitions into the SYCL language, but if you have a use case for this and can verify it, please do so. I think the PIE-related lines are [here](https://github.com/Kitware/CMake/blob/ef9c38ccf8b6bf12e8d8b539a316a304076948c8/Modules/CMakeCXXInformation.cmake#L105).

> The brunt of CMake implicit properties still need to be properly defined

Can you elaborate? I'm not sure I understand what you mean. All required include paths and libraries are appended by `syclcc` currently; setting them through exported targets like `hipsycl::rocm` would however certainly feel "cleaner" and more CMake-esque.

> Bit of a minor. you mention that you took the `add_sycl_to_target` nomenclature from some other implementation... [...] I feel it's a bit of a weird function name

Yes, that's what ComputeCpp and triSYCL are using (albeit with slightly different parameters). Regarding the name: If at some point we're doing more than just setting the source language, e.g. explicitly linking to libhipSYCL, I think it isn't too bad. In any case, I think it's pretty great that I can now target both ComputeCpp and hipSYCL within my CMakeLists.txt using the exact same line!

VileLasagna:
> Okay, sure! Why do you have that branching on HIPSYCL_ENV == hip though? syclcc should accept hip just fine (which is a bit of a mess btw, I wish there was only one definitive parameter for each backend...)

I did it initially because it felt like you were narrowing it down to "cuda, cpu and rocm" for your options and link libraries.... so I assumed the thing to do would be narrow down the synonyms to those during this phase. Could be that I'm jumping the gun a bit

> Can you elaborate? I'm not sure I understand what you mean (about implicit properties)

So, in addition to  `LINK_DIRECTORIES`, `LINK_LIBRARIES`, `COMPILE_DEFINITIONS` there's an "IMPLICIT" set of a bunch of these variables. They represent "compiler auto", so to speak, such as the link and include directories for your implementation of the STL. I bumped into this one with the previous point about properties. Because I still needed to explicitly list the hipsycl include and link directories (unless I'm misremembering this/ made some other weird stuff during various code mangling).

Ultimately, yeah, this is just a list of stuff to go over that automates SYCL in the same way that other languages get, say, if your project is CUDA, you don't even need to link cudart, it's implicitly done so. And if you set `POSITION_INDEPENDENT_CODE` to `ON` it will know how to forward this to nvcc properly, whereas if someone, say Qt's config files, want to force you to explicitly `-fPIC`, nvcc WILL throw a hissy fit and just refuse to compile anything (lessons were learned the hard way here XD)

> I think it's pretty great that I can now target both ComputeCpp and hipSYCL within my CMakeLists.txt using the exact same line!

I 100% agree that this is the most crucial aspect. Especially from a build perspective, the last thing you need is yet another set of "Ah, but if it's THIS one you find then...."
I actually want to give ComputeCpp a go myself now that I've managed to grab an nVidia card (AMD's drivers for Vega don't have SPIR support, wth?). My test SYCL program has a weird bug that at this point I'm tempted to believe is some weird implementation thing in hipSYCL itself but I've still got to see it crash with a different implementation and swallow my hubris. So this is definitely a nice step in that direction.

VileLasagna:
Hey, @illuhad , the code I'm working on is [this one](https://gitlab.com/VileLasagna/toyBrot). It's essentially a "baby's first mandelbrot" generator that I've been reimplementing using several different technologies as a way to get familiar with them. SYCL is the newest implementation and isn't super clean at this point (there's some straight up hard paths in the CMake, I'm adapting as we go along with the integration...). 

The actual work is done in [FracGen.cpp](https://gitlab.com/VileLasagna/toyBrot/blob/master/SYCL/FracGen.cpp), rest of the code is window and input management using SDL2. The issue in question is that application always works as expected for the first generation but I get a crash if I call generate again (most times, but not always, on the first call). Leads me to believe that there's either some cleanup I'm doing incorrectly or destructors aren't being called properly. As I said, however, I'm getting my feet wet with SYCL just now so I still want to give it a go using a different implementation of it to make sure it's some random dumb thing on my end (wouldn't be the first time)

> I would suggest that, before we put a lot of work into hacking stuff in syclcc, we better wait for the result of the discussion in issue #42, as this has the potential to massively change (and simplify) syclcc.

Yeah, I've been keeping an eye on it and it is, indeed, interesting.

> I'm going to let you duke it out for the most part 

I'm happy to help. @psalz did a lot of the heavy lifting already, so it's a bunch of the not quite as terrible parts left =P

psalz:
> I did it initially because it felt like you were narrowing it down to "cuda, cpu and rocm" for your options and link libraries.... so I assumed the thing to do would be narrow down the synonyms to those during this phase.

We can do that (even though as I said it should currently not be necessary, but I'm all for narrowing those down), but then you'd have to account for all the other synonyms as well, i.e.,
    
    cuda, nvidia
    rocm, amd, hip, hcc
    cpu, host, hipcpu

> there's an "IMPLICIT" set of a bunch of these variables. They represent "compiler auto", so to speak, [...]

I've looked at the CMakeLists.txt of your toyBrot project and setting hipSYCL [libraries and paths like this](https://gitlab.com/VileLasagna/toyBrot/blob/b452b802a47b2b7771cc3bcba4567949e62fd884/SYCL/CMakeLists.txt#L24-31) should most definitely not be required, as they are set by `syclcc` (e.g. [here](https://github.com/illuhad/hipSYCL/blob/9334e2703ef5ea03904df4992d7d3fd8c0f993a9/bin/syclcc#L646)).

---

> I actually want to give ComputeCpp a go myself now that I've managed to grab an nVidia card (AMD's drivers for Vega don't have SPIR support, wth?). 

Yes they removed SPIR support from their drivers some months ago - it's extremely annoying. That's why I'm still running a driver from September 2018 on my RX 580, with that it still works.

psalz:
I rebased this onto master and cherry-picked @VileLasagna's changes in. I also added proper handling of compiler flags for different build types (e.g. `CMAKE_SYCL_FLAGS_MINSIZEREL`).

One thing we should probably have a look at is the fact that as of right now the `HIPSYCL_PLATFORM` environment variable takes precedence over command line parameters (i.e., `-DHIPSYCL_PLATFORM=<>`), which is exactly opposite of how `syclcc` handles it (where `--hipsycl-platform=` takes precedence).

VileLasagna:
Nice one. I'll see if I can have a pick at some more of these features in the near future (sometime next week?), probably fix the caching of HYPSYCL_PLATFORM

psalz:
I just pushed a small change that turns `HIPSYCL_PLATFORM` into a proper cache variable and changes the precedence of the environment variable to be overridden by any explicit command line arguments.

My issue is that an internal release of mine actually somewhat depends on this PR being merged, so I'd like to move it forward a bit more quickly. The good thing about this PR is that it shouldn't break the already established use case of simply setting `CMAKE_CXX_COMPILER=syclcc`. In fact, the changes to `-isystem` and the out-of-source builds should only improve its reliability.

@VileLasagna do you think there currently exist any more glaring problems that should be addressed before this could be merged? Otherwise I'd suggest you simply create a new PR for any additional improvements to this setup, such as the PIE/PIC compatibility!

illuhad:
@psalz Just let me know when you think that it's ready for merging!

VileLasagna:
@psalz I'll take the new tweaks for a spin but I can't immediately think of _glaring_ issues. It's certainly a work in progress but I feel like it's good enough for an initial release with a disclaimer on the README or something of the like

psalz:
Really? It works fine for me! What error are you getting?

I just pushed a small fix for how `HIPSYCL_RESTRICT_DEVICE_HEADER_PATH` is handled, as it previously didn't work properly for the empty path (which was the use case that made me handle it differently to other variables in the first place...). Also I rebased onto master.

VileLasagna:
It looked like nvcc was getting the wrong host compiler, but I got pulled away before I had a chance to look closely in the compiler calls. I'll try and get back on it tomorrow

VileLasagna:
Hey, psalz. I've gone through this and.. I'm not sure why things were working before or exactly what changed that made it break suddenly? But it turns out I was right in that it was inadequate forwarding of variables to NVCC. Nothing on the stuff that we've tweaked, it seems as I got it while looking at illuhad's PR for the accessor leak. I'm making a separate pull request for this, should help the build for nvcc in general, but we seem to be good here

psalz:
Alright, thanks for the feedback @VileLasagna. For a final round of tweaks I've just added proper support for `RPATH`. In addition to defining the corresponding language flags, this also adds exports for the libhipSYCL targets, such as `hipSYCL::hipSYCL_cpu` and so on. These are then linked to the target that is passed into `add_sycl_to_target`, which allows CMake to add the location of the hipSYCL library to the target's `RPATH` (making it much more convenient to use, if hipSYCL is installed in a non-standard location).

illuhad:
Great work guys, thank you! :)

Issue #43, opened 2019-04-12
building throws: syclcc fatal error: hipSYCL platform has been specified multiple times


KestutisMa:
.../hipsycl-cuda-git/src/hipSYCL/src/hipsycl_rewrite_includes/InclusionRewriter.cpp
syclcc fatal error: hipSYCL platform has been specified multiple times

Arch Linux, AUR package https://aur.archlinux.org/packages/hipsycl-cuda-git
that clones hipsycl-cuda-git 20190411.add57fc-1

illuhad:
Thanks for reporting! The AUR PKGBUILD is outdated, I will fix it ASAP. To fix it, try removing the line `export HIPSYCL_PLATFORM=...` from the PKGBUILD. If you have CUDA installed, cmake should automatically enable the CUDA backend for you.

illuhad:
Please try the PKGBUILD from PR #44

KestutisMa:
Thanks for quick reply! `makepkg` was successful using PKGBUILD from #44 branch. Only thing: openmp should also be in dependencies list.

illuhad:
Thanks, PR #44 is merged and OpenMP added as dependency. AUR PKGBUILD will be updated with the new PKGBUILD.

PR #44, opened 2019-04-13
Fix PKGBUILDs for Arch Linux to account for current build system

illuhad:
This fixes issue #43, updating the Arch Linux PKGBUILDs for the current state of the build system. Also taken account is the recent switch to CUDA 10.1 in Arch Linux which is not yet supported by clang 8. So, we are now using cuda 9.2 as dependency which is available in the Arch Linux AUR.

As a sidenote, I noticed that compilation of hipSYCL with nvcc seems to be broken at the moment, likely since the addition of the explicit copy API. However, I think the reason is probably the fault of general nvcc weirdness, not the recent code additions. This is best discussed in a separate issue.

### DIFF
--- a/install/archlinux/CUDA/PKGBUILD
+++ b/install/archlinux/CUDA/PKGBUILD
@@ -8,7 +8,7 @@ url="https://github.com/illuhad/hipSYCL"
 makedepends=("cmake")
 provides=("hipSYCL" "SYCL")
 license=("BSD")
-depends=("cuda-9.2" "llvm" "clang" "boost")
+depends=("cuda-9.2" "llvm" "clang" "boost" "openmp")
 
 source=('hipSYCL::git+https://github.com/illuhad/hipSYCL.git')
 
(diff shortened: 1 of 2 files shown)


psalz:
> As a sidenote, I noticed that compilation of hipSYCL with nvcc seems to be broken at the moment, likely since the addition of the explicit copy API.

Really? It works for me (both for compiling the library itself with `-DUSE_NVCC=1`, as well as compiling e.g. the unit tests with `export HIPSYCL_PLATFORM=nvcc`.

illuhad:
Interesting! I get an error at 
https://github.com/illuhad/hipSYCL/blob/9334e2703ef5ea03904df4992d7d3fd8c0f993a9/include/CL/sycl/accessor.hpp#L196-L198
claiming that using `accessor` without template brackets is C++17. This is obviously nonsense, as the brackets are there, so I suspect that this is nvcc weirdness caused by the particuar combination of CUDA version and host compiler version on Arch Linux.

PR #46, opened 2019-04-15
Fix various minor compiler warnings

psalz:
Currently hipSYCL produces a lot of warnings when compiling with Clang 8 (without even using `-Wall`). Well, actually its not that many, but the diagnostic output Clang produces for each of them is about a page long, making it almost impossible to spot any actual issues within user code. As most of the warnings are due to a few minor issues, I just fixed them in one go.

This mostly converts a bunch of functions that were defined within headers as `static` to `inline` (which has very similar semantics from a linker perspective, but doesn't produce a bunch of "unused function" warnings).

However, there is one more really pesky warning that is not yet addressed by this: Clang complains that `warning: conversion function converting 'cl::sycl::item<3, true>' to itself will never be used`. This is particularly annoying as it not only seems to be a faulty diagnostic (the `cl::sycl::item<Dims, false>::operator item<Dims, true>` doesn't convert to itself...), but there also appears to be no (publicly documented) compiler flag for turning it off!  It certainly isn’t mentioned in the [diagnostics reference](https://clang.llvm.org/docs/DiagnosticsReference.html). I’ve actually written a script that tries to narrow it down from all 700-something flags using a binary search, but no dice.

The only way I could get rid of this warning is to split the definitions of `item<Dims, true>` and `item<Dims, false>` into separate template specializations that inherit their shared members from a common base class. This is of course a larger change and I'm not sure if you'd like to go down that route @illuhad. This also means that e.g. the `detail::item_offset_storage` would no longer be required. Let me know if you'd want me to add this to this PR as well.

### DIFF
--- a/include/CL/sycl/h_item.hpp
+++ b/include/CL/sycl/h_item.hpp
@@ -34,12 +34,12 @@ namespace cl {
 namespace sycl {
 
 template<int dimensions>
-class group;
+struct group;
 
 template <int dimensions>
 struct h_item
 {
-  friend class group<dimensions>;
+  friend struct group<dimensions>;
 
   __device__
   h_item(){}
--- a/include/CL/sycl/accessor.hpp
+++ b/include/CL/sycl/accessor.hpp
@@ -301,10 +301,10 @@ access::target::constant_buffer)) && dimensions > 0 */
            int D = dimensions,
            std::enable_if_t<(P == access::placeholder::false_t &&
                              T == access::target::host_buffer) ||
-                            (P == access::placeholder::true_t &&
+                            ((P == access::placeholder::true_t &&
                             (T == access::target::global_buffer ||
                              T == access::target::constant_buffer)) &&
-                            (D > 0) >* = nullptr>
+                            (D > 0)) >* = nullptr>
   accessor(buffer<dataT, dimensions> &bufferRef,
            range<dimensions> accessRange,
            id<dimensions> accessOffset = {})
--- a/include/CL/sycl/context.hpp
+++ b/include/CL/sycl/context.hpp
@@ -63,8 +63,10 @@ public:
 
     _platform = deviceList.front().get_platform();
 
-    for(const auto dev : deviceList)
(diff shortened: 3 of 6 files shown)


illuhad:
Thank you! Interesting you get such a long diagnostic output.. I've only seen a couple lines with warnings here and there.
I'm tending towards not using (larger) workarounds in our code to circumvent bugs in compiler diagnostics. So, we could wait and see what future clang versions do, or go the other way and, independently try to simplify the code (which is always good), remove the no-offset no-storage optimization and hope that the registers get freed up by dead code elimination passes in the optimizer stage. I would like to see some tests though to check if that actually happens :)

include/CL/sycl/h_item.hpp
```diff
@@ -34,12 +34,12 @@ namespace cl {
 namespace sycl {
 
 template<int dimensions>
-class group;
+struct group;
```
illuhad: I've never seen any warnings related to this forward declaration. What's the problem with forward declaring classes?

psalz: It's not so much about being a `class`, but the actual declaration of `group` being a `struct`. The warning is `warning: 'group' defined as a struct template here but previously declared as a class template; this is valid, but may result in linker errors under the Microsoft C++ ABI`

psalz: (here references the actual definition and the previous declaration is the forward declaration)

psalz:
Interesting that you don't get as many warnings. A typical compilation with Clang for me emits several pages of this: image

include/CL/sycl/h_item.hpp
illuhad: Okay, I see. That makes sense.

illuhad:
Wow! Then it is really time to fix this. No idea why I don't see this..

illuhad:
You're right, especially with a merged `item_base` and `item_impl` this actually seems more elegant than the current implementation. Feel free to implement this :)

PR #48, opened 2019-04-18
Refactor item implementation

psalz:
As discussed in #46, this refactors the implementation of `cl::sycl::item` to use partial template specializations over SFINAE, with the primary goal of silencing a faulty Clang diagnostic about the conversion operator from `item<dims, false>` to `item<dims, true>` never being used.

To construct items, instead of relying on a `detail::item_impl` type that is inaccessible to users, a new `detail::make_item()` can now be used.

I've also removed the global-size-only constructor (or rather, I didn't migrate it from `item_impl` to the new `item_base`), so that an `id` always has to be explicitly provided, as I think it makes for a cleaner API. New calls to `detail::get_global_id` have thus been added in `handler`.

The only thing I'm not too certain about are the changes I had to make to `h_item`. I didn't have time to delve into the semantics of these functions (and according to your comment the spec might be a bit ambiguous here anyway), however I'm pretty sure these functions haven't been getting much use either way. Case in point: Both `h_item::get_global()` and `h_item::get_physical_local()` were broken since I removed/changed the constructors they've been using in #37 (oops!). In any case, the whole parallel for hierarchical invoke stuff is in desperate need of some unit tests!

### DIFF
--- a/include/CL/sycl/h_item.hpp
+++ b/include/CL/sycl/h_item.hpp
@@ -56,7 +56,10 @@ public:
   __device__
   item<dimensions, false> get_global() const
   {
-    return item<dimensions, false>{};
+    return detail::make_item<dimensions>(
+      detail::get_global_id<dimensions>(),
+      detail::get_global_size<dimensions>()
+    );
   }
 
   __device__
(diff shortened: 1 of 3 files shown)


illuhad:
Thanks! Your changes to `h_item` preserve the current behavior, which I believe is correct since `h_item` is available at work-item level inside `parallel_for_work_item`. So, having the global id refer to the "actual" global id is the only thing that makes sense to me.
The main issue I see with the current `h_item` implementation is that it doesn't correctly distinguish between physical and logical local id. But this is of course unrelated to your PR. And yeah, I agree, hierachical dispatch urgently needs more love. 
I was planning on working on it anyway in conjunction with some optimizations of the CPU backend since hierarchical dispatch allows for a more efficient parallelization when compiling for the host (i.e. using SYCL as a library instead of relying on special compilers). This better portability is also why the hierarchical approach seems to be somewhat preferred in modern SYCL :)

PR #51, opened 2019-04-19
Fix nvcc compilation of libhipSYCL

psalz:
It looks like I broke the compilation of libhipSYCL using `nvcc` with #48, as apparently it has difficulties to find the `cl::sycl::range` type, even when disambiguated with the `sycl` namespace. I've added a workaround that creates an alias, then everything works. I sure love nvcc...

By the way: It would be great if something like this could also be caught by Travis (as I frankly don't try `-DUSE_NVCC=1` very often) - maybe we should add another build config?

### DIFF
--- a/include/CL/sycl/item.hpp
+++ b/include/CL/sycl/item.hpp
@@ -118,8 +118,10 @@ struct item<dimensions, true> : detail::item_base<dimensions>
   }
 
 private:
+  template<int d>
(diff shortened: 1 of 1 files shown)


illuhad:
Thanks! nvcc is always good for nice surprises :) I agree that a Travis build for nvcc would be great, at least as long as we officially claim to support nvcc with the source-to-source transformation.

Issue #49, opened 2019-04-18
Silent cudaLaunch failures when compiling with clang's CUDA implementation


illuhad:
Moving the discussion from #42 to a dedicated issue. As mentioned in #42, @psalz found out that this simple code:

```cpp
#include <CL/sycl.hpp>

int main() {
    cl::sycl::queue queue;
    cl::sycl::buffer<float, 1> buf(10);

    queue.submit([&](cl::sycl::handler& cgh) {}); // The culprit

    queue.submit([&](cl::sycl::handler& cgh) {
        auto acc = buf.get_access<cl::sycl::access::mode::discard_write>(cgh);
        cgh.parallel_for<class fail>(buf.get_range(), [=](cl::sycl::item<1> item) {
            acc[item] = 1.f;
        }); 
    }); 

    return 0;
}
```

Silently fails to run the kernel. The error is however not restricted to this particular code and can also seemingly "strike at random". An error can only be seen if using `cuda-memcheck`, which reveals error `cudaErrorInvalidDeviceFunction (error 8) due to "invalid device function"  on CUDA API call to cudaLaunch`.

We know:
* the problem disappears when changing small seemingly unrelated bits of code (in this case, either removing the first empty command group or moving it after the second command group solves the issue)
* Generated device code and cuda launch code on the host side is the same for both working and non-working versions
* Reproducible (at least?) with clang 8
* Everything works fine if compiled with nvcc.

Things to try:
* Does adding an explicit `queue.wait_and_throw()` at the end change anything? Terminating a program without synchronization either via `queue` or by creating a host accessor is not allowed by spec, although hipSYCL historically has handled that well. The question is: Could it happen that hipSYCL in some destructor tries to run the kernel while CUDA runtime has already started shutting down? **EDIT: No, explicit synchronization doesn't help**

illuhad:
The kokkos guys may have a similar (same?) problem: kokkos/kokkos#1547

illuhad:
Here it's suggested that such errors can be caused by using data after it has left scope: kokkos/kokkos#1173

psalz:
I have some news on this: We discovered that the bug appears to have been fixed in the current Clang trunk (i.e., Clang 9), however an assertion is still thrown in debug builds. I've also narrowed the fix down to a particular commit, and created an issue about it in the LLVM bug tracker: https://bugs.llvm.org/show_bug.cgi?id=41597.

illuhad:
Excellent, thank you!

[...]

Issue #47, opened 2019-04-15
Potential hipSYCL related crash in Mandelbrot fractal viewer


illuhad:
@VileLasagna, I'm moving the discussion here so that we don't clutter the cmake discussion :) I could not reproduce any crashes so far. Your SYCL code looks okay at first glance. However, I don't see a fractal - there's a black screen and if I click and drag the mouse, white rectangles are drawn on top. Is this expected (e.g. unfinished rendering code)?

EDIT: Nevermind - it seems that although the command group handler runs, the actual kernel is never running. This is likely a problem in hipSYCL.

VileLasagna:
The white rectangle is a zoom selection area. Curious that it's not rendering for you, that is weird. Expected output is [like so] (https://photos.app.goo.gl/AR1uSjquvDULrZNo8)

And the error I get on subsequent runs is [this one](https://photos.app.goo.gl/7DZx9XmUwwqJEZfe9)

illuhad:
Okay, the kernel invocation problem seems to be a clang issue - doesn't happen when using nvcc. It's now working for me, without crashes.
Can you please try commenting out the lines causing the assertion failure. I've just checked the code and I'm skeptical if that assertion statement is actually correct.

VileLasagna:
I'll give it a go tonight. I was remoting in from work but it seems my home PC has either crashed or was turned off somehow.

psalz:
Without having tested it myself, this

>the kernel invocation problem seems to be a clang issue

sounds a lot like the weird compiler bugs I'd previously run into while trying to get Catch2 to compile with Clang. Try running it through `cuda-memcheck` - is the error something like `invalid call to cudaLaunch`? I actually wanted to find a small reproducible case of this for #42 today.

VileLasagna:
Hello @illuhad I've tried commenting out that assertion and sure enough, I seem unable to reproduce the crash now, so it seems you were on the money. Looks like I've got a working SYCL thingie for now =P

With that being the case, do you think there's still value in getting the crash back and trying to push it through cuda-memcheck, @psalz? I can try and get to it later this week but I'll need to massage my build into working for either CUDA or NVCC.

psalz:
If you know that the kernel actually executes now it's fine, but as far as I can remember the problem with the Clang bug is that it fails silently, only `cuda-memcheck` will report the issue.

[...]

PR #52, opened 2019-04-20
Add SYCL extension: Mechanism for automatic calls to require() for placeholder accessors 

illuhad:
This is an initial (experimental!) implementation of a SYCL extension that allows to `require()` placeholder accessors automatically, as seen in the following example:

```cpp

cl::sycl::queue q;
cl::sycl::buffer<int, 1> buff{/* Initialize somehow */};
cl::sycl::accessor<int, 1, cl::sycl::access::mode::read_write, 
      cl::sycl::access::target::global_buffer, cl::sycl::access::placeholder::true_t> acc{buff};

// This will call handler::require(acc) for each command group subsequently
// launched in the queue, until the `automatic_requirement` object is destroyed
// or `automatic_requirement.release()` is called.
auto automatic_requirement = cl::sycl::vendor::hipsycl::automatic_require(q, acc);
// The member function `is_required()` can be used to check if
// if the automatic requirement object is active:
assert(automatic_requirement.is_required());

q.submit([&] (cl::sycl::handler& cgh) {
  // No call to require() is necessary here!
  cgh.parallel_for<class kernel1>([=] (cl::sycl::item<1> idx){
    /* use acc according to the requested access mode here */
  });
});

q.submit([&] (cl::sycl::handler& cgh) {
  cgh.parallel_for<class kernel2>([=] (cl::sycl::item<1> idx){
    /* use acc according to the requested access mode here */
  });
});

// After release(), acc will not be required anymore by kernels submitted to the queue
// afterwards.
automatic_requirement.release();
assert(!automatic_requirement.is_required());

/* Kernel submissions that do not need acc could go here */

// acc will now be required again
automatic_requirement.reacquire();

q.submit([&] (cl::sycl::handler& cgh) {
  cgh.parallel_for<class kernel3>([=] (cl::sycl::item<1> idx){
    /* use acc according to the requested access mode here */
  });
});

// If the automatic_requirement object goes out of scope, it will release the auto requirement
// if it is active.
```

This extension is always activated. The macro `HIPSYCL_NO_EXT_AUTO_PLACEHOLDER_REQUIRE` is defined if the given hipSYCL version supports this extension.

This extension serves two purposes:
1. Avoid having to call `require()` again and again if we want to use the same accessor in many subsequent kernels. This can lead to a significant reduction of boilerplate code.
2. Simplify code when working with SYCL libraries that accept lambda functions or function objects. Imagine there's a `sort()` function in a SYCL library and you want to pass a comparator to that function. Currently, there's no easy way to access some independent data in that comparator because accessors must be requested in the command group handler. This would not be possible in that case since the command group would be spawned internally by `sort`, and the user has no means of accessing it.

Note that, with this extension, you could potentially end up with lots of command groups that have no content except for the kernel launch. This offers the opportunity for some additional removal of boilerplate code, i.e. we could introduce some shortcut for kernel launch like so:
```cpp
// q,acc are queue/accessor
auto automatic_requirement = cl::sycl::automatic_require(q, acc);

q.submit_parallel_for<class kernel3>([=] (cl::sycl::item<1> idx){
   ...
});
```
where `submit_parallel_for()` would call `submit()` with just the kernel launch inside.

PS: I've added a test case for the extension, however the unit tests now seem to be affected by issue #49. I've also noticed that when compiling with clang, right after the statement `acc[0]=1` has executed in line 1004, actually `acc[0]==2`. I've verified this both with printf and by copying back to host and checking there. This looks like another wtf-class clang bug to me.
When moving the test case out of `unit_tests.cpp` in its dedicated file without other tests, both problems disappear and things work (which again indicates some invalid state, potentially in CUDA), although at the end of the test issue #47 manifests itself. Finally, I can reproduce that problem as well in its full glory :)


illuhad:
Fixed remaining issues, and now using convention of having extension functions residing in `cl::sycl::vendor::hipsycl`. Should now be ready for merging.

PR #53, opened 2019-04-21
Fix memory leak caused by shared_ptr dependency cycle

illuhad:
There is a cyclic `shared_ptr` dependency causing the memory leak mentioned in issue #47:

* `buffer_ptr` maintains a list of related `task_graph_node_ptr` objects in the access log for dependency calculation
* `task_graph_node_ptr` contains a closure for asynchronous execution, such as a kernel launch
* kernel launch closure will in general capture accessors
* accessors maintain a `shared_ptr` to the `buffer_ptr` to which they refer via the `accessor_tracker`.
... and the cycle (which prevents any buffer from being freed!) is complete.

This PR fixes this by resetting the closure in the `task_graph_node_ptr` to a default-constructed function object after submission, thus breaking the cycle.

### DIFF
--- a/src/libhipSYCL/task_graph.cpp
+++ b/src/libhipSYCL/task_graph.cpp
@@ -134,6 +134,13 @@ task_graph_node::submit()
 
     _stream->activate_device();
     task_state state = _tf();
+    
+    // Remove the task functor after execution to avoid
+    // cyclic dependencies between buffer objects (that store
+    // task_graph_node_ptrs for dependency calculation) and
+    // the captured accessors
+    this->_tf = task_functor{};
+
     _submitted = true;
 
     if(state == task_state::enqueued)
@@ -158,6 +165,7 @@ task_graph_node::submit()
     // Submitted must be set to true to avoid
     // subsequent submissions
     _submitted = true;
+    this->_tf = task_functor{};
     // ToDo: Should we also consider the task as done here?
     // Or at least trigger the callback?
 


psalz:
Huh, that's certainly an interesting bug :)

psalz:
Well, it looks like this broke something, or at least made an existing issue more prevalent for me: I was just getting ready to play around some more with #49, when I noticed that this change causes both Clang and nvcc to error with `Program hit cudaErrorCudartUnloading (error 29) due to "driver shutting down" on CUDA API call to cudaStreamSynchronize.` (again only reported through `cuda-memcheck`). This is both with and without the empty command group.

I actually ran into this exact error before, while trying to construct a minimal demo case for #49, which didn't have any buffers, like this: 
```c++
#include <CL/sycl.hpp>

int main() {
    cl::sycl::queue queue;

    queue.submit([&](cl::sycl::handler& cgh) {
        cgh.parallel_for<class fail>(cl::sycl::range<1>(10), [=](cl::sycl::item<1> item) {
            printf("Foo\n");
        });
    });

    return 0;
}
```

(This is of course before f2175db, but it also still happens afterwards).

I briefly tried to debug this, but then decided it's a fringe case and the other error was more pressing. Well, it looks like there is more to it. Maybe this and #49 are related after all?

illuhad:
I know this one and I believe it's mostly harmless, at least as long as we only get it in conjunction with `cudaStreamSynchronize`  and `cudaStreamDestroy` calls.

I believe the underlying issue is that the unloading order of shared libraries is undefined. So, it can be that the CUDA runtime gets unloaded before `libhipSYCL`. In that case, when the destructors of our static objects run, `stream_manager::~stream_manager()` tries to to synchronize one last time before destroying the stream. This then causes this error message.
This should not create any problems for the user, because if he's interested in the result of his computation at all, he will have some host accessor which already synchronizes the work or will have an explicit call to `queue::wait()` which also makes sure that all is done.

Why we don't get error messages from hipSYCL I don't know, perhaps this is so late post-`main` that printing doesn't work anymore or perhaps cuda just reports `cudaSuccess` in that case because it assumes that this is not so important?

I'm not sure how to fix this. Perhaps by making `libhipSYCL` a static library? In any way, I don't think it's that urgent for now.

illuhad:
Ah, I think the reason why we don't get error messages is because
https://github.com/illuhad/hipSYCL/blob/f2175db7cbdc19e0e767f905f5cbc61d8f406eb5/src/libhipSYCL/queue.cpp#L65-L66
are not surrounded by `detail::check_error()` :D

psalz:
I see, and I presume we only get this error now that objects are actually being released within static destructors, instead of them leaking ;-). However: We are linking the user program and libhipSYCL itself with `-lcudart_static` (i.e., statically), so shouldn't the CUDA runtime be guaranteed to exist until libhipSYCL and the user programs have exited...?

illuhad:
I'm not sure how it works in detail (and of course nvidia doesn't reveal anything..) but as far as I know the CUDA Runtime relies on `libcuda.so` provided by the GPU driver itself (and likely loads it explicitly with `dlopen`). So, even if you link the *runtime* statically, I believe that the program still loads the driver API in `libcuda.so` dynamically. Perhaps the error message `"driver shutting down"` refers exactly to that: That it's a problem of the driver portion, not the runtime.
To verify, you can try having a look at the mapped virtual memory ranges of a running CUDA program linked against `libcudart_static` (both the GNOME system monitor and ksysguard can do that). I would expect a `libcuda.so` to show up there.

Issue #50, opened 2019-04-18
error: __float128 is not supported on this target


jeffhammond:
I am trying to build hipSYCL for the Radeon GPU (`Advanced Micro Devices, Inc. [AMD/ATI] Polaris 22 [Radeon RX Vega M GH] (rev c0)`) that is associated with the Intel Hades Canyon NUC (`Intel(R) Core(TM) i7-8809G CPU`).

I cannot figure out how to get past the `__float` error.

# System C++ compiler
`
$  /usr/bin/c++ -v
Using built-in specs.
COLLECT_GCC=/usr/bin/c++
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/8/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-linux-gnu
Configured with: ../src/configure -v --with-pkgversion='Ubuntu 8.2.0-7ubuntu1' --with-bugurl=file:///usr/share/doc/gcc-8/README.Bugs --enable-languages=c,ada,c++,go,brig,d,fortran,objc,obj-c++ --prefix=/usr --with-gcc-major-version-only --program-suffix=-8 --program-prefix=x86_64-linux-gnu- --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --with-sysroot=/ --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-gnu-unique-object --disable-vtable-verify --enable-libmpx --enable-plugin --enable-default-pie --with-system-zlib --with-target-system-zlib --enable-objc-gc=auto --enable-multiarch --disable-werror --with-arch-32=i686 --with-abi=m64 --with-multilib-list=m32,m64,mx32 --enable-multilib --with-tune=generic --enable-offload-targets=nvptx-none --without-cuda-driver --enable-checking=release --build=x86_64-linux-gnu --host=x86_64-linux-gnu --target=x86_64-linux-gnu
Thread model: posix
gcc version 8.2.0 (Ubuntu 8.2.0-7ubuntu1) 
`

# CMake output
```
-- The C compiler identification is GNU 8.2.0
-- The CXX compiler identification is GNU 8.2.0
-- Check for working C compiler: /usr/bin/cc
[...]

PR #55, opened 2019-04-22
Minor build fixes in cmake

VileLasagna:
Three minor cmake fixes in hipsycl itself:

1 - Explicitly set `CMAKE_CXX_STANDARD` to 14, in case it doesn't get automagically set 

2 - Set and forwarded `CMAKE_CUDA_HOST_COMPILER` and `CMAKE_CUDA_STANDARD` to the CUDA target if building with NVCC, as those don't always get correctly inherited to match the C++ side of the build

3 - Probed the system for the `ROCM_PATH` variable and used it, if present, to add ROCM's include path to the build. I myself had always been adding the full path manually, but ROCM sets this on a file in profile.d when it install itself, so probing this env variable is an option

None of these changes should break pre-existing builds. The one that could would be no1 if the compiler doesn't support it but it seems to me SYCL doesn't even build with a standard lower than 14 due to lambda shenanigans

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,6 +70,8 @@ set(WITH_CPU_BACKEND true CACHE BOOL "Build hipSYCL with support for host execut
 cmake_policy(SET CMP0005 NEW)
 add_definitions(-DHIPSYCL_DEBUG_LEVEL=${HIPSYCL_DEBUG_LEVEL})
 
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 add_subdirectory(src)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,8 +70,6 @@ set(WITH_CPU_BACKEND true CACHE BOOL "Build hipSYCL with support for host execut
 cmake_policy(SET CMP0005 NEW)
 add_definitions(-DHIPSYCL_DEBUG_LEVEL=${HIPSYCL_DEBUG_LEVEL})
 
-set(CMAKE_CXX_STANDARD 14)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 add_subdirectory(src)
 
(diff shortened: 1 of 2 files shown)


illuhad:
Thank you! :) However, it seems that this PR breaks the Travis CI build (at least with clang-cuda), interestingly with the same error as reported in #50. I've just rechecked and it builds fine without this PR. Could you please investigate? Does this PR depend on PR #40 ?

VileLasagna:
It should build straight off of current master. Can you  give me the console output of your CI so I can have a look?

illuhad:
See [here](https://travis-ci.com/illuhad/hipSYCL/builds/109106693). For testing, you can use the Dockerfiles or singularity definition files in `install` (replace repository with your fork and checkout your branch), as the CI just tries to build those.

VileLasagna:
That's interesting. I run into this specific issue myself quite a bit with clang and nvcc. The workaround is to pass `-D__STRICT_ANSI__` into CXX_FLAGS. Curious it showed up only now in that build. I'll have a play but my guess is that the culprit would be me enforcing CXX_STD to 14. If it just works otherwise in CI I might just revert that one commit.

VileLasagna:
It indeed was the CXX standard... kind of curious that it seemed I bumped into needing it on my side. Good to come in contact with the docker stuff to check before pushing though. I've just pushed a revert commit but can, if you'd prefer to have a cleaner tree, do some evil rebasing trickery and get rid of the commit(s) altogether

psalz:
Without having looked at the CI output our tried building this myself, it might be due to CMake defaulting to GNU standards (i.e., `gnu++14`) on Linux. Try setting `set_property(TARGET <target> PROPERTY CXX_EXTENSIONS OFF)` - at least that's what I previously had to do for a similar sounding issue.

illuhad:
Thanks for all the ideas. I think we should try getting C++14 enabled by default because of the feedback from issue #56, where a missing `-std=c++14` caused compilation failures. @VileLasagna, could you please have a look at @psalz's proposals?

VileLasagna:
Sure thing. Will try and massage that into conformity

PR #57, opened 2019-04-25
Enforce cxx standards in build

VileLasagna:
This should be the correct version of that fix. Passed both docker build tests. I've also removed the now extraneous `--std=c++14` build flag from the ROCm build and used the opportunity to multithread the build in both that dockerfile and the one for CUDA

Thanks @psalz for giving me a reason to revisit every single build at work now that you've shown me CMAKE_CXX_EXTENSIONS =P

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,6 +71,10 @@ cmake_policy(SET CMP0005 NEW)
 add_definitions(-DHIPSYCL_DEBUG_LEVEL=${HIPSYCL_DEBUG_LEVEL})
 
 
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
+
 add_subdirectory(src)
 
 
--- a/install/docker/CUDA/Dockerfile
+++ b/install/docker/CUDA/Dockerfile
@@ -20,4 +20,4 @@ ENV CXX=clang++-8
 ENV PATH=/usr/local/cuda/bin:$PATH
 ENV LIBRARY_PATH=/usr/local/cuda/lib64:$LIBRARY_PATH
 RUN cmake -DCMAKE_INSTALL_PREFIX=/usr -DWITH_CPU_BACKEND=ON -DWITH_CUDA_BACKEND=ON /tmp/hipSYCL
-RUN make install
+RUN make -j$(($(nproc) -1)) install
--- a/install/docker/ROCm/Dockerfile
+++ b/install/docker/ROCm/Dockerfile
@@ -14,6 +14,5 @@ USER root
 ENV CXX=clang++-6.0
 ENV PATH=/opt/rocm/bin:$PATH
 ENV HIPSYCL_GPU_ARCH=gfx900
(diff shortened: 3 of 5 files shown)


illuhad:
Awesome, thank you! Could you please also put implement the changes you did to the Dockerfiles in the singularity definition files (also in `install`). They do exactly the same as the Dockerfiles, so we should try to keep them in sync.

VileLasagna:
Sure thing. Have already shit everything down for the night, but will do it tomorrow morning, should be easy

VileLasagna:
Having a go at this right now and it's not liking a class from hipcpu... wonder if that got bumped while we weren't looking?

```c++
[ 38%] Building CXX object src/libhipSYCL/CMakeFiles/hipSYCL_cpu.dir/application.cpp.o
cd /home/rocm-user/hipSYCL/build/src/libhipSYCL && /home/rocm-user/hipSYCL/bin/syclcc   -DHIPSYCL_DEBUG_LEVEL=1 -DhipSYCL_cpu_EXPORTS -I/home/rocm-user/hipSYCL/include -I/home/rocm-user/hipSYCL/contrib/hipCPU/include  -O3 -DNDEBUG -fPIC   --hipsycl-platform=cpu --hipsycl-bootstrap -std=c++14 -o CMakeFiles/hipSYCL_cpu.dir/application.cpp.o -c /home/rocm-user/hipSYCL/src/libhipSYCL/application.cpp
In file included from /home/rocm-user/hipSYCL/contrib/hipCPU/include/hipCPU/hip/hip_runtime.h:62:0,
                 from /home/rocm-user/hipSYCL/bin/../include/CL/sycl/detail/../backend/backend.hpp:45,
                 from /home/rocm-user/hipSYCL/bin/../include/CL/sycl/detail/task_graph.hpp:32,
                 from /home/rocm-user/hipSYCL/bin/../include/CL/sycl/detail/runtime.hpp:35,
                 from /home/rocm-user/hipSYCL/bin/../include/CL/sycl/detail/application.hpp:33,
                 from /home/rocm-user/hipSYCL/src/libhipSYCL/application.cpp:28:
/home/rocm-user/hipSYCL/contrib/hipCPU/include/hipCPU/hip/detail/runtime.hpp:271:25: error: enclosing class of constexpr non-static member function 'std::size_t hipcpu::device::get_max_shared_memory() const' is not a literal type
   constexpr std::size_t get_max_shared_memory() const
                         ^
/home/rocm-user/hipSYCL/contrib/hipCPU/include/hipCPU/hip/detail/runtime.hpp:184:7: note: 'hipcpu::device' is not literal because:
 class device
       ^
/home/rocm-user/hipSYCL/contrib/hipCPU/include/hipCPU/hip/detail/runtime.hpp:184:7: note:   'hipcpu::device' has a non-trivial destructor
```

VileLasagna:
Double posting as I'm doing stuff now:

Going through his (@misos1) steps and how he bumped into this it would appear that syclcc is the culprit, in a way. It's reading the correct clang when it's set through `CC` and `CXX` but it doesn't get forwarded straight through CMake's variables... The reason It's only being picked up here could be because the standard CC in Bionic is good enough whereas the one in Xenial is (predictably) not.

I'll push a minor tweak that introduces the multithreaded build and all, but I'll leave the environment variables for the time being. Still don't like them so I'll see if I can figure out a way to do this part semi-smartly

illuhad:
Great, thanks for investigating!

PR #59, opened 2019-04-29
Move implementation of application members to CPP

psalz:
Following some additional discussion in #35 I'm moving all the inline static member functions of `application` to the CPP source file, in order to avoid potential linker issues when targeting Windows in the future.


illuhad:
Great, thank you!

Issue #42, opened 2019-04-11
Strategy: The future of the source-to-source transformation


illuhad:
Since we'll have a clang plugin hopefully soonish that will directly allow the clang CUDA/HIP frontend to ingest SYCL code (see issue #34), we could in principle drop the source-to-source transformation entirely.
I'd like to start a discussion here with hipSYCL users and developers to get some feedback on possible futures of the hipSYCL compilation toolchain. Is the source-to-source transformation important to you and we should support both source-to-source and the new clang plugin? Do you need nvcc support? Or is clang support (with the plugin) sufficient for you?

Here are some pros of the source-to-source transformation that come to my mind:
* Allows compilation of SYCL code with nvcc. This can be interesting from a marketing position ("you can do anything that nvcc can and can use the newest CUDA features right away")
* Possible to specify areas in the code with preprocessor definitions that hipSYCL shouldn't modify. This could be beneficial if you're interested in mix-and-match with SYCL and CUDA/HIP code.

The new clang plugin on the other hand gives us:
* Much more robustness and reliability (there are edge cases in the source-to-source transformation...)
* Faster compilation speed
* Solid SYCL support
* Paves the way for runtime selection whether a kernel should be executed on host or device (could in principle however also be implemented with source-to-source and clang's CUDA implementation [but likely not with nvcc])
* Potentially even some parts of the C++ standard library could be used in kernels (although not yet with the initial version of the plugin)
* Implementation of specific optimizations in the future since we have access to the IR

While not impossible, it may require some additional effort to support both the current source-to-source transformation and the new clang plugin approach because the clang plugin treats any function without attributes implicitly as `__host__ __device__`. This means that all functions for SYCL kernels (e.g. math functions) must also support compilation for both host and device. At the moment, we assume in the runtime that everything used in kernels is `__device__` only. This is also assumed by the current source-to-source transformation.
Also, if we still have to support source-to-source, it may limit our ability to implement things with IR transformations.

VileLasagna:
Dropping the transformation step does look attractive from a compiling time standpoint and increased robustness is always good. As much as I'm tempted to just get into a "light the torches" mood, there are a couple of niggles with dropping nvcc:

 - Might become a semi-non-issue in the future as it catches up better but being tied to older versions of CUDA because clang can't handle them yet can get nasty.
[...]

PR #60, opened 2019-04-29
Initial implementation of clang plugin

illuhad:
# Clang Plugin - Initial implementation
This PR implements an initial, working version of the clang plugin as discussed in issue #42 as a **technological preview**. It cannot yet be used for actual SYCL code with the hipSYCL runtime (see caveat 1) **EDIT: It can now be used for actual SYCL code :)**.
I would like to split the changes over several PRs to simplify review, so this is only the first PR of the ongoing clang-plugin work. The main purpose of this PR is not to deliver a solution that can be used in production right away, but to gather feedback. The next PRs will then progressively add in the remaining bits to allow this plugin to work with the rest of hipSYCL.

## Caveats
1. The plugin in this PR cannot yet be used in conjunction with the hipSYCL runtime/device library because we need to change all `__device__` functions in hipSYCL to `__host__ __device__` (see below). **EDIT: This is now done, and the plugin can be used to compile real SYCL code**
2. Not yet tested on AMD and only tested with clang 8.
3. hierarchical parallel for support is implemented but not yet tested because this requires the hipSYCL runtime, which is not yet compatible with the plugin. See issue 1.
4. No integration with `syclcc` because of issue 1. This PR doesn't change the current toolchain at all, it merely adds the plugin as a demo for people to play around with.

## How it works - important implications
I've found that are two main strategies to get the clang CUDA frontend accept SYCL code:
1. Make `__host__` and `__device__` attribute changes before CUDA Sema has a chance to complain that the code is not valid CUDA code. Unfortunately, this approach doesn't work reliably as there seem to be cases where it is not possible to inject plugin code before Sema due to the tight coupling between parsing and semantic analysis in C++.
2. Initially parse SYCL kernels as host code (which works because SYCL doesn't use any non-C++ constructs), and only start modifying stuff once semantic analysis is done and cannot complain anymore.

We're using the second approach here. This has important consequences:
* Since the kernel is initially treated as host code, any builtins from libhipSYCL as well as custom target-specific CUDA/HIP optimizations must be `__host__ __device__` functions. This leads to the current caveat 1 from above. **This is important: It will change the way CUDA/HIP interoperability works with hipSYCL - only `__host__ __device__` code can be called directly from SYCL kernels**. However, I feel that this is okay since the SYCL spirit is to have your code dynamically decide at runtime where it wants to run, so `__host__ __device__` is the natural choice anyway. If you want to have target specific optimizations, the following is still possible:
```cpp
__host__ __device__ int super_optimized_func(){
#ifdef SYCL_DEVICE_ONLY // can also use __CUDA_ARCH__ on cuda
  return optimized_stuff;
#else
  return 0; // dummy value or implement host fallback here
#endif
}
```
* Since kernels are treated as host code initially, we cannot launch kernels with the regular CUDA-style `<<<>>>` syntax, because in that case Sema would complain about illegal calls from device code to host code. However, as it turns out, you can also launch a kernel by just calling it like a regular function (as long as it has the `__global__` attribute, clang will emit the kernel launch code). The only thing that's missing is the kernel call configuration (how many threads etc). Fortunately, both CUDA and HIP provide separate functions to specify that manually for the next kernel launch, as shown in the example below.

Once Sema has finished and the whole translation unit is parsed, we go through the AST:
* anything that's marked with our custom `__sycl_kernel` attribute [we're hijacking the diagnose_if attribute for that at the moment - a real custom attribute is cumbersome] receives a `__global__` attribute.
* All other functions get marked as `__host__ __device__`, except for variadic functions which are not allowed in device code, and functions that already had `__host__`, `__device__`, or `__global__`  attributes to begin with.
* Additional stuff for parallel hierarchical for

Because clang already starts emitting code before the whole translation unit is parsed, our modifications won't take effect because clang has already decided not to emit all the functions which were initially `__host__`. So, we rerun CodeGen over the all functions that we have modified.

After this stage, we have *everything* in device IR. Since this will in general lead to linking errors because libraries are of course unavailable in device code, we strip all functions that we have marked as `__host__ __device__` from the IR if they are not used
* by a kernel
* or by a function that had explicit `__device__` or `__host__ __device__` attributes to begin with.

That's it! We now have exactly the stuff in device code that we need without having to mark it as `__device__` in the input code :D And yes, we can now also use functions from e.g. the STL in kernels as long as they do not need external libraries.

## What it does - Example
The plugin allows to compile SYCL-like CUDA/HIP code like the following:
```cpp

#include <iostream>
#include <cstdlib>
#include <cuda_runtime_api.h>

// ***************** The following code would be part of libhipSYCL / the hipSYCL runtime ******
// Haven't yet found a way to set macro definitions from the plugin. In future, this
// definition could be either in backend.hpp or passed by syclcc
#define __sycl_kernel __attribute__((diagnose_if(false,"hipsycl_kernel","warning"))) 

// We need these calls for configuring CUDA kernel calls - on AMD there's a similar function
// called hipConfigureCall() that we can use instead.
extern "C" unsigned __cudaPushCallConfiguration(dim3 gridDim, dim3 blockDim,
                                                size_t sharedMem = 0,
                                                void *stream = 0);

extern "C" unsigned __cudaPopCallConfiguration();

// host device functions can still be accessed from kernels
__host__ __device__
void mysync()
{ 
#if defined(__CUDA_ARCH__)
  __syncthreads();
#endif
}

// host/device overloads work correctly
__host__ void whereami()
{
  printf("host!\n");
}

__device__ void whereami()
{
  printf("device!\n");
}

__host__ __device__
int get_thread_id()
{
#if defined(__CUDA_ARCH__)
  return threadIdx.x;
#endif
  return 0;
}

void __hipsycl_push_kernel_call(dim3 grid, dim3 block, int shared, void* stream)
{
  __cudaPushCallConfiguration(grid, block, shared, stream);
}

void __hipsycl_pop_kernel_call()
{
  __cudaPopCallConfiguration();
}


template<class F>
__sycl_kernel void run_kernel(F f)
{
  f();
}

template<class F>
__host__
void dispatch(dim3 grid, dim3 block, F f)
{
  __hipsycl_push_kernel_call(grid, block, 0, 0);
  // Note that we're not using CUDA-style kernel launch syntax with <<<>>>
  run_kernel(f);
  __hipsycl_pop_kernel_call();
}
//*************** user code starts here ****************

void forward_declared();

void forward_declared()
{
}

void g() {}

void myfunc()
{
  g();
}

template<class T>
void execute_kernel()
{
  T x;
  dispatch(1, 10, [x](){
    // We can even use STL functions like std::min in kernels!
    printf("%d\n", std::min(1,get_thread_id()));
    whereami();
    myfunc();
  });
}


int main()
{
  std::cout << "We can still do stuff in host code!" << std::endl;
  // template instantiations also work
  execute_kernel<int>();
  cudaDeviceSynchronize();
}


```

Note that with the plugin it is even possible to **use code from the STL in kernels**, as long as
* the code doesn't rely on symbols from libraries like `libstdc++`
* the code doesn't use features not allowed in device code like exceptions (since these also need libstdc++).
In both cases an error at link time will be the result.

## How to use
```
clang++ -x cuda -fplugin=/path/to/libhipsycl_clang.so ...
```
or

```
clang++ -x hip -fplugin=/path/to/libhipsycl_clang.so ...
```


illuhad:
(Travis CI build failure seems unrelated to this PR)

PR #61, opened 2019-05-07
Fix hardcoded interpreter path in shabang

nazavode:
This PR fixes `syclcc` in environments where the `python3` interpreter isn't installed system-wide. The change should be backward compatible since `env` would still pick `/usr/bin/python3` when it is the default one.


illuhad:
Awesome, thanks for the contribution! This PR should be very useful e.g. for HPC clusters :)

PR #62, opened 2019-05-07
clang plugin: Make hipSYCL runtime library compatible with clang plugin

illuhad:
This is an initial implementation of the required runtime changes to make hipSYCL
work with the clang plugin.

The changes are mostly necessary because the clang plugin expects that everything
called inside kernels must be `__host__ __device__`. To retain compatibility
with the current source-to-source transformation the following macros are introduced:
* `HIPSYCL_KERNEL_TARGET`: Everything that should be callable exclusively from kernels is marked 
  with this macro. Expands to `__host__ __device__` for the clang plugin and `__device__`
  otherwise
* `HIPSYCL_UNIVERSAL_TARGET`: Everything that should be callable everywhere is marked with this
  macro. Expands to `__host__ __device__` always.
* `__HIPSYCL_DEVICE_CALLABLE__`: Will be defined when compiling for a target that will have
  device intrinsics (e.g. `__syncthreads()`) available. This is not the same as `SYCL_DEVICE_ONLY`
  because hipCPU also implements these intrinsics without actually compiling for device. 
  Calls to such functions inside `HIPSYCL_KERNEL_TARGET` functions should be guarded with this macro.

The main change introduced by this commit is the consistent use of those macros (plus changes to add uniform kernel invocation for both the clang plugin path and the legacy source-to-source toolchain)

This commit assumes that `HIPSYCL_CLANG` will be defined when compiling with the clang plugin. This must be done by the user at the moment. Later on, `syclcc` could take care of that.

Since functions with kernel target get called exclusively in kernels, and kernels currently
run exclusively on `__device__`, the host case will never actually be called. While those warnings
certainly don't look nice, they can be safely ignored for now.

With these changes, I have been able to:
* Compile the unit tests with clang plugin for CUDA
* Run the unit tests with clang plugin for CUDA successfully, without errors

Known limitations:
* Some builtins like the native math functions are not yet adapted to the changes and cannot
  be used with the clang plugin at the moment.
* **EDIT: The following issue is fixed now using dummy return statements.** During host compilation warnings about missing return statements will be emitted because we
  use the following idiom as workaround to the requirement that kernels can only call `__host__ __device__`
  functions:
```cpp
HIPSYCL_KERNEL_TARGET T some_from_device_library(){
#ifdef __HIPSYCL_DEVICE_CALLABLE__
	return result_calculated_using_device_functions;
#endif
}
```


illuhad:
(please consider this a draft - wanted to create a draft PR but misclicked)

PR #65, opened 2019-05-13
Fix linkage for two functions in clang backend

psalz:
Small fix for two functions defined in clang backend header that cause linker errors when compiling more than one translation unit.


illuhad:
Whoopsie, thanks! :)

PR #66, opened 2019-05-13
clang plugin: Fix IR function pruning for cycles in call graph

illuhad:
The previous method of determining whether a function can be pruned from the IR relied on determining "backwards" if one of the callers of the function is an entrypoint.
This approach is prone to infinite loops when there are cycles in the callgraph.

This commit fixes this by creating an explicit callgraph, finding all children of the entrypoints in the callgraph, and removing functions if they are not among the children of entrypoints.

### DIFF
--- a/src/hipsycl_clang_plugin/IR.hpp
+++ b/src/hipsycl_clang_plugin/IR.hpp
@@ -42,7 +42,66 @@
 
 #include "CL/sycl/detail/debug.hpp"
 
+#include <unordered_set>
+#include <unordered_map>
+#include <vector>
+
 namespace hipsycl {
+
+class CallGraph
+{
+public:
+  void addNodeIfNotPresent(llvm::Function* F)
+  {
+    if(Callees.find(F) == Callees.end())
+    {
+      Callees[F] = std::vector<llvm::Function *>{};
+    }
+  }
+
+  void addCallee(llvm::Function* Node, llvm::Function* Callee)
+  {
+    Callees[Node].push_back(Callee);
+  }
+
+  const std::vector<llvm::Function*>& getCallees(llvm::Function* Node) const
+  {
+    return Callees.at(Node);
+  }
+
+  /// Finds all nodes in the subgraph reachable from F and stores
+  /// the result in \c Children.
+  /// If \c Children is non-empty, its contents will be interpreted
+  /// as functions of which we are already certain that they are children.
+  ///
+  /// This behaviour can be used to use several calls to this function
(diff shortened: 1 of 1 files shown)


psalz:
Yes! This fixes the infinite recursion mentioned in #42.

I'm however still getting errors for undefined functions being emitted into the PTX (error is `ptxas fatal   : Cannot take address of function '_ZNKSt13runtime_error4whatEv'`, i.e., for `std::runtime_error::what() const`). Curiously, while the frontend plugin mentions marking it as `__host__ __device__`, `FunctionPruningIRPass::runOnFunction` seems to never be called for this function.

illuhad:
Could you please provide a minimal test case to reproduce it? (and perhaps open a new issue, since this seems unrelated)

illuhad:
I've tried playing around with both exceptions and virtual functions, but so far could not reproduce the issue.

psalz:
Yes, I'll see if I can come up with a minimal test case!

psalz:
It wasn't possible to find a small test case, but we found a fix instead :-) I'll clean it up and file a PR hopefully tomorrow!

Sorry for hijacking this issue, just a couple of quick questions:
- Any particular reason your using a `llvm::FunctionPass` instead of a `llvm::ModulePass`? The docs seem to indicate that function passes are quite strict in what they allow, and it might not even be legal to construct the list of functions and entrypoints (although I'm not sure about that). 
- Why are you calling `markAsHostDevice` within `FrontendASTVisitor::processFunctionDecl`? You indicate in `applyAttributes` that it is not necessary, and it does seem to cause some trouble.

illuhad:
Okay, thanks for the news :)

As far as your questions are concerned:
1. While I think using a function pass should be fine because we're not removing any functions in `runOnFunction()` itself, a module pass may be cleaner now. Originally, a function pass made more sense because we would determine for each function individually whether we can prune it. That was basically the one-function-at-a-time-analysis use case that function passes are for. With the call graph, things have changed a bit now. Why would it not be legal to construct the list of functions and entrypoints?
2. What sort of trouble does it cause? Actually, I'm not sure if it really is not necessary in all cases. This call makes sure that the function becomes `__host__ __device__` later on. Originally, I thought that a function must be marked as `__device__` in order to have it emitted during device compilation. To my surprise, this however turned out not to be true in general, but it still may play a role for deciding which overload gets emitted.. I have stopped digging deeper into clang's codebase to find out when exactly what gets emitted because things just started to work ;)

PR #68, opened 2019-05-17
clang plugin: Fix two IR pruning edge cases, refactor call graph

psalz:
This took a bit longer than expected. As mentioned in #66, I've encountered some issues with the new Clang plugin while trying to compile a larger code base. I've tried to determine the root causes -- with limited success. Nevertheless, I have two fixes in place that allow me to compile said code base.

First, the pruning is now performed in an LLVM `ModulePass`, instead of a `FunctionPass`. This is because the latter's `runOnFunction` callback seemingly does not always receive all functions (possibly related to
delayed instantiation of templates?) which in turn leads to those functions not being pruned, causing `ptxas` errors. First I thought this might be related to the time the function pass was being executed (`EP_EarlyAsPossible`), but interestingly running it at e.g. `EP_OptimizerLast` doesn't seem to change anything -- so I have no idea why this is happening. In any case, I think doing the pruning in a `ModulePass` makes semantically more sense. Also, as I mentioned before, I'm not even sure if the use of a `FunctionPass` was entirely legal -- at least the restrictions mentioned [here](http://llvm.org/docs/WritingAnLLVMPass.html#the-functionpass-class) seem rather harsh ("FunctionPass subclasses are not allowed to: [...] Maintain state across invocations of runOnFunction (including global data).").

The second issue is unfortunately even more obscure and the root cause is hard to pinpoint. Symptoms again include various ptxas errors, for example `ptxas fatal   : Unresolved extern function '_ZSt17__throw_bad_allocv'`. As I mentioned in #66 I first thought this was somehow related to the (potentially superfluous?) calls to `markAsHostDevice` within `FrontendASTVisitor::processFunctionDecl`. However, further investigation indicated that those calls only caused additional "entrypoints" to be emitted to the IR, and I suspect those are the real culprits. In my case this was a ton (50+) of extremely long Boost template function instantiations, that apparently are also explicitly marked as `__device__` functions (Boost does have CUDA support, after all). I honestly didn't want to dig any further into Boost, as each of those template types was at least a screen full of gibberish.

I did however find a workaround that I think is cleaner than just not calling `markAsHostDevice`: The pruning pass now only considers kernels (i.e., functions marked as `__global__`) as "entrypoints", instead of also including any function explicitly marked as `__device__`.  This change should not be a problem, as to my knowledge, Clang currently does not support device object linking (at least that's what [these](http://lists.llvm.org/pipermail/llvm-dev/2017-August/116946.html) [threads](http://clang-developers.42468.n3.nabble.com/linking-relocatable-device-object-code-with-clang-td4061009.html) seem to indicate, and I couldn't get it working myself in a brief test). Was there any other reason for including explicit device functions as entrypoints?

Lastly, while switching to a `ModulePass` I took the liberty to also cleanup and refactor the pass and to replace the custom `CallGraph` with LLVM's implementation. The main difference between the two is that the latter does not include nodes for LLVM intrinsics. For this reason, I simply don't prune any intrinsics at all -- which seems to work fine.

**Edit:** I forgot to mention: With these changes I sometimes unfortunately still get a few warnings such as `ptxas warning : Unresolved extern variable '_ZTISt13runtime_error' in whole program compilation, ignoring extern qualifier`, where `_ZTISt13runtime_error`  is a pruned function. This only occurs during debug builds, and I'm not sure how I could get rid of them.

### DIFF
--- a/src/hipsycl_clang_plugin/HipsyclClangPlugin.cpp
+++ b/src/hipsycl_clang_plugin/HipsyclClangPlugin.cpp
@@ -45,9 +45,12 @@ static void registerFunctionPruningIRPass(const llvm::PassManagerBuilder &,
 }
 
 static llvm::RegisterStandardPasses
-  RegisterFunctionPruningIRPass(llvm::PassManagerBuilder::EP_EarlyAsPossible,
-                                registerFunctionPruningIRPass);
+  RegisterFunctionPruningIRPassOptLevel0(llvm::PassManagerBuilder::EP_EnabledOnOptLevel0,
+                                         registerFunctionPruningIRPass);
 
+static llvm::RegisterStandardPasses
+  RegisterFunctionPruningIRPassOptimizerLast(llvm::PassManagerBuilder::EP_OptimizerLast,
+                                             registerFunctionPruningIRPass);
 
 } // namespace hipsycl
 
--- a/src/hipsycl_clang_plugin/IR.hpp
+++ b/src/hipsycl_clang_plugin/IR.hpp
@@ -1,7 +1,7 @@
 /*
  * This file is part of hipSYCL, a SYCL implementation based on CUDA/HIP
  *
- * Copyright (c) 2019 Aksel Alpay
+ * Copyright (c) 2019 Aksel Alpay and contributors
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -29,6 +29,7 @@
 #define HIPSYCL_IR_HPP
 
 
+#include "llvm/Analysis/CallGraph.h"
 #include "llvm/Pass.h"
 #include "llvm/IR/Function.h"
 #include "llvm/Support/raw_ostream.h"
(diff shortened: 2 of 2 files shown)


illuhad:
Thanks for all the work! :)

> First, the pruning is now performed in an LLVM ModulePass, instead of a FunctionPass. This is because the latter's runOnFunction callback seemingly does not always receive all functions (possibly related to delayed instantiation of templates?) 

Good change. I also don't understand why `runOnFunction` would not be run on all functions. After all, my understanding is that it's there to apply some optimization to all functions. Potential LLVM bug?

> I did however find a workaround that I think is cleaner than just not calling markAsHostDevice: The pruning pass now only considers kernels (i.e., functions marked as __global__) as "entrypoints", instead of also including any function explicitly marked as __device__. This change should not be a problem, as to my knowledge, Clang currently does not support device object linking (at least that's what these threads seem to indicate, and I couldn't get it working myself in a brief test). Was there any other reason for including explicit device functions as entrypoints?

No, I just wanted to be ready to support `SYCL_EXTERNAL` as soon as clang implements device linking. Instead of just removing/commenting out the code that causes explicit device functions to be treated as entrypoints, could we perhaps surround it with some `HIPSYCL_EXPERIMENTAL_DEVICE_LINKAGE` macro or similar? In that way, we can easily reenable it once clang gains support for that.
I'm interested in trying to reproduce these issues. Could you please tell me which boost library this is about? This may point to an issue concerning the interoperability between hipSYCL and CUDA libraries. As this is one of the features distinguishing hipSYCL from other SYCL implementations, I would like to make sure that it works reliably ;)

> Lastly, while switching to a ModulePass I took the liberty to also cleanup and refactor the pass and to replace the custom CallGraph with LLVM's implementation. The main difference between the two is that the latter does not include nodes for LLVM intrinsics. For this reason, I simply don't prune any intrinsics at all -- which seems to work fine.

Sounds good :)

> Edit: I forgot to mention: With these changes I sometimes unfortunately still get a few warnings such as ptxas warning : Unresolved extern variable '_ZTISt13runtime_error' in whole program compilation, ignoring extern qualifier, where _ZTISt13runtime_error is a pruned function. This only occurs during debug builds, and I'm not sure how I could get rid of them.

hm, I'm familiar with this error in the context of missing global variable pruning. Demangling reveals that this is `typeinfo for std::runtime_error`. Are you sure it's a pruned *function*? My first suspicion would be that this could point to global variable pruning not working correctly with RTTI. Is this `_ZTISt13runtime_error ` still in use somewhere in the LLVM IR?

psalz:
> Good change. I also don't understand why runOnFunction would not be run on all functions. After all, my understanding is that it's there to apply some optimization to all functions. Potential LLVM bug?

I wouldn't rule it out, but since all of this stuff is barely documented, I guess we might as well just be missing something...

> could we perhaps surround it with some HIPSYCL_EXPERIMENTAL_DEVICE_LINKAGE macro or similar?

Done!

> I'm interested in trying to reproduce these issues. Could you please tell me which boost library this is about?

Okay, it took some time but I've boiled the issue down to a very minimal example! Here you go:

```c++
#include <boost/graph/adjacency_list.hpp>
// As of Boost 1.70, this includes a header which contains a __noinline__ attribute
// for __GNUC__ == 4 (which Clang (8) apparently also identifies as).
// This breaks CUDA compilation with Clang, as the CUDA (10) headers define __noinline__
// in an incompatible manner. As a workaround we thus simply undefine it altogether.
// Potentially related to https://svn.boost.org/trac10/ticket/9392
#if defined(__clang__) && defined(__CUDA__)
#undef __noinline__
#endif
#include <boost/graph/graphviz.hpp>

struct vertex_props {
  std::string label;
};

using my_dag = boost::adjacency_list<boost::vecS, boost::vecS,
                                     boost::bidirectionalS, vertex_props>;

int main() {
  my_dag dag;
  std::stringstream ss; 
  boost::write_graphviz(
      ss, dag, boost::make_label_writer(boost::get(&vertex_props::label, dag)));
  return 0;
}
```

The issue might be related to the `__noinline__` workaround - not sure. I've had that workaround in place for quite a while (I discovered it back when we originally added Clang support for hipSYCL).

> Demangling reveals that this is typeinfo for std::runtime_error. Are you sure it's a pruned function?

You're right, I shouldn't have trusted my mental demangler ;-) I've also seen this warning for `vtable for <...>`, so it appears to be some kind of stale metadata thing. Again, only during debug builds.

---

By the way, with the transition to the Clang plugin I'm now getting a lot of deprecation warnings for certain CUDA library calls. The problem: Those are being called from within HIP. Do you get these as well, or is it just me again? ;-)

illuhad:
Thanks!

I didn't yet have time to try to reproduct the issue, but will try to do so in the coming days.

I do see some deprecation warnings (related to CUDA context management), but I think that they had already been there when compiling with clang-cuda and good old `syclcc`. I would be very surprised if the clang plugin can trigger new warnings, since at the sema stage (where warnings are emitted) the plugin hasn't yet changed anything. So, apart from the fact that kernels are now parsed as host code, it should be exactly the same as before.

PR #69, opened 2019-05-17
Small improvements on cmake compiler and library discover:

hbsnmyj:
There are mainly three changes regarding the build file:

a) changed the libhipSYCL compiler to use CMAKE_CXX_COMPILER instead
of default cc/cxx by supplying a --force-alternative-compiler option.
In some systems, the default cc/cxx points to an ancient compiler, and
a regular user has no control over it. If the syclcc script happens to find
these things, the compiling of the software will choke.

b) change reference of -lLLVM to ${LLVM_LIBS}, as the latter is discovered
by find_package(LLVM), the former is not. As a result, if the llvm libraries
are not in the default path, the linker will not be able to find the llvm libraries.

c) by similar rationale, link to the target exported by find_pacakge(Clang)
instead of explicit specify -lclangAST etc.


hbsnmyj:
A comment on the CI status:

The build on clang-6/Rocm failed. After some digging, it seems the failure is due to the clangConfig.cmake/clangTarget.cmake files were moved manually by the Ubuntu maintainer, causing the config file to be undiscoverable and not functioning. The package will still fail to build if we manually add "/usr/share/llvm-6.0" to the CMAKE_PREFIX_PATH.

In fact, the current master branch relies on the linking directory presets of clang++-6.0 and will fail to build if CXX is set to anything other than clang++-6.0 in the docker image.

To fix this behavior, we can supply a FindClang.cmake or use some ad-hoc find_library() cmake calls. But I really don't think this is the concern of hipSYCL to fix a broken package in a particular distribution.

illuhad:
Hi!

thanks for the contribution and thank you for investigating the CI issue :)
* Thanks for your changes b) and c), those look great :)
* I'm a bit skeptical about a). `--force-alternative-compiler` was never intended to be used apart from specific cases for debugging. It replaces the actual hipSYCL toolchain with whatever compiler you specify. This might work for the CPU backend, but for the ROCm and CUDA backend I expect that the corresponding `libhipSYCL` libraries will actually be compiled against the CPU backend, not against ROCm/CUDA (because these codepaths get activated only if you compile `libhipSYCL` with a HIP/CUDA compiler). This *may* still work to some extent as most of the HIP runtime API calls happen in headers that are included by the user's SYCL application. However, there are some types shared between the user application and libhipSYCL (e.g. `hipStream_t`). These types can be different between backends. So, linking e.g. user SYCL application compiled for CUDA with `libhipSYCL_cuda` that was actually compiled for CPU will likely break things.

I do however agree about your intention of making it possible to specify the compiler used by `syclcc` :) There are several options for us to approach this:
* The `syclcc-clang` compiler wrapper (which is intended to replace current `syclcc` in the near future) already has a mechanism for that. It reads a configuration file (that is initially generated by cmake) and gets its compiler paths from there. We could backport this mechanism to `syclcc` to pass `CMAKE_CXX_COMPILER` to `syclcc` via this configuration file.
* The `syclcc` CPU backend already takes into account the `CXX` environment variable. We could also implement this environment variable for the other backends. (or potentially add a command-line argument?)

In general, it is however probably not worth investing a lot of time into current `syclcc`, as it will be replaced by `syclcc-clang` anyway.

As far as the CI is concerned: I agree that this is not primarily our problem. I would still like to have a working CI though, because it will be kind of useless if it always fails, even if it is not our fault ;)
The CUDA CI build already installs the clang-8 package from apt.llvm.org. We could do the same for ROCm and just use this package (which hopefully provides working cmake files). We will have to do this eventually anyway when we switch to `syclcc-clang`, because it needs at least clang-8 (better clang-9) on ROCm.

PR #73, opened 2019-06-24
Mutex protect accessor_tracker:find_accessor

psalz:
This fixes a race condition (which I encountered) within `accessor_tracker` by locking the mutex inside `find_accessor()`.

### DIFF
--- a/include/CL/sycl/detail/runtime.hpp
+++ b/include/CL/sycl/detail/runtime.hpp
@@ -88,6 +88,8 @@ public:
   buffer_ptr find_accessor(
       const accessor_base* accessor_ptr) const
   {
+    std::lock_guard<mutex_class> lock{_lock};
+
     const void* object_ptr = reinterpret_cast<const void*>(accessor_ptr);
     auto it = _buffer_map.find(object_ptr);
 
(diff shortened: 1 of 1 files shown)


illuhad:
Thanks! I was aware of this issue, but only for the CPU backend. Did you also encounter this for the GPU backend? I hadn't implemented this fix yet because this is of course massively inefficient on CPU, where we currently have one thread per work item and each of them will try locking this mutex, so I was thinking more about how to restructure the accessor tracker. But this will serve well as a preliminary solution :)

psalz:
Yes, this was on CUDA! Maybe a `std::shared_mutex` would suit your needs?

illuhad:
Okay, I don't really understand why this can happen on CUDA. Need to think about this some more. `shared_mutex` would indeed work, in fact @camaclean has implemented such a mechanism in his hipSYCL fork to fix this issue. My main problem with this approach is that this would require C++17, whereas SYCL 1.2.1 is still using C++14 as baseline. This should probably become irrelevant though with the next major revision of the spec.

psalz:
I guess it's happening because I'm submitting command groups from two separate threads!

psalz:
Also, regarding `shared_mutex` and C++17: I'm pretty sure there's a Boost alternative to that!

illuhad:
Ah, okay. If you're submitting from several threads such issues are probably expected. Thread safety is something that still requires some work in hipSYCL :)

PR #74, opened 2019-06-24
[CMake] Fix dependency tracking, migrate to Clang plugin

psalz:
This includes two distinct changes/improvements to the hipSYCL CMake module/config file:

First, I discovered that the CMake setup with the custom `SYCL` language had some issues with dependency tracking in that changing a header file would not cause a re-build of all dependent translation units. At first I suspected that we're not correctly defining the CMake variables for producing dependency files (i.e. `-MMD` and friends), and that turned out to be the case. However, even after copying those flags from the `CXX` language, CMake would still not correctly rebuild. Then I made a terrible discovery: As it turns out, when generating GNU Makefiles for certain languages, CMake determines the dependencies on its own **using a regex**. It's true - CMake doesn't understand something like this `#define MY_HEADER "file.h" \n #include MY_HEADER`. Not only that, but the list of languages for which it does this is hardcoded somewhere within a C++ module, i.e., out of reach for our custom `SYCL` language.

My workaround now is to recommend using Ninja over GNU Makefiles, as the former thankfully uses compiler-generated dependency files (which are now correctly being generated with this first commit). It's not ideal, but it works.

Secondly, I migrated the CMake module to use the new Clang plugin-based compiler. I dropped support for all the old options (e.g. `HIPSYCL_RESTRICT_DEVICE_HEADER_PATH` and so), and added support for the new ones. Since the CMake module wasn't being advertised anywhere so far, I guess this change won't affect too many people - and even then, it's easy to migrate to the new module.


illuhad:
Thanks! I'm starting to be quite happy that I'm not the one who has to fiddle around with the cmake integration ;) Thanks for all the time you put into debugging that! Any feedback from other hipSYCL cmake users? @VileLasagna ? If not I'm going to merge.

VileLasagna:
Interesting to hear about the cause of the dependency woes. In a lot of projects at work, I just can't use ninja at all, precisely due to it not generating dependencies correctly (don't quite remember the specifics of the situation... maybe it's addExternalProjects? but can look it up if you want).

This looks pretty exciting, I actually haven't played with the new stuff yet but will give it a go tonight and let you guys know how it goes

illuhad:
@VileLasagna, you're on ROCm, right? Using `syclcc-clang` on ROCm is still a bit complicated, because as far as I know AMD doesn't ship the required hip-clang compiler with the usual ROCm distributions. You'll probably have to manually install that compiler as described [here](https://github.com/illuhad/hipSYCL/issues/67#issuecomment-492571954).

psalz:
> I just can't use ninja at all, precisely due to it not generating dependencies correctly (don't quite remember the specifics of the situation... maybe it's addExternalProjects? but can look it up if you want).

Yes, please! I sure hope we can find a working solution for all setups.

> Did you follow this issue up on CMake's own bug tracker, @psalz ? I'm curious to see how this goes and if we can massage some sort of workaround

No, I didn't have time yet - I'll add it to my list ;-). I guess ideally we'd like to have some variable that allows enabling the C/CXX behavior for custom languages...

VileLasagna:
Hey, sorry it took me a bit to come back. Last few weeks have been... a bit intense... But I'm playing with this right now. Didn't get to automagically build, sadly, so now I'm trying to go through what's going on.
First issue is it seems to have not detected GPU_ARCH and apparently the CMake option didn't get forwarded properly. Got fixed by setting the ENV variable. It's now complaining about `no member named 'int_fast8_t' in the global namespace` and some other nonsense like this. Could be libstdc++ issues?
Need to have a closer look and understand if it's a problem on my side and/or some additional setup that I need to go through.

Full command line that spits out the errors is:

```
[ 25%] Building SYCL object SYCL/CMakeFiles/tbSYCL.dir/main.cpp.o
cd /home/lasagna/Work/Repos/lasagna/builds/toyBrot/Desktop_Qt_5_12_2_CLANG-Release/SYCL && \
/opt/hipsycl/bin/syclcc-clang \
-DSDL_DISABLE_IMMINTRIN_H -D__STRICT_ANSI__ \
-I/opt/hipsycl/include/hipSYCL -I/opt/hipsycl/include \
-I/opt/rocm/hsa/include -I/opt/rocm/hcc/include \
-I/usr/include/SDL2  --hipsycl-platform=rocm \
-O3 -DNDEBUG   -fcxx-exceptions -std=c++14 \
-o CMakeFiles/tbSYCL.dir/main.cpp.o -c /home/lasagna/Work/Repos/lasagna/toyBrot/SYCL/main.cpp
```

illuhad:
Not sure what's going on with `int_fast8_t`, but a remark concerning the GPU architecture: `syclcc-clang` behaves differently than the old `syclcc` by reading the configuration file `/etc/hipSYCL/syclcc.json`. That file is generated when building hipSYCL. If you set the `DEFAULT_GPU_ARCH` cmake variable when building hipSYCL, `syclcc-clang` should automatically use that architecture as default. You can also use this mechanism instead of the cmake integration to set your GPU architecture.

Also, I'm a bit surprised by those include paths - `syclcc-clang` should already take care of them (and it's probably better to let it figure those paths out, since they may vary based on the backend), so explicitly specifying them should not be necessary.

illuhad:
@VileLasagna, any news? Can we merge?

VileLasagna:
Hey there, man. Sry I've been away for a while. I have the time to bang my head against this today so will let you know if I manage to massage this all into working tonight and, if so, how so we can either tweak or provide instructions

VileLasagna:
All right... so now I've come back to this with more calm and got a better understanding. As you thought, yeah, I'm using the old hcc based backend. I'll shuffle my own plans over here and will get a better solution for this but if you guys have managed to get it working I don't think you should hold back because of me (the builds I need to update take about half an eternity on each go, even on my 1920X). I did manage to get it building with cuda using the new compiler, but something was not quite correct. My guess is that it's on my side, I had to point it to an older version of CUDA because I'm on clang 8.0 and my system cuda is 10.1 so I guess I'm a bit all over the place with my own environment.

Over the next few days I'll do, at the very least, a revamping of my rocm setup and if after that I can't get it working, I'll come here and scream again =P

I have a fork of the [Experimental_ROC](https://github.com/VileLasagna/Experimental_ROC/tree/archlinux) which I used to add Arch support for building from source. I also did a similar fork at work for OpenSUSE 15.0. I'm going to merge the OpenSUSE into my personal fork and, seeing as that repo specifically has been forgotten, I was already planning to go ahead and update everything to ROCm 2.5 (it's stuck at 2.0) and will make sure I have the new clang stuff you mentioned in the other issue there if it's not by default. After this is done I'll have another go at the new system.

Other than that... I guess the only question is if we want to have a CMake option for "Use Legacy syclcc" or something like that because, as much as this is a much better system, for now it's still a bit fiddly to set up

psalz:
> My guess is that it's on my side, I had to point it to an older version of CUDA because I'm on clang 8.0 and my system cuda is 10.1 so I guess I'm a bit all over the place with my own environment.

Yes, afaik Clang 8.0 doesn't support CUDA 10.1!

You mentioned further up that you had troubles with Ninja not generating dependencies for certain setups, did you have time to look into what that issue was?

VileLasagna:
Hey @psalz 
So, i went back to some of the projects I ran into this issue and found what happens with Ninja.

if you `ExternalProject_Add(someLib ...)` and later on `target_link_libraries(SomeProject ${CMAKE_BINARY_DIR}/internal/deploy/libSomeLib.so)`,  the standard MakeFiles generator will pick up that file as being generated by the intermediary setp. Ninja, however, will say something like:

```
ninja: error: 'Linux-x86_64/ThirdParty/RelWithDebInfo/lib/libmpi.so', needed by 'src/ProjectMain/CMakeFiles/ProjectMainTest.dir/cmake_device_link.o', missing and no known rule to make it
```
It MIGHT not be an issue relevant for hipSYCL specifically and it might be workable if you define additional targets etc, but it's just something I've bumped into whilst dealing with larger projects (and also part of why I don't really like having ExternalProject as part of the main build rather than an initial boostrap step).

IMO this is an issue on the CMake side, maybe listing those as build_byproducts will cause them to be listed for Ninja too? Not sure, never dove super deep into this

illuhad:
Merging now in order to not let this hang around indefinitely, please submit another PR if it turns out that changes are necessary.

PR #76, opened 2019-07-09
clang plugin: Fix NVPTX debug symbols by marking functions as host/device more selectively

psalz:
It looks like Clang 9 will introduce support for emitting debug symbols into NVPTX targets (in [4e9db1b](https://github.com/llvm/llvm-project/commit/4e9db1beff1cb17d7121e66c0257fa0e68fe20b6)). While that is great, it turns out that the previous approach of eagerly marking most functions as host/device and subsequently pruning them in IR appears to leave *some* residue behind that trips up `ptxas`, causing it to segfault. We previously talked about the warnings that are sometimes being generated about missing symbols - I suspect that to be related. Unfortunately this means currently it is not possible to do debug builds for any hipSYCL program using Clang 9.

I looked into how this could be remedied and together with @peterth came up with a solution:  Instead of marking everything as host/device and pruning it later, we only mark those functions that are actually being called by user kernels. This is done by first finding all user-provided kernel functions within the AST, and then only marking functions as host/device that appear within the call graph of those kernel functions.

---

The approach appears to be working well, and in fact I suspect that it could make IR pruning obsolete altogether. At least the tests as well as some of my own programs appear to compile just fine without the pruning pass. The only thing I'm not too sure about is the pruning of global variables, do you know of any specific cases of global variables that would cause issues if not pruned?

### DIFF
--- a/src/hipsycl_clang_plugin/Frontend.hpp
+++ b/src/hipsycl_clang_plugin/Frontend.hpp
@@ -48,6 +48,69 @@
 
 namespace hipsycl {
 
+namespace detail {
+
+///
+/// Utility type to generate the set of all function declarations
+/// implictly or explictly reachable from some initial declaration.
+///
+class CompleteCallSet : public clang::RecursiveASTVisitor<CompleteCallSet> {
+  public:
+    using FunctionSet = std::unordered_set<clang::FunctionDecl*>;
+
+    CompleteCallSet(clang::Decl* D)
+    {
+      TraverseDecl(D);
+    }
+
+    bool VisitFunctionDecl(clang::FunctionDecl* FD)
+    {
+      visitedDecls.insert(FD);
+      return true;
+    }
+
+    bool VisitCallExpr(clang::CallExpr* CE)
+    {
+      if(auto Callee = CE->getDirectCallee())
+        TraverseDecl(Callee);
+      return true;
+    }
+
+    bool VisitCXXConstructExpr(clang::CXXConstructExpr* CE)
+    {
+      if(auto Callee = CE->getConstructor())
+      {
+        TraverseDecl(Callee);
+        // Since for destructor calls no explicit AST nodes are created, we simply use this opportunity to
+        // find the corresponding destructor for all constructed types (since we assume that every
+        // type that can be constructed on the GPU also can and will be destructed).
+        if(auto Ptr = llvm::dyn_cast_or_null<clang::PointerType>(Callee->getThisType()->getCanonicalTypeUnqualified()))
+          if(auto Record = llvm::dyn_cast<clang::RecordType>(Ptr->getPointeeType()))
+            if(auto RecordDecl = llvm::dyn_cast<clang::CXXRecordDecl>(Record->getDecl()))
+              if(auto DtorDecl = RecordDecl->getDestructor())
+                TraverseDecl(DtorDecl);
+      }
+      return true;
+    }
+
+    bool TraverseDecl(clang::Decl* D)
+    {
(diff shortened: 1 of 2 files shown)


illuhad:
Thank you! I'm a bit skeptical about building a callgraph from the AST. Based on the old source-to-source transformation (which does exactly that), my experience was that it's cumbersome to build an accurate callgraph from the AST. In particular, just looking at `CallExpr` is not enough since an effective function call in C++ can take many forms, i.e. implicit calls to constructors and destructors (which can also take many different forms) and so on. To avoid these issues, I decided to only work with callgraphs at IR level where they are very easy to build.
Is there a way to remove these debug residues instead?

global variables as far as I know may cause warnings when they refer to some external symbols (e.g. `std::cout`). They do not cause errors in my experience.

psalz:
Fair point! In fact it turned out that the existing `clang::CallGraph` was already too simplistic in that it apparently didn't reach nested function calls at all (or maybe I've been using it wrong..? No idea.). We changed it to a custom `CompleteCallSet` type that builds a set of all reachable function declarations on its own. It does support implicit constructors, destructors and conversions already, and I suspect we should be able to add any additional C++ contraptions with relative ease.

I've added an integration test that showcases the different types of explicit and implicit function calls that this approach can currently deal with, however I haven't included it yet in the test CMakeLists, as it is intended for the new toolchain (maybe it's time to also migrate the unit tests to use `syclcc-clang`?).

> Is there a way to remove these debug residues instead?

Presumably, however I have no idea where to start. I suspect you'd have to try and follow the emission of debug symbols inside Clang/LLVM in reverse and figure out where the superfluous ones come from... However that sounds quite painful.

I do still think the selective marking is a promising approach, as removing the IR pruning pass would not only improve compilation speeds, but also make the whole toolchain much less brittle.

psalz:
I just realized that this would also fix another problem that I've been having with the Clang plugin. Consider this program:

```c++
#include <emmintrin.h>
#include <CL/sycl.hpp>

void unrelated_fn() {
    int data[] = { 42, 13, 21, 52 };
    auto mm = _mm_load_si128((const __m128i*)data);
}

int main() {
    cl::sycl::queue queue;
    queue.submit([](cl::sycl::handler& cgh) {
        cgh.single_task([](){ printf("test\n"); });
    });
}
```

So here we have a function somewhere that isn't related to a kernel at all, which is using SSE builtins. For me this came up while trying to load some image data using [stb_image](https://github.com/nothings/stb). While that particular library supports disabling SIMD using a macro, that's not always an option.

Currently, using Clang 8, this will produce the following error:

```
error: always_inline function '_mm_load_si128' requires target feature 'sse2', but would be inlined into function 'unrelated_fn' that is compiled without support for 'sse2'
```

while with this PR, it just compiles!

illuhad:
Okay, I can see that this is a grave issue. I'm always up for changes if they make things more robust :) Before merging, I would like the new approach to be able compile at least the same code as previously. As far as I can see, the following is not yet implemented: 
* Add (implicit) constructors of base classes to call graph
* Add function calls/constructors in initialization lists to call graph

These would be needed to compile e.g. SYCL parallel STL, and are common cases that we can expect will be hit frequently for object oriented SYCL code (which is kind of what SYCL targets).

illuhad:
Also, we should fix CI, which seems to still be on clang 6.0 which the new code doesn't seem to be compatible with (which is fine - we should probably use clang 8.0 as new supported baseline) ;)

illuhad:
Okay, this seems to work far better than I had expected. Very impressive! I've tested it with SYCL parallel STL and it seems to work fine. I'll try to get the CI fixed and then we can merge.

psalz:
Great :) I attribute this to @PeterTh's 10 years of compiler development experience 🎉

PR #77, opened 2019-07-12
Work around Clang name mangling bug by using custom kernel names

psalz:
(Note: This PR builds on top of #76 with the expectation that it will be merged before this).

Great news! I think this is a definitive fix for #49, without requiring any changes in Clang - which means it also works for Clang 8.

First, some bad news: After submitting the potential fix for #49 to the LLVM developers I found another instance of that bug (or a very related form of it), which would be much harder to fix. Essentially, the whole issue boils down to Clang using a simple counter to hand out unique mangled names for anonymous structs (i.e., lambdas). Since it's a counter, the name of any particular struct depends on the names of all of the other structs that have been mangled before it. As it turns out, this order is not always the same, leading to hard to find bugs where the name of a kernel function (again, a lambda) differs between host and device side binaries, leading to the `cudaLaunchError` in issue #49. The patch I submitted essentially fixed this issue whenever it occurs within a single Clang invocation. However as it turns out, unfortunately this bug can also happen between two entirely distinct Clang processes (Clang spawns copies of itself, e.g. one for targeting `x86`, and another for `nvptx64`). In this case it is much harder to ensure consistent naming, and a more deterministic way of handing out these anonymous struct IDs is required.

We looked around for an alternative solution, and found that there is an AST attribute, `clang::AsmLabelAttr`, that can be used to short-circuit the name mangler to simply use whatever the attribute specifies as the mangled name of a declaration.

This PR adds support for the Clang plugin to determine unique kernel names that will be used for all `__global__` entry points emitted into PTX. Conveniently SYCL already anticipated such an API -- the template specializations passed into e.g. `parallel_for<>` -- and that's exactly what this is built on. This of course means that unlike before, hipSYCL now also requires distinct names for each kernel.

I've marked this as a draft as it probably needs more testing, although it already successfully compiled everything I could throw at it. More importantly though, this should also be verified to work with AMD backends -- for which I don't have the hardware.


illuhad:
Holy shit, that's awesome! I don't have time to look at it in depth today, but I'd quickly like to mention that I expect SYCL kernel naming requirements to be relaxed in the future, and there's also some work going on in Intel LLVM SYCL on that matter, i.e. there's definitely a push in that direction. So, if possible at all, I would prefer to not rely on that name so much.
I'm just brainstorming, but would something like this perhaps be possible?
* At AST level, generate a random name and use this for the kernel using the attribute you mentioned.
* Attach llvm metadata to the kernel node with the name that it should have. That metadata will then end up in the device IR.
* At IR level, rename functions with that metadata to the name as specified in the metadata.

PeterTh:
Hi! (I worked with Philip on this)

From my experience in C++ frontends (with **a lot** of time spent with Clang and C++ mechanics), I believe you shouldn't rely on anything other than a user-defined name in this case for the purpose of external linking. 

E.g. consider an anonymous struct as introduced by a lambda expression in the common SYCL use case. This lambda can be arbitrarily nested in templates which might be instantiated in different contexts, resulting in multiple anonymous kernels generated from the exact same file/line/column. In our current code (this PR) that situation can be resolved by the user by providing a templated "name" for the kernel, including only the information required to uniquely identify it. 

The only other option would be to basically traverse the entire declaration context stack around the point where the lambda is introduced, and mangle all the template arguments into the generated identifier. This, however, could re-introduce the problem of discrepancies in naming between the PTX/GPU run of Clang and the host-side run as soon as another, anonymous struct is part of this template instantiation stack.

illuhad:
thanks for sharing your insights and for helping us with that! I understand that it is a problem for template instantiation if the template arguments are not retained in the mangled name. What about a combination, e.g. if we only replace the anonymous struct numbering with line/col? The Intel SYCL guys seem to follow a similar approach with the goal of making SYCL kernel naming optional, so there seems to be a way. We should definitely have a closer look at [their approach](https://github.com/intel/llvm/pull/250).

psalz:
Can we merge this for now and revisit kernel naming later on? I unfortunately don't have time to look into Intel's approach right now, and this PR both fixes a real and very painful bug while also being fully conformant with the current SYCL spec.

illuhad:
I see your point and I agree. However, can we add some preprocessor macro to conditionally enable this (can be on by default for now) ? After all, it's a bug in clang's CUDA implementation that will hopefully get fixed some day (e.g. when Intel's `__unique_stable_name()` is merged) so this workaround will then hopefully become unnecessary.

psalz:
Sure thing, I called it `HIPSYCL_USE_KERNEL_NAMES` for lack of a better name.

illuhad:
Thank you! I had actually hoped we could also reenable default-named kernels in `handler.hpp` on CPU, where the clang plugin is not needed, but I see that this would probably be a bit more involved, so we can also revisit that later.

I noticed that with this PR, compilation of the current version of the unit tests with the addition from PR #94 fails due to
```
include/CL/sycl/handler.hpp:63:20: error: definition with same mangled name '__hipsycl_kernel_$math_binary' as another definition
__sycl_kernel void single_task_kernel(Function f)
```
Do you also get this error?

Also, current master has docmentation on all available macros in `docs/macros.md` - please add your macro there :)

PeterTh:
@illuhad I have fixed the naming issue (for the binary math tests and the new tests added by that pull) in #97. So please merge that one first, and then this should work out.

illuhad:
I see, thank you!

psalz:
I've rebased this onto master and added documentation for `HIPSYCL_USE_KERNEL_NAMES`. I also fixed  a bug where the mangled names of modified types such as `unsigned int` would contain a space.

>  I had actually hoped we could also reenable default-named kernels in handler.hpp on CPU, where the clang plugin is not needed, but I see that this would probably be a bit more involved, so we can also revisit that later.

Yes, maybe this could be handled through a pragma, or a command line option for the plugin.

illuhad:
Thank you! Sorry to bring up another issue, but I think I may have found a problem. Consider the following code, which currently fails to compile with this PR but according to the kernel naming requirements in section 4.8.5 of the spec is legal:
```
#include <CL/sycl.hpp>

class Kernel
{
public:
  void operator()(cl::sycl::id<1> idx)
  {}
};

int main()
{
  cl::sycl::queue q;
  
  q.submit([&](cl::sycl::handler& cgh){
    cgh.parallel_for(cl::sycl::range<1>{128}, Kernel());
  });
  
  q.wait();
}
```
since the spec states that the type of the kernel function should be automatically used as kernel name if the kernel is a globally visible named function object. SYCL parallel STL actually makes use of that (which is how I noticed the issue), so I think we should try to fix that.

So, reintroduce a default argument for the kernel name and use the actual kernel type if the kernel is not a lambda or the kernel name is the default argument? This would then also allow us to support unnamed lambda kernels right away if the clang bug is fixed.

psalz:
Alright, I've reintroduced the default kernel name and implemented a check inside the Clang plugin to ensure that it can only be used for kernel functors that are not lambdas. That is slightly less restrictive than the spec (in that you could define a local functor type and use it without a name), but I guess that shouldn't be a problem if the goal is to eventually not require any names at all.

illuhad:
Excellent, thank you! That looks like a good solution to me. Currently unit tests don't compile on CPU because the host dispatch functions take the kernel functor as universal reference. When the kernel is a function object that has a non-const `operator()`, this errors - I will fix that in a subsequent PR. Apparently I didn't think about non-lambda kernels myself when I wrote that code :)

PR #79, opened 2019-07-13
Support only clang 8 and newer

illuhad:
This PR bumps the officially minimum supported clang version to 8. Included is an update to the Readme as well as an update to the singularity and dockerfiles to compile with clang 8, i.e. CI will now be done with clang 8 on all platforms.
This should also fix the CI build failure of PR #76.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,10 +21,20 @@ FORCE)
   endif()
 endif()
 
+find_package(LLVM 9 CONFIG)
+if(NOT LLVM_FOUND)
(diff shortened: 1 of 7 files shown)


psalz:
Maybe this could also include explicit versions in the CMake `find_package(LLVM)` calls? I currently have to edit these to 8 manually as I also have a Clang 6 installation which it defaults to otherwise.

illuhad:
Good idea! I've added some code to use the newest available LLVM out of 8 or 9.

PR #88, opened 2019-07-24
Add atomic float extension, fix atomic volatile compilation error

illuhad:
This adds the floating point atomic extension. If `HIPSYCL_EXT_FP_ATOMICS` is defined
before `CL/sycl.hpp` is included, hipSYCL will allow atomic operations on floating point
types.
This is an extension not covered by the spec, so SYCL code using that extension may become unportable.

It is advised to check with the `__HIPSYCL__` macro if compiling for hipSYCL, and enabling a
fallback implementation otherwise.
Additionally, note that not all of hipSYCL's backends support the same set of atomic float
operations. For example, CUDA supports `double` atomics while HIP does not. It is the user's
responsibility to ensure that the code remains portable.

This commit also modifies the `atomic` class such that it stores a raw pointer internally
instead of a `multi_ptr`. This fixes compilation errors related to `multi_ptr::get()` not being volatile,
while called from the volatile `atomic` member functions.

### DIFF
--- a/include/CL/sycl/atomic.hpp
+++ b/include/CL/sycl/atomic.hpp
@@ -42,6 +42,14 @@ enum class memory_order : int
   relaxed
 };
 
+#ifdef HIPSYCL_EXT_FP_ATOMICS
+  #define HIPSYCL_CONDITIONALLY_ENABLE_ATOMICS(template_param) \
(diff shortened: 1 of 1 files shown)


reguly:
That does enable atomics properly in my code, thank you for the help!

PR #90, opened 2019-07-29
Improve performance on CPU by changing execution and memory model

illuhad:
Currently, we rely entirely on hipCPU for parallelization and execution on CPU. Since `__syncthreads()` can occur at any time, hipCPU has to spawn as many threads as there are work items in a group to guarantee correct barrier semantics.
This is (usually) massively inefficient. Since there are many varieties of kernel invocations in SYCL where no barriers can occur, we have more flexibility in this regard than hipCPU.

This PR implements a rework of the CPU execution model to use OpenMP multithreading across work groups (instead of work items), potentially combined with (hopefully vectorized) inner loops over the work items and should eliminate this massive performance issue for many cases.

Additionally, this PR switches to a unified memory model on CPU. Previously, CPU was treated like hipSYCL's other targets (GPUs), i.e it was implicitly assumed that it has its own dedicated memory and requires data transfers between a host memory region and device memory regions. These unnecessary copies are now eliminated.

Notable changes in the code include:
* `nd_item`, `item` etc now explicitly store the iteration space information (ids and sizes) as member variables instead of querying them ondemand from the HIP builtins. This is necessary, because on the CPU we now have work items represented as different iterations of a loop, so we need to be able to change ids in each iteration. On GPU, the `HIPSYCL_ONDEMAND_ITERATION_SPACE_INFO` macro is enabled by default which restores the "old" behavior of using the HIP builtins instead of member variables. Using this macro is only allowed on GPU, since it is incompatible with the new CPU execution model. Benchmarking is needed to test if there are any performance penalties to the new approach on GPU at all (e.g. register pressure) -- if not we could just drop support for the old approach.
* Custom OpenMP kernel execution/dispatch functions in `handler.hpp`. To separate those from the HIP ones, they are now in two different namespaces `sycl::detail::dispatch::device` and `sycl::detail::dispatch::host`. This also made some changes necessary to the clang plugin/source transformation where the code tries to identify the dispatch function by their name, which now has changed.
* When reworking the `group` and `h_item` classes, I've also taken this opportunity to lay the foundation for the support of flexible work group ranges. The corresponding parallel for functions are not yet enabled, but most of the work should now already be done.
* For helper functions related to `id<>`, the `detail::id` namespace has been introduced. This has required all uses of `id<>` inside `detail` to be changed to `sycl::id<>`, in analogy to what is already done for `range<>`.

Caveats/Regressions:
* ndrange parallel for does still use the old execution model, due to the barriers that can occur here. The plan is to also transition this to the new model in a subsequent PR. This would require code in the clang plugin's IR pass that splits the inner loop over the work items whenever a `barrier()` is called.
* ~~dynamic local memory inside hierarchical parallel for does not work correctly. This is because local memory is managed by hipCPU which still assumes that parallelization is only over the work items, i.e. there can only be once work group active at any given time. We need to change the local memory buffer to one large buffer, segmented into the local memory portions of each individual work group. This will be addressed in a subsequent commit.~~ This regression should be fixed as of 17560f2

---
This PR changes code related to the heart of kernel execution, and impacts any backend. We should therefore test this with as much code as possible before merging.

Unit tests have been run and passed for (the code should not differentiate between ROCm and CUDA backends, only between CPU and GPU):

- [x] `syclcc-clang` + CPU
- [x] `syclcc-clang` + GPU + `HIPSYCL_ONDEMAND_ITERATION_SPACE_INFO`
- [x] `syclcc-clang` + GPU without `HIPSYCL_ONDEMAND_ITERATION_SPACE_INFO`
- [ ] `syclcc` + CPU
- [ ] `syclcc` + GPU + `HIPSYCL_ONDEMAND_ITERATION_SPACE_INFO`
- [ ] `syclcc` + GPU without `HIPSYCL_ONDEMAND_ITERATION_SPACE_INFO`

(note that the CPU backend is not expected to pass the explicit copy tests since hipCPU does not implement the multidimensional variants of `hipMemcpyAsync`)

### DIFF
--- a/bin/syclcc-clang
+++ b/bin/syclcc-clang
@@ -487,10 +487,12 @@ class pure_cpu_compiler:
 
   def run(self):
     command = [self._compiler] + self._compiler_args
+
+    command += self._args
+    # For GCC, linked libraries must follow the object files
     if self._contains_linking_stage:
       command += self._linker_args
     
-    command += self._args
     return run_or_print(command, self._is_dry_run)
 
 def print_usage(config):
--- a/doc/macros.md
+++ b/doc/macros.md
@@ -10,6 +10,7 @@ Most of these are managed by `detail/backend/backend.hpp`.
 * `__HIPSYCL_DEVICE__` - defined if generating code for GPU
 * `SYCL_DEVICE_ONLY` - defined if generating code for GPU
 * `HIPSYCL_CLANG` - defined by `syclcc-clang` when compiling with the clang plugin
+* `HIPSYCL_SVM_SUPPORTED` - defined when the backend supports unified memory
 
 ## Mainly for hipSYCL developers
 * `__HIPSYCL_TRANSFORM__` defined by legacy `syclcc` during the source-to-source transformation step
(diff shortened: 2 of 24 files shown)


illuhad:
It seems that legacy `syclcc` currently has issues unrelated to this PR in that it wrongly marks the command group function as `__device__`. I have therefore decided to skip testing with legacy `syclcc` for now.

PR #94, opened 2019-08-01
Fix genfloat binary math operations and add unit testing for them

PeterTh:
Fixes an oversight in binary operators that renders them unusable, and adds unit tests to hopefully prevent regressions in the future.

A lot more math unit tests would be needed for full coverage -- this fully covers the implemented binary operations on all types classified as `genfloat`.


PeterTh:
Correction: actually, this doesn't cover all `genfloat`s in the test case, just up to a width of 4. Going all the way to 16 would be no problem in principle, but require quite annoying code in practice (in the absence of C++17).
Anyway, I think testing up to 4 is better than not testing.

illuhad:
Thank you!

Issue #93, opened 2019-08-01
syclcc-clang fails to build libhipsycl_rocm: missing basic cstdint types (uint8_t, etc..)


VileLasagna:
So, I've updating things on my end and testing out the new syclcc-clang stuff now that the change by psalz on the CMake side. Took me a while to sort out the whole rocm side, updating everything to 2.6 (still not quite done there), understanding and migrating to the hip-clang plugin... But after some back and forth I've done it and quite like not having to rely on hcc.

Still not 100% and this issue is about one problem which I actually found a fix to. As I was getting this up, the issue mentioned in #74 where clang would fail the build complaining about not knowing basic int types was back. After banging my head in frustration for a while, I've managed to sort it out and am going to open a PR for it, would just like to confirm direction for the fix.

The root cause of the issue is that clang was trying to load incorrect internal headers. In my case, as I've built the packages from source from my fork of Experimental_ROC, rocm's raw clang gets put into `<rocm_path>/llvm/` with the executables in `bin` and the headers located in `lib/clang/9.0.0/include` 

This is relevant because in line 434, syclcc-clang looks for these files in `<rocm_path>/hcc/lib/clang`, which are still there because I have hcc as well (it's still required to build HC and is it's final release). Turns out those headers don't quite match up and the fix for the build is correcting that path.

My doubt here, though, is that I'm somewhat surprised I seem to be the first one bumping into this issue given how it happens. Not returning that path explicitly seems to cause troubles as well (I've tried to return empty or `/usr/include` and build did not succeed).

As such I'm trying to both understand a bit more how come no one else seems to have this issue as well as what would be the ideal fix to put in the PR.

My suggestion would be walking back one dir up from `hipsycl-clang`'s location and, from there, look into `/lib/clang` instead of assuming the full path. Again, only unsure because I seem to be the odd one out here =P

ps: I've mentioned that this is not the only issue I'm having, and while this is true, the others are not related to this one and aren't things that I've explored. Just for curiosity's sake, here they are:
 - My performance on CPU is very very bad. My own version of toybrot with OpenMP takes about 1-1.5 secs, and this is taking 15 and not behaving. But I think this is on the included project so not quite hipsycl's scope
 - I've also been trying to build CUDA from the same clang. On it's own it seems to work fine but my project through hipsycl just crashes out. I haven't begun investigating this yet
[...]

PR #95, opened 2019-08-01
Correct syclcc clang includes

VileLasagna:
As mentioned in #93 

Issue:
syclcc-clang adds an incorrect include path for clang's internal includes (it's leftover looking from when we used hcc)

Fix:
Instead try to look for the directory relative to whatever clang gets passed in (either through the json or through --hipsycl-clang=...)

Additionally:

 I've added a couple of missing find_package statements for hipsycl_rewrite_includes. Without them, linkage fails. I assume the reason for this is because llvm's clang does not normally adds itself to system paths, as it'd interfere with system packages and, without these explicit looks, linker will go through the regular system paths instead (in my case, I build with rocm's clang 9.0 but system has clang 8 from Arch Linux' regular packages)

### DIFF
--- a/bin/syclcc-clang
+++ b/bin/syclcc-clang
@@ -111,7 +111,7 @@ class syclcc_config:
     config_file_path = os.path.abspath(
       os.path.join(self.hipsycl_installation_path,
                   "etc/hipSYCL/syclcc.json"))
-    
+
     self._config_file = None
     # First check silently if we can open the default config file.
     # If that fails, we try later on after argument parsing if a
@@ -119,7 +119,7 @@ class syclcc_config:
     # (This happens when compiling hipSYCL)
     if os.path.exists(config_file_path):
       self._config_file = config_file(config_file_path)
-      
+
     self._args = args
 
     # Describes different representations of options:
@@ -128,14 +128,14 @@ class syclcc_config:
     # 3.) the field in the config file.
     self._options = {
       'platform': option("--hipsycl-platform", "HIPSYCL_PLATFORM", "default-platform",
-"""  The platform that hipSYCL should target. Valid values: 
+"""  The platform that hipSYCL should target. Valid values:
     * cuda: Target NVIDIA CUDA GPUs
     * rocm: Target AMD GPUs running on the ROCm platform
     * cpu: Target only CPUs"""),
 
       'clang': option("--hipsycl-clang", "HIPSYCL_CLANG", "default-clang",
 """  The path to the clang executable that should be used for compilation
-    (Note: *must* be compatible with the clang version that the 
+    (Note: *must* be compatible with the clang version that the
      hipSYCL clang plugin was compiled against!)"""),
 
       'cuda-path': option("--hipsycl-cuda-path", "HIPSYCL_CUDA_PATH", "default-cuda-path",
@@ -149,32 +149,32 @@ class syclcc_config:
 
       'cpu-compiler': option("--hipsycl-cpu-cxx", "HIPSYCL_CPU_CXX", "default-cpu-cxx",
 """  The compiler that should be used when targeting only CPUs."""),
-      
+
       'config-file' : option("--hipsycl-config-file", "HIPSYCL_CONFIG_FILE", "default-config-file",
 """  Select an alternative path for the config file containing the default hipSYCL settings.
     It is normally not necessary for the user to change this setting.""")
     }
     self._flags = {
       'is-dryrun': option("--hipsycl-dryrun", "HIPSYCL_DRYRUN", "default-is-dryrun",
-"""  If set, only shows compilation commands that would be executed, 
+"""  If set, only shows compilation commands that would be executed,
   but does not actually execute it."""),
       'is-bootstrap': option("--hipsycl-bootstrap", "HIPSYCL_BOOTSTRAP", "default-use-bootstrap-mode",
-"""  Enter bootstrap mode. This is only required when building hipSYCL itself and 
+"""  Enter bootstrap mode. This is only required when building hipSYCL itself and
   should not be set by the user.""")
     }
 
     self._hipsycl_args = []
     self._forwarded_args = []
-    
+
     for arg in self._args:
       if self._is_hipsycl_arg(arg):
         self._hipsycl_args.append(arg)
       else:
         self._forwarded_args.append(arg)
-        
+
     if self._is_option_set_to_non_default_value("config-file"):
       self._config_file = config_file(self._retrieve_option("config-file"))
-      
+
     if self._config_file == None:
       # If the config file is still None at this point, probably no alternative
       # config file was supplied and the default one doesn't exist.
@@ -230,7 +230,7 @@ class syclcc_config:
         v == "false"):
       return False
     return True
-      
+
 
   def _is_flag_set(self, flag_name):
     flag = self._flags[flag_name]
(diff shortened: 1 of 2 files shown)


illuhad:
Thank you! I found that when several versions of clang are available, this does not necessarily pick the same version of clang and llvm (where it uses the newest), which can cause trouble. Is there a way to make it pick the same clang and llvm versions? It also seems that there is an issue with the clang 9 package from `apt.llvm.org`. cmake looks for `/usr/lib/llvm-9/lib/libclang_shared.so.9` which is not provided by this clang distribution. Since we need very new clang versions to support the newest CUDA releases, we should expect our users to use the nightly `apt.llvm.org` packages. 

If the clang cmake module causes trouble, I would also be fine with just disabling compilation of the source transformation tools by default to advance the deprecation process. No need to spend time on compiling something that users are discouraged to use :)

illuhad:
Any news?

VileLasagna:
Hey, man. Sorry I've been away, was planning on going over this tomorrow and hadn't had a proper look at your question yet. 
Hmm... the way to do this, I guess, would then to get the full path of the clang executable used on the build, up one directory and add that to CMAKE_PREFIX_PATH. Such that, if you're building from the system, it would add `/usr` to it (essentially a non-op) but if, say, your clang was `/opt/rocm/llvm/bin/clang`, it would add `/opt/rocm/llvm/` to it.

I was under the impression it was picking the correct one when I tested but can double check, for sure. I'm currently rebuilding all of my rocm stuff as I'm almost finished with polishing up my upgrade to rocm 2.6 (famous last words right here). Once that is done and pushed I can get back to this issue and make sure it's working as expected

illuhad:
What's the current state? We don't need the cmake stuff right away (compilation of the source transformation tools that were affected are now disabled anyway), but I'd like to merge the fix regarding the includes very soon as it's a blocker for my plans of extended testing on ROCm. 

By the way, I will likely move to using [aomp](https://github.com/ROCm-Developer-Tools/aomp/releases) as new rocm compiler, since I suspect this is the one clang fork that is actually getting AMD's love. There are binary packages, so it's also very easy to install.

With aomp, things work fine if you just remove the problematic include in `syclcc-clang` entirely - so that would also be a good option moving forward.

VileLasagna:
It shouldn't be hard to implement either a tweak to CMAKE_PREFIX_PATH or try and ask for a specific version.  Alternatively I can just remove the "offending" commit. I was waiting on you to decide which way you'd prefer this sorted

illuhad:
I'm sorry, I wasn't aware you've been waiting for my feedback. I would suggest waiting with the fix for clang cmake integration until the aomp integration is done. The reason is that cmake seems to be unable to correctly deduce the clang version used by aomp (since aomp adds a suffix to the version string) which will already require a rework of how we are currently looking for llvm. When we have the aomp integration I believe it is easier to discuss possible solutions and check how well they work on our supported platforms once we have the basics in place.

Regarding the toolchain: My main issue with it is that you're not really pointed to which distribution you should pick. aomp is not really a new compiler, it's effectively just a branch of AMD's clang/llvm fork ("hip-clang") bundled with matching ROCm device and runtime libraries. So, it is effectively just a distribution of hip-clang, with the benefit that it comes with binary packages (good for our CI) and is easier to install since it already comes with everything it needs. Additionally, its OpenMP support opens the door for us to add a targeted OpenMP backend in the future.
HC C++ would have gone anyway (with all variants of hip-clang) - in my mind it has always been clear that it's not going to be kept around in the future. If I recall correctly, the hcc readme advises people to move their HC code to HIP or OpenCL.

VileLasagna:
Yeah, HC C++ and HCC have been marked for deprecation for quite a while. It's just that they haven't phased it out entirely yet. To my understanding the version in ROCm 2.6 should be the final release for HCC, from their deprecation notice a while back. So waiting on the dust settling on what's coming to replace it for real is probably the way to go for the future.

illuhad:
Yeah.. I'm moving in the direction now of using mainline clang for ROCm. Was the option that worked best for me. See PR #113.
Also, I've implemented an alternative solution in PR #113 for the include path. I'm using a cmake variable and let cmake do the guessing of the path instead of `syclcc-clang` trying to be clever. I think this may better since include paths can be quite different for diffierent clang distributions. What do you think?
Sorry for not building on your work for now, time has been pressing ;)

VileLasagna:
Eh, it's all good. I'll give that PR a go once I have the time, see how I fare with that. No hard feelings if it turns out there's a better solution =P

illuhad:
Excellent, looking for forward to that :)

VileLasagna:
So.. I've just built it. Had to massage it a little bit due partially to the way i've set up my environment, but might have run into a new detail that needs to be tweaked in CMake.

So, I've got both the classic HIP through HCC as well as the new HIP-CLANG installed (and the default).

Now, the place I ran into a rough edge was that it looks like rather than just having the one new libhipSYCL, we have both a libhipSYCL_clang and a libhipSYCL_rocm, with the latter building on the older hcc toolchain. This one wasn't building because it didn't automagically pick up libhip_hcc (this is more or less expected on my end)

If that's correct (we want both libs):
 - I'd need to think how to generalise the type of fix I've done here, maybe do search and enable thing if the person for whatever reason ONLY has hip-clang?
 - The CMake package files reference only three variants of libhipsycl, which means that we'd need to add `clang` to the possible list of hipsycl platforms and add the target to the config files (this is easy)

If that's incorrect (should be either or):
 - Then the `libhipsycl_rocm` target needs to be switched on/off depending on probed/passed HIP_PLATFORM.
 - BUT then we'd need to do similar changes on the CMake config files to expose hipSYCL::hipSYCL_rocm instead (also: should we consider aliasing these to `hipSYCL::rocm` 
 and the like?)

I can apply changes for either of these paths and open a new PR/issue

illuhad:
How did you manage to get `libhipSYCL_rocm` compile with hcc? For quite some time now, cmake exclusively uses the new `syclcc-clang` to build hipSYCL, and `syclcc-clang` doesn't know anything about hcc. `syclcc-clang` should also automatically link against `libhip_hcc` (which has a confusing name - it's also needed for HIP-clang, without hcc involved at all).

`libhipSYCL_rocm` and `libhipSYCL_clang` are both required as they do two entirely different things (which is also why cmake doesn't treat `libhipSYCL_clang` as a libhipSYCL variant):
* `libhipSYCL_rocm` is the hipSYCL runtime library containing SYCL scheduling, memory management and so on. `syclcc-clang` will link SYCL applications against this library on ROCm. This library is needed at runtime.
* `libhipSYCL_clang` is the clang plugin containing clang AST and IR passes (so, compiler stuff). It's loaded by `syclcc-clang` into clang's plugin system during compilation, but is not needed at runtime.

What exactly is your setup and what was your fix?

VileLasagna:
OIC...

So this is adding to my confusion from the AMD side... If you build hip-clang `libhip_hcc` doesn't get built at all. The spat out CMake config files also strip it from the targets. I was under the impression that when I was doing the hello-world tests for hip-clang, libhip_hcc wasn't being linked at all and when I was trying to convert a larger project, it got linked in and I was getting some weird segfaults. 

So if this is all as expected then I need to look at the ROCm side and figure out what's happening there... ¬¬

I was tackling this yesterday and got a bunch of missing symbols so just realised I was wrong somewhere and decided to have another look today. So.. yeah... disregard, this is all me, although I still need to understand what's happening to the hip libs in that case

PR #97, opened 2019-08-02
Add "Common Functions" as per spec, including unit tests

PeterTh:
Also refactors testing and extends test coverage to vec8 and vec16.

Note that the functions are mostly implemented in a straightforward fashion. There might be more efficient mappings of some of them to some backends (e.g. CUDA), but I think for now it's most important to be standard compliant.

### DIFF
--- a/include/CL/sycl/common_functions.hpp
+++ b/include/CL/sycl/common_functions.hpp
@@ -36,12 +36,26 @@
 namespace cl {
 namespace sycl {
 
-template<typename T>
+namespace traits {
+  template<typename T>
+  struct is_genfloat {
+	  static constexpr bool value = std::is_floating_point<T>::value;
+  };
+  template<typename E, int N>
+  struct is_genfloat<vec<E, N>> {
+	  static constexpr bool value = std::is_floating_point<E>::value;
+  };
+  template<typename T>
+  constexpr bool is_genfloat_v = is_genfloat<T>::value;
+}
+
+template<typename T, std::enable_if_t<traits::is_genfloat_v<T>, int> = 0>
 HIPSYCL_KERNEL_TARGET
 inline auto clamp(T x, T minval, T maxval) {
   return fmin(fmax(x, minval), maxval);
 }
-template<typename T, typename D, std::enable_if_t<!std::is_same<D,T>::value, int> = 0>
+template<typename T, typename D,
+  std::enable_if_t<!std::is_same<D,T>::value && traits::is_genfloat_v<T>, int> = 0>
 HIPSYCL_KERNEL_TARGET
 inline auto clamp(T x, D minval, D maxval) {
   return fmin(fmax(x, T{minval}), T{maxval});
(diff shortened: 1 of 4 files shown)


include/CL/sycl/common_functions.hpp
```diff
@@ -0,0 +1,140 @@
...
+
+#include "vec.hpp"
+#include "math.hpp"
+
+namespace cl {
+namespace sycl {
+
+namespace detail {
+  template<class T>
+  constexpr T pi = T(3.1415926535897932385L);
```
illuhad: In `math.hpp` we use `M_PI` - we should probably be consistent here and either use `detail::pi` everywhere or use `M_PI` everywhere. I'm fine with both :)

include/CL/sycl/common_functions.hpp
```diff
@@ -0,0 +1,140 @@
...
+}
+
+template<typename T>
+inline auto clamp(T x, T minval, T maxval) {
```
illuhad: please mark all functions intended to be used inside kernels with `HIPSYCL_KERNEL_TARGET` (expands to `__host__ __device__` currently). This is required for the support of "manual mode" and "legacy source-to-source transformation", although the latter is deprecated. See [here](https://github.com/illuhad/hipSYCL/issues/42#issuecomment-483312569) for more details.

PeterTh:
Done. 
I decided to just use M_PI (even though it's not-100%-standard) since in practice it works everywhere.

illuhad:
Thanks! I'm still getting
```
unit_tests.cpp:1291:24: error: call of overloaded ‘min(cl::sycl::vec<float, 8>&, cl::sycl::vec<float, 8>&)’ is ambiguous
```
(and lots of the same, and same for `max`). This is because your new `min()` and `max()` using `fmin` and `fmax` collide with the ones from `builtin.hpp` for integers. So, we should probably only enable the `fmin` and `fmax` variants for floating point arguments.

PeterTh:
To repro and fix this I added a unit test for integer min/max. 

However, even if I limit my implementations to `genfloat` I'm facing the problem that the min/max in `builtin.hpp` also applies to `float` -- should this be the case?

On the other hand, if I remove the ambiguity by removing my versions of `min(T,T)` and `max(T,T)` entirely, I run into the issue that the version provided in `builtin.hpp`, while entering the overload set and causing ambiguity, doesn't actually correctly instantiate for all the cases I test.

I don't fully understand the intention behind the definitions/macros in `builtin.hpp`, so I'm not sure how to proceed. Should I rewrite builtin.hpp to only provide the `geninteger` versions?

illuhad:
You're right, there's a bug! `builtin.hpp` was indeed intended to be for integer builtins (contrary to `math.hpp` for floats), but the `HIPSYCL_DEFINE_BUILTIN` macro doesn't correctly limit itself to integer types. Additionally, that header seems to only have partially made the transition to the new clang plugin where the `HIPSYCL_KERNEL_TARGET` macro was used. `builtin.hpp` still uses straight `__device__` attributes.

So, I see two ways to proceed:
* You fix the `builtin.hpp` header to only accept integer types
* Or you comment out the `min/max` unit tests and I fix the `builtin.hpp` header, and also bring it up to speed with the current clang plugin setup. I would then reenable the unit tests once `builtin.hpp` is fixed.

PeterTh:
I've resolved the ambiguities and added some basic testing for `geninteger` min/max.  Note that the existing implementation broke when I added `char` (not unsigned), and vectors thereof, to the list of types tested -- for now I just skipped that.

illuhad:
Excellent, thank you!

PR #98, opened 2019-08-07
Silence deprecation warnings in hip which occur for newer CUDA versions

PeterTh:
This is just a small change to silence deprecation warnings in the hip runtime (which is a big quality of life improvement since it allows you to build hipSYCL programs targeting newer CUDA platforms without getting a lot of warnings in platform code).

This should of course be removed if/when the hip runtime is updated.


illuhad:
I had almost forgot how nice life can be without all those CUDA warnings :) Thank you!
(CI failure is because since yesterday, there seem to be some network connectivity issues on the side of Travis CI - can be ignored until they fix it)

PR #99, opened 2019-08-07
Add geometric functions, unit test improvements

PeterTh:
This adds the SYCL geometric functions as per spec, tests them, and also includes various unit test enhancements.


illuhad:
Thank you, looks great!

PR #119, opened 2019-08-23
Remove unneeded dependencies after task graph node submission

illuhad:
After we have submitted a task graph node, we don't need to track its dependencies anymore.
Purging the dependencies allows task graph nodes to be destroyed early, which can prevent
destructor cascades that potentially can even cause stack overflows when many operations
have been submitted by the user.

### DIFF
--- a/src/libhipSYCL/task_graph.cpp
+++ b/src/libhipSYCL/task_graph.cpp
@@ -134,6 +134,13 @@ task_graph_node::submit()
 
     _stream->activate_device();
     task_state state = _tf();
+
+    // Remove unnecessary dependencies to allow task graph nodes
+    // to be destroyed.
+    // We know that we don't need the dependencies anymore, because
+    // we know that assert(is_ready()) passes - so all dependencies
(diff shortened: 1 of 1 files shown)


src/libhipSYCL/task_graph.cpp
```diff
@@ -134,6 +134,13 @@ task_graph_node::submit()
 
     _stream->activate_device();
     task_state state = _tf();
+
+    // Remove unnecessary dependencies to allow task graph nodes
+    // to be destroyed.
+    // We know that we don't need the dependencies anymore, because
+    // we know that assert(is_ready()) passes - so all dependencies
+    // are satisified anyway, and do not need to be considered anymore.
+    _requirements = std::vector<task_graph_node_ptr>{};
```
keryell: You cannot use https://en.cppreference.com/w/cpp/container/vector/clear instead?

illuhad: Sure, you're right, this would have been nicer!

PR #146, opened 2019-09-27
Clang plugin: Improve shared memory detection for parallel for hierarchical invoke

psalz:
This improves the detection of shared memory variable declarations for the `parallel_for_work_group` hierarchical invoke within the Clang plugin by properly handling compound statements as well as marking memory as shared inside certain functions called from within the work-group functor.

What prompted these changes was me trying to wrap a work-group functor inside another function. This broke shared memory declarations, so I started looking into it. Initially I thought it might be sufficient to simply walk the complete call set (re-using the data structure we added for device function marking) and mark all variable declarations in those functions as shared as well. However, that is not actually desirable, as this might include functions that should very much not share their state across work items (for example state inside `parallel_for_work_item`). I then realized that having shared memory only really makes sense when you can subsequently access it from individual work items. So, as a simple heuristic, I'm only considering functions that receive a `cl::sycl::group` as one of their parameters (which would allow them to call `parallel_for_work_item`).

As mentioned above, the `storeLocalVariablesInLocalMemory` function now also handles compound statements, which I think is a rather uncontroversial change. However, there remain a few open questions regarding certain other statement types: For example, how should a for-loop be handled? Should the loop variable(s) also be marked as shared? Most likely not. But what about the body of the loop then? Since I don't have a proper usage scenario available I did not consider these cases for now.

This also includes a change that restructures the device compilation tests into a more unit-testy format. While the main objective here remains to ensure that this file compiles in the first place, there is now also some tests to check for the correct behavior of the resulting executable.

### DIFF
--- a/tests/device_compilation_tests.cpp
+++ b/tests/device_compilation_tests.cpp
@@ -25,26 +25,43 @@
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include "unit_tests.hpp"
+
 #include <initializer_list>
 
-#include <CL/sycl.hpp>
+// NOTE: Since these tests are concerned with device compilation, the first
+// and arguably most important "test" is that this file compiles at all. We
+// then also include a few unit tests to assert the correct behavior of the
+// resulting programs.
 
-// What we're testing here is different forms of function calls, including
-// implicit and templated functions and constructors as well as destructors.
-// The hipSYCL Clang plugin must recognize that each of these is being called
-// by the user-provided kernel functor, and mark them as __device__ functions
-// accordingly.
-//
-// Additionally, we test that complex kernel names are correctly supported.
+BOOST_FIXTURE_TEST_SUITE(device_compilation_test_suite, reset_device_fixture)
 
-void foo4(int v) {
-  printf("foo4: %d\n", v);
+int add_five(int v) {
+  return v + 5;
 }
 
-struct Conversion {
-  Conversion(float value) : value(value) {
-    foo4(*this);
+void call_lambda(int& v) {
+  ([&](){ v = add_five(v); })();
+}
+
+struct MyParent {
+  MyParent(std::initializer_list<int> init) {
+    for(auto&& e : init) {
+      parent_value += e;
+    }
+    call_lambda(parent_value);
   }
+
+  virtual ~MyParent() {
+    *increment_in_dtor += parent_value + 7;
+  };
+
+  int parent_value = 0;
+  int* increment_in_dtor;
+};
+
+struct Conversion {
+  Conversion(float value) : value(value) {}
 
   operator int() {
     return static_cast<int>(value);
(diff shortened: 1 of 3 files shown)


illuhad:
Thank you! I will have a more in-depth look at it on monday. To answer your questions, here are my thoughts about how hierarchical parallel for should be ideally implemented, i.e. what we should work towards.
In general, hierarchical parallel for as in the spec is complex and not well specified.

> What prompted these changes was me trying to wrap a work-group functor inside another function. This broke shared memory declarations, so I started looking into it. Initially I thought it might be sufficient to simply walk the complete call set (re-using the data structure we added for device function marking) and mark all variable declarations in those functions as shared as well. However, that is not actually desirable, as this might include functions that should very much not share their state across work items (for example state inside parallel_for_work_item). I then realized that having shared memory only really makes sense when you can subsequently access it from individual work items. So, as a simple heuristic, I'm only considering functions that receive a cl::sycl::group as one of their parameters (which would allow them to call parallel_for_work_item).

Yes, implementing hierarchical parallel for correctly is extremely tricky. Another issue to consider that we haven't even touched (and which makes it even more complicated) is that everything inside the work group scope (including code executed from function calls there) should be executed only once.
Currently, we completely ignore that which requires us to be extra careful about what we make `__shared__`: If we have code in the work group scope, that code is currently executed by all threads - if the variables it operates on are now shared, race conditions may occur that lead to incorrect results. This is why currently the cases where variables are marked as `__shared__` are very limited.

I think, the long-term, spec-conformant solution should work something like this:
* Walk the complete call set in the work group scope (but don't descend into `parallel_for_work_item` calls and `parallel_for_work_item` arguments - don't want to have kernel lambdas as `__shared__`) to find any calls to `parallel_for_work_item`. One could argue that the spec would also require pfwi arguments to be `__shared__` since they are still in the work group scope - this would however make little sense IMHO.
* Find out which variables are captured by `parallel_for_work_item` kernel lambdas
* Only mark those as `__shared__` (if they are not `private_memory`), which increases performance and reduces local memory usage
* Once in IR, inline everything except calls to `parallel_for_work_item()`
* Surround everything before and after `parallel_for_work_item()` calls with `if(local_id == 0){...}`.

Only marking those variables as `__shared__` that are also used in pfwi would also guarantee correct behavior with the current situation where we do not yet have the `if(local_id == 0){...}` in the group scope.

I also suspect that, in general, marking stuff as `__shared__` may have to be done in IR and not AST. If the work group scope calls a function, then everything inside that function will also have to be declared as `__shared__`. If this function is also used by another (e.g. non-hierarchical) kernel in the same compilation unit, then the variables there may be incorrectly marked as `__shared__`. 
So, in the long run, we need a way to generate different code for the same function when called from different kernels.

tests/device_compilation_tests.cpp
```diff
@@ -25,82 +25,98 @@
...
     for(auto&& e : init) {
       parent_value += e;
     }
-    foo1(parent_value);
+    call_lambda(parent_value);
   }
 
   virtual ~MyParent() {
-    printf("~MyParent: %d\n", parent_value);
+    call_on_dtor(parent_value + 7);
```
illuhad: When running auto-test framework (e.g., `HIPSYCL_NO_RUNTIME_TESTING=rocm python3 build.py psalz improve-shmem-detection`), device compilation tests fail on ROCm:

```
error: <unknown>:0:0: in function _ZN29device_compilation_test_suite8MyParentD2Ev void (%"struct.device_compilation_test_suite::MyParent"*): unsupported indirect call to function <unknown>
```
I'm not sure if this is a new problem introduced by your changes. Could well be that the more elaborate tests now expose an already existing problem. I'm also not sure if an `std::function` in kernel code is allowed by the spec. In principle, the spec doesn't really specify the interaction between STL and SYCL. I could well imagine that `std::function` breaks some assumptions (that are legal to make for SYCL) of device compilers that compile for architectures without stack/function pointers/proper functions and have to inline everything.
In general, I think we should make sure that the tests always run, and potentially change the tests.

psalz: Ah yes, I didn't consider that the way `std::function` works might be illegal on some architectures. Luckily it's not integral to the test, I just wanted to assert somehow that the destructor of `MyParent` was correctly being called. I changed the mechanism - please try again!

psalz:
> I think, the long-term, spec-conformant solution should work something like this:
> ...

Yes, that sounds reasonable!

> Surround everything before and after `parallel_for_work_item()` calls with `if(local_id == 0){...}`.

Hmm, I think to make this work we would then also require `parallel_for_work_item()` to be executed using dynamic parallelism - because how would something like a loop or branching otherwise be done? For example:

```cpp
// Iterate over all tiles
for(size_t K = 0; K < SIZE; K += TILE_SIZE) {
  // Copy data
  group.parallel_for_work_item([&](cl::sycl::h_item<2> h_item) {
    ...
  });
  // Do some work
  group.parallel_for_work_item([&](cl::sycl::h_item<2> h_item) {
    ...
  });
}
```

Apart from the fact that its not quite clear how one would wrap the loop head in such a conditional; suppose it was, then all other threads would execute the inner functions only once.

By the way, such a program currently only works because we're not considering the loop variable for marking as shared memory. If `K` were declared before the loop, things would probably break once executed on more than one warp/wavefront.

> I also suspect that, in general, marking stuff as `__shared__` may have to be done in IR and not AST. If the work group scope calls a function, then everything inside that function will also have to be declared as `__shared__`. If this function is also used by another (e.g. non-hierarchical) kernel in the same compilation unit, then the variables there may be incorrectly marked as `__shared__`.

Yep, I actually ran into this issue with my initial overzealous approach. Not sure how difficult it would be to duplicate some functions in LLVM, but I'd expect some pain :)

illuhad:
> Hmm, I think to make this work we would then also require parallel_for_work_item() to be executed using dynamic parallelism - because how would something like a loop or branching otherwise be done? 

Unfortunately, I don't think we can use dynamic parallelism. As far as I know it's currently not implemented in clang, and if it were, we couldn't use it: The CUDA documentation very clearly states that the launched child kernel has its own shared memory and any attempt to access the shared memory of the parent kernel is not allowed.
This of course defeats one of the main purposes of the `parallel_for_work_group` scope where one would like to allocate shared memory that is then used by the child kernel in `parallel_for_work_item`. We might be able to use a section of global memory to emulate local memory, but this would of course be massively inefficient.

> Apart from the fact that its not quite clear how one would wrap the loop head in such a conditional; suppose it was, then all other threads would execute the inner functions only once.

hm, good point. If the approach with conditional turns out to be infeasible we could also try the following approach (which may perhaps be easier to implement, at least initially when ignoring the complexities described by the last two points):
* No conditional, but we mask out all write accesses to shared memory and global memory such that they only take place in the main thread (so, in effect, we surround each write with its own conditional).
* Each write must also be surrounded by synchronization primitives to make sure all threads see the changes from the main thread.
* The same rules as for writes must also be applied to externally defined functions since these may also perform writes internally (think e.g. of atomics).
* Apart from variables captured by `parallel_for_work_item()`, variables that depend on return values from externally defined functions must also become `__shared__` because the return value may depend on the execution of a write (e.g. `atomicAdd()`)

So, compared to the conditional solution we would have potentially a lot more conditionals and synchronization in the worst case, i.e. a lot of writes in the work group scope. But, in the regular case, where the user maybe allocates some shared memory, initializes it and then spawns some `parallel_for_work_item()` calls (potentially in a loop) there should be little overhead... Unless the iteration variable is captured by `parallel_for_work_item()`. Then it will get really messy because of additional synchronization in each loop iteration..

> Yep, I actually ran into this issue with my initial overzealous approach. Not sure how difficult it would be to duplicate some functions in LLVM, but I'd expect some pain :)

Do we really need a duplication? Couldn't we just add an additional pass that just inlines everything inside a kernel? This is allowed because recursion, function pointers etc in kernels is not allowed, so all SYCL kernels should be inline-able. I haven't checked, but most likely this is easy to do in llvm.
Once everything inside kernels is inlined, we would only need to edit the content of the individual kernels. We would then only need functionality to turn a local variable into a `__shared__` variable (probably along the lines of making it global and giving it the right address space)

illuhad:
Looks good now! Thanks & sorry for the delay!

PR #148, opened 2019-10-01
Handle lack of devices more gracefully

psalz:
Just a minor change; if no devices are available we'd like for hipSYCL to return an empty list from `device::get_devices`, instead of just throwing.

Since `hipGetDeviceCount` can [only either return](http://rocm-developer-tools.github.io/HIP/group__Device.html#ga8555d5c76d88c50ddbf54ae70b568394) `hipSuccess` or `hipErrorNoDevice`, I simply removed the call to `detail::check_error` inside `device::get_num_devices` (btw, that method being public appears to be non-standard).

### DIFF
--- a/src/libhipSYCL/device.cpp
+++ b/src/libhipSYCL/device.cpp
@@ -65,7 +65,7 @@ vector_class<device> device::get_devices(
 int device::get_num_devices()
 {
   int num_devices = 0;
-  detail::check_error(hipGetDeviceCount(&num_devices));
+  hipGetDeviceCount(&num_devices);
   return num_devices;
 }
 


illuhad:
Thank you. I agree that this function should not throw if there are now devices, but the current implementation should already take that into account. Is it not working? Current implementation in `detail::check_error()`:
https://github.com/illuhad/hipSYCL/blob/2daf8407e49dd32ebd1c266e8e944e390d28b22a/src/libhipSYCL/exception.cpp#L54-L61

psalz:
You're right, I should've checked the definition of `detail::check_error` - sorry about that. I poked around some more and it turns out the call to `hipGetDeviceCount` actually returns a `hipErrorUnknown` - which causes hipSYCL to throw. The problem appears to be that CUDA is not actually throwing a `cudaErrorNoDevice`, but a `cudaErrorInsufficientDriver` (which `cudaGetDeviceCount` [is allowed to return](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1g18808e54893cfcaafefeab31a73cc55f)).

HIP [doesn't seem to handle this](https://github.com/ROCm-Developer-Tools/HIP/blob/8ff7a99a4cc431cb4feb0e7cbf02b967917f4aa4/include/hip/nvcc_detail/hip_runtime_api.h#L200) error in the version currently included by hipSYCL, however [it looks like support has since been added](https://github.com/ROCm-Developer-Tools/HIP/blob/108992428d8b8f09e675127061ae3dd85d0668aa/include/hip/nvcc_detail/hip_runtime_api.h#L215).

So I guess the best approach would be to update HIP and then handle this error within `detail::check_error` as well!

The context of this is by the way that I was trying to run a hipSYCL executable on the head node of a GPU cluster, after building it there.

Issue #158, opened 2019-10-15
Reverse mapping of SYCL work item indices to HIP/CUDA dimensions


illuhad:
Currently, we map:
* id.get(0) -> hipThreadIdx_x
* id.get(1) -> hipThreadIdx_y
* id.get(2) -> hipThreadIdx_z

This is problematic, because it is inconsistent with linear ids as required by the spec. In SYCL, linear ids use the highest dimension as fastest index. This affects not only linear ids themselves but also statements like `accessor[id]` where a linear id is internally used.
In HIP/CUDA on the other hand, the x dimension (or dimension 0, in OpenCL terminology) is assumed to be the fastest moving index. As coalescing of memory accesses happens along the x dimensions, accessing multi-dimensional memory buffers the way SYCL expects it will yield suboptimal performance.

To reconcile the SYCL world and the HIP/CUDA world, we should change the mapping such that
* 1D case:
  * id.get(0) -> hipThreadIdx_x
* 2D case:
  * id.get(0) -> hipThreadIdx_y
  * id.get(1) -> hipThreadIdx_x
* 3D case:
  * id.get(0) -> hipThreadIdx_z
  * id.get(1) -> hipThreadIdx_y
  * id.get(2) -> hipThreadIdx_x

This should be fixed ASAP as the current implementation is arguably a performance bug.

keryell:
We are using heterogeneous computing for compute performance and power efficiency... so this is important!

illuhad:
Indeed. I have done most of the performance testing with 1D benchmarks up to now which is probably how this could have gone unnoticed. This is top priority, I have already started working on a fix right after I learned of the issue today afternoon - hopefully this will be fixed in the next few days.

PR #160, opened 2019-10-16
Reverse mapping of SYCL ids to backend ids (fixes issue #158)

illuhad:
This causes SYCL work item/group ids to be mapped to the HIP/CUDA backend in reverse order:
2D example:
id[0] -> hipThreadIdx_y
id[1] -> hipThreadIdx_x

See issue #158 for more details.

This can yield performance benefits since SYCL encourages the user to use dimensions-1 as fastest moving index, while HIP/CUDA uses the x component as fastest moving index.
While this primarily affects work items, the reversal is also done for group ids to remain consistent.

To make everything work again, ranges (in particular related to kernel call configurations) are also reversed.

While this PR passes the unit tests, due to the fundamental nature of this change and the great potential for breakage, I will leave this PR open a couple more days than usual to give everybody a chance to test.

### DIFF
--- a/include/CL/sycl/detail/thread_hierarchy.hpp
+++ b/include/CL/sycl/detail/thread_hierarchy.hpp
@@ -41,7 +41,7 @@ namespace detail {
 // should only be used in the implementation of more 
 // high-level functions in this file since they do
 // not take into the transformation needed to map
-// the fastest sycl index to the fastest hardware index:
+// the fastest SYCL index to the fastest hardware index:
 // Per SYCL spec, the highest dimension (e.g. dim=2 for 3D)
 // is the fastest moving spec. In HIP/CUDA, it is x.
 // Consequently, any id or range that is actually used
@@ -401,7 +401,7 @@ inline size_t get_group_id<3>(int dimension)
 
 /// Flips dimensions such that the range is consistent with the mapping
 /// of SYCL index dimensions to backend dimensions.
-/// When launching a sycl kernel, grid and blocksize should be transformed
+/// When launching a SYCL kernel, grid and blocksize should be transformed
 /// using this function.
 template<int dimensions>
 inline dim3 make_kernel_launch_range(dim3 range);
--- a/include/CL/sycl/detail/thread_hierarchy.hpp
+++ b/include/CL/sycl/detail/thread_hierarchy.hpp
@@ -37,7 +37,15 @@ namespace cl {
 namespace sycl {
 namespace detail {
 
-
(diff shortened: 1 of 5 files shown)


psalz:
I ran this on the [Celerity examples](https://github.com/celerity/celerity-runtime/tree/master/examples) and it looks like everything still works :). Also the speedup for the naive matrix multiplication appears to be pretty massive (anywhere from 2x on 8 GPUs to 5x on a single GPU). I only did some very quick and dirty benchmarking, so take it with a grain of salt.

illuhad:
Great news, thanks for the feedback! There's an issue for benchmarking (#161), so hopefully we'll have some reliable results in the future :)

include/CL/sycl/detail/thread_hierarchy.hpp
```diff
@@ -37,7 +37,15 @@ namespace cl {
 namespace sycl {
 namespace detail {
 
-
+// The get_global_id_* and get_global_size_* functions 
+// should only be used in the implementation of more 
+// high-level functions in this file since they do
+// not take into the transformation needed to map
+// the fastest sycl index to the fastest hardware index:
```
keryell:
```suggestion
// the fastest SYCL index to the fastest hardware index:
```

include/CL/sycl/detail/thread_hierarchy.hpp
```diff
@@ -198,97 +207,222 @@ sycl::range<dimensions> get_global_size()
...
   case 2:
-    return hipBlockIdx_z;
+    return hipBlockIdx_x;
   }
   return 0;
 }
 
+/// Flips dimensions such that the range is consistent with the mapping
+/// of SYCL index dimensions to backend dimensions.
+/// When launching a sycl kernel, grid and blocksize should be transformed
```
keryell:
```suggestion
/// When launching a SYCL kernel, grid and blocksize should be transformed
```

include/CL/sycl/handler.hpp
```diff
@@ -933,7 +933,9 @@ void set_args(Ts &&... args);
 #endif
       {
         __hipsycl_launch_kernel(detail::dispatch::device::parallel_for_kernel<KernelName>,
-                              grid, block, shared_mem_size, stream->get_stream(),
+                              detail::make_kernel_launch_range<dimensions>(grid),
```
keryell: All the indentation in this change batches seems weird to me.

illuhad: you're right, thank you! I'll go over the indentation again and prettify things.

PR #163, opened 2019-10-21
Add dedicated data transfer stream for queues

psalz:
This adds a secondary HIP stream to each SYCL queue, which is intended for submitting data transfers.  The stream is created for the same device as the primary stream, and can be accessed using `[queue|handler]::get_transfer_stream`. Since modern GPUs support "concurrent copy and execution", this allows hipSYCL to perform some data transfers concurrently with kernel executions, potentially reducing launch latency for subsequent kernel tasks. Both implicit (through device accessors) and explicit (through copy operations) data transfers are changed to make use of the new stream.

Now I'm not sure (and I haven't tested this), whether this allows uploads and downloads to be executed concurrently as well (as modern GPUs typically have two copy engines, one for upload and one for download). I suspect not, but I think this change should at least cover typical applications. Ultimately I think hipSYCL should move away from the `SYCL queue == HIP stream` correspondence, and instead use some kind of stream pool. This would then also allow for concurrent kernel executions (for better occupancy).

Note that I didn't add corresponding `get_hip_transfer_stream` members to `queue` and `handler`, as `get_hip_stream` doesn't seem to be used anywhere.

----

Here's an example that demonstrates the changes:

```cpp
#include <cstdio>
#include <list>

#include <CL/sycl.hpp>

	cl::sycl::queue queue;

	cl::sycl::buffer<uint64_t, 1> buf_a(1024);
	cl::sycl::buffer<uint64_t, 1> buf_b(1024);
	std::list<std::pair<std::string, cl::sycl::event>> events;

	queue
	    .submit([&](cl::sycl::handler& cgh) {
		    auto dw_a = buf_a.get_access<cl::sycl::access::mode::discard_write>(cgh);
		    auto dw_b = buf_b.get_access<cl::sycl::access::mode::discard_write>(cgh);
		    cgh.parallel_for<class Initialize>(buf_a.get_range(), [=](cl::sycl::item<1> item) {
			    dw_a[item] = 5e7;
			    dw_b[item] = 5e7;
		    });
	    })
	    .wait();

	auto kernel_e = queue.submit([&](cl::sycl::handler& cgh) {
		auto rw_a = buf_a.get_access<cl::sycl::access::mode::read_write>(cgh);
		cgh.parallel_for<class Kernel>(buf_a.get_range(), [=](cl::sycl::item<1> item) {
			// Do some work
			const auto max = rw_a[item];
			double sum = 0.f;
			while(sum < max) {
				sum += 1.0;
			}
			rw_a[item] = sum;
		});
	});

	uint64_t copied_value = 1337;
	auto copy_e = queue.submit([&](cl::sycl::handler& cgh) {
		auto r_b = buf_b.get_access<cl::sycl::access::mode::read>(cgh, 1);
		cgh.copy(r_b, &copied_value);
	});

	events.push_back(std::make_pair("kernel", kernel_e));
	events.push_back(std::make_pair("copy", copy_e));

	auto it = events.begin();
	while(true) {
		const auto status = it->second.get_info<cl::sycl::info::event::command_execution_status>();
		if(status == cl::sycl::info::event_command_status::complete) {
			printf("%s done\n", it->first.c_str());
			if(it->first == "copy") { printf("copied value: %ld\n", copied_value); }
			it = events.erase(it);
		} else {
			++it;
		}

		if(it == events.end()) {
			if(events.empty()) break;
			it = events.begin();
		}
	}
}
```

Output before the patch:
```
Oct 21 13:55:45 kernel done
Oct 21 13:55:45 copy done
Oct 21 13:55:45 copied value: 50000000
```

Output after the patch:
```
Oct 21 13:56:54 copy done
Oct 21 13:56:54 copied value: 50000000
Oct 21 13:56:56 kernel done
```

While explicit copy operations are my primary use case for this, the same works for the implicit transfers generated by `buffer::get_access` calls.

### DIFF
--- a/include/CL/sycl/handler.hpp
+++ b/include/CL/sycl/handler.hpp
@@ -571,7 +571,7 @@ void set_args(Ts &&... args);
       return detail::task_state::enqueued;
     };
 
-    this->submit_task(kernel_launch);
+    this->submit_task(kernel_launch, stream);
   }
 
   template <typename KernelName = class _unnamed_kernel,
@@ -797,6 +797,7 @@ void set_args(Ts &&... args);
 
 
   detail::stream_ptr get_stream() const;
+  detail::stream_ptr get_transfer_stream() const;
 private:
   void add_access(detail::buffer_ptr buff,
                   access::mode access_mode,
@@ -940,7 +941,7 @@ private:
       return detail::task_state::enqueued;
     };
 
-    this->submit_task(kernel_launch);
+    this->submit_task(kernel_launch, stream);
 
   }
 
@@ -986,7 +987,7 @@ private:
       return detail::task_state::enqueued;
     };
 
-    this->submit_task(kernel_launch);
+    this->submit_task(kernel_launch, stream);
   }
 
 
@@ -1038,7 +1039,7 @@ private:
       return detail::task_state::enqueued;
     };
 
-    this->submit_task(kernel_launch);
+    this->submit_task(kernel_launch, stream);
   }
 
   template <typename KernelName, typename WorkgroupFunctionType, int dimensions>
@@ -1081,7 +1082,7 @@ private:
       return detail::task_state::enqueued;
     };
 
-    this->submit_task(kernel_launch);
+    this->submit_task(kernel_launch, stream);
   }
 
   template <typename T, int dim, access::mode mode, access::target tgt,
@@ -1200,7 +1201,7 @@ private:
     using namespace detail;
     using T = std::remove_pointer_t<decltype(get_raw_pointer(src))>;
     debug_print_copy_kind(kind);
-    const auto stream = this->get_stream();
+    const auto stream = this->get_transfer_stream();
     auto copy_launch = [=]() -> task_state
     {
       stream->activate_device();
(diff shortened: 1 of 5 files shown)


illuhad:
Thank you! I've had the idea of adding several streams to one queue too, but had discarded the idea in favor of what I'm working on at the moment: A major rewrite of the scheduler/runtime, which among other things will decouple the queue from the underlying stream. Basically, the scheduler will create and manage as many streams as there are DMA engines + concurrent kernel execution capabilities. The scheduler will then distribute work automatically across all the streams. This will not only allow for improved data transfer overlapping, but also for automatic concurrent kernel execution with a single queue.
While it may make sense to merge your changes as a temporary solution, eventually it will have to be replaced by the features of the new runtime.

psalz:
Sounds great! I only need this for some experimental stuff I'm working on right now, so if you don't want to merge I can also just work of this branch!

illuhad:
Let's see how fast I can make progress. Obviously this is a very useful feature - the changes I want to  make are quite massive and not restricted to this aspect (I can describe more details in an issue if you are interested). I think it still makes sense to merge your changes if it turns out that it takes me quite some time to implement all my changes.

illuhad:
I think we can close this now as similar functionality is already contained in recent hipSYCL.

Issue #149, opened 2019-10-08
Segfault in __cudaPopCallConfiguration when using CUDA 10.1


n-io:
Trying to execute the `vector add` example from the githib may page results in segfault after compilation succeeds.

Installing into user space to avoid using sudo access, then build and run the test app:
```
$ cmake -DCMAKE_INSTALL_PREFIX=/data/nico/local/ /data/nico/code/hipSYCL/
$ make install
$ cd /tmp/foo
$ /data/nico/local/bin/syclcc-clang --hipsycl-gpu-arch=sm_52 test.cpp
$ LD_LIBRARY_PATH=/data/nico/local/lib ./test
Segmentation fault (core dumped)
```
LLVM/Clang environment variables are all set as suggested.

`gdb` backtrace says:
```
(gdb) run
Starting program: /tmp/foo/test 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7ffff663a700 (LWP 7740)]
[New Thread 0x7ffff4a74700 (LWP 7741)]
[New Thread 0x7ffff4273700 (LWP 7742)]
[New Thread 0x7ffff3a72700 (LWP 7743)]
[New Thread 0x7ffff3271700 (LWP 7744)]

Thread 2 "test" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff663a700 (LWP 7740)]
0x00007ffff7b67b90 in __cudaPopCallConfiguration () from /usr/local/cuda-10.1/targets/x86_64-linux/lib/libcudart.so.10.1
(gdb) backtrace
#0  0x00007ffff7b67b90 in __cudaPopCallConfiguration () from /usr/local/cuda-10.1/targets/x86_64-linux/lib/libcudart.so.10.1
#1  0x0000000000403a69 in __hipsycl_pop_kernel_call() ()
[...]

PR #164, opened 2019-11-09
Fixing Issue #149 by removing __cudaPopCallConfiguration() call

sbalint98:
As discussed in issue #[149](https://github.com/illuhad/hipSYCL/issues/149) the removal of the  __cudaPopCallConfiguration() call fixes the segmentation fault that surfaced when using Cuda 10.1.


illuhad:
Thank you! Now that we can in principle also support cuda 10.1, we might want to upgrade the cuda installation scripts for 10.1. You can create a new issue for that, if you like :)

PR #167, opened 2019-11-15
Update SYCL targets image

illuhad:
Update SYCL targets image to reflect current status of SYCL implementations


fodinabor:
I think you can remove the ? from other OpenCL SPIR-V devices for Intel SYCL implementation. They seem to support anything that has OpenCL 1.2 and SPIR-V support.. (also theoretically other targets if using another backend than spir-unknown-unknown)
They do also support Intel FPGAs.

illuhad:
thanks, I think I've added the ? because pretty much the only vendor supporting OpenCL SPIR-V ingestion is Intel, so in practice it's not clear if there are other devices.. Maybe we can just remove the entire line because it's pure speculation. Will add Intel FPGA, thanks!

fodinabor:
I _think_ AMD on ROCm supports SPIR-V, too.. Didn't check again though.

Fri Nov 15 13:47:31 GMT+01:00 2019 Aksel Alpay <notifications@github.com>:

illuhad:
Added Intel FPGA, layout changes, changed color schemes to better match implementations/implementors.

If ROCm supports SPIR-V now, it must be a very new (and probably experimental) feature. I'm sure they didn't a few months ago. The lack of SPIR-V support is also why some people where not so happy when AMD dropped SPIR in ROCm support since it prevented ComputeCpp from working..

fodinabor:
Yeah.. I See.. Was a bit out of date then.. And probably mixed spir and spir-v

Fri Nov 15 14:22:07 GMT+01:00 2019 Aksel Alpay <notifications@github.com>:

keryell:
@illuhad I let you tweet this image again... :-)

Issue #166, opened 2019-11-15
Which C++ standards are supported in device code?


invexed:
First of all I'd just like to say thank you for this amazing project. I'm looking forward to testing it out.

Before I get too deep into researching SYCL, may I ask which C++ standards are supported in device code with the clang plugin and CUDA backend? I have a lot of C++17 device code (that currently compiles with clang) that I'm hoping to port to SYCL. Is that supported by hipSYCL?

fodinabor:
As hipSYCL uses a mostly standard clang frontend and only adds modifiers to some functions, almost all C++ features, that are supported by Clang, should be supported in the device code as well.
I guess you're using Clang for your CUDA code already? Then I don't see any issues with using that code in hipSYCL.

illuhad:
Thanks for your interest in the project! As @fodinabor says, C++17 should be no problem in principle. Of course, there are some restrictions on the allowed subset of C++ in SYCL device code, similarly to other heterogeneous programming models:
* nothing that requires runtime support like exceptions, new/delete, rtti, ... 
* recursion is not allowed by the SYCL spec in device code; cuda is a bit more forgiving here. While hipSYCL does allow the same constructs when compiling for nvidia GPU as CUDA, the AMDGPU LLVM backend has no concept of functions, so recursive code will most likely fail when compiling for AMD.
* The SYCL spec does not define how SYCL interacts with the STL; simple things like `std::plus` or static functionality (type traits and such) will work in device code, but in general it is hard to foresee if STL will work in device code because it depends on STL implementation which is beyond our control. For example, some STL implementations may incorrectly pull in SSE intrinsics or similar for some headers which will of course fail in device code.

There's one additional caveat: Since the hipSYCL headers require at least C++14, the `syclcc-clang` compiler wrappers pass `-std=c++14` to clang to enforce a standard that can compile our header files. From playing around it seems that clang will use the standard that is the last one on the command line - in this case we will most likely be fine if the user additionally sets the standard.

@sbalint98, can you please investigate how a proper solution could look like? Maybe we can add a mechanism in `syclcc-clang` to only add `-std=c++14` if no other, newer standard has been specified (which might be a good idea anyway)?

Until we have fixed this on the master branch, you can just manually replace the `-std=c++14` in `bin/syclcc-clang` for your experimentation if it turns out to be an issue - it should be easy with find/replace :)

invexed:
[...]

PR #168, opened 2019-11-20
Implemented cpp standard checking in syclcc compiler wrapper 

sbalint98:
Ass discussed in issue #166, now -std=c++14 will only be automatically added when the user does not define a cpp standard. If the user tries to use a standard earlier than c++14 an error is thrown.

### DIFF
--- a/bin/syclcc-clang
+++ b/bin/syclcc-clang
@@ -173,9 +173,10 @@ class syclcc_config:
 """  Enter bootstrap mode. This is only required when building hipSYCL itself and 
   should not be set by the user.""")
     }
-
+    self._insufficient_cpp_standards = ['98', '03', '11', '0x']
     self._hipsycl_args = []
     self._forwarded_args = []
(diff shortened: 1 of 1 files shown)


bin/syclcc-clang
```diff
@@ -528,7 +549,8 @@ class pure_cpu_compiler:
     hipsycl_library_path = os.path.join(config.hipsycl_installation_path,"lib/")
 
     self._linker_args = []
-    self._compiler_args = ["-fopenmp", "-std=c++14"]
+    self._compiler_args = ["-fopenmp"]
+    self._common_compiler_args = config.common_compiler_args
```
illuhad: You're not actually using `self._common_compiler_args` for the CPU compiler, right? So, you would need to either also append it to `command` in `run()` or (I think preferably) just do `self._compiler_args += config.common_compiler_args` here.

sbalint98: Yes, you're right.

illuhad:
Wonderful, thank you :)

PR #170, opened 2019-11-23
Add multi-GPU example

senior-zero:
If I understand it correctly, memory for buffers is allocated in buffer object construction. The memory is allocated for the device that was specified in queue construction. I haven't found a different way of specifying the buffer device, so I'm constructing buffers immediately after queue construction. If there is a better way, I'll be glad to hear about it and update the example. Also, I would like to use an event profiler, but it seems that it hasn't implemented yet. If profiling and events are not in progress, I would like to implement it.

### DIFF
--- a/examples/multi_gpu/multi_gpu_vec_add_async.cpp
+++ b/examples/multi_gpu/multi_gpu_vec_add_async.cpp
@@ -1,0 +1,96 @@
+#include <memory>
+#include <cassert>
+#include <iostream>
+
+#include <CL/sycl.hpp>
+
+using data_type = float;
+
+class device_context
+{
+public:
+  device_context (
+    size_t size_arg,
+    const data_type * a,
+    const data_type * b,
+    const data_type * c,
+    cl::sycl::device &device)
+  : size (size_arg)
+  , queue (device)
+  , buff_a (a, size)
+  , buff_b (b, size)
+  , buff_c (c, size)
+  {
+    assert(a.size() == b.size());
+  }
+
+  const size_t size {};
+  cl::sycl::queue queue;
+  cl::sycl::buffer<data_type> buff_a;
+  cl::sycl::buffer<data_type> buff_b;
+  cl::sycl::buffer<data_type> buff_c;
+};
+
+void add(std::vector<std::unique_ptr<device_context>> &contexts)
+{
+  for (auto &context: contexts)
+    {
+      cl::sycl::range<1> work_items{context->size};
+
+      auto e = context->queue.submit([&](cl::sycl::handler& cgh){
+        auto access_a = context->buff_a.get_access<cl::sycl::access::mode::read>(cgh);
+        auto access_b = context->buff_b.get_access<cl::sycl::access::mode::read>(cgh);
+        auto access_c = context->buff_c.get_access<cl::sycl::access::mode::write>(cgh);
+
+        cgh.parallel_for<class vector_add>(work_items,
+                                           [=] (cl::sycl::id<1> tid) {
+                                             access_c[tid] = access_a[tid] + access_b[tid];
+                                           });
+      });
+    }
+}
+
+int main()
+{
+  std::vector<cl::sycl::device> devices = cl::sycl::device::get_devices ();
+
+  const unsigned int n = 100'000'000;
+
+  std::vector<data_type> a (n, 1);
+  std::vector<data_type> b (n, 2);
+  std::vector<data_type> c (n);
+
+  double single_gpu_elapsed_time {};
+
+  for (unsigned int devices_count = 1; devices_count <= devices.size (); devices_count++)
+    {
+      std::vector<std::unique_ptr<device_context>> contexts;
+      const unsigned int chunk_size = n / devices_count;
+
+      for (unsigned int dev_id = 0; dev_id < devices_count; dev_id++)
+        {
+          const unsigned int beg = chunk_size * dev_id;
+          const unsigned int end = dev_id == devices_count - 1 ? n : chunk_size * (dev_id + 1);
+          const unsigned int size = end - beg;
+
+          contexts.push_back (std::make_unique<device_context> (
+            size,
+            a.data () + beg,
+            b.data () + beg,
+            c.data () + beg,
+            devices[dev_id]));
+        }
+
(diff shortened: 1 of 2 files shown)


examples/multi_gpu/multi_gpu_vec_add_async.cpp
```diff
@@ -0,0 +1,96 @@
...
+            size,
+            a.data () + beg,
+            b.data () + beg,
+            c.data () + beg,
+            devices[dev_id]));
+        }
+
+      auto begin = std::chrono::system_clock::now (); // TODO Implement event profiler
+      add(contexts);
+      auto end = std::chrono::system_clock::now ();
```
psalz: You can't just measure the time like that, because at this point the kernels are not guaranteed to be finished. You can use the `cl::sycl::event` returned by `cl::sycl::queue::submit` to block (`cl::sycl::event::wait`) until it is done. However what you'd probably want to do in this case is to collect all events in an array and poll for their completion status using something like `e.get_info<cl::sycl::info::event::command_execution_status>() == cl::sycl::info::event_command_status::complete`.

psalz:
> I haven't found a different way of specifying the buffer device, so I'm constructing buffers immediately after queue construction.

I think hipSYCL's buffer implementation needs some love in general, as per spec buffers aren't really tied to any particular device at all. I've run into this issue as well, and I think in the long run buffers should probably be allocated lazily on the device side whenever they're being used.

examples/multi_gpu/multi_gpu_vec_add_async.cpp
psalz: Now that I think about it, you could also just block on each event as long as you do it after submitting all the kernels first - you're going to wait for the longest running one anyways.

illuhad: No need to poll the completion state, `wait()` will already do essentially the same thing on an internal atomic bool that is set once the event has finished.

illuhad:
Thank you for this contribution. We could add some examples. I'm not 100% sure though if maybe the multi-gpu case might be a better candidate for unit tests though, since, (at least once the new runtime lands, see below) this is nothing special about hipSYCL 

The way a buffer should work per the SYCL spec is that you create one buffer for all devices. If you use the buffer inside a command group that is scheduled for execution on a particular device, the buffer will allocate memory on that device and make sure that data is present on the device. At the moment there are some limitations in the hipSYCL implementation, most notably that the buffer is hardcoded for managing one host-side allocation and exactly one device-side allocation (which is why you are using one buffer per device, the proper way to do it would be with a single buffer for all devices). On NVIDIA at least, you can circumvent this issue by telling it to allocate CUDA unified memory. The GPU driver will then automatically migrate the data to wherever it's needed. This feature is not particularly well tested at the moment though.

>  I think hipSYCL's buffer implementation needs some love in general, as per spec buffers aren't really tied to any particular device at all. I've run into this issue as well, and I think in the long run buffers should probably be allocated lazily on the device side whenever they're being used.

This is coming with the new scheduler/runtime that is currently being developed; it will bring several new features to the buffer implementation:
* support for managing arbitrary numbers of allocations on different devices (i.e., a single buffer covering many devices);
* support for lazy allocation on command group scheduling;
* support for data management at sub-buffer granularity using a 3D "page" table.

> Also, I would like to use an event profiler, but it seems that it hasn't implemented yet. If profiling and events are not in progress, I would like to implement it.

The reason this isn't implemented yet is that it would force us to insert a HIP/CUDA event into the stream after each SYCL operation which can add measurable overhead. On the other hand, since a hipSYCL program behaves like a HIP/CUDA program, the regular NVIDIA and AMD profiling tools can be used which yield better profiling data with as low overhead as possible.
Unless you need event profiling super urgently, I would suggest to simply use existing profiling tools and only implement event profiling once we have the new runtime, since a lot of things will change then. I would expect an implementation of event profiling in hipSYCL to be of similar precision as measuring the time between command group submission and the end of a `wait()` call (as suggested by @psalz) with the current runtime implementation anyway.

senior-zero:
> Thank you for this contribution. We could add some examples. I'm not 100% sure though if maybe the multi-gpu case might be a better candidate for including it in the unit tests though, since, (at least once the new runtime lands, see below) this is nothing special about hipSYCL.
> 
> The way a buffer should work per the SYCL spec is that you create one buffer for all devices. If you use the buffer inside a command group that is scheduled for execution on a particular device, the buffer will allocate memory on that device and make sure that data is present on the device. At the moment there are some limitations in the hipSYCL implementation, most notably that the buffer is hardcoded for managing one host-side allocation and exactly one device-side allocation (which is why you are using one buffer per device, the proper way to do it would be with a single buffer for all devices). On NVIDIA at least, you can circumvent this issue by telling it to allocate CUDA unified memory. The GPU driver will then automatically migrate the data to wherever it's needed. This feature is not particularly well tested at the moment though.
> 
> > I think hipSYCL's buffer implementation needs some love in general, as per spec buffers aren't really tied to any particular device at all. I've run into this issue as well, and I think in the long run buffers should probably be allocated lazily on the device side whenever they're being used.
> 
> This is coming with the new scheduler/runtime that is currently being developed; it will bring several new features to the buffer implementation:
> 
> * support for managing arbitrary numbers of allocations on different devices (i.e., a single buffer covering many devices);
> * support for lazy allocation on command group scheduling;
> * support for data management at sub-buffer granularity using a 3D "page" table.
> 
> > Also, I would like to use an event profiler, but it seems that it hasn't implemented yet. If profiling and events are not in progress, I would like to implement it.
> 
> The reason this isn't implemented yet is that it would force us to insert a HIP/CUDA event into the stream after each SYCL operation which can add measurable overhead. On the other hand, since a hipSYCL program behaves like a HIP/CUDA program, the regular NVIDIA and AMD profiling tools can be used which yield better profiling data with as low overhead as possible.
> Unless you need event profiling super urgently, I would suggest to simply use existing profiling tools and only implement event profiling once we have the new runtime, since a lot of things will change then. I would expect an implementation of event profiling in hipSYCL to be of similar precision as measuring the time between command group submission and the end of a `wait()` call (as suggested by @psalz) with the current runtime implementation anyway.

Thank you for your answer. It seems that it's better not to add the examples until buffers are not supporting different devices. I'll reopen this request when the new features are supported by hipSYCL.

PR #178, opened 2019-12-12
Fix race condition in task graph node

PeterTh:
We recently discovered the source for an intermittent, but rather frequent crash we experienced in hipSYCL. By bisecting we traced it back to 2adb3cd8.

I'm not 100% sure about the best way to fix this issue, since I don't know the internal details of hipsycl task scheduling. What I did was add a spinlock to protect access to that vector. (We used a spinlock as that should have basically 0 initialization overhead when creating a new task graph node)

### DIFF
--- a/include/CL/sycl/detail/task_graph.hpp
+++ b/include/CL/sycl/detail/task_graph.hpp
@@ -32,6 +32,7 @@
 #include "../backend/backend.hpp"
 
 #include "stream.hpp"
+#include "util.hpp"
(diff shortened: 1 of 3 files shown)


illuhad:
Thank you, I can see the issue. I think that your solution is certainly good enough since that code will be replaced by the new scheduler anyway. At the moment, there is some networking trouble here so I cannot access my test machine where I usually play with PRs - I'll merge once I've had a chance to try this :)

PR #180, opened 2019-12-17
Fixed syclcc bug, preventing compilation against recent rocm

sbalint98:
Hip headers for the Cuda build are now not included when we compile for HiP. 

This has caused issues when compiling hipSYCL against more recent versions of rocm, because instead of the headers from the rocm installation, the headers from the contrib/HIP directory were considered. 

Now we only include these headers when we are compiling with Cuda support


illuhad:
Thanks!

Issue #127, opened 2019-08-27
Compiler error when indexing sycl::accessor with dimensions > 1


bluebear94:
When compiling [this code](https://gist.github.com/bluebear94/25d7e2b787d9e880b0d2c2f3ef7b5c4b):

```
$ /opt/hipSYCL/bin/syclcc-clang mandelbrot.cpp -o mandelbrot --hipsycl-platform=cuda --hipsycl-gpu-arch=sm_62 --std=c++17 -Wall -Wextra -Wpedantic

[snip]

In file included from mandelbrot.cpp:4:
In file included from /opt/hipSYCL/bin/../include/CL/sycl.hpp:42:
In file included from /opt/hipSYCL/bin/../include/CL/sycl/queue.hpp:41:
In file included from /opt/hipSYCL/bin/../include/CL/sycl/handler.hpp:37:
/opt/hipSYCL/bin/../include/CL/sycl/accessor.hpp:528:76: error: calling a private constructor of class 'cl::sycl::accessor<unsigned char, 1, cl::sycl::access::mode::read,
      cl::sycl::access::target::host_buffer, cl::sycl::access::placeholder::false_t>'
    accessor<dataT, dimensions-1, accessmode, accessTarget, isPlaceholder> sub_accessor;
                                                                           ^
mandelbrot.cpp:74:19: note: in instantiation of function template specialization 'cl::sycl::accessor<unsigned char, 2, cl::sycl::access::mode::read, cl::sycl::access::target::host_buffer,
      cl::sycl::access::placeholder::false_t>::operator[]<2, void>' requested here
                        uint8_t idx = I[y][x];
                                       ^
/opt/hipSYCL/bin/../include/CL/sycl/accessor.hpp:598:3: note: declared private here
  accessor(){}
  ^
/opt/hipSYCL/bin/../include/CL/sycl/accessor.hpp:529:18: error: '_range' is a private member of 'cl::sycl::accessor<unsigned char, 1, cl::sycl::access::mode::read,
      cl::sycl::access::target::host_buffer, cl::sycl::access::placeholder::false_t>'
    sub_accessor._range = detail::range::omit_first_dimension(this->_range);
                 ^
/opt/hipSYCL/bin/../include/CL/sycl/accessor.hpp:602:21: note: declared private here
  range<dimensions> _range;
                    ^
[...]

PR #186, opened 2020-01-10
Fix the nested subscript operator data access for accessors using the new subscript_proxy class

illuhad:
This fixes issue #127 by implementing nested subscript operator access for the accessor class by introducing a new `subscript_proxy` class. The proxy stores the `this` pointer of the original accessor and collects indices until it has as many as the dimension requires. It then uses the stored `this` pointer to invoke `accessor::operator[](id<dimension>)`.

### DIFF
--- a/include/CL/sycl/accessor.hpp
+++ b/include/CL/sycl/accessor.hpp
@@ -116,6 +116,87 @@ size_t get_pointer_offset(const sycl::accessor<dataT, dimensions,
 }
 
 
+
+template <typename dataT, int dimensions, access::mode accessmode,
+          access::target accessTarget, access::placeholder isPlaceholder,
+          int current_dimension = 1>
+class subscript_proxy
+{
+  HIPSYCL_UNIVERSAL_TARGET
+  static constexpr bool can_invoke_access(int current_dim, int dim) {
+    return current_dim == dim - 1;
+  }
+public:
+  static_assert(dimensions > 1, "dimension must be > 1");
+  
+  using accessor_type = sycl::accessor<dataT, dimensions, accessmode,
+                                       accessTarget, isPlaceholder>;
+
+  using next_subscript_proxy =
+      subscript_proxy<dataT, dimensions, accessmode, accessTarget,
+                      isPlaceholder, current_dimension+1>;
+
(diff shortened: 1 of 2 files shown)


tomdeakin:
This looks like a great approach, thanks for getting round to it so fast!

I've built this branch with the CUDA and host (`cpu`) backends and it builds fine.

I did a quick sanity check on the ordering of the construction of `_access_id` and I *think* it's perfect! Do you think the unit test could be made to compare the access pattern more accurately if the kernel set the value with the `get_linear_id` function instead of manually calculating it? Is the below any different? I'm confusing myself thinking about it!

```
  q.submit([&](s::handler& cgh){
    auto acc = buff2.get_access<s::access::mode::discard_read_write>(cgh);

    cgh.parallel_for<class nested_subscript2d>(buff_size2d, [=](s::id<2> idx){
      size_t x = idx[0];
      size_t y = idx[1];
      // Writing
      acc[x][y] = static_cast<int>(idx.get_linear_id()); // <---- instead of manually calculating it
      // Reading and making sure access id the same as with operator[id<>]
      if(acc[x][y] != acc[idx])
        acc[x][y] = -1;
    });
  });
```

illuhad:
It should be exactly the same, the manual calculation just implements the linear id equation. My thinking was that this way, we also test that `get_linear_id()` is actually doing what it's supposed to do. Later on, we have the checks:

```c++
        size_t linear_id3d = x*buff_size3d[1]*buff_size3d[2] + y*buff_size3d[2] + z;
        s::id<3> id3d{x,y,z};
        BOOST_CHECK(host_acc3d[id3d] == linear_id3d);
        BOOST_CHECK(host_acc3d.get_pointer()[linear_id3d] == linear_id3d
```
Since `get_linear_id()` (or rather, the actual index calculation function that is invoked by `get_linear_id()`) is actually used internally for `operator[](sycl::id<d>)`, this way we check
* that access with `operator[][][]` is identical to `operator[](sycl::id<3>)`
* that `operator[](sycl::id<3>)` correctly accesses the elements according to the linear id equation.

tomdeakin:
Ah yes I see that now. Looks good to me.

PR #194, opened 2020-01-15
Pull request to review code for subbuffer implementation

jon-chuang:
I am struggling with some issues, namely memory deallocation.
Everything works fine until deallocation, when is subbuffer should be called and if returns true, deallocation does not take place. How can I implement this?

### DIFF
--- a/include/CL/sycl/buffer.hpp
+++ b/include/CL/sycl/buffer.hpp
@@ -48,6 +48,7 @@
 
 namespace cl {
 namespace sycl {
+
 namespace property {
 namespace buffer {
 
@@ -185,7 +186,7 @@ public:
   {
     _alloc = allocator;
 
-    constexpr bool is_const_iterator = 
+    constexpr bool is_const_iterator =
         std::is_const<
           typename std::remove_reference<
             typename std::iterator_traits<InputIterator>::reference
@@ -211,14 +212,14 @@ public:
       // the memory range [first, last) as writeback buffer
       this->_buffer->enable_write_back(true);
 
-      
+
       auto buff = this->_buffer;
       this->_cleanup_trigger->add_cleanup_callback([first, last, buff](){
         const T* host_data = reinterpret_cast<T*>(buff->get_host_ptr());
 
         std::copy(host_data, host_data + std::distance(first,last), first);
       });
-      
+
     }
   }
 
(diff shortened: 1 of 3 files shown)


include/CL/sycl/buffer.hpp
```diff
@@ -211,27 +212,53 @@ class buffer : public detail::property_carrying_object
...
+
+      // Let's do some void* pointer arithmetic!
+      T* buff = static_cast<T*>(b_->get_buffer_ptr());
+      T* host = static_cast<T*>(b_->get_host_ptr());
+
+      std::cerr << "Status ok" << std::endl << std::flush;
+
+      _buffer = detail::buffer_ptr{
+        new detail::buffer_impl{
+          (void*)(buff + offset),
```
illuhad: What happens if `dimensions > 1`?

jon-chuang:
I have managed to perform deallocation, however, now the buffer is calling `cl::sycl::invalid_parameter_error` upon allocation or copying.

include/CL/sycl/detail/buffer.hpp
```diff
@@ -130,6 +130,22 @@ class buffer_impl
               void* host_ptr,
               bool is_svm_ptr = false);
 
+  buffer_impl(
+    void* _buffer_pointer_source,
+    void* _host_memory_source,
```
illuhad: This is not really a good approach:

In general, you would expect a SYCL object to manage memory and bind memory lifetime to object lifetime using reference counting semantics.

With this implementation, when the last copy of the original buffer goes out of scope, the subbuffer will suddenly point to invalid memory.
When the subbuffer goes out of scope no cleanup will ever happen because memory is still managed by the original buffer.

You need to use the same `buffer_impl` as the orignal buffer.

illuhad: (This is also important for data dependency calculation, as this is done per `buffer_impl` object)

include/CL/sycl/buffer.hpp
jon-chuang: This is just a stopgap meant to support d=1

include/CL/sycl/detail/buffer.hpp
jon-chuang: I assumed that the contract was that the subbuffers are subservient to the parent buffer, so that when the parent buffer dies all the subbuffers do too. 

I suppose the subbuffers may live longer the parents if one passes their reference out of the scope.

jon-chuang: I don't completely understand what you are saying, let me think about it and understand the current implementation.

jon-chuang: In the new implementation, will only the memory region owned by the subbuffer continue to live after the parent dies, or the entire buffer?

illuhad: Since we cannot partially deallocate an allocated memory block on all APIs that I am aware of, allocation and deallocation will always have to be done for the entire buffer.

jon-chuang:
I don't think I will be able to work on this pull request as the new runtime will seemingly solve my problems, I look forward to it.

PR #199, opened 2020-01-28
Apply hotfix to allow sycl::min() and sycl::max() compile correctly on CUDA

illuhad:
Currently it seems that `sycl::min()` and `sycl::max()` are broken on CUDA. While the exact root cause of the issue is not yet entirely clear to me, this change seems to fix it.
CPU backend has worked before and is unaffected by this change. Not sure about ROCm, but likely it's the same as CUDA.


psalz:
Works for me!

PR #200, opened 2020-02-03
Fix typo in CMake module

psalz:
The typo caused the math expression to expand to " + 1" for the first available backend, which appears to evaluate to 1 in newer versions of CMake, however causes an error for CMake 3.5.2.


illuhad:
Thanks!

PR #205, opened 2020-02-19
add libraries to ROCM_LINK_LINE

DieGoldeneEnte:
Add some libraries required to link hipSYCL programs. This is the same line as used in the install/scripts/install-hipsycl.sh script except `-lamd_comgr -lamd_hostcall -latmi_runtime`, because they require additional libraries, which are not required for the minimal vector add example in the README.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -145,7 +145,7 @@ add_subdirectory(src)
 
 set(DEFAULT_GPU_ARCH "" CACHE STRING "Optional: Default GPU architecture to compile for when targeting GPUs (e.g.: sm_60 or gfx900)")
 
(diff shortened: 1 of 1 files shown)


illuhad:
As discussed, please confirm that this library selection corresponds to the libraries provided by the AMD binary packages so that we don't break hipSYCL by default for users who are using AMD packages.

DieGoldeneEnte:
I checked the package and can confirm the presence of `libhsa-runtime64.s`, `libmcwamp.so` and `libhc_am.so` in the searched directories. A package for `amd_comgr` is also present in the rocm repository (and installed when following the instructions [here](https://rocm.github.io/ROCmInstall.html#ROCm-Installation)). To get `amd_hostcall` and `atmi_runtime` the aomp package needs to be installed (which is not in the rocm repository).

Since adding `-lamd_comgr` shouldn't be a problem, I added it as well.

illuhad:
Wonderful, thank you!

PR #207, opened 2020-02-25
Fix kernel outlining for forward declared functions

illuhad:
Issue found through discussion with @AndreiCNitu. Consider the following code, which currently does not compile:

```
void forward_declared2();
void forward_declared1();

...

q.submit([&](cl::sycl::handler& cgh){
  cgh.single_task<class test>([=](){
    forward_declared1();
  });  
});
...

void forward_declared1(){forward_declared2();}
void forward_declared2(){}
```

The hipSYCL clang plugin currently fails to correctly outline such kernels. This is because the call expression in the kernel then references only the forward declared function, which causes 
* the `CompleteCallSet` to miss the portion of the call graph that is invoked by the forward declared function
* the clang plugin not to emit the forward declared function to device code, causing linker errors.

This PR fixes the issue by making the `CompleteCallSet` attempt to find the definition for each function that it investigates, and if possible, investigate the definition instead of the declaration.

### DIFF
--- a/src/hipsycl_clang_plugin/Frontend.hpp
+++ b/src/hipsycl_clang_plugin/Frontend.hpp
@@ -34,6 +34,8 @@
 #include <regex>
 #include <sstream>
 
+#include "clang/AST/Decl.h"
+#include "clang/Basic/LLVM.h"
 #include "clang/Frontend/FrontendPluginRegistry.h"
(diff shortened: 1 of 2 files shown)


AndreiCNitu:
Thanks!

Issue #174, opened 2019-12-02
atomic::compare_exchange_strong()


mrzv:
I noticed that `atomic::compare_exchange_strong()` is marked as unimplemented. Doesn't the following implementation work, or am I missing something?
```
template<class T, cl::sycl::access::address_space addressSpace>
HIPSYCL_KERNEL_TARGET
bool
cl::sycl::atomic<T,addressSpace>::
compare_exchange_strong(T &expected, T desired,
                        memory_order successMemoryOrder,
                        memory_order failMemoryOrder) volatile
{
#ifdef __HIPSYCL_DEVICE_CALLABLE__
    T old = expected;
    expected = atomicCAS(_ptr, expected, desired);
    return old == expected;
#else
    return detail::invalid_host_call_dummy_return<T>();
#endif
}   
```

illuhad:
I think this is correct. I'm not sure if there is maybe another subtle reason why I left this unimplemented (e.g. small differences in the required memory model), but as a start this looks fine.

Alkeryn:
I'm really waiting on that one as it is kind of the only way to add atomic operation for unsuported stuff.

illuhad:
@Alkeryn please feel free to submit a PR ;)

illuhad:
Fixed by PR #209

PR #209, opened 2020-02-28
Add implementation of atomics on CPU using __atomic

### LINKED ISSUE #172: atomics on CPU
I have the following simple atomic counter example that compiles and runs fine for CUDA, but doesn't compile on the CPU.
```
#include <iostream>
#include <CL/sycl.hpp>

namespace s = cl::sycl;

int main()
{
    s::queue q;

    int counter = 0;
    {
        s::buffer<int> counter_buf(&counter, 1);

        q.submit([&](cl::sycl::handler& cgh)
        {
            auto access_counter = counter_buf.get_access<cl::sycl::access::mode::read_write>(cgh);

            cgh.parallel_for<class atomic_increment>(s::range<1>(1 << 30), [=] (cl::sycl::id<1> tid)
            {
                s::atomic<int> atomic_counter { s::global_ptr<int> {&access_counter[0]} };
                atomic_counter.fetch_add(1);
            });
        });
    }

    std::cout << "Counter: " << counter << std::endl;
}
```

```
/opt/hipSYCL/CUDA/bin/syclcc-clang --hipsycl-gpu-arch=sm_52 -O3 atomic-counter.cpp -o atomic-counter-cuda
```
compiles and runs fine.

```
/opt/hipSYCL/CUDA/bin/syclcc-clang --hipsycl-platform=cpu -g atomic-counter.cpp -o atomic-counter-cpu
In file included from atomic-counter.cpp:3:
In file included from /opt/hipSYCL/CUDA/bin/../include/CL/sycl.hpp:58:
/opt/hipSYCL/CUDA/bin/../include/CL/sycl/atomic.hpp:103:12: error: use of undeclared identifier 'atomicAdd'
    return atomicAdd(_ptr, operand);
           ^
atomic-counter.cpp:23:32: note: in instantiation of function template specialization 'cl::sycl::atomic<int, cl::sycl::access::address_space::global_space>::fetch_add<int, nullptr>' requested here
                atomic_counter.fetch_add(1);
```

I know atomics are not fully supported (they are listed as a limitation in README), but other issues suggest that only minor features of atomics are missing. Is the above error by design, or is it something that can be easily fixed?

mrzv:
This is implementation of atomics on CPU using __atomic built-ins. For now, it's WIP for discussion. It's meant to fix #172.

### DIFF
--- a/include/CL/sycl/atomic.hpp
+++ b/include/CL/sycl/atomic.hpp
@@ -68,11 +68,33 @@ public:
   /// \todo unimplemented
   HIPSYCL_KERNEL_TARGET
   void store(T operand, memory_order memoryOrder =
-      memory_order::relaxed) volatile;
+      memory_order::relaxed) volatile
+  {
+#ifdef __HIPSYCL_DEVICE_CALLABLE__
+#ifdef SYCL_DEVICE_ONLY
+    return detail::invalid_host_call_dummy_return<T>();     // really should be invalid_device_call
+#else
+    return __atomic_store_n(_ptr, operand, __ATOMIC_RELAXED);
+#endif
+#else
+    return detail::invalid_host_call_dummy_return<T>();
+#endif
+  }
 
   /// \todo unimplemented
   HIPSYCL_KERNEL_TARGET
-  T load(memory_order memoryOrder = memory_order::relaxed) const volatile;
+  T load(memory_order memoryOrder = memory_order::relaxed) const volatile
+  {
+#ifdef __HIPSYCL_DEVICE_CALLABLE__
+#ifdef SYCL_DEVICE_ONLY
+    return detail::invalid_host_call_dummy_return<T>();     // really should be invalid_device_call
+#else
+    return __atomic_load_n(_ptr, operand, __ATOMIC_RELAXED);
+#endif
+#else
+    return detail::invalid_host_call_dummy_return<T>();
+#endif
+  }
 
   HIPSYCL_KERNEL_TARGET
   T exchange(T operand, memory_order memoryOrder =
--- a/include/CL/sycl/atomic.hpp
+++ b/include/CL/sycl/atomic.hpp
@@ -86,7 +86,6 @@ public:
   }
 
   /* Available only when: T != float */
-  /// \todo unimplemented
   HIPSYCL_KERNEL_TARGET
   bool compare_exchange_strong(T &expected, T desired,
                                memory_order successMemoryOrder = memory_order::relaxed,
@@ -97,7 +96,7 @@ public:
     expected = atomicCAS(_ptr, expected, desired);
     return old == expected;
 #else
-    return detail::invalid_host_call_dummy_return<T>();
+    return __atomic_compare_exchange_n(_ptr, &expected, desired, false, __ATOMIC_RELAXED, __ATOMIC_RELAXED);
 #endif
   }
 
--- a/include/CL/sycl/atomic.hpp
+++ b/include/CL/sycl/atomic.hpp
@@ -90,7 +90,16 @@ public:
   HIPSYCL_KERNEL_TARGET
   bool compare_exchange_strong(T &expected, T desired,
                                memory_order successMemoryOrder = memory_order::relaxed,
-                               memory_order failMemoryOrder = memory_order::relaxed) volatile;
+                               memory_order failMemoryOrder = memory_order::relaxed) volatile
+  {
+#ifdef __HIPSYCL_DEVICE_CALLABLE__
+    T old = expected;
+    expected = atomicCAS(_ptr, expected, desired);
+    return old == expected;
+#else
+    return detail::invalid_host_call_dummy_return<T>();
+#endif
+  }
 
   /* Available only when: T != float */
   template<class t = T,
--- a/include/CL/sycl/atomic.hpp
+++ b/include/CL/sycl/atomic.hpp
@@ -65,36 +65,14 @@ public:
                   "Invalid pointer type for atomic<>");
   }
 
-  /// \todo unimplemented
+  /// \todo unimplemented on GPU
   HIPSYCL_KERNEL_TARGET
   void store(T operand, memory_order memoryOrder =
-      memory_order::relaxed) volatile
-  {
-#ifdef __HIPSYCL_DEVICE_CALLABLE__
-#ifdef SYCL_DEVICE_ONLY
-    return detail::invalid_host_call_dummy_return<T>();     // really should be invalid_device_call
-#else
-    return __atomic_store_n(_ptr, operand, __ATOMIC_RELAXED);
-#endif
-#else
-    return detail::invalid_host_call_dummy_return<T>();
-#endif
-  }
+      memory_order::relaxed) volatile;
 
-  /// \todo unimplemented
+  /// \todo unimplemented on GPU
   HIPSYCL_KERNEL_TARGET
-  T load(memory_order memoryOrder = memory_order::relaxed) const volatile
-  {
-#ifdef __HIPSYCL_DEVICE_CALLABLE__
-#ifdef SYCL_DEVICE_ONLY
-    return detail::invalid_host_call_dummy_return<T>();     // really should be invalid_device_call
-#else
-    return __atomic_load_n(_ptr, operand, __ATOMIC_RELAXED);
-#endif
-#else
-    return detail::invalid_host_call_dummy_return<T>();
-#endif
-  }
+  T load(memory_order memoryOrder = memory_order::relaxed) const volatile;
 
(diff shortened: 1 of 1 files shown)


mrzv:
@jeffhammond What's the right way to implement `fetch_min` and `fetch_max`? There is not an equivalent `__atomic`. The only way I can think of is via `compare_exchange`. Is there a better way?

mrzv:
@jeffhammond Also the memory order on success and failure for compare_exchange_strong deserves attention. I just picked relaxed without knowing what I'm doing.

mrzv:
This PR includes implementation listed in #174.

include/CL/sycl/atomic.hpp
```diff
@@ -81,16 +81,24 @@ class atomic {
 #ifdef __HIPSYCL_DEVICE_CALLABLE__
     return atomicExch(_ptr, operand);
 #else
-    return detail::invalid_host_call_dummy_return<T>();
+    return __atomic_exchange_n(_ptr, operand, __ATOMIC_RELAXED);
```
illuhad: I think the `#ifdef` code paths should be structured slightly differently. There are currently several compilation paths when compiling with hipSYCL:

* GPU compilation pass
* CPU compilation pass of the CUDA/ROCm backends. This is where we currently have those `invalid_host_call_dummy_return()` calls. Because using CPU and GPU simultaneously is not yet supported (due to limitations in the runtime which will be fixed by the new runtime), these dummy calls should never be actually executed and are in principle something like a `assert(false)`.
* CPU compilation pass when compiling for the CPU backend.

Now, the `__HIPSYCL_DEVICE_CALLABLE__` macro is defined
* by CUDA/ROCm backend if currently in the GPU compilation pass
* by the CPU backend because hipCPU defines many of those HIP backends.
In other words, it's defined when HIP intrinsics like atomicExch are available.

This means that the code here will never execute in practice because it will be compiled for the host compilation pass for the CUDA/ROCm backend.

Instead, we want to have those atomic functions run whenever we compile for CPU. Solution: Check with the `SYCL_DEVICE_ONLY` macro if we are in a GPU compilation pass:
```
#ifdef SYCL_DEVICE_ONLY
  return atomicExch(...);
#else
  return __atomic_exchange_n(...)
#endif
```

mrzv:
One more thing that's unclear: `load` and `store` are not implemented. There are obvious `__atomic` functions to use, but I don't see what the right function is in HIP. Any ideas?

include/CL/sycl/atomic.hpp
mrzv: How about now?

illuhad: Fine now, we could even drop the surrounding `#ifdef __HIPSYCL_DEVICE_CALLABLE__` to make the code simpler. There are enough other places that would prevent running CPU code in kernels when compiling for GPU. In the future I want to make the CPU backend always available (even when compiling for GPU), so then we would have to remove this anyway.

illuhad:
@mrzv I think you may have discovered the reason why those functions were unimplemented ;) I would suggest to only implement those on CPU for now.
I will investigate how to implement those functions on GPU. If I remember correctly, the spec was very unclear for those anyway because it was not clear what values for the template parameter `T` were allowed.

mrzv:
@illuhad Ok, done, I think.

include/CL/sycl/atomic.hpp
```diff
@@ -68,11 +68,33 @@ class atomic {
   /// \todo unimplemented
   HIPSYCL_KERNEL_TARGET
   void store(T operand, memory_order memoryOrder =
-      memory_order::relaxed) volatile;
+      memory_order::relaxed) volatile
+  {
+#ifdef __HIPSYCL_DEVICE_CALLABLE__
+#ifdef SYCL_DEVICE_ONLY
+    return detail::invalid_host_call_dummy_return<T>();     // really should be invalid_device_call
```
illuhad: Better not call this on device. This function calls `assert()` which may not work on all GPUs (I think newer NVIDIA GPUs/newer CUDA versions may be able to do that, not sure about AMD).

You could instead surround the entire function with `#ifdef HIPSYCL_PLATFORM_CPU` to make it only visible when compiling for CPU.

mrzv: That would require moving it outside the class declaration, right?

mrzv: Would a `static_assert` be better?

mrzv: Actually, never mind, it's not a template function, `static_assert` would get triggered every time.

illuhad: I think an out-of-line definition outside of the class would make sense. In this case the declaration could be visible both on CPU and GPU, but the definition would be invisible on GPU, leading to a linker error.

mrzv: How about now?

illuhad: looks good, thank you!

jeffhammond:
> @jeffhammond What's the right way to implement `fetch_min` and `fetch_max`? There is not an equivalent `__atomic`. The only way I can think of is via `compare_exchange`. Is there a better way?

Funny timing - I spent a bunch of time on this recently. The right implementation does CAS but only after an early-exit check for when no update is required.  This is a fundamental difference between min/max and arithmetic ops like add (all credit to @jcownie-intel for teaching me this via https://github.com/LLNL/RAJA/issues/447).

If example code helps, see https://github.com/jeffhammond/HPCInfo/blob/master/atomics/min-max/gcc-min-max.c.  I don't claim this code is perfect, so please verify correctness before using.

mrzv:
@jeffhammond Why do you issue an extra `__atomic_load_n` inside your do-while loop? Doesn't `__atomic_compare_exchange_n` fill `old` in case of failure?

mrzv:
@jeffhammond Also, shouldn't the `if (old > value)` check live inside the do-while loop?

mrzv:
@jeffhammond I pushed an implementation of `fetch_min` and `fetch_max`. Can you take a look and critique?

jeffhammond:
> @jeffhammond Why do you issue an extra __atomic_load_n inside your do-while loop? Doesn't __atomic_compare_exchange_n fill old in case of failure?
> 
Good point. It shouldn’t be there if the CAS refreshes it.

I meant it when I said my code wasn’t perfect! 😂

include/CL/sycl/atomic.hpp
```diff
@@ -79,18 +79,34 @@ class atomic {
       memory_order::relaxed) volatile
   {
 #ifdef __HIPSYCL_DEVICE_CALLABLE__
+#ifdef SYCL_DEVICE_ONLY
     return atomicExch(_ptr, operand);
+#else
+    return __atomic_exchange_n(_ptr, operand, __ATOMIC_RELAXED);
```
illuhad: It's fine if we only support relaxed memory order for now, but could we maybe at least add `assert(memoryOrder == memory_order::relaxed)` on CPU? This way the user would notice that what he does is unsupported rather than the code just using a potentially different memory order.

mrzv: Sure. I'll fix this. Currently, the only `memory_order` that exists is `relaxed`, so the assertion is vacuous, but I get the point about making it future-proof.

illuhad:
Thank you! We also need test cases for the `atomic` class, but I can also do this later. Unless you want to, of course ;)

mrzv:
I agree that we desperately need test cases; I've been testing this stuff in a one-off project, which is a bit of a pain. The trouble is I can't figure out your test framework. It's setup as a separate CMake project, and I haven't figured out how to get it to work. Why not just setup `make test` to run the tests?

illuhad:
I think this probably has historic reasons.. It should be relatively easy to use though:
```
cmake -DSYCLCC=/path/to/syclcc-clang /path/to/tests
HIPSYCL_PLATFORM=cpu|cuda|rocm HIPSYCL_GPU_ARCH=<your-gpu-arch> make
```
This was written before we had our cmake integration, we should probably update it to use our current cmake integration, then we could select the target platform etc from cmake.

mrzv:
Ok, this works for me. But let me express my enthusiastic support for updating cmake integration. (I'm afraid I'm not qualified to create a pull request for this one.)

jeffhammond:
I will test it in isolation and then hopefully write some SYCL atomic tests, but I don't see how the atomic_min/max implementation can be correct.  I'm happy to be wrong, but see my comment above for details.

In any case, I will be happy to provide the tests and fixes once I get a few hours to work on this.

mrzv:
@jeffhammond Can you say a little more? What do you think is wrong with it?

jeffhammond:
I think I am just stupid.  Let me do some tests and then I'll follow-up if necessary.

PR #210, opened 2020-02-28
Add accessor::operator[] for atomic mode

mrzv:
While I'm fixing things that are getting in my way, here's `operator[]` in atomic mode.

### DIFF
--- a/include/CL/sycl/accessor.hpp
+++ b/include/CL/sycl/accessor.hpp
@@ -608,6 +608,9 @@ accessMode == access::mode::discard_read_write) && dimensions == 1) */
       return atomic<dataT, access::address_space::global_space> { global_ptr<dataT>(_ptr + detail::linear_id<dimensions>::get(index, _buffer_range)) };
   }
 
+  template<int D = dimensions,
+           access::mode M = accessmode,
+           typename = std::enable_if_t<(D == 0) && (M == access::mode::atomic)>>
   atomic<dataT, access::address_space::global_space> operator[](size_t index) const
   {
       return atomic<dataT, access::address_space::global_space> { global_ptr<dataT>(_ptr + index) };
--- a/include/CL/sycl/accessor.hpp
+++ b/include/CL/sycl/accessor.hpp
@@ -593,6 +593,7 @@ accessMode == access::mode::discard_read_write) && dimensions == 1) */
   template<int D = dimensions,
            access::mode M = accessmode,
            typename = std::enable_if_t<M == access::mode::atomic && D == 0>>
+  HIPSYCL_UNIVERSAL_TARGET
   operator atomic<dataT, access::address_space::global_space> () const
   {
       return atomic<dataT, access::address_space::global_space> { global_ptr<dataT>(_ptr) };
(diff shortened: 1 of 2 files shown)


mrzv:
This doesn't do the right thing. I.e., it's WIP for now.

mrzv:
Ok, that was just a minor problem. I think it's Ok now.

illuhad:
Thank you, looks good on first glance, I'll play with your PRs a bit more on monday when I'm back at work and get paid to do so ;)

include/CL/sycl/accessor.hpp
```diff
@@ -589,14 +590,31 @@ accessMode == access::mode::discard_read_write) && dimensions == 1) */
 
 
   /* Available only when: accessMode == access::mode::atomic && dimensions == 0*/
-  //operator atomic<dataT, access::address_space::global_space> () const;
+  template<int D = dimensions,
+           access::mode M = accessmode,
+           typename = std::enable_if_t<M == access::mode::atomic && D == 0>>
+  operator atomic<dataT, access::address_space::global_space> () const
```
illuhad: Can you please mark these functions as `HIPSYCL_UNIVERSAL_TARGET`, like the other member functions? This causes it to become available both inside kernels and in host code (`__host__ __device__` in CUDA terminology)

include/CL/sycl/accessor.hpp
```diff
@@ -589,14 +590,34 @@ accessMode == access::mode::discard_read_write) && dimensions == 1) */
...
-  //atomic<dataT, access::address_space::global_space> operator[](
-  //    id<dimensions> index) const;
+  template<int D = dimensions,
+           access::mode M = accessmode,
+           typename = std::enable_if_t<(D > 0) && (M == access::mode::atomic)>>
+  HIPSYCL_UNIVERSAL_TARGET
+  atomic<dataT, access::address_space::global_space> operator[](
+      id<dimensions> index) const
+  {
+      return atomic<dataT, access::address_space::global_space> { global_ptr<dataT>(_ptr + detail::linear_id<dimensions>::get(index, _buffer_range)) };
```
illuhad: Final nitpick: Can you introduce a linebreak here? This line is twice as long as the other lines in this file ;) Also: We use an indentation of 2 spaces.

mrzv: Done, I think.

include/CL/sycl/accessor.hpp
```diff
@@ -589,14 +590,35 @@ accessMode == access::mode::discard_read_write) && dimensions == 1) */
...
+  {
+    return atomic<dataT, access::address_space::global_space>
+             { global_ptr<dataT>(_ptr + detail::linear_id<dimensions>::get(index, _buffer_range)) };
+  }
 
-  //atomic<dataT, access::address_space::global_space> operator[](
-  //    size_t index) const;
+  template<int D = dimensions,
+           access::mode M = accessmode,
+           typename = std::enable_if_t<(D == 0) && (M == access::mode::atomic)>>
```
illuhad: After thinking about a bit, I think there's a typo here. I think this should be enabled if `D==1` and not if `D==0`, because indexing 0-dimensional data doesn't make any sense.

mrzv: Good catch. And the other comparison should be `D > 1`. This should be fixed now.

illuhad: Why `D>1` for the other overload? A SYCL id can be one-dimensional, so I think `D>0` was correct. Am I missing something?

mrzv: Ah, good point. Fixed.

illuhad:
Thank you :)

PR #214, opened 2020-03-05
Silence warning in HIP headers (when using -Wall)

psalz:
This silences a warning emitted from HIP headers, when compiling with `-Wall`. This is only required until we update the HIP submodule to include https://github.com/ROCm-Developer-Tools/HIP/pull/1497.

Not sure if there are any breaking changes in HIP that we have to account for, also I cannot test on ROCm - otherwise I'm of course all for updating HIP :)


illuhad:
Thank you. On ROCm we always use the HIP headers provided by the ROCm installation for compatibility reasons (instead of our submodule). If CUDA compiles with the new HIP headers, there should be no issue with updating our submodule. On ROCm, since we cannot control the exact HIP version, we will probably still need this PR though.

PR #215, opened 2020-03-05
Fix the type check in atomic::fetch_*

mrzv:


illuhad:
Thank you, this seems to work fine!

PR #216, opened 2020-03-06
Restructure code into runtime, sycl, compiler subdirs and corresponding namespaces, implement Xilinx proposal for multiple entrypoints

illuhad:
This PR is a major restructuring of the hipSYCL code base. The motivation is that, with the new runtime, we will have a very strict separation between code from the SYCL interface headers and the SYCL runtime. I would also like to underline this via namespaces and the directory structure. Therefore, this PR moves to the following code structure:
| Component | Header files | Source files | Namespace |
|------------------|-------------------|------------------|------------------|
| SYCL Interface | `include/hipSYCL/sycl` | `src/sycl` | `hipsycl::sycl` |
| SYCL Runtime | `include/hipSYCL/runtime` | `src/runtime` | `hipsycl::rt` |
| Compiler (clang plugin) | `include/hipSYCL/compiler` | `src/compiler` | `hipsycl::compiler` |

This will also give me a good structure to develop the new runtime against.

Note that I have not separated all the existing runtime code from the SYCL interface in this PR; this will be done when the new runtime comes where I will need to remove the old runtime bits anyway.

Because the SYCL headers are now in the generic namespace `hipsycl::sycl`, this has allowed me to implement the Xilinx proposal to generalize namespaces and allow a SYCL program to use multiple SYCL implementations simultaneously (see here: https://github.com/triSYCL/triSYCL/pull/249).
In short:
* `#include <CL/sycl.hpp>` exposes the SYCL api in the namespace `cl::sycl::`
* `#include <SYCL/sycl.hpp>` exposes the SYCL api in the namespace `sycl::`
* `#include <hipSYCL/sycl.hpp>` exposes the SYCL api in the namespace `hipsycl::`.
Note that this feature is mostly a side effect of the larger reshuffling of code, I have not yet tested it fully :-)

This is still marked as WIP because I would like to look at this some more to make sure it's good to go.


keryell:
That sounds like a very good idea! ;-)
Soon we will be able to have 5 different SYCL implementations running at the same time and handling various accelerators! :-)

illuhad:
@keryell 5 Implementations? Is sycl-gtx also going to implement this (I had the impression the project was mostly dormant), or is there a SYCL implementation that I am not aware of?

keryell:
> @keryell 5 Implementations? Is sycl-gtx also going to implement this (I had the impression the project was mostly dormant), or is there a SYCL implementation that I am not aware of?

OK, let's say that 5 was an optimistic upper bound estimate today. :-)

psalz:
A possibly unintended consequence of this change is that it's no longer possible to forward declare SYCL types within the `cl::sycl` namespace, if they are actually defined within `hipsycl::sycl`.

See: https://www.compiler-explorer.com/z/4dEQkL

illuhad:
What would be the use case for this? Why can't the user just `#include <CL/sycl.hpp>`?

triSYCL implements this exactly the same way, and will also be affected by this. I don't see a lot options around this, if SYCL should become more general than OpenCL as per the generalization proposal (making the `cl::` namespace deprecated) while still retaining some degree of backwards compatibility.

psalz:
I've been forward declaring SYCL types in Celerity headers whenever I don't need the actual implementation, for example when I'm only using pointers to SYCL types -- to reduce compile times.

keryell:
@psalz for the same reason I think you could not forward declare `std::string`...

Perhaps this is something to add to SYCL spec... Do not do this. :-)

You are playing here with some implementation details...

psalz:
I actually did some googling on whether forward declaring library types was considered bad practice, and I doesn't seem like there is any broad consensus on it. However apparently the C++ standard [explicitly forbids it](https://github.com/cplusplus/draft/blob/61763ed574e73949323008b8c8479c793408ea06/source/lib-intro.tex#L2568), so maybe it really is something worth considering adding to the spec.

It's not a big deal, just something I noticed when trying to compile Celerity against the newest version of hipSYCL.

keryell:
@psalz since you seem at speed on this subject, you can send a PR on the SYCL specification. :-)

PR #218, opened 2020-03-13
Update SYCL ecosystem picture

keryell:
Add CUDA backend for Intel SYCL.
Precise that ComputeCPP generates OpenCL + PTX for Nvidia GPU.
Add TBB for triSYCL on CPU.


illuhad:
Thanks, Ronan!

PR #224, opened 2020-03-18
Implement atomic::{load,store} using compare_exchange_strong

### LINKED ISSUE #213: Atomic load and store on GPU
This issue is to make sure we don't forget that `atomic::load()` and `atomic::store()` are not yet implemented on the GPUs. I have a hack that implements them using `compare_exchange_strong`, but that's not a good long-term solution.

mrzv:
Not suggesting to actually merge this, but this does implement `load`/`store` via CAS, so it would fix #213.


illuhad:
Thank you! Is there any reason not to merge that I'm not seeing? To me it seems a potentially inefficient solution using CAS is clearly better than no implementation at all. We can imlrove on it later on if we find a better solution.

mrzv:
It's a horrendously inefficient solution. No other reason. I'm a little worried that we'll mentally check this issue off, only to stumble later into some terrible performance.

I unfortunately don't understand the device memory model, but from the [CUDA solution](https://stackoverflow.com/questions/32341081/how-to-have-atomic-load-in-cuda) I linked in #213, and what the docs say about [HIP and __threadfence](https://github.com/ROCm-Developer-Tools/HIP/blob/master/docs/markdown/hip_kernel_language.md#memory-fence-instructions), it sounds like there is a simple clean solution that combines those two.

illuhad:
This has been superseded by PR #584, which has introduced atomic load/store support using compiler builtins.

PR #229, opened 2020-03-22
Fixed the nbody bruteforce example

sbalint98:
This PR should fix the nbody example. Now it compiles without warnings.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,12 +1,8 @@
-
-find_program(SYCLCC NAMES syclcc CACHE STRING)
-if(SYCLCC MATCHES SYCLCC-NOTFOUND)
-  message(SEND_ERROR "hipSYCL syclcc compiler not found, exiting.")
-  return()
-endif()
-
-set(CMAKE_CXX_COMPILER ${SYCLCC})
 project(hipsycl-examples)
+
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR}/../cmake)
+
+find_package(hipSYCL CONFIG REQUIRED)
 
 if(NOT CMAKE_BUILD_TYPE)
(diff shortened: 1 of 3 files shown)


illuhad:
Thanks, great to have this example finally fixed :)

examples/bruteforce_nbody/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
 add_executable(bruteforce_nbody bruteforce_nbody.cpp)
-target_link_libraries(bruteforce_nbody hipSYCL)
+target_link_libraries(bruteforce_nbody)
```
illuhad: You can remove this line entirely - in the current formulation, it just says, for target `bruteforce_nbody` link with libraries in `<empty set>`. If we are at it, maybe we should think about doing a proper cmake integration using `find_package(hipSYCL)` and `add_sycl_to_target()`. I can give you some examples if you are interested.

sbalint98: Thank you, I'd like to do a proper CMake integration, so can you give me some examples?

examples/CMakeLists.txt
```diff
@@ -1,13 +1,10 @@
...
-  message(SEND_ERROR "hipSYCL syclcc compiler not found, exiting.")
-  return()
-endif()
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR}/../cmake)
 
-set(CMAKE_CXX_COMPILER ${SYCLCC})
-project(hipsycl-examples)
+find_package(hipSYCL CONFIG REQUIRED)
 
+set(CMAKE_CXX_COMPILER ${SYCLCC})
```
illuhad: This should not be necessary. `add_sycl_to_target` should take care of this.

sbalint98: got it!

illuhad:
Thank you!

Issue #223, opened 2020-03-18
PRK nstream compilation crashes with CUDA platform


jeffhammond:
I think you know where the PRK source code is.  I am using 15bb69ebd6ab7c0f3b85bb2622e3a1d0e2f904a2.

There is no issue using `--hipsycl-platform=cpu`.  Attempts with `sm_35` and `sm_50` failed similarly.

```sh
jrhammon@klondike:~/prk-repo/Cxx11$ make sycl
/opt/hipSYCL/bin/syclcc-clang -DPRKVERSION="2.16" -std=c++17 -O3 -DHIPSYCL --hipsycl-platform=cuda --hipsycl-gpu-arch=sm_60 -Wl,-rpath=/opt/hipSYCL/llvm/lib -DUSE_2D_INDEXING=0 nstream-sycl.cc -o nstream-sycl
Stack dump:
[...]

PR #230, opened 2020-03-23
Reworked packaging system, added option for clang9

sbalint98:
Update for the packaging scripts, the base package is now created with LLVM9 for stability reasons #223 .


devops/repos/common/init.sh
```diff
@@ -1,7 +1,10 @@
 #!/bin/bash
 
-HIPSYCL_REPO_STAGE_DIR=${HIPSYCL_REPO_STAGE_DIR:-./stage}
+REPO_STAGE_DIR=${REPO_STAGE_DIR:-./stage}
```
illuhad: Can you comment on why this variable needed to be renamed? To be it seems good practice to have our environment variables namespaced using the `HIPSYCL_` prefix.

devops/repos/update_repo.sh
```diff
@@ -0,0 +1,50 @@
...
+CONTAINER_DIR=$SCRIPT_DIR_ABS/$CONTAINER_DIR_SUFFIX
+[ "$LLVM_9_BUILD" = "ON" ] && CONTAINER_DIR=$SCRIPT_DIR_ABS/clang9-$CONTAINER_DIR_SUFFIX
+
+export CONTAINER_DIR
+export BUILD_HIPSYCL
+export BUILD_CUDA
+export BUILD_ROCM
+export BUILD_BASE
+export NO_BUILD
+export LLVM_9_BUILD
```
illuhad: Rather than querying for LLVM 9 explicitly, wouldn't it be more beneficial to have the llvm branch as an environment variable? Otherwise, we may have to change the code here once LLVM 9 is obsolete and the discussion is whether to use LLVM 10 or 11.

install/scripts/install-llvm.sh
```diff
@@ -1,11 +1,26 @@
...
 set -e
 BUILD_DIR=$HOME/git/llvm-vanilla
 rm -rf $BUILD_DIR
 
 #git clone -b release/9.x https://github.com/llvm/llvm-project $BUILD_DIR
-git clone -b release/10.x https://github.com/llvm/llvm-project $BUILD_DIR
+if [ "$LLVM_9_BUILD" = "OFF" ]
+then
+	echo "Cloning LLVM 10"	
+	git clone -b release/10.x https://github.com/llvm/llvm-project $BUILD_DIR
```
illuhad: As I mentioned, why not have the branch as environment variable?

devops/repos/update_repo.sh
sbalint98: Thank you, This is a good point. I will change it to be so.

devops/repos/common/init.sh
sbalint98: I renamed it because this folder might contain the base and the rocm packages as well. Therefore I thought naming it `HIPSYCL_` would be misleading in a scenario where the rocm and base packages are put in a different stage directory in order to carry out further testing on them. However

illuhad:
Is this now ready for final review?

sbalint98:
yes it is :)

install/scripts/install-hipsycl.sh
```diff
@@ -15,6 +17,19 @@ git clone --recurse-submodules -b $HIPSYCL_REPO_BRANCH https://github.com/$HIPSY
...
+	release/9.x)
+		LLVM_INCLUDE_PATH=$INSTALL_PREFIX/llvm/lib/clang/9.0.1/include
+		;;
+	release/10.x)
+		LLVM_INCLUDE_PATH=$INSTALL_PREFIX/llvm/lib/clang/10.0.1/include 
+		;;
+	*)
+		echo "LLVM version $HIPSYCL_PKG_LLVM_REPO_BRANCH is not supported"
+		exit -1
+		;;
```
illuhad, line 30: I wonder if we could improve this in the future such that we only have one set of variables for the clang version e.g. $MAJOR.$MINOR.$PATCH, such that it always includes `$INSTALL_PREFIX/llvm/lib/clang/$MAJOR.$MINOR.$PATH/include ` and pulls from `release/$MAJOR.x` or something similar. Or we just have a variable for the branch and a variable for the version that goes into the include path.

For now this case statement is good enough, but it's only going to get longer in the future ;)

install/scripts/install-llvm.sh
```diff
@@ -1,11 +1,32 @@
...
 
-#git clone -b release/9.x https://github.com/llvm/llvm-project $BUILD_DIR
-git clone -b release/10.x https://github.com/llvm/llvm-project $BUILD_DIR
+
+case $HIPSYCL_PKG_LLVM_REPO_BRANCH in
+
+	release/9.x)
+		echo "Cloning LLVM $HIPSYCL_PKG_LLVM_REPO_BRANCH"	
+		git clone -b $HIPSYCL_PKG_LLVM_REPO_BRANCH https://github.com/llvm/llvm-project $BUILD_DIR
+		sed -i 's/CHECK_SIZE_AND_OFFSET(ipc_perm, mode);//g' $BUILD_DIR/compiler-rt/lib/sanitizer_common/sanitizer_platform_limits_posix.cc
```
illuhad, line 15: This is only necessary for clang 9? Wouldn't it be easier to just clone in any case, and then, if we are on clang 9, additionally apply the patch? This would simplify the code compared to a switch statement and we wouldn't have to modify the code if we want to transition to a new clang version.

install/scripts/install-rocm.sh
```diff
@@ -6,6 +6,9 @@ rm -rf $BUILD_DIR
 
 #git clone -b hipsycl-0.8 https://github.com/illuhad/aomp $BUILD_DIR/aomp
 git clone https://github.com/ROCm-Developer-Tools/aomp $BUILD_DIR/aomp
+cd $BUILD_DIR/aomp
+git fetch --all
+git checkout rel_0.7-7
```
illuhad, line 11: We also may want to have a variable for this in the future.

install/scripts/rebuild-base-images.sh
```diff
@@ -1,7 +1,25 @@
...
+cp $HIPSYCL_PKG_SCRIPT_DIR/*.def /tmp/hipsycl-pkg-builder
+cp $HIPSYCL_PKG_SCRIPT_DIR/*.sh /tmp/hipsycl-pkg-builder
+cd /tmp/hipsycl-pkg-builder
+
+echo $HIPSYCL_PKG_CONTAINER_DIR
+
+# Workaround for exposing variable inside singularity post script
+sed -i "/%post/a export\ HIPSYCL_PKG_LLVM_REPO_BRANCH=$HIPSYCL_PKG_LLVM_REPO_BRANCH" base-ubuntu-18.04.def
+sed -i "/%post/a export\ HIPSYCL_PKG_LLVM_REPO_BRANCH=$HIPSYCL_PKG_LLVM_REPO_BRANCH" base-archlinux-rolling.def
+sed -i "/%post/a export\ HIPSYCL_PKG_LLVM_REPO_BRANCH=$HIPSYCL_PKG_LLVM_REPO_BRANCH" base-centos-7.def
```
illuhad, line 16: Does `SINGULARITYENV_XYZ` not work to have environment variable `$XYZ` in the post script? Or is this only available after build time?

install/scripts/install-llvm.sh
sbalint98: You're right, I'll change that

install/scripts/rebuild-base-images.sh
sbalint98: As far as I know, it will only be available after build time, unfortunately...

install/scripts/install-rocm.sh
sbalint98: Sure thing, I can add this to this commit right away. Is there a reason why we would want this done in a separate commit?

install/scripts/install-hipsycl.sh
sbalint98: You're right, but If you agree I would change this right now, so we don't have to touch this code soon again

illuhad: It's fine if you want to do this change now, I just wanted to give you the option to get some closure with this PR for now since it has already been open for quite some time ;)

install/scripts/install-rocm.sh
illuhad: You can do this now if you prefer - see my comment above.

install/scripts/rebuild-base-images.sh
illuhad: I wonder if there's a clean solution.. This all sounds very hacky to me.

sbalint98: I am sorry, you were right with the SINGULARITYENV_XYZ variable one can expose vars for the postscript as well

illuhad:
Thank you!

PR #234, opened 2020-04-09
Add HIPSYCL_PLATFORM_ROCM macro

illuhad:
define `HIPSYCL_PLATFORM_ROCM` when compiling for ROCm. This allows us to deprecate `HIPSYCL_PLATFORM_HCC` in the future since hcc itself is deprecated (any not supported anymore by hipSYCL anyway)

### DIFF
--- a/doc/macros.md
+++ b/doc/macros.md
@@ -5,7 +5,7 @@ Most of these are managed by `detail/backend/backend.hpp`.
 
 * `__HIPSYCL__` - defined if compiling with hipSYCL
 * `HIPSYCL_PLATFORM_CUDA` - defined if compiling for CUDA
-* `HIPSYCL_PLATFORM_HCC` - defined if compiling for ROCm
+* `HIPSYCL_PLATFORM_HCC` (**deprecated**) - defined if compiling for ROCm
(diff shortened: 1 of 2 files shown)


fodinabor:
Why don't you mark it deprecated immediately?

illuhad:
@fodinabor Good point.. I guess I just didn't think that far, I simply wanted to have the `HIPSYCL_PLATFORM_ROCM` macro ;)

fodinabor:
Markung something deprecated mostly tells people that this is not the recommended thing to use...
Aaand as HCC itself is deprecated, I consider it a logical step to not encourage people to continue using the HCC macro, especially not for new projects.
Not that it changes much, but I don't see why new projects should use this macro, and a deprecation mark will at least prevent (some) new users from using it ;-)

illuhad:
> Marking something deprecated mostly tells people that this is not the recommended thing to use...
And as HCC itself is deprecated, I consider it a logical step to not encourage people to continue using the HCC macro, especially not for new projects.

I agree. I have marked `HIPSYCL_PLATFORM_HCC` as deprecated in the macro documentation.

PR #236, opened 2020-04-22
Initial implementation of stream class

illuhad:
This adds the initial implementation of the stream class using `printf()`.

On ROCm, I think this requires at least clang 10. Need to do some more testing.

Of the stream manipulators, only `endl` is supported for now.

### DIFF
--- a/include/hipSYCL/sycl/stream.hpp
+++ b/include/hipSYCL/sycl/stream.hpp
@@ -1,0 +1,283 @@
+/*
+ * This file is part of hipSYCL, a SYCL implementation based on CUDA/HIP
+ *
+ * Copyright (c) 2018-2020 Aksel Alpay
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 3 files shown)


sbalint98:
Would you like me to have a look at whether it requires clang10? There is also a new release of aomp which might be interesting.

illuhad:
@sbalint98 I know that it doesn't work on AMD with clang 9, but it would be great if you could verify if it works with clang 10. However, I need that urgently, so if you do this, please do it ideally today (latest tomorrow) as we need that code for the IWOCL/SYCLcon tutorial :)

illuhad:
Okay, it seems `printf()` on ROCm is still experimental when compiling with clang and probably requires very recent versions of clang and ROCm. 
I have therefore changed this PR such that it emits a warning when using stream `operator<<` on ROCm describing the experimental nature. `printf` on ROCm is now only actually invoked if the user explicitly defines `HIPSYCL_EXPERIMENTAL_ROCM_PRINTF`. Otherwise, `operator<<` will just be an empty function to get code to still compile.

PR #242, opened 2020-04-29
Add custom link line for CPU platform

nazavode:
This PR adds the ability to specify custom link flags to be used when targeting the `cpu` platform in the same way it is now supported both for `cuda` and `hip`. This allows to overcome situations where the targeted `clang` needs additional link flags, e.g.: when using `brew`-ed `llvm` on macOS:

```console
$ cmake -DCPU_LINK_LINE="-L/usr/local/opt/llvm/lib -Wl,-rpath,/usr/local/opt/llvm/lib" ..
```

### DIFF
--- a/bin/syclcc-clang
+++ b/bin/syclcc-clang
@@ -491,11 +491,6 @@ class clang_plugin_compiler:
       self._compiler_args = []
       # In non-bootstrap mode, automatically link hipSYCL
       if not config.is_bootstrap:
-        self._linker_args += [
-          "-rpath", hipsycl_library_path,
-          "-L"+hipsycl_library_path,
-          "-lhipSYCL_cuda"
-        ]
         self._compiler_args += [
           # Use the bundled HIP installation
           "-I" + os.path.join(config.hipsycl_installation_path, "include/hipSYCL/contrib")
@@ -509,16 +504,7 @@ class clang_plugin_compiler:
       self._target = "hip"
       
       self._linker_args = config.rocm_link_line
-
-      # In non-bootstrap mode, automatically link hipSYCL
-      if not config.is_bootstrap:
-        self._linker_args += [
-          "-rpath", hipsycl_library_path,
-          "-L"+hipsycl_library_path,
-          "-lhipSYCL_rocm"
-        ]
-      
-      
+            
       self._compiler_args = [
         "-x", "hip",
         "--cuda-gpu-arch=" + config.target_arch,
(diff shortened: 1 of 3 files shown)


bin/syclcc-clang
```diff
@@ -69,8 +69,7 @@ class config_file:
     if not key in self._data:
       return False
     if isinstance(self._data[key], str):
-      if(self._data[key].endswith("-NOTFOUND") or
-        self._data[key] == ""):
+      if(self._data[key].endswith("-NOTFOUND")):
```
nazavode, line 72: @illuhad I feel that this change could end up breaking a lot of stuff downstream, just consider it provisional: having an empty string as the value for a specific key in the configuration file could be considered correct to allow users explicitly set it to empty at configure time (e.g.: `-DCUDA_LINK_LINE=''`)?

bin/syclcc-clang
```diff
@@ -271,6 +273,9 @@ class syclcc_config:
         flag.commandline, flag.environment
     ))
 
+  def _get_cpu_substitution_vars(self):
+    return {}
```
illuhad, line 277: I feel we should at least expose `$HIPSYCL_PATH` and `$HIPSYCL_LIB_PATH` in analogy to the cuda and rocm substitution vars. This can be convenient for e.g. manually rpath-ing the hipSYCL library path.

bin/syclcc-clang
illuhad: hm.. for CUDA and ROCm an empty string cannot be correct because you always need to link against some CUDA/ROCm libraries.. But I think I see how you arrived at this change for the CPU backend, where no additional linker arguments are necessary. I suspect the assumption that empty string means missing key may be quite important in order to detect missing settings in syclcc and also to make working with environment variables more comfortable (not requiring to explicitly having to unset them when they are not needed anymore). 
I wonder if a better course of action might not be to also move the `-lhipSYCL_cpu`, `-lhipSYCL_rocm` etc to the link line. In that case, the link line would never be empty and the user also has control over how the SYCL runtime is linked, if desired.

nazavode: Yep, it felt a bit too much to relax that check. I tried to think about your suggestion to move `-lhipSYCL_<platform>` flags out of `syclcc-clang` and into the config file but I didn't come up with a decent solution yet: given the fact that `-lhipSYCL_<platform>` flags make sense only when _not_ in bootstrap mode, having them in the config file (that right now contains only mandatory flags instead, valid both in and out of bootstrap mode) would require a pass to purge those known flags from linker args when in bootstrap mode. If you think that something like this could be acceptable I would go down this path tho.

nazavode: I tried to give a chance to the second solution:

* all hipSYCL linker flags are now pulled in from the config file instead of being hardcoded in `syclcc-clang`
* a separate config file to be used only during bootstrap is created at configure time (`syclcc-bootstrap.json`), this file contains no hipSYCL linker flags and unfortunately the empty string check (reintroduced) passes only because CMake leaves an empty space in the configuration template string (a bit sketchy imho, but works)
* the former `syclcc.json` is unused during bootstrap and contains all the hipSYCL linker flags needed by the deployed `syclcc-clang`

nazavode:
Note to self: if this gets merged, fix spack package for tricky compiler mixes (e.g.: missing `libgomp.so` at runtime for non system-shipped toolchains, add toolchain `-rpath` via `-DCPU_LINK_LINE=...`).

illuhad:
So, in general, I think a solution that allows for the removal of the bootstrap process in the future is the right way to go, since this is kind of where I'm heading: The new runtime will only contain runtime code, and is designed such that it could in principle be compiled with a regular C++ compiler, so `syclcc` will not need to be invoked when building hipSYCL.
I wonder if, instead of a separate config file, it would be easier if we had `$EMPTY` in our substitution dict and just overwriting settings using command line arguments..

illuhad:
Superseded by #392

PR #256, opened 2020-05-25
Fixes #254: change __atomic_*_fetch to __atomic_fetch_*

### LINKED ISSUE #254: spack installation atomic functions
I installed hipSYCL using spack (0.14.2-691-950180ab0) with the command `spack install hipsycl@master%gcc@8.4.0 build_type=RelWithDebInfo cuda=True ^gettext@0.19.8.1`.

This sets the "default-cpu-cxx" to "/usr/bin/g++-8" (which uses libstdc++). This leads to a problem with the atomic functions used on the cpu.

Example:
```cpp
#include <vector>
#include <iostream>
#include <numeric>

#include <CL/sycl.hpp>

namespace sycl = cl::sycl;

int main() {
    std::vector<int> before(10);
    std::iota(before.begin(), before.end(), 1);
    std::vector<int> after(10, 0);

    // print buffer BEFORE kernel call
    for (int i : before) {
        std::cout << i << " ";
    }
    std::cout << std::endl;
    for (int i : after) {
        std::cout << i << " ";
    }
    std::cout << std::endl << std::endl;    

    {
        sycl::queue queue(sycl::default_selector{});
        
        sycl::buffer<int, 1> buffer_before(before.data(), before.size());
        sycl::buffer<int, 1> buffer_after(after.data(), after.size());

            auto acc_before = buffer_before.get_access<sycl::access::mode::atomic>(cgh);
            auto acc_after  = buffer_after.get_access<sycl::access::mode::discard_write>(cgh);
            
            cgh.parallel_for<class test_kernel>(sycl::range<>(before.size()), [=](sycl::item<> item) {
                const auto idx = item.get_linear_id();
                // acc_after[idx] = acc_before[idx]++;
                acc_after[idx] = acc_before[idx].fetch_add(1);
            });
        });
    }

    // print buffer AFTER kernel call
    for (int i : before) {
        std::cout << i << " ";
    }
    std::cout << std::endl;
    for (int i : after) {
        std::cout << i << " ";
    }
    std::cout << std::endl;    

    return 0;
}
```
## Expected result:
```
1 2 3 4 5 6 7 8 9 10 
0 0 0 0 0 0 0 0 0 0 

2 3 4 5 6 7 8 9 10 11 
1 2 3 4 5 6 7 8 9 10 
```
I can get this expected result if I explicitly specify the GPU as target using `--hipsycl-platform=cuda`. 
## Actual result:
```
1 2 3 4 5 6 7 8 9 10 
0 0 0 0 0 0 0 0 0 0 

2 3 4 5 6 7 8 9 10 11 
2 3 4 5 6 7 8 9 10 11 
```
However if I explicitly specify the CPU as target using `--hipsycl-platform=cpu` the result is wrong, i.e. `fetch_add(1)` **doesn't** return the value **before** the call to the atomic function (as specified in the SYCL standard).

The problem is that the builtin intrinsic function `__atomic_add_fetch` used in GCC's libstdc++ **doesn't** return the value before the function call. Instead it returns the value **after** the function call ([see](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html)).

arkantos493:
Changed `__atomic_*_fetch` to `__atomic_fetch_*` in atomic.hpp to be compliant with the SYCL 1.2.1 standard.

Currently the atomic functions `fetch_*()` functions return the updated value while targeting the CPU, but as of SYCL 1.2.1 they should return the value **before** the call to `fetch_*()` (i.e. **not** the updated value). 
Changing `__atomic_*_fetch` to `__atomic_fetch_*` fixes this issue in the CPU.

### DIFF
--- a/include/hipSYCL/sycl/atomic.hpp
+++ b/include/hipSYCL/sycl/atomic.hpp
@@ -125,7 +125,7 @@ public:
     return atomicAdd(_ptr, operand);
 #else
     assert(memoryOrder == memory_order::relaxed);
-    return __atomic_add_fetch(_ptr, operand, __ATOMIC_RELAXED);
+    return __atomic_fetch_add(_ptr, operand, __ATOMIC_RELAXED);
 #endif
 #else
     return detail::invalid_host_call_dummy_return<T>();
@@ -144,7 +144,7 @@ public:
     return atomicSub(_ptr, operand);
 #else
     assert(memoryOrder == memory_order::relaxed);
-    return __atomic_sub_fetch(_ptr, operand, __ATOMIC_RELAXED);
+    return __atomic_fetch_sub(_ptr, operand, __ATOMIC_RELAXED);
 #endif
 #else
     return detail::invalid_host_call_dummy_return<T>();
@@ -163,7 +163,7 @@ public:
     return atomicAnd(_ptr, operand);
 #else
     assert(memoryOrder == memory_order::relaxed);
-    return __atomic_and_fetch(_ptr, operand, __ATOMIC_RELAXED);
+    return __atomic_fetch_and(_ptr, operand, __ATOMIC_RELAXED);
 #endif
 #else
     return detail::invalid_host_call_dummy_return<T>();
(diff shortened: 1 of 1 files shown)


mrzv:
Oh, man, good catch. This just reinforces that #209 left the need for tests open.

illuhad:
Thank you! Yes, we should probably open an issue to remind us that we still lack tests here..

Issue #243, opened 2020-04-29
buffer.hpp broken with libc++


nazavode:
Current `libc++` support for `std::reinterpret_pointer_cast` [is still missing](https://reviews.llvm.org/D70754), this makes any attempt to build SYCL sources with `-stdlib=libc++` to fail:

```console
$ syclcc -stdlib=libc++ -std=c++17 sycl.cpp
In file included from sycl.cpp:4:
In file included from /hipSYCL/bin/../include/CL/sycl.hpp:31:
In file included from /hipSYCL/bin/../include/CL/../hipSYCL/sycl/sycl.hpp:47:
/hipSYCL/bin/../include/CL/../hipSYCL/sycl/buffer.hpp:365:36: error: no member named
      'reinterpret_pointer_cast' in namespace 'std'
    this->_writeback_buffer = std::reinterpret_pointer_cast<T>(other._writeback_buffer);
```

illuhad:
hm.. hipSYCL now requires C++17. `std::reinterpret_pointer_cast` is part of C++17. So, I guess we should just wait for `libc++` to support that..

nazavode:
Yep that makes sense, I was really disappointed to discover that `libc++` still lacks some C++17 stuff... Anyway, do you see something like a back port into hipSYCL (it's a matter of 4 lines of code: https://en.cppreference.com/w/cpp/memory/shared_ptr/pointer_cast) as a no no?

fodinabor:
one could surely add a shim header, that defines the function if necessary and otherwise just import it into a hipSYCL namespace...

nazavode:
@fodinabor exactly, if @illuhad says fine I'm obviously willing to provide a PR :)

illuhad:
@nazavode If you file a PR I will merge it to improve the situation on Mac which is already non-optimal, but  don't want to spend much time on this myself as the SYCL world is now rapidly moving to C++17.

nazavode:
@illuhad that makes perfect sense, and any `+` in a diff is a line that must be maintained. I will provide a PR but obviously feel free to reject it!

arkantos493:
I've added a PR since I had the same problem after installing hipSYCL via spack.

illuhad:
Fixed by PR #257. Thanks, @arkantos493 !

nazavode:
Thank you so much @arkantos493 for having this one fixed!

PR #257, opened 2020-05-25
PR to issue #243: fixed std::reinterpret_pointer_cast with older libc++ versions

arkantos493:
Currently `buffer.hpp` file uses `std::reinterpret_pointer_cast`. This works with `libstdc++` in C++17 mode.
However it doesn't work with `libc++` prior to the current llvm master branch.

So I added an additional header (`detail/reinterpret_pointer_cast.hpp`) which checks if `libc++` is used via the `_LIBCPP_VERSION` macro (which is not set in `libstdc++`). If the value of this macro is less than `11000` (the value set in the current master branch, clang 10.0.0 sets the value to `10000`), the header file defines the `reinterpret_pointer_cast` function as shown on [cppreference](https://en.cppreference.com/w/cpp/memory/shared_ptr/pointer_cast). Otherwise it uses the std implementation.

After including the newly added header, we only have to change the lines in `buffer.hpp` from using `std::reinterpret_pointer_cast` to `detail::reinterpret_pointer_cast`.

### DIFF
--- a/include/hipSYCL/common/reinterpret_pointer_cast.hpp
+++ b/include/hipSYCL/common/reinterpret_pointer_cast.hpp
@@ -32,8 +32,8 @@
 #include <memory>
 
 namespace hipsycl {
-namespace sycl {
-namespace detail {
+namespace common {
+namespace shim {
 
 #ifdef _LIBCPP_VERSION
 // libc++ has std::reinterpret_pointer_cast since version 11000
--- a/include/hipSYCL/sycl/buffer.hpp
+++ b/include/hipSYCL/sycl/buffer.hpp
@@ -44,7 +44,7 @@
 #include "range.hpp"
 
 #include "detail/buffer.hpp"
-#include "detail/reinterpret_pointer_cast.hpp"
+#include "../common/reinterpret_pointer_cast.hpp"
 
 #include "accessor.hpp"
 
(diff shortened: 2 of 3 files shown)


fodinabor:
LGTM, just a little comment: personally I'd move it in a special `shim` namespace to make clear that it's really just a placeholder..

illuhad:
Thank you! Since we might want to use `reinterpret_pointer_cast` also in other components apart from the SYCL interface (e.g. runtime or compiler), I also wonder whether it might be better to place it in `hipsycl::common` (or maybe `hipsycl::common::util/shim`) and `include/hipSYCL/common/`?

I'm afraid that we will probably have to support clang 10 still for a substantial amount of time, so I don't think this will just be a placeholder that we can get rid of quickly...

arkantos493:
If you mean that I should add a new folder `common/` in the same directory as `detail/` and move `reinterpret_pointer_cast.hpp` from `detail/` to `common/` and change the namespace from `detail::reinterpret_pointer_cast` to `common::shim::reinterpret_pointer_cast`, then I can do that.

illuhad:
> If you mean that I should add a new folder common/ in the same directory as detail/ and move reinterpret_pointer_cast.hpp from detail/ to common/ and change the namespace from detail::reinterpret_pointer_cast to common::shim::reinterpret_pointer_cast, then I can do that.

The new directory would not be in the same directory as `detail`. You see, I have recently started to shuffle code around to separate the three main components of hipSYCL (SYCL interface, compiler, runtime).
This means that we now have the subdirectories
* `src/compiler` - compiler component
* `src/sycl` - SYCL interface
* `src/runtime` - SYCL runtime (not there yet, will come with the new runtime that's currently in development).
The same for header files:
* `include/hipSYCL/compiler` - compiler component
* `include/hipSYCL/sycl` - SYCL interface
* `include/hipSYCL/runtime` - not there yet, will come with the new runtime.

The `detail` subdirectory is part of the `include/hipSYCL/sycl` directory, i.e. part of the SYCL interface code base. What I'm suggesting is to give equal access to this function for all three components by creating `include/hipSYCL/common`, i.e. moving it out of the code tree of the SYCL interface.

Because of a similar reasoning, I suggest to not put it in the `sycl` namespace since it is reserved for the SYCL interface. Just put it in `hipsycl::common::shim` as you say.

arkantos493:
That's a good idea. I changed it accordingly.

illuhad:
Thank you!

PR #264, opened 2020-06-12
Added cmake integration to examples, and tests

sbalint98:
This PR makes the example and test projects use the hipSYCL CMake integration. It is checked in the default installation directory of the hipSYCL package in case CMake fails to find it.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,10 +1,21 @@
 project(hipsycl-examples)
 
-set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR}/../cmake)
-
-find_package(hipSYCL CONFIG REQUIRED)
-
-if(NOT CMAKE_BUILD_TYPE)
+if(HIPSYCL_INSTALL_DIR)
+  # We make sure that we don't override an already set hipSYCL_DIR variable
+  if (hipSYCL_DIR)
+    message(FATAL_ERROR "Please only use HIPSYCL_INSTALL_DIR or hipSYCL_DIR")
+  endif() 
+  set(hipSYCL_DIR ${HIPSYCL_INSTALL_DIR}/lib/cmake)
+  # We don't set the PATHS var here since we don't want to find something other than the user intends to.
+  find_package(hipSYCL CONFIG REQUIRED)
+else()
+  # The user might have set hipSYCL_DIR or some other var manually,
+  # therefore we first try to find hipSYCL with default settings 
+  find_package(hipSYCL CONFIG)
+  # If found in the previous step nothing will be searched here
+  find_package(hipSYCL CONFIG PATHS /opt/hipSYCL/lib/cmake/ REQUIRED)
+endif()
+message(STATUS "Found hipSYCL in ${hipSYCL_DIR}")if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE Release)
 endif()
 
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -3,13 +3,22 @@ cmake_minimum_required(VERSION 3.5)
 
 find_package(Boost COMPONENTS unit_test_framework REQUIRED)
 
-find_program(SYCLCC NAMES syclcc-clang CACHE STRING)
-if(SYCLCC MATCHES SYCLCC-NOTFOUND)
-  message(SEND_ERROR "hipSYCL syclcc-clang compiler not found, exiting.")
-  return()
+if(HIPSYCL_INSTALL_DIR)
+  # We make sure that we don't override an already set hipSYCL_DIR variable
+  if (hipSYCL_DIR)
+    message(FATAL_ERROR "Please only use HIPSYCL_INSTALL_DIR or hipSYCL_DIR")
+  endif() 
+  set(hipSYCL_DIR ${HIPSYCL_INSTALL_DIR}/lib/cmake)
+  # We don't set the PATHS var here since we don't want to find something other than the user intends to.
+  find_package(hipSYCL CONFIG REQUIRED)
+else()
+  # The user might have set hipSYCL_DIR or some other var manually,
+  # therefore we first try to find hipSYCL with default settings 
+  find_package(hipSYCL CONFIG)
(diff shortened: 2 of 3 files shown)


examples/CMakeLists.txt
```diff
@@ -2,7 +2,15 @@ project(hipsycl-examples)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR}/../cmake)
 
-find_package(hipSYCL CONFIG REQUIRED)
+find_package(hipSYCL CONFIG)
+
+if(NOT hipSYCL_FOUND)
+  message(WARNING "Not found hipSYCL cmake integration retrying with default path: /opt/hipSYCL/lib/cmake/")
```
illuhad, line 8: Maybe we can correct the grammar here and make it an actual sentence ;)

examples/CMakeLists.txt
```diff
@@ -2,7 +2,15 @@ project(hipsycl-examples)
...
+  set(hipSYCL_DIR /opt/hipSYCL/lib/cmake/)
+  find_package(hipSYCL CONFIG REQUIRED)
```
illuhad, line 10: I think we can omit that step if we give cmake a hint to where the package is found in the first `find_package` using the `PATHS` argument. See https://cmake.org/cmake/help/latest/command/find_package.html

tests/CMakeLists.txt
```diff
@@ -1,15 +1,18 @@
...
-find_program(SYCLCC NAMES syclcc-clang CACHE STRING)
-if(SYCLCC MATCHES SYCLCC-NOTFOUND)
-  message(SEND_ERROR "hipSYCL syclcc-clang compiler not found, exiting.")
-  return()
-endif()
+find_package(hipSYCL CONFIG)
 
-set(CMAKE_CXX_COMPILER ${SYCLCC})
+if(NOT hipSYCL_FOUND)
+  message(WARNING "Not found hipSYCL cmake integration retrying with default path: /opt/hipSYCL/lib/cmake/")
```
illuhad, line 10: same here

tests/CMakeLists.txt
```diff
@@ -1,15 +1,18 @@
...
+  set(hipSYCL_DIR /opt/hipSYCL/lib/cmake/)
+  find_package(hipSYCL CONFIG REQUIRED)
```
illuhad, line 12: and here

examples/CMakeLists.txt
```diff
@@ -2,7 +2,15 @@ project(hipsycl-examples)
...
+endif()
+set(hipSYCL_DIR /opt/hipSYCL/lib/cmake/ CACHE PATH
```
illuhad, line 12: It may be more intuitive for the user to just have that point to `/opt/hipSYCL` and then add `lib/cmake` automatically. After all, the subdirectory where the cmake files are stored will be the same for every hipSYCL installation.

examples/CMakeLists.txt
sbalint98: Ohh I see. I should have read the manual thoroughly beforehand. I think using the PATHS hint makes this change very straightforward. In my opinion it elminates all the other problems. I pushed a new version using the PATHS hint.

examples/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ project(hipsycl-examples)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR}/../cmake)
 
-find_package(hipSYCL CONFIG REQUIRED)
+find_package(hipSYCL CONFIG PATHS /opt/hipSYCL/lib/cmake/ REQUIRED)
```
illuhad, line 5: Do we need to make the path user configurable if the user has installed hipSYCL in a custom location that is not found by cmake? I would propose to only make the hipSYCL installation prefix configurable, i.e.
```
find_package(hipSYCL CONFIG PATHS ${HIPSYCL_INSTALL_DIR}/lib/cmake/ REQUIRED)
```

tests/CMakeLists.txt
```diff
@@ -1,15 +1,10 @@
 project(hipsycl-tests)
 cmake_minimum_required(VERSION 3.5)
 
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR}/../cmake)
```
illuhad, line 4: Is this still necessary?

examples/CMakeLists.txt
sbalint98: In the section of the search procedure of the CMake find_pacakge documentation, it says: 

`9. Search paths specified by the PATHS option. These are typically hard-coded guesses.`

If I understand correctly, the steps that come before 9 take precedence over `PATHS`, but that means the user should pass the `-DhipSYCL_DIR` flag which should point to the `HIPSYCL_INSTALL_DIR/lib/cmake` directory. if he or she wants to use hipSYCL installed to a custom location

I think we should check if the user set the `HIPSYCL_INSTALL_DIR` and then set the `hipSYCL_DIR`, before looking for the hipSYCL package.

tests/CMakeLists.txt
sbalint98: No, I dont think so

examples/CMakeLists.txt
illuhad: Ah, I forgot about `-DhipSYCL_DIR`. It's fine then. I think for the unit tests it's okay if the user has to remember to point it to `..../lib/cmake` since it will mostly be used by developers anyway.

sbalint98:
I pushed a version that, in my opinion, is the most error safe for the user, and also provides the option of specifying the installation root dir without pointing to the CMake integration files.

- In case `HIPSYCL_INSTALL_DIR` is specified that the directory will be searched first, and the default path won't be searched.
- In case both  `HIPSYCL_INSTALL_DIR` and `hipSYCL_DIR` are set, an error is emitted. 
- If there  `HIPSYCL_INSTALL_DIR` is not specified, first, we will look for the installation without specifying the path, and, in case we haven't found the installation, we will check with the default installation path. This should guarantee that if the user sets additional variables which influence the searched paths, the installation intended by the user is found.

illuhad:
Okay, that looks good so far. Can you check if it still works with the autotest script? I think this might need some very small changes then.

Issue #265, opened 2020-06-16
ROCm 3.5 link error/ROCm hipStreamAddCallback crash


tomdeakin:
Have you seen this error before?
```
[ 50%] Linking CXX shared library libhipSYCL_rocm.so
/opt/rh/devtoolset-8/root/usr/lib/gcc/x86_64-redhat-linux/8/../../../../bin/ld: cannot find -lmcwamp
/opt/rh/devtoolset-8/root/usr/lib/gcc/x86_64-redhat-linux/8/../../../../bin/ld: cannot find -lhc_am
clang-9: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [src/sycl/libhipSYCL_rocm.so] Error 1
make[1]: *** [src/sycl/CMakeFiles/hipSYCL_rocm.dir/all] Error 2
make: *** [all] Error 2
```

I'm not inside the `devtoolset-8`, so there might be some hard coded paths curtesy of the ROCm install via the `yum` repository.

illuhad:
No.. But AMD's ROCm packages for various distributions can vary quite a bit. Linking against these libraries used to be necessary, so this is done by default by `syclcc`. I suspect that these libraries are either not there entirely with your ROCm distribution, or their content is statically linked into some other library.
Try removing these libraries from the `ROCM_LINK_LINE` setting inside `$HIPSYCL_INSTALL_PREFIX/etc/hipSYCL/syclcc.json`. This is from where `syclcc` draws its default linker flags.

tomdeakin:
Thanks - I had to remove it directly in `CMakeList.txt` but it managed to build hipSYCL.

I find a different error when building SYCL code:
```
error: unable to load plugin '/nfs/software/x86_64/hipsycl/master-d4064619/bin/../lib/libhipSYCL_clang.so': '/nfs/software/x86_64/hipsycl/master-d4064619/bin/../lib/libhipSYCL_clang.so: undefined symbol: _ZNK4llvm10ModulePass17createPrinterPassERNS_11raw_ostreamERKSs'
```

illuhad:
This happens when there is an ABI mismatch between the clang plugin and the clang you are loading it into. E.g., clang plugin is compiled against a different clang version, or cmake has not correctly detected settings that LLVM was built with (e.g. `-fno-rtti`, `-fno-exceptions`). The latter I have seen with some AMD provided clang distributions that don't seem to correctly advertise their build flags.

[...]

PR #269, opened 2020-06-23
Added updated install script for aomp 11.6-1

sbalint98:
This pull request updates the install scripts in order to accommodate the update for clang 10 and for ROCm 3.5 with the newest aomp release. 

The updated packages are available under the: http://repo.urz.uni-heidelberg.de/sycl/test_llvm10_rocm11/ link for the supported distros. Please note that these packages are experimental, and they need to be tested before merging this PR. 

Furthermore, it contains minor fixes for the packaging system


illuhad:
It seems this is affected by the `hipStreamAddCallback` crash from issue #265

illuhad:
Following the discussion in #265, we either need to wait for ROCm 3.6 or try an earlier ROCm version since AMD has now confirmed that `hipStreamAddCallback` is broken for the default stream in ROCm 3.5.

illuhad:
@sbalint98 How would you like to proceed? Downgrade to earlier ROCm/aomp?

sbalint98:
Sure that seems like a good way to proceed in my opinion. It used to work with the one earlier release (which pulls from rocm-3.3 ) but it needed a lot of tweaking. I will have a look

illuhad:
Closing now, please file new PR with different ROCm version :) Also, because of PR #281 we should really push out clang 10 now...

PR #271, opened 2020-06-28
Fix Compile with AppleClang on macOS

### LINKED ISSUE #228: MacOS CMake issues
Prompted by https://github.com/illuhad/hipSYCL/issues/222#issuecomment-601104011, I tried to build on Mac but it seems to ignore my explicit specification of Clang location and instead finds `/usr/bin/clang++`.

It is possible that the issue here is that the Homebrew installs of LLVM don't have the required components, but CMake is not doing a proper job detecting that.

```sh
jrhammon-mac02:build jrhammon$ cmake .. -DCMAKE_INSTALL_PREFIX=/opt/hipsycl -DCMAKE_CXX_COMPILER=/usr/local/Cellar/llvm/9.0.1/bin/clang++ -DCMAKE_C_COMPILER=/usr/local/Cellar/llvm/9.0.1/bin/clang -DDISABLE_LLVM_VERSION_CHECK=ON
-- The C compiler identification is Clang 9.0.1
-- The CXX compiler identification is Clang 9.0.1
-- Check for working C compiler: /usr/local/Cellar/llvm/9.0.1/bin/clang
-- Check for working C compiler: /usr/local/Cellar/llvm/9.0.1/bin/clang -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/local/Cellar/llvm/9.0.1/bin/clang++
-- Check for working CXX compiler: /usr/local/Cellar/llvm/9.0.1/bin/clang++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Warning (dev) at CMakeLists.txt:17 (set):

-- Could NOT find LLVM (missing: LLVM_DIR)
-- Building hipSYCL against LLVM configured from LLVM_DIR-NOTFOUND
-- Selecting clang: /usr/bin/clang++
CMake Error at CMakeLists.txt:77 (message):
  clang include path CLANG_INCLUDE_PATH-NOTFOUND does not exist.  Please
  provide clang's internal include path manually.


-- Using clang include directory: CLANG_INCLUDE_PATH-NOTFOUND
-- Configuring incomplete, errors occurred!
See also "/Users/jrhammon/Work/SYCL/hipSYCL/build/CMakeFiles/CMakeOutput.log".
```

ax3l:
This adds a GitHub Action scripts for CI on macOS and fixes occurring issues with AppleClang.

macOS dependencies are installed via brew, especially including `libomp` for OpenMP support. Clang on macOS needs further a `-Xclang` or `-Xpreprocessor` flag before its `-fopenmp` flag, which is now added in the compiler wrapper script.

This requires CMake 3.16+ *on macOS* in order to query `OpenMP_CXX_INCLUDE_DIRS`: https://cmake.org/cmake/help/v3.16/module/FindOpenMP.html

Note that I only extract the include path - other properties such as flags will be found for the CC/CXX compiler used to build hipsycl and are not necessary the ones for the `clang` in `CLANG_EXECUTABLE_PATH`.

Also avoids [unknown linker flags](https://github.com/illuhad/hipSYCL/issues/228#issuecomment-601316674) on macOS.

`hipSYCL_cpu` test: [passes](https://github.com/ax3l/hipSYCL/actions) :)

Fix #228


bin/syclcc-clang
```diff
@@ -558,7 +558,11 @@ class pure_cpu_compiler:
     hipsycl_library_path = os.path.join(config.hipsycl_installation_path,"lib/")
 
     self._linker_args = []
-    self._compiler_args = ["-fopenmp"]
+    self._compiler_args = []
+    if sys.platform == "darwin":
+        # y'all still need libomp installed
+        self._compiler_args += ["-Xclang"]
+    self._compiler_args += ["-fopenmp"]
```
illuhad, line 565: I have very little experience on Mac - I know that the default compiler there is clang, but I wonder if this is really the only choice for an OpenMP compiler? The hipSYCL CPU backend can work with any C++ OpenMP compiler. Would Mac users be interested in e.g. using gcc on Mac? If so, we would need to change this so that it doesn't break other compilers.

install/docker/CUDA/Dockerfile
```diff
@@ -5,10 +5,12 @@ ARG hipsycl_origin=https://github.com/illuhad/hipSYCL
 ENV hipsycl_branch=$hipsycl_branch
 ENV hipsycl_origin=$hipsycl_origin
 RUN apt-get update
-RUN apt-get install -y wget
+RUN apt-get install -y apt-transport-https ca-certificates gnupg software-properties-common wget
 RUN echo "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-8 main" >> /etc/apt/sources.list.d/llvm.list
 RUN echo "deb-src http://apt.llvm.org/bionic/ llvm-toolchain-bionic-8 main" >> /etc/apt/sources.list.d/llvm.list
 RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
+RUN wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | gpg --dearmor - | tee /etc/apt/trusted.gpg.d/kitware.gpg >/dev/null
+RUN apt-add-repository 'deb https://apt.kitware.com/ubuntu/ bionic main'
```
illuhad, line 13: okay.. these dockerfiles are deprecated anyway, so no need to put work into it. While it doesn't hurt keeping them up to date, they will probably be removed entirely soonish.

install/scripts/base-ubuntu-18.04.def
```diff
@@ -2,16 +2,18 @@ BootStrap: docker
 From: ubuntu:18.04
 
 %setup
+cp ./install-cmake.sh ${SINGULARITY_ROOTFS}/install-cmake.sh
```
illuhad, line 5: *if* we need to enforce cmake 3.16 globally we will also have to install it for the other supported distributions (in particular CentOS) in order to not break the packaging pipeline.

src/compiler/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.16)
```
illuhad, line 1: Do we really need to raise it here as well? The clang compiler plugin does not use OpenMP and is compiled with a regular, pure C++ compiler.

src/sycl/CMakeLists.txt
```diff
@@ -74,10 +74,12 @@ if(WITH_ROCM_BACKEND)
 endif()
 
 if(WITH_CPU_BACKEND)
+  find_package(OpenMP REQUIRED)
   add_library(hipSYCL_cpu SHARED ${HIPSYCL_SOURCES})
   target_compile_definitions(hipSYCL_cpu PRIVATE HIPSYCL_BOOTSTRAP=1)
   target_compile_options(hipSYCL_cpu PRIVATE --hipsycl-platform=cpu ${COMMON_SYCLCC_OPTIONS})
-  target_include_directories(hipSYCL_cpu PRIVATE ${INCLUDE_DIRS} ${HIPSYCL_SOURCE_DIR}/contrib/hipCPU/include)
+  target_include_directories(hipSYCL_cpu PRIVATE ${INCLUDE_DIRS} ${OpenMP_CXX_INCLUDE_DIRS} ${HIPSYCL_SOURCE_DIR}/contrib/hipCPU/include)
```
illuhad, line 81: Again, I wonder if we can make this specific only to Apple, so that the other platforms can continue to support older cmake versions.

illuhad:
@sbalint98 You may also want to have a look at it since some of the changes can impact packaging.

src/sycl/CMakeLists.txt
illuhad: ... and can you elaborate why we need to add the OpenMP include directories when building hipSYCL, but not when later compiling programs using `syclcc`? There it seems to be sufficient to add `-Xclang`?

.github/workflows/ubuntu.yml
```diff
@@ -0,0 +1,22 @@
...
+        submodules: 'recursive'
+    - name: install dependencies
+      run: |
+        sudo apt-get -qqq update
+        sudo apt-get install libboost-dev libclang-9-dev llvm-9-dev python3
+    - name: build
+      run: |
+        mkdir build && cd build
+        cmake ..
+        make -j 2
```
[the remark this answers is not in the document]
ax3l: No probs, just added this to test if Ubuntu works and then expanded to macOS.
ShallI remove the Ubuntu GH action for now and keep the macOS one?

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 3.5)
+cmake_minimum_required (VERSION 3.16)
```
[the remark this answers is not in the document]
ax3l: I understand this, yet installing CMake is a few-liner on all systems we face today as it comes with pre-compiled binaries and is fully self-contained.

- x86 binaries: https://cmake.org/download/ - wget & untar awaway :)
- pip (yeah, even pip): https://pypi.org/project/cmake/
- apt repos (developer laptops): https://apt.kitware.com/
- three lines from source: https://cmake.org/install/
- Linux/macOS user-level package managers: Spack, brew
- more user-level package managers: conan, conda, NuGet, vcpkg, [snap](https://snapcraft.io/install/cmake/rhel) ...

We can, of course, just keep it with 3.5 and throw a warning. But imho there is no reason to stick with old CMake releases.
Or ship a backport for `FindOpenMP.cmake`. Or try to find `omp.h` via [`find_file`](https://cmake.org/cmake/help/latest/command/find_file.html).

Personally, I would just document the few lines to install CMake without admin rights in user-facing docs :)

bin/syclcc-clang
ax3l: I am also a Linux users and just support a lot of developers in our team running on macs. Generally, most popular open source projects just focus on supporting & testing the native system compiler on macOS (and Windows) - AppleClang is the primary compiler there.

Anything else (e.g. GCC stacks on macOS) are more challenging and get quickly into nifty support problems, simply because most upstream projects don't test against it. I don't see how the current logic works with other compilers than clang right now and I think this would need on all systems more abstraction as yet implemented, independent of macOS.

install/docker/CUDA/Dockerfile
ax3l: No probs, just tried to catch what I saw. Any other workflow files you run that need updates?

install/scripts/base-ubuntu-18.04.def
ax3l: Absolutely, can you please point me to the files that need updating and how I can test them? :)

src/compiler/CMakeLists.txt
ax3l: Will do, no prob.

src/sycl/CMakeLists.txt
ax3l: will just check here for cmake 3.16+ for now.

we need them in both cases on macOS, in which line did I miss to add the includes?

CMakeLists.txt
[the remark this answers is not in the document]
ax3l: > could just remain at 3.5 globally and only require 3.16 when compiling for Mac?

implemented: )

src/sycl/CMakeLists.txt
```diff
@@ -74,10 +74,20 @@ if(WITH_ROCM_BACKEND)
...
+    if(CMAKE_VERSION VERSION_LESS "3.16")
+      message(FATAL_ERROR "CMake 3.16.0+ is required for macOS OpenMP support!")
+    endif()
+  endif()
   add_library(hipSYCL_cpu SHARED ${HIPSYCL_SOURCES})
   target_compile_definitions(hipSYCL_cpu PRIVATE HIPSYCL_BOOTSTRAP=1)
   target_compile_options(hipSYCL_cpu PRIVATE --hipsycl-platform=cpu ${COMMON_SYCLCC_OPTIONS})
   target_include_directories(hipSYCL_cpu PRIVATE ${INCLUDE_DIRS} ${HIPSYCL_SOURCE_DIR}/contrib/hipCPU/include)
+  if(APPLE)
+    target_include_directories(hipSYCL_cpu PRIVATE ${OpenMP_CXX_INCLUDE_DIRS})
```
ax3l, line 88: potentially, I would like to add them to the `syclcc` compiler itself as intrinsic flags it passes to clang but did not find a quick way to achieve that.
I tried hacking the Python script but this did not change the compile error in this line.

install/docker/CUDA/Dockerfile
ax3l: no updates for linux needed anymore.

install/scripts/base-ubuntu-18.04.def
ax3l: no updates for linux needed anymore.

install/scripts/install-cmake.sh
```diff
@@ -0,0 +1,10 @@
+export INSTALL_PREFIX=${INSTALL_PREFIX:-/opt/cmake}
```
ax3l, line 1: this script is now unused but might be handy in the future.
shall I keep or remove it?

sbalint98:
Thanks for this PR! regarding the packages with the optional cmake 3.16 it looks good to me :)

bin/syclcc-clang
illuhad: Okay, in that case I think it's fine for the time being. On Linux, it just works with multiple compilers by virtue of `-fopenmp` being the only thing necessary to get it working with gcc, clang, icc etc. Thus switching the compiler with `HIPSYCL_CPU_CXX` just works. I think the "correct" way to generalize this to handle such cases would be to to give the user the ability to customize the compilation command as we have already done for the linking steps on CUDA and ROCm with the `CUDA_LINK_LINE` and `ROCM_LINK_LINE` settings.

src/sycl/CMakeLists.txt
illuhad: Yes, we would need a `syclcc` configuration setting for the CPU compilation flags as described above.

src/sycl/CMakeLists.txt
illuhad: Nevermind, I get it now - `syclcc` doesn't add the includes, these get added by cmake, and when compiling with `syclcc` later on, the user must manually specify the OpenMP include paths.

illuhad:
Thank you! :clap:

src/sycl/CMakeLists.txt
ax3l: Yes, I hard-coded this in between but was still not working, so I left it for the time being as this.

PR #274, opened 2020-07-09
cmake: Don't look for specific versions of LLVM

illuhad:
Remove the old cmake behavior of by default trying to look for the newest LLVM it can find.
* cmake LLVM version detection is sometimes unreliable
* the old behavior can cause cmake to silently ignore if the user has explicitly specified `-DLLVM_DIR`
* make it easier to use hipSYCL with very new unreleased LLVM versions without us having to maintain an ever-growing list of LLVM versions in cmake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,28 +20,7 @@ FORCE)
   endif()
 endif()
 
-if(DISABLE_LLVM_VERSION_CHECK)
-  # just use whatever we are given - some llvm distributions do
-  # not correctly advertise their versions, and if users explicitly point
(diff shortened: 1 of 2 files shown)


illuhad:
Package maintainers, note that this removes the cmake `DISABLE_LLVM_VERSION_CHECK` flag - you can remove this cmake option from your packages if you use it.
@sbalint98 
@nazavode
@acxz

acxz:
Yay! I like this very much!

nazavode:
Updating Spack package as soon as possible, thanks for the ping!

PR #278, opened 2020-07-23
Rename master branch -> stable

illuhad:
This renames any mention of the master branch to `stable`, which is more in line with reality as it is now more of a "latest stable" branch, while development is carried out on the `develop` branch.

See https://github.com/illuhad/hipSYCL/blob/develop/CONTRIBUTING.md for our new branch structure.

Once this is merged, I will
* Rename the `master` branch to `stable`
* Make `develop` the new default branch.

Heads up to package maintainers: You may need to update your packages
* if you explicitly reference the master branch or
* if you clone the default branch and would prefer to build packages from `stable` instead of `develop`.

@sbalint98  (I have already changed all mentions of "master" in our scripts, but it is better if you check again)
@nazavode 
@acxz


acxz:
Shouldn't affect the Arch packages since we pull from the default branch anyway. Thx for the heads up.

nazavode:
Do you plan to keep both `master` and `stable` around for some time or just rename the former right on the spot? The Spack package points at `master` right now, I would like to figure out the right timing to minimize the window in which the upstream package will be actually broken.

illuhad:
@nazavode If it helps, of course I can leave a copy of current master around to avoid breakage for users (but probably won't update it anymore). Let me know when the changes to the spack package have landed so that I can remove the old branch for good.

nazavode:
@illuhad thanks, that would be great, I can keep an eye on this PR and ping back you when the downstream change is in place.

illuhad:
@nazavode great, thanks!

sbalint98:
Looks good to me :)

illuhad:
@nazavode stable branch is now alive, you can now start to deploy your changes :)

nazavode:
@illuhad https://github.com/spack/spack/pull/17689 merged!

illuhad:
@nazavode excellent, that was quick! I will remove the master branch today or in the next couple of days.

PR #279, opened 2020-07-23
Drop support for nd_range parallel for on CPU

illuhad:
This PR drops supports for `nd_range` parallel for on the CPU backend.

`nd_range` parallel for **cannot** be implemented efficiently as defined by the specification on pure-library implementations such as the hipSYCL CPU backend. **No user desiring performance portability should therefore ever use it anyway.**
By "not efficient" I don't mean 20% or 40% slower, I mean four orders of magnitude increase in runtime are expected. In other words, it is unusable anyway - this PR will hence only point users explicitly to the problem instead of letting them figure out the problem on their own.

Code using `nd_range` parallel for will still compile, but throw an exception at runtime.

Supporting `nd_range` on CPU means we need to support an entirely different execution model just for CPU `nd_range` which considerably complicates the implementation, especially as we start implementing SYCL 2020 features and might even make some of them impossible.

Only supporting the (superior) model of mapping one thread to an entire work group as used by hierarchical and scoped parallelism will allow us to clean up the code considerably in the future.

In the future we might reenable support for some corner cases that can be implemented without a different execution model. In particular `nd_range` parallel for with a work group size of 1 would be possible, although I doubt that this will be useful for many users.

Recommendations for users who need `nd_range` on CPU:
* please verify that you really need the features of `nd_range` parallel for. If local memory is not used, you can use basic parallel for.
* users targeting SYCL 1.2.1 may use hierarchical parallel for, which can express the same algorithms, but may have functionality caveats in hipSYCL and/or other SYCL implementations.
* if you use hipSYCL exclusively, you are encouraged to use scoped parallelism:
  https://github.com/illuhad/hipSYCL/blob/develop/doc/scoped-parallelism.md

@DieGoldeneEnte This might be interesting for you as it can greatly simplify the implementation of group algorithms on CPU :)

### DIFF
--- a/include/hipSYCL/sycl/handler.hpp
+++ b/include/hipSYCL/sycl/handler.hpp
@@ -345,24 +345,6 @@ void parallel_for_kernel(Function f,
     }
 }
 
-// This must still be executed by hipCPU's hipLaunchKernel for correct
-// synchronization semantics
-template<int dimensions, class Function>
-inline
-void parallel_for_ndrange_kernel(Function f, sycl::id<dimensions> offset) noexcept
-{
-  nd_item<dimensions> this_item{
-    &offset,
-    detail::get_group_id<dimensions>(),
-    detail::get_local_id<dimensions>(),
-    detail::get_local_size<dimensions>(),
-    detail::get_grid_size<dimensions>()
-  };
-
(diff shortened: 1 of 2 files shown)


illuhad:
Closing in favor of #294

PR #280, opened 2020-07-26
CPU backend nd_range parallel for: Add prototype implementation using fibers as work items for massive speedup

illuhad:
Alternative to PR #279.

This replaces the `nd_range` execution model with an execution model based on lightweight user threads (`boost.fiber`) for work items and multithreading across work groups. This is considerably faster than the old execution model:
* previously, `nd_range` parallel for was ~10000x slower than hierarchical/basic parallel for
* this PR improves performance such that `nd_range` is "only" 40x slower.

So, in summary, it's much better than previously, but the fact remains that `nd_range` parallel for cannot be implemented efficiently in pure-library SYCL implementations. Even with this PR, performance cannot be competitive for `nd_range` parallel for.

Since with this PR, multithreading is now happening across work groups for all kernel invocation mechanism, it would also unify the execution models.

This would also add new dependencies for hipSYCL, namely `boost.context` and `boost.fiber`.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -17,6 +17,7 @@ jobs:
         brew install cmake
         brew install libomp
         brew install llvm
+        brew install boost
         set -e
     - name: build
       run: |
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,6 +45,8 @@ endif()
 message(STATUS "Building hipSYCL against LLVM configured from ${LLVM_DIR}")
 #find_package(Clang REQUIRED)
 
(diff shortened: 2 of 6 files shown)


illuhad:
@sbalint98 , we may need to add boost.context and boost.fiber as dependencies when building packages and make sure they're installed in the singularity images. Can you have a look?

illuhad:
Alternatively, we could also try to deploy a minimal fiber library based on `setcontext/swapcontext` of our own instead of relying on boost. After all, the semantics we need are in principle we very simple.

illuhad:
Closing in favor of #294

PR #283, opened 2020-08-07
Updated build system in order to allow building packages separately

### LINKED ISSUE #260: Upgrade stack to Clang 10 by default?
Hello. May I ask if there are any plans to upgrade the repos to Clang 10? If so, is there anything I can do to help speed things up? One reason for being eager to upgrade is that ```-ffast-math``` causes compilations errors with Clang 9 ([https://bugs.llvm.org/show_bug.cgi?id=45034](url)). Of course a Clang 10 stack could be installed manually, but I think it'd benefit users to have it available in the repos. As I say, let me know if there's anything I can do to help :)

sbalint98:
This update accommodates building the base package without building the rocm packages. Since the new rocm version has a bug ( #265 and [this](https://github.com/ROCm-Developer-Tools/HIP/issues/2119#issuecomment-652556974) ) we want to keep the old rocm version when updating to clang 10. 

The new base packages containing clang 10 are available [here](http://repo.urz.uni-heidelberg.de/sycl/test_llvm10/) After this PR is merged the new clang packages are going to be pushed out to the official package repositories. Therefore this PR should close #260


illuhad:
It seems that at least in conjunction with this ROCm version, clang 10 does not work very well. When targeting AMD devices, we get errors such as 
```
Memory access fault by GPU node-1 (Agent handle: 0x55b65dc6b120) on address 0x400000007000. Reason: Unknown.
unknown location(0): fatal error: in "device_test_suite/dynamic_local_memory": signal: SIGABRT (application abort requested)
/home/aksel/data/storage/src/cs/cpp/sycu/SYCU/tests/unit_tests.cpp(181): last checkpoint: "dynamic_local_memory" test entry
```
for multiple unit tests. I don't know what exactly causes this. This problem does not appear with clang 9. We could check if clang 11 or 12 fix this issue..

illuhad:
Also an interesting thought, if it turns out newer versions of ROCm work better with clang 10: The new runtime should not be affected by #265, meaning we could also attempt again to upgrade ROCm if we push out the new runtime simultaneously.

illuhad:
Issue persists with
* clang 10 + ROCm 3.5
* clang 11 + ROCm 3.5 (clang 11 has some additional quirks)
Going to try a kernel upgrade next

illuhad:
Issues are fixed with PR #287. We can roll out clang 10 once those are merged :)

PR #289, opened 2020-08-22
Initial PR for new runtime (lite version)

illuhad:
This PR adds the new runtime in a "lite" version. Meaning that certain features are implemented, but not yet exposed for the sake of stability, allowing a quicker merge. Other features (in particular full DAG lazy/predictive scheduler) are known to be bugged and therefore disabled by default or will be redesigned in the future.

Unit test coverage:
* CUDA backend: All tests pass
* ROCm backend: All tests pass, except for 2D and 3D explicit memcopies from device to host - I suspect this is a HIP/ROCm bug because same code works on CUDA and on ROCm if direction of memcpy (host to device) is changed.
* CPU backend: All tests pass 
   -  ~~except `nd_range` parallel for tests. `nd_range` parallel for is not implemented because it will be replaced either by a fiber based implementation (see #280) or an implementation based on coroutines.~~ *See PR #294*
   -  ~~There are occasional failures in 2 explicit copy tests. Since they only occur occasionally, this might point to a multithreading or UB issue. This requires more investigation.~~ *This seems to have been fixed by aaead19*

# Features (probably incomplete list)
This is a major redesign of large portions of the project, but I feel the long list of new features justifies the effort:

* **Single library for all backends** (`libhipSYCL-rt`) instead of libraries for each backend (`libhipSYCL_cpu`, `libhipSYCL_cuda` etc)
* **Strict seperation between backend specific code and generic code**, clear, simple interface to add new backends, making it easy to add additional backends in the future
* **Multiple runtime backends can now be active at the same time and interact** (note: to submit kernels to multiple backends you would additionally need the compiler to support compiling for both simultaneously, which is not yet possible for ROCm and CUDA due to clang limitations. Compiling for GPU backend + CPU needs some cleanup in `syclcc` and maybe SYCL interface).
* SYCL interface is now header-only; bootstrap mode in `syclcc` is no longer required and has been removed. **When building hipSYCL, only the runtime needs to be compiled which can be done with any regular C++ compiler.** This should simplify the build process greatly.
* Architecture supports **arbitrary execution models in different backends** - queue/stream based, task graphs etc.
* **CUDA and CPU backends do not depend on HIP API anymore**. The CUDA backend now goes directly to CUDA without going through HIP, and the CPU backend goes directly to OpenMP without going through hipCPU. Once the SYCL interface has been cleaned up, we will not need to ship those submodules anymore. Only the ROCm backend still uses HIP. **This means that hipSYCL now effectively is hipcudaopenmpSYCL** :)
* **Strict separation between SYCL interface and runtime**, making it easy to expose new features (e.g. SYCL 2020) in the SYCL interface by leveraging the SYCL runtime interfaces underneath.
* For each operation, SYCL interface can pass additional information to runtime/scheduler using hints framework. **Device on which an operation is executed is just another hint for the runtime.**
* **Support for lazy DAG execution** (Note: Only partially activated by default)
* **Almost entirely callback-free execution model in CUDA/ROCm backends for potentially higher task throughput** (currently hipSYCL relies heavily on callbacks to signal task completion, which can add a lot of latency for small tasks)
* **New memory management system**
   - manages arbitrary allocations on multiple devices
   - manages memory potentially below buffer granularity, using 3D page table to track invalid memory regions (Note: By default, page size is equal to buffer size currently, so this feature is not yet fully exposed)
* Backend queues (e.g. CUDA streams) are maintained by the backend in a pool, the scheduler then distributes operations across the queues. **No matter how many `sycl::queues` exist, compute/memory-overlap always works equally well.** (Note: CUDA/ROCm backends currently just have 1 compute and 1 memcpy stream instead of querying device capabilities). This means a `sycl::queue` is now nothing more than an interface to the runtime.
* **Vastly improved error handling**. Proper implementation of async errors/error handlers. Task execution will be cancelled when an error is detected, so unlike earlier hipSYCL versions, the application will not stall because events don't complete when things start to go wrong.
* **ROCm backend: Add support for 3D data transfers**
* Colored `HIPSYCL_DEBUG` output!!!! That's right, most important feature of all -- errors are red, warnings purple, info is green.
* New `settings` framework, allowing configuring behavior of certain aspects of hipSYCL using environment variables. For example, debug level now controlled by environment variable `HIPSYCL_DEBUG_LEVEL`
* CPU backend is now always required (even when only targeting GPU), hipSYCL will always be compiled with CPU backend enabled.

# Steps to be done before merging
- [x] Test with proper SYCL applications.
- [x] Check if there are circular dependencies causing memory leaks
- [x] Investigate CPU unit test failures
- [x] Merge new CPU `nd_range` parallel for implementation
- [x] Maybe add more documentation, although that could be done post-merge


keryell:
What a revolution!

bd4:
Very exciting, I will try building and running some of my example programs! Sounds like with the new memory management system, this is a good foundation for USM support? Most of my more interesting examples use USM.

illuhad:
@bd4 The new memory management layer itself is independent from USM as USM allocations are handled by backend/driver - however, related USM features (explicit dependencies, ordered queues) benefit greatly from the new runtime. Long story short, I will start implementing USM in the next few weeks.

illuhad:
Package maintainers, be advised of the following changes this PR introduces that might affect packaging:

1) former cmake variable `ROCM_LINK_LINE` is now built internally by cmake based on the `ROCM_LIBS` variable. To customize which ROCm libraries need to be linked `ROCM_LIBS` should now be used.
2) `boost.context` and `boost.fiber` are now new dependencies
3) cmake variable `WITH_CPU_BACKEND` is gone; CPU backend is always required, even if you are only interested in CUDA or ROCm. This also means that OpenMP is a hard dependency for all backends for now.
4) hipSYCL now comes with a unified runtime library for all backends that have been enabled at build time. Meaning that `libhipSYCL_cuda`, `libhipSYCL_rocm`, `libhipSYCL_cpu` don't exist anymore - there is now only a single `libhipSYCL-rt`. This may be relevant if you have separate packages for ROCm, CUDA, CPU variants of hipSYCL and have relied on the libraries not clashing when installing them to the same directory.
5) Maybe more implications I have overlooked, so please test ;)

In general this PR introduces massive changes, so there's a chance for new bugs. If you are cloning from `sycl/2020` or `stable` branches you won't be it right away as we first merge into `develop` for additional testing. `stable` and `sycl/2020` are even with current `develop`, so if you prefer to switch away from `develop` for now because of this you won't lose anything.

@sbalint98 
@nazavode 
@acxz
(maybe we should maintain a list of packagers to which people can add themselves. Maybe an issue or a markdown file in the repository.. I'm always worried I might forget somebody)

PR #290, opened 2020-08-25
Add nightly packages

sbalint98:
This PR updates the packaging system with the ability to build nightly packages from the develop branch. This will give the user the option to either use the stable, well-tested stack where hipSYCL is built from the stable branch (`hipSYCL`, `hipSYCL-rocm`, `hipSYCL-base` ) or the development stack where hipSYCL is built from the develop branch (`hipSYCL-nightly`, `hipSYCL-rocm-nightly`, `hipSYCL-base-nightly`)

Currently, the `hipSYCL-base-nightly` package contains the installation of clang10 whereas `hipSYCL-base` still has clang9 . the `hipSYCL-rocm-nightly` package is the same as the `hipSYCL-rocm` package, but that might change in the future.

PR #291, opened 2020-09-01
Fix buffer/id operators, add accessor operator==

psalz:
See ⬆.

As I've noted in the test for `accessor::operator==`, a strict reading of the 1.2.1 Rev 7 spec, section 4.3.2 would imply that two accessors that were independently created from the same buffer (i.e., not copied) should not be equal. As the two are however indistinguishable, I don't think there is any harm in this current behavior.

### DIFF
--- a/tests/unit_tests.cpp
+++ b/tests/unit_tests.cpp
@@ -505,6 +505,54 @@ auto make_test_value(const T<1>& a, const T<2>& b, const T<3>& c) {
   return std::get<dimensions - 1>(std::make_tuple(a, b, c));
 }
 
+template <template<int D> class T, int dimensions>
+void test_id_range_operators() {
+  const auto test_value = make_test_value<T, dimensions>({ 5 }, { 5, 7 }, { 5, 7, 11 });
+  const auto other_test_value = make_test_value<T, dimensions>({ 3 }, { 3, 4 }, { 3, 4, 9 });
+
+  {
+    // T + T
+    const auto result = test_value + other_test_value;
+    if(dimensions >= 1) BOOST_TEST(result[0] == 8);
+    if(dimensions >= 2) BOOST_TEST(result[1] == 11);
+    if(dimensions == 3) BOOST_TEST(result[2] == 20);
+  }
+
+  {
+    // T + size_t
+    const auto result = test_value + 2;
+    if(dimensions >= 1) BOOST_TEST(result[0] == 7);
+    if(dimensions >= 2) BOOST_TEST(result[1] == 9);
+    if(dimensions == 3) BOOST_TEST(result[2] == 13);
+  }
+
+  {
+    // T += T
+    auto result = test_value;
+    result+= other_test_value;
+    if(dimensions >= 1) BOOST_TEST(result[0] == 8);
+    if(dimensions >= 2) BOOST_TEST(result[1] == 11);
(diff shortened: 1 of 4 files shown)


illuhad:
@ax3l It seems that the Mac OS check cannot find LLVM for some reason. Can you have a look?

illuhad:
Thank you!

tests/unit_tests.cpp
```diff
@@ -326,6 +326,75 @@ BOOST_AUTO_TEST_CASE(buffer_versioning) {
...
+
+    BOOST_REQUIRE(acc_a1 == acc_a1);
+    BOOST_REQUIRE(acc_a2 == acc_a2);
+    BOOST_REQUIRE(acc_a1 != acc_a3);
+    BOOST_REQUIRE(acc_a1 != acc_a4);
+    BOOST_REQUIRE(acc_a1 != acc_b1);
+    // NOTE: A strict reading of the 1.2.1 Rev 7 spec, section 4.3.2 would imply
+    // that these should not be equal, as they are not copies of acc_a1.
+    BOOST_REQUIRE(acc_a1 == acc_a5);
+    BOOST_REQUIRE(acc_a1 == acc_c1);
```
illuhad, line 367: When merging this into the new runtime, I realized that these two tests cause trouble here. With the new runtime, an accessor does the following:
1. On construction, binds itself to a buffer by storing a `shared_ptr` on the host side to the `rt::data_region` (which is basically what `buffer_impl` used to be)
2. When binding to a `handler` object, stores a pointer to the `rt::requirement` handle that uniquely describes the requirement in the DAG
3. On kernel submission, this pointer is exchanged for the actual device pointer, after the scheduler has decided on which device the operation should run.

The problem is that these tests are run in command group scope (not kernel scope), so between steps 2 and 3. In that moment, device accessors for the same buffer and access range are uniquely identifiable because of the `rt::requirement` pointer that they hold. After step 3, they are again indistinguishable again.
Host accessors are different however: Since constructing a host accessor is a synchronous operation, after construction they are already after step 3) and hold a data pointer. They are hence indistinguishable from other host accessors to the same data region.

This means that for the device accessor tests, the compared accessors will be different while for host accessors they are not.

Is comparison of device accessors in command group scope important to you? This might be a grey area of the spec, as using device accessors on the host is illegal, and it is not entirely clear if this might already be a form of "use".
If this is not critical for you, I would propose to remove those tests for now - the proper fix imo would be to compare device accessors in a kernel, which could be implemented in the tests in a subsequent PR.

psalz: No, it's not at all important to me! I've chosen to do the comparison within the CGF because it was easier ;-). While reading up on accessors again for this PR I've also noticed that the whole concept of "use" is very vague, and I've been thinking about opening a corresponding issue in the spec repository.

illuhad: Thanks, I will then comment out these two tests for now in the branch for the new runtime.

> I've been thinking about opening a corresponding issue in the spec repository.

Please do :) It's not the first time this vagueness of the spec has caused some issue, e.g. [here](https://stackoverflow.com/questions/61741313/sycl-dpc-accessor-vs-global-ptr-in-kernel-function-object) it was attempted to use `accessor::get_pointer()` in command group scope, which also caused trouble..

ax3l:
@illuhad looks like this fails since #290 was merged.

PR #292, opened 2020-09-08
CMake: No Int Type

ax3l:
Fix the warning:
```
CMake Warning (dev) at CMakeLists.txt:17 (set):
```

Ref.: https://cmake.org/cmake/help/v3.18/command/set.html


illuhad:
Thanks!

PR #294, opened 2020-09-10
New implementation of CPU nd_range parallel for for new runtime with better performance

illuhad:
This adds support for nd_range parallel for on CPU to the new runtime, with a twist:

The implementation relies on `boost.fiber` (which is a new dependency once this is merged). However, work groups are initially processed only by a single fiber, which iterates across the work items. This is also the execution model at the heart of scoped and hierarchical parallelism and very efficient.
As soon as a barrier is encountered, dedicated fibers per work items are spawned automatically to guarantee that all work items can reach the barrier. The original fiber only processes the first work item onwards, and the newly spawned fibers process the other items.

This means that as long as no barriers are in `nd_range` parallel for kernels, they are now close to the performance of hierarchical and scoped parallelism. For a triad, I'm measuring between 85 and 95% percent performance of the other models - I suspect the remaining performance gap is because of slightly larger overheads (e.g. spawning the fiber).

Of course, once barriers are involved it will be not competitive performance-wise vs hierarchical and scoped because of the context switching overhead of the fibers.

Supersedes #279 and #280


illuhad:
@sbalint98 fyi, this introduces `boost.fiber` and `boost.context` as new dependencies.

keryell:
Nice, this lazy Boost.Fiber creation on demand!

illuhad:
@keryell yes, it effectively acts like an automatic version of a `NO_BARRIER` macro :)

PR #295, opened 2020-09-11
Add boost.context, boost.fiber as dependencies for the packages

sbalint98:
Adds the dependencies on boost.fiber and boost.context


install/scripts/packaging/make-ubuntu-pkg.sh
```diff
@@ -21,7 +21,7 @@ Version: ${HIPSYCL_VERSION_STRING}
 Section: base
 Priority: optional
 Architecture: amd64
-Depends: hipsycl-base${HIPSYCL_PKG_NAME_SUFFIX}  (>= 0.8), python3 (>= 3.0)
+Depends: hipsycl-base${HIPSYCL_PKG_NAME_SUFFIX}  (>= 0.8), python3 (>= 3.0), libboost-fiber1.65-dev, libboost-context1.65-dev, libboost-filesystem1.65-dev
```
keryell, line 24: Does it make sense to specifically require these old versions today?

illuhad: Same question for CentOS. Is it possible to also allow anything newer?

sbalint98: Well, for centos the default boost package is 1.53,  one can choose to install even older one (1.49) or 1.69. and this is from epel, so I am not sure if we can get anything newer using the centos repos. 

For Ubuntu, you're right de default version is 1.65 so we do not have to specify the version, and we can hope that newer version will be added. 

We could consider installing our own boost in the hipSYCL-base packages

illuhad: > Well, for centos the default boost package is 1.53, one can choose to install even older one (1.49) or 1.69. and this is from epel, so I am not sure if we can get anything newer using the centos repos.

Okay, I see.

> For Ubuntu, you're right de default version is 1.65 so we do not have to specify the version, and we can hope that newer version will be added.

That seems like a good idea to me.

> We could consider installing our own boost in the hipSYCL-base packages

hmm.. unless there are issues with the already existing packages of the distributions, I don' think we should try to maintain yet another large pile of software ;)

sbalint98: Okay, I removed the version from the Ubuntu version.

illuhad: Thank you!

PR #220, opened 2020-03-14
Implementing hidden friends

sbalint98:
Implementing spec maintenance change described in issue #173 (https://github.com/KhronosGroup/SYCL-Docs/pull/42)

PR #298, opened 2020-09-12
Don't return local variable by reference

aTom3333:
Some of the operator of `multi_ptr` return reference to local variables, this is bad.


illuhad:
Whopsie, thanks. It seems this is a regression that was introduced in PR #220. In any case I think `multi_ptr` is rarely used in hipSYCL because hipSYCL can use regular pointers just fine.

Issue #296, opened 2020-09-12
Cmake POSITION_INDEPENDENT_CODE support


aTom3333:
Adding support for cmake's `POSITION_INDEPENDENT_CODE` property for syclcc would be nice.
As of now, with the following setup I get a linker error.

CMakeLists.txt:
```cmake_minimum_required(VERSION 3.16)
project(hipsycl_pic)

set(HIPSYCL_PLATFORM cpu)
find_package(hipsycl CONFIG PATHS /path/to/hipSYCL/install/lib/cmake/)

add_library(hipsycl_pic_lib SHARED add.h add.cpp)
add_sycl_to_target(TARGET hipsycl_pic_lib SOURCES add.cpp)

set_target_properties(hipsycl_pic_lib PROPERTIES
    POSITION_INDEPENDENT_CODE ON
)

include(GenerateExportHeader)
generate_export_header(hipsycl_pic_lib
        BASE_NAME mylib
        )
target_include_directories(hipsycl_pic_lib PUBLIC ${CMAKE_CURRENT_BINARY_DIR})
```

add.h:
```#ifndef ADD_H
#define ADD_H

#include "mylib_export.h"

#ifdef __cplusplus
extern "C" {
#endif

void MYLIB_EXPORT add(float* a, float* b, float* out, unsigned int n);
    
#ifdef __cplusplus
};
#endif
#endif // ADD_H
```

add.cpp:
```#include "add.h"
#include <CL/sycl.hpp>

using namespace cl::sycl;
void add(float* a, float* b, float* out, unsigned int n)
{
    static queue q;
    buffer<float> aBuf(a, n);
    buffer<float> bBuf(b, n);
    buffer<float> outBuf(out, n);
    q.submit([&](handler& h){
        auto aAcc = aBuf.get_access<access::mode::read>(h);
        auto bAcc = bBuf.get_access<access::mode::read>(h);
        auto outAcc = outBuf.get_access<access::mode::write>(h);
        h.parallel_for<class sycl_kernel>(range<1>(n), [=](id<1> id) {
            outAcc[id] = aAcc[id] + bAcc[id];
        });
    });
}
```

Resulting linker error:
[...]

PR #299, opened 2020-09-13
Inform cmake how to produce PIC code with the SYCL compiler

aTom3333:
Relates to #296 
Easier than I thought but I'm not sure that is the proper way to do it, seems to work fine


illuhad:
Thank you!
@psalz Can you have a look?

psalz:
The various files for the SYCL language defined here try to follow the same structure as the CXX language that comes predefined with CMake. However these mechanisms aren't well documented, so I wasn't always sure which file does what, when (and it's been a while, so I remember even less of it now). I've so far adopted a lazy approach, only porting flags from CXX to SYCL when something doesn't work.

That being said, if you take a look at your `<path/to/cmake>/Modules/CMakeCXXInformation.cmake`, you can see that the PIC-related flags are set there, so for us they should also go into `CMakeSYCLInformation.cmake` (I've tried to keep them in the same order as in the CXX file, so I think [line 184](https://github.com/illuhad/hipSYCL/blob/6c06179b5400d04150fd84926bf25d1d133b7dc5/cmake/CMakeSYCLInformation.cmake#L184) would be a good place to add this). Note that there are several additional flags for compilation and linking related to PIC/PIE defined there, so if you want to add this feature, I think it'd be best to copy all flags. Additionally, instead of hard-coding flags, so far I've usually copied the value from the corresponding `CXX` variable, for the off chance that a particular SYCL compiler uses a different flag (unlikely, but still). The `CXX` flags mostly do the same, copying the corresponding values from their `C` counterparts.

aTom3333:
You're right, I don't how I missed the various the lines
```
if(NOT CMAKE_CXX_COMPILE_OPTIONS_PIC)
  set(CMAKE_CXX_COMPILE_OPTIONS_PIC ${CMAKE_C_COMPILE_OPTIONS_PIC})
endif()
```
and the like when looking through cmake source, I just found the definition in the compiler files (like in `<path/to/cmake>/Modules/Compiler/GNU.cmake`)
```
set(CMAKE_${lang}_COMPILE_OPTIONS_PIC "-fPIC")
```
As such i thought of those flags to be more related to the compiler than the language.

I'll make the change to forward to the CXX options and include the other PIE related ones.

psalz:
LGTM!

illuhad:
Thanks guys :)

PR #300, opened 2020-09-13
Don't discard HIPSYCL_PLATFORM on the first run

### LINKED ISSUE #275: HIPSYCL_PLATFORM not settable in CMakeLists.txt
Setting `-DHIPSYCL_PLATFORM=cpu` on the command line works fine but using `set(HIPSYCL_PLATFORM "cpu")` in the CMakeLists.txt file gives the following error:
```CMake Error at /opt/hipSYCL/lib/cmake/CMakeSYCLInformation.cmake:53 (message):
  More than one hipSYCL target platform is available.

  Please specify HIPSYCL_PLATFORM=cpu|cuda|rocm
  /opt/hipSYCL/lib/cmake/hipsycl-config.cmake:9 (enable_language)
  CMakeLists.txt:15 (find_package)
```
I could reproduce this issue using the [example](https://github.com/illuhad/hipSYCL/blob/master/doc/examples.md) from the docs. If required, I can provide the CMakeLists.txt used.

aTom3333:
Fixes #275

### DIFF
--- a/cmake/CMakeSYCLInformation.cmake
+++ b/cmake/CMakeSYCLInformation.cmake
@@ -39,7 +39,7 @@ if(HIPSYCL_ROCM_BACKEND_AVAILABLE)
   endif()
 endif()
 
-set(HIPSYCL_PLATFORM "" CACHE STRING "The platform that hipSYCL should target. One of ${HIPSYCL_PLATFORMS_STRING}.")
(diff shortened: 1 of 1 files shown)


illuhad:
Thanks! 
@psalz This looks uncontroversial to me, but as resident cmake wizard you might want to look at it anyway ;)

psalz:
Yes, I think that should be fine! Note that it was previously already possible to set the platform from within a CMake config, by using `set(HIPSYCL_PLATFORM "cpu" CACHE STRING "" FORCE)`.

PR #305, opened 2020-09-15
Change the add_sycl_to_target cmake function to link with the new runtime

aTom3333:
Make `add_sycl_to_target` with new runtime. Client cmake is unchanged

### DIFF
--- a/cmake/hipsycl-config.cmake.in
+++ b/cmake/hipsycl-config.cmake.in
@@ -21,13 +21,7 @@ function(add_sycl_to_target)
     ${ARGN}
   )
 
-  if(HIPSYCL_PLATFORM_CANONICAL STREQUAL "cpu")
-    target_link_libraries(${ADD_SYCL_TARGET} PUBLIC hipSYCL::hipSYCL_cpu)
-  elseif(HIPSYCL_PLATFORM_CANONICAL MATCHES "cuda")
-    target_link_libraries(${ADD_SYCL_TARGET} PUBLIC hipSYCL::hipSYCL_cuda)
-  else()
-    target_link_libraries(${ADD_SYCL_TARGET} PUBLIC hipSYCL::hipSYCL_rocm)
-  endif()
+  target_link_libraries(${ADD_SYCL_TARGET} PUBLIC hipSYCL::hipSYCL-rt)
 
   # For now we have to link using syclcc, as CMake doesn't know about all
   # the required libraries that need to be linked with the target.


illuhad:
Ah, thank you. It seems I forgot that :) Why is it actually necessary to link against the hipSYCL libraries from cmake at all? `syclcc` should automatically link any hipSYCL libraries when a linking operation takes place.
Thinking about it, I think this might be why have seen issues like it trying to link both `hipSYCL_cuda` and `hipSYCL_cpu`. This might happen when the CMake cache says something different than syclcc settings..

aTom3333:
That's a good question, I had errors because of cmake trying to link againt `hipSYCL_cpu`/`hipSYCL_cuda` so I went straight to replace that with `hipSYCL-rt` without testing without it.

I just tried without and I get an error at runtime `error while loading shared libraries: libhipSYCL-rt.so: cannot open shared object file: No such file or directory` with `HIPSYCL_PLATFORM=cpu`, it works fine with `HIPSYCL_PLATFORM=cuda`. The problem is that `-rpath` isn't given by the `pure_cpu_compiler`, I can fix that.

What do you prefer, fixing syclcc and not linking in cmake? Still link in cmake even if it is redundant? (imo linking in cmake is still useful if someone look at the cmake files to see what is being linked even if otherwise useless but it is ultimately as you prefer)

illuhad:
Thanks for investigating :) The reason this works for CUDA/HIP backends and not for CPU is because for CPU we support in principle any OpenMP compiler. But `-rpath` is not as standardized as basic linker options like `-L`, so it might be called differently in different compilers.
Therefore, since we cannot know how to do with in every conceivable compiler, the `pure_cpu_compiler` currently only adds `-rpath` if the compiler is detected as clang. For CUDA and HIP the compiler is always clang, so we always know how to `rpath`.

So, I would say that this runtime error is expected at the moment and unrelated to this. The proper fix I think would be to introduce a way to customize the CPU link line in cmake as we have for ROCm and CUDA (see e.g. `CUDA_LINK_LINE` cmake variable). Then we can just have `-rpath` by default in there and the user can change it if necessary.

As a workaround users can of course always use `LD_LIBRARY_PATH`. I'm inclined to think that adjusting `LD_LIBRARY_PATH`, while a slight inconvenience, would probably be easy and obvious for most users.

At the moment I tend towards not linking using cmake, and separately introducing `CPU_LINK_LINE` so that we can have `rpath` with `syclcc` for the CPU backend.

@psalz, is there anything we are missing that requires linking those libraries from cmake?

illuhad:
It seems this is superseded by PR #309

aTom3333:
I guess it is, I haven't even had time to look at that

aTom3333:
Closing cause #309 seems superior

PR #313, opened 2020-09-18
Only require LLVM and build clang plugin for GPU backends

illuhad:
This removes the dependency on LLVM and clang if only the CPU backend is built. The clang plugin is now also only compiled when either CUDA or ROCm backends are enabled (which need it).

This can significantly simplify the deployment of hipSYCL on systems where GPU support is not necessary, unavailable or undesired, because the CPU backend is a pure C++ library and therefore easy to build.

I believe our Mac users will like this ;) @jeffhammond @ax3l


jeffhammond:
Now all you gotta do is add an OpenMPtarget GPU backend :-)

illuhad:
@jeffhammond I have some plans regarding that ;)

jeffhammond:
@illuhad If you want to be really crazy, implement a Kokkos back-end.  It will likely be easier than OpenMP target and will get you a bunch of targets, although not too much more than you have right now.

illuhad:
It's an interesting idea, but I'm not sure if that brings much benefit. My understanding is that in order to have all Kokkos backends users need to mark their lambdas as `KOKKOS_LAMBDA` to make CUDA/HIP happy which need `__device__` attributes here. To make that work with SYCL where such attributes are not necessary we would need dedicated compiler support (the same thing we already have for our CUDA/HIP backends basically). If we just pass regular Lambdas to Kokkos we won't be able to target Kokkos CUDA/HIP backends and we would basically get OpenMP support. 
And I suppose we have more control for optimization in that case if we go directly to OpenMP instead of funneling it through Kokkos.

illuhad:
Alright, building (CPU-only) on Mac should now be as easy as `brew install cmake libomp boost; cmake $SRC; make install`

ax3l:
Wuhu, nice one!

PR #318, opened 2020-09-23
Restructure sycl interface, remove HIP and hipCPU submodules, enable simultaneous compilation for CPU and GPU

illuhad:
This PR

* moves all parts of the SYCL interface that need to be compiled for device into `libkernel`. Some work has started to separate different backends in `libkernel`. My vision is that ultimately, we want to have a directory for common code shared between all backends and individual directories for backend-specific code in `libkernel` (similarly to the runtime). This should simplify developing new backends.
* Removes old and unused macros
* Does some cleanup and, as a consequence, removes the now unnecessary HIP and hipCPU submodules
* Moves away from the static compilation model: The host pass of the clang CUDA/HIP backends can now in principle be used to target CPUs, so that a SYCL program can have both CPU and GPU available
   - this is not yet exposed in `syclcc`, so if you want to experiment with this you need to construct your compiler invocation manually. Target the GPU backend and add `-fopenmp -D__HIPSYCL_ENABLE_HOST_OMP_TARGET__`.
   - there seems to be a conflict between the `-fopenmp` flag and the clang cuda wrapper headers - this is a clang toolchain issue. I'm not sure if it affects CUDA as a whole or just my combination of clang 10 with libstdc++ 10. HIP seems to work fine.
   - Note that this introduces a subtle change of meaning to the `HIPSYCL_PLATFORM_CPU/CUDA/ROCM` macros: They are no longer fully exclusive. For example, if you are in the host pass of the CUDA backend, both `HIPSYCL_PLATFORM_CPU` and `HIPSYCL_PLATFORM_CUDA` are set. Rather these macros should now be interpreted as whether the compiler understands native C++/CUDA/ROCm constructs. To make this clearer, there are also new macros of the form `HIPSYCL_LIBKERNEL_COMPILER_SUPPORTS_CUDA/HIP/HOST` and `HIPSYCL_LIBKERNEL_IS_DEVICE_PASS_CUDA/HIP/HOST`. Those are always defined to either 0 or 1.


include/hipSYCL/glue/generic/hiplike/hiplike_kernel_launcher.hpp
```diff
@@ -30,24 +30,27 @@
...
+
 
 namespace hipsycl {
 namespace glue {
 
 namespace hiplike_dispatch {
 
 template<int dimensions>
-__device__
+__host__ __device__
```
fodinabor, line 53: any reason for not using `HIPSYCL_KERNEL_TARGET` here?

illuhad: It's not really necessary. `HIPSYCL_KERNEL_TARGET` is there to share code between backends that require specific attributes (looking at you, CUDA/HIP) and backends that don't. Since the kernel launchers are the heart of backend-specific kernel launching logic, we don't need to be more general here and can instead just use directly the attributes that we need.

PR #336, opened 2020-10-13
 Added safer patterns to the install scripts

### LINKED ISSUE #333: safer patterns in install scripts - part 1
I am of the view that `rm -rf`-ing something in a user's home directory is not an appropriate thing to do without explicit confirmation.

A better pattern would be to check for the existence of the directory and only use it if it does not already exist.

https://github.com/illuhad/hipSYCL/blob/develop/install/scripts/install-llvm.sh
```sh
set -e
BUILD_DIR=$HOME/git/llvm-vanilla
rm -rf $BUILD_DIR

echo "Cloning LLVM $HIPSYCL_PKG_LLVM_REPO_BRANCH"
git clone -b $HIPSYCL_PKG_LLVM_REPO_BRANCH https://github.com/llvm/llvm-project $BUILD_DIR
```

### LINKED ISSUE #334: safer patterns in install scripts - part 2
hipSYCL install scripts use `BUILD_DIR` and `INSTALL_PREFIX`, which default to `/opt/hipsycl`.  On machines where that directory isn't writeable, I'd like to be able to define a global environment variable where stuff should go, but I don't want to give these particular variables global scope because other code might use them.

It would be ideal if all hipSYCL environment variables like this were named `HIPSYCL_*`.

sbalint98:
This PR should fix issues #334  and #333 

#333 
* install scripts the `INSTALL_PREFIX` varibale has been changed to `HIPSYCL_INSTALL_PREFIX`
* The `BUILD_DIR` variable has been renamed to `HIPSYCL_BUILD_DIR` in the `install-hipsycl.sh` and to `HIPSYCL_<ROCM/LLVM>_BUILD` for the other two files. This distinction should prevent confusion when building multiple componenets. 

#334 
The `rm -rf ` has been removed from all the scipts. Now if the install directory is present the user is asked if they want to use the directory if the user does not want to use that director the script exits, otherwise, a correct clone of the given software is assumed and the script proceeds with configuration.

### DIFF
--- a/install/scripts/install-hipsycl.sh
+++ b/install/scripts/install-hipsycl.sh
@@ -4,38 +4,48 @@ HIPSYCL_PKG_LLVM_VERSION_MINOR=${HIPSYCL_PKG_LLVM_VERSION_MINOR:-0}
 HIPSYCL_PKG_LLVM_VERSION_PATCH=${HIPSYCL_PKG_LLVM_VERSION_PATCH:-1}
 HIPSYCL_PKG_LLVM_REPO_BRANCH=${HIPSYCL_PKG_LLVM_REPO_BRANCH:-release/${HIPSYCL_PKG_LLVM_VERSION_MAJOR}.x}
 
-export INSTALL_PREFIX=${INSTALL_PREFIX:-/opt/hipSYCL}
+export HIPSYCL_INSTALL_PREFIX=${HIPSYCL_INSTALL_PREFIX:-/opt/hipSYCL}
 
 set -e
-BUILD_DIR=/tmp/hipSYCL-installer
-
+HIPSYCL_BUILD_DIR=${HIPSYCL_BUILD_DIR:-/tmp/hipSYCL-installer}
 HIPSYCL_REPO_USER=${HIPSYCL_REPO_USER:-illuhad}
 HIPSYCL_REPO_BRANCH=${HIPSYCL_REPO_BRANCH:-stable}
 HIPSYCL_WITH_CUDA=${HIPSYCL_WITH_CUDA:-ON}
 HIPSYCL_WITH_ROCM=${HIPSYCL_WITH_ROCM:-ON}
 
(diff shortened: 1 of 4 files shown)


illuhad:
Thank you!

install/scripts/install-hipsycl.sh
```diff
@@ -4,38 +4,48 @@ HIPSYCL_PKG_LLVM_VERSION_MINOR=${HIPSYCL_PKG_LLVM_VERSION_MINOR:-0}
...
--DROCM_PATH=$INSTALL_PREFIX/rocm \
--DCUDA_TOOLKIT_ROOT_DIR=$INSTALL_PREFIX/cuda \
--DCLANG_EXECUTABLE_PATH=$INSTALL_PREFIX/llvm/bin/clang++ \
+-DLLVM_DIR=$HIPSYCL_INSTALL_PREFIX/llvm/lib/cmake/llvm \
+-DROCM_PATH=$HIPSYCL_INSTALL_PREFIX/rocm \
+-DCUDA_TOOLKIT_ROOT_DIR=$HIPSYCL_INSTALL_PREFIX/cuda \
+-DCLANG_EXECUTABLE_PATH=$HIPSYCL_INSTALL_PREFIX/llvm/bin/clang++ \
 -DCLANG_INCLUDE_PATH=$LLVM_INCLUDE_PATH \
--DCMAKE_INSTALL_PREFIX=$INSTALL_PREFIX \
+-DCMAKE_HIPSYCL_INSTALL_PREFIX=$HIPSYCL_INSTALL_PREFIX \
```
illuhad, line 48: This seems like an auto replace error ;) This should have stayed `CMAKE_INSTALL_PREFIX`. Since cmake doesn't know `CMAKE_HIPSYCL_INSTALL_PREFIX`, install prefix now cannot be set anymore. @sbalint98 can you fix that? (or maybe @jeffhammond wants to quickly submit a PR?)

Issue #315, opened 2020-09-21
CMake: Define semantics when CMAKE_CXX_COMPILER is not Clang-compatible


fknorr:
We currently don't have a concept how to deal with a heterogeneous compiler environment, e.g. with MSVC set as the "host compiler" set via `CMAKE_CXX_COMPILER` and clang being invoked through syclcc. CMake identifies the "compiler id" during its configuration phase, and since syclcc is now largely invisible to CMake, it assumes the same properties for SYCL targets and passes the same command line arguments. This will break in the heterogeneous case since Clang does not accept MSVC flags. I realize that MSVC is not supported currently, but I believe we should define the semantics in general.

When compiling against the CPU backend, we could pass `CMAKE_CXX_COMPILER` as the `--cpu-compiler` to syclcc. But when building against CUDA or HIP, we need to ensure CMake is passing GNU-compatible flags. CMake does not support selecting multiple compilers for the same language (e.g. on a per-target basis), so the simplest option I see is to have CMake error out when `HIPSYCL_PLATFORM=hip|cuda` is used with a GNU-incompatible `CMAKE_CXX_COMPILER_ID`.

This was not supported in the CMake setup before #309 either since all flags are simply copied from CXX, although having different compiler would have been possible with the additional SYCL language. I would argue that enforcing a compatible compiler in the front-end is a sane decision.

PR #337, opened 2020-10-13
Restructure syclcc to expose simultaneous compilation for multiple backends

illuhad:
* Restructure `syclcc` to not have independent compiler drivers for different backends; instead different backends just add compiler flags to a common command line. This allows us in principle to compile for multiple backends. Note that there are some limitations to what works in practice:
   * HIP cannot work with CUDA because of clang limitations
   * CUDA won't work with OpenMP because of a clang bug
* `HIPSYCL_PLATFORM, HIPSYCL_GPU_ARCH` are deprecated and replaced by `HIPSYCL_TARGETS/--hipsycl-targets` which allows specification of multiple backends or targets, e.g. `--hipsycl-targets="omp;hip:gfx900,gfx906"` would compile for OpenMP as well as Vega 10 and Vega 20.
   * The old `HIPSYCL_PLATFORM/HIPSYCL_GPU_ARCH` still work, so build systems and our cmake integration should not be immediately affected. In the mid term, we should however switch our cmake integration to the new `HIPSYCL_TARGETS` mechanism. @psalz, @fknorr can you maybe comment on this? Are there any obstacles to be aware of?
* Even when not explicitly compling for OpenMP, `syclcc` now automatically enables a minimal sequential variant of the OpenMP backend (basically OpenMP without `-fopenmp`). This will allow us to implement SYCL 2020 host tasks

### DIFF
--- a/bin/syclcc-clang
+++ b/bin/syclcc-clang
@@ -36,10 +36,6 @@ import sys
 import subprocess
 import string
 
-class hipsycl_platform:
-  PURE_CPU = "cpu"
-  CUDA = "cuda"
-  HIP = "hip"
 
 class OptionNotSet(Exception):
   def __init__(self, msg):
@@ -130,7 +126,7 @@ class syclcc_config:
     # 3.) the field in the config file.
     self._options = {
       'platform': option("--hipsycl-platform", "HIPSYCL_PLATFORM", "default-platform",
-"""  The platform that hipSYCL should target. Valid values: 
+"""  [deprecated] The platform that hipSYCL should target. Valid values: 
     * cuda: Target NVIDIA CUDA GPUs
     * rocm: Target AMD GPUs running on the ROCm platform
     * cpu: Target only CPUs"""),
@@ -147,7 +143,7 @@ class syclcc_config:
 """  The path to the ROCm installation directory"""),
 
       'gpu-arch': option("--hipsycl-gpu-arch", "HIPSYCL_GPU_ARCH", "default-gpu-arch",
-"""  The GPU architecture that should be targeted when compiling for GPUs.
+"""  [deprecated] The GPU architecture that should be targeted when compiling for GPUs.
     For CUDA, the architecture has the form sm_XX, e.g. sm_60 for Pascal.
     For ROCm, the architecture has the form gfxYYY, e.g. gfx900 for Vega 10, gfx906 for Vega 20."""),
 
@@ -165,7 +161,10 @@ class syclcc_config:
 
       'config-file' : option("--hipsycl-config-file", "HIPSYCL_CONFIG_FILE", "default-config-file",
 """  Select an alternative path for the config file containing the default hipSYCL settings.
-    It is normally not necessary for the user to change this setting.""")
+    It is normally not necessary for the user to change this setting. """),
+    
+      'targets': option("--hipsycl-targets", "HIPSYCL_TARGETS", "default-targets",
+"""  Targets to compile for. Example: --hipsycl-targets='omp;hip:gfx900,gfx906'""")
     }
     self._flags = {
       'is-dryrun': option("--hipsycl-dryrun", "HIPSYCL_DRYRUN", "default-is-dryrun",
@@ -175,6 +174,7 @@ class syclcc_config:
     self._insufficient_cpp_standards = ['98', '03', '11', '14', '0x']
     self._hipsycl_args = []
     self._forwarded_args = []
+    self._targets = None
     
     for arg in self._args:
       if self._is_hipsycl_arg(arg):
@@ -344,26 +344,58 @@ class syclcc_config:
             raise RuntimeError("Insufficient c++ standard '{}'".format(std_args[0]))
         return []
 
+  def _parse_targets(self, target_arg):
+    platform_substrings = target_arg.replace("'","").replace('"',"").split(';')
+    
+    result = {}
+    for p in platform_substrings:
+      platform_target_separated = p.split(':')
+      if len(platform_target_separated) > 2 or len(platform_target_separated) == 0:
+        raise RuntimeError("Invalid target description: " + p)
+      
+      platform = platform_target_separated[0].strip().lower()
+
+      if not platform in result:
+        result[platform] = []
+
+      if len(platform_target_separated) > 1:
+        targets = [t.strip().lower() for t in platform_target_separated[1].split(",")]
+        for t in targets:
+          if not t in result[platform]:
+            result[platform].append(t)
+
+    return result
+  
   @property
-  def target_platform(self):
-    platform = self._retrieve_option("platform")
+  def targets(self):
+    
+    if self._targets == None:
+      raw_target_string = ""
+      try:
+        raw_target_string = self._retrieve_option("targets")
+      except OptionNotSet:
+        # Legacy compatibility args
+        platform = self._retrieve_option("platform")
 
-    hip_platform_synonyms      = set(["rocm", "amd", "hip"])
-    cuda_platform_synonyms     = set(["nvidia", "cuda"])
-    pure_cpu_platform_synonyms = set(["host", "cpu", "hipcpu"])
+        hip_platform_synonyms      = set(["rocm", "amd", "hip"])
+        cuda_platform_synonyms     = set(["nvidia", "cuda"])
+        pure_cpu_platform_synonyms = set(["host", "cpu", "hipcpu", "omp"])
 
-    if platform in hip_platform_synonyms:
-      return hipsycl_platform.HIP
-    elif platform in cuda_platform_synonyms:
-      return hipsycl_platform.CUDA
-    elif platform in pure_cpu_platform_synonyms:
-      return hipsycl_platform.PURE_CPU
 
-    raise RuntimeError("Invalid hipSYCL platform: '{}'".format(platform))
+        if platform in hip_platform_synonyms:
+          target_arch = self._retrieve_option("gpu-arch")
+          raw_target_string = "hip:" + target_arch
+          
+        elif platform in cuda_platform_synonyms:
+          target_arch = self._retrieve_option("gpu-arch")
+          raw_target_string = "cuda:" + target_arch
+          
+        elif platform in pure_cpu_platform_synonyms:
+          raw_target_string = "omp"
 
-  @property
-  def target_arch(self):
-    return self._retrieve_option("gpu-arch")
+      self._targets = self._parse_targets(raw_target_string)
+
+    return self._targets
 
   @property
   def cuda_path(self):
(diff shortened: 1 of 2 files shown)


fknorr:
With this change in place, would it be possible to have a `syclcc` flag to emit compiler/linker flags on stdout instead of invoking Clang?

@psalz and I realized last week that the CMake integration could be greatly simplified by setting the appropriate compiler and linker flags in the CMake script and calling Clang directly instead of deferring the invocation to `syclcc`. However, we did not really have any idea how to do this without duplicating the `syclcc` logic in the CMake script.

If such a flag were implemented, CMake could invoke `syclcc` once during its configuration step to fill its `CXX_FLAGS` / `LINKER_FLAGS` variables and then call Clang directly for the compilation step, allowing us to get rid of the compiler launcher workaround and additional dependency tracking introduced in #309 and #317. We would then require the `CMAKE_CXX_COMPILER` to be Clang whenever `add_sycl_to_target` is used, resolving #315 in the process.

illuhad:
We have always had `--hipsycl-dryrun` / `HIPSYCL_DRYRUN` which prints the compiler invocation instead of executing it. This also works with this PR. Would we need something in addition?

I still wonder whether we really want to make such strong guarantees as you propose regarding the behavior of `syclcc`. In the future it might be interesting to add compilation modes where `syclcc` does more complex things with the toolchain. For example, manually compiling the kernels for each backend with separate compiler invocations and embedding them in the host binary with a custom mechanism.

bin/syclcc-clang
```diff
@@ -448,135 +494,239 @@ def run_or_print(command, print_only):
...
+      "-I" + os.path.join(self._hipsycl_path, "include/"),
+      "-D__HIPSYCL__"
+    ] + self._common_compiler_args
+
+    return args
+
+  @property
+  def common_linker_flags(self):
+    linker_args = [
+      "-Wl,-rpath", self._hipsycl_lib_path,
```
sbalint98: I think this line should be `"-Wl,-rpath=" + self._hipsycl_lib_path,` because currently syclcc emits arguments in the following form: ` -Wl,-rpath /opt/hipSYCL/bin/../lib/` It seems that clang (10.0.1) does not forward the path, but only the `-rpath` to the linker causing an error, since clang tries to compile `self._hipsycl_lib_path`

fodinabor: :+1:
and for some reason, I also had to add quotes around the parameter...

psath:
I'm not by any means a cmake expert, but I've taken a stab at adding support for HIPSYCL_TARGETS in #362, which is working on my OpenMP+gfx803+gfx900 project.

bin/syclcc-clang
```diff
@@ -448,135 +494,239 @@ def run_or_print(command, print_only):
...
+      ]
+      
+    return flags
+
+  def get_linker_flags(self):
+    linker_args = [
       "-lboost_context",
-      "-lboost_fiber"
+      "-lboost_fiber",
+      "-lomp"
```
psath, line 644: While temporary, this presumes the LLVM implementation of OpenMP, which is not by default included in some LLVM/Clang installs, and will likely conflict if the user selects something other than Clang as the host compiler. As a primarily ROCm user, I'm not expecting a SYCL implementation to make any particular demands of the OpenMP implementation just to run GPU code, though GCC 8 should suffice for the C++17 OpenMP (4.5 at least) requirements.

I'm working on a patch to infer the OpenMP flags that correspond with whatever `cpu-compiler` is set to. Though it could also be implemented as an explicit parameter if there is a strong need to have it differ from the one corresponding to the host compiler.

illuhad: > While temporary, this presumes the LLVM implementation of OpenMP, which is not by default included in some LLVM/Clang installs, and will likely conflict if the user selects something other than Clang as the host compiler

No - this is the omp sequential backend which is not actually used when compiling for CPU. It is used when compiling for *GPU* backends so that we are guaranteed that we always have a CPU backend available that can for example run SYCL 2020 host tasks. Effectively it is the OpenMP backend without `-fopenmp`.

The reason we don't use the regular OpenMP backend here is that there is a bug in clang CUDA that prevents `clang -x cuda -fopenmp` from working.

Because it is only used when also compiling for GPU, we are always guaranteed that we are using clang.

>  I'm not expecting a SYCL implementation to make any particular demands of the OpenMP implementation just to run GPU code

hipSYCL requires OpenMP always and this is enforced by cmake. As mentioned previously even when you just want to run on GPU hipSYCL still needs a CPU backend, for example for SYCL 2020 host tasks. Or for the case when a user requests data copies between two host accessors. This is also why you cannot build hipSYCL without the CPU backend enabled.

My plan for the future is to add something like a TBB backend that we can use as default CPU backend instead of openmp sequential, which is basically a hack.

psath: Hmmm

> > While temporary, this presumes the LLVM implementation of OpenMP, which is not by default included in some LLVM/Clang installs, and will likely conflict if the user selects something other than Clang as the host compiler
> 
> No - this is the omp sequential backend which is not actually used when compiling for CPU. It is used when compiling for _GPU_ backends so that we are guaranteed that we always have a CPU backend available that can for example run SYCL 2020 host tasks. Effectively it is the OpenMP backend without `-fopenmp`.

So I can see that *if* OpenMP is selected as one of the targets, it uses the full `-fopenmp` rather than getting the link flag from `omp_sequential_invocation`. But even then it does not use the OpenMP of the user-requested host compiler, at least in my testing. `fopenmp` always presumes `libomp` unless I modify it to `-fopenmp=libgomp`, which the HIP backend then accepts. Thus the idea to pin the OpenMP flags to match the host compiler.

> 
> The reason we don't use the regular OpenMP backend here is that there is a bug in clang CUDA that prevents `clang -x cuda -fopenmp` to work.

This makes perfect sense until that upstream clang bug is resolved. Until then LLVM OpenMP probably needs to be added as an explicit dependency for the CUDA backend.
> 
> Because it is only used when also compiling for GPU, we are always guaranteed that we are using clang.
> 
> > I'm not expecting a SYCL implementation to make any particular demands of the OpenMP implementation just to run GPU code
> 
> hipSYCL requires OpenMP always and this is enforced by cmake. As mentioned previously even when you just want to run on GPU hipSYCL still needs a CPU backend, for example for SYCL 2020 host tasks. Or for the case when a user requests data copies between two host accessors. This is also why you cannot build hipSYCL without the CPU backend enabled.

Yeah the need for a host implementation is explicit in the SPEC, so the need for OpenMP in lieu of TBB, pthreads, etc makes sense. Cmake is finding my GCC OpenMP so it's not failing, but not identifying the lack of LLVM OpenMP. Paths corresponding to the GCC implementation don't get forwarded to clang during the GPU passes, so it chokes looking for headers.

> 
> My plan for the future is to add something like a TBB backend that we can use as default CPU backend instead of openmp sequential, which is basically a hack.

Yep, no worries, the comment above `omp_sequential_invocation` makes this clear.

illuhad: >  So I can see that if OpenMP is selected as one of the targets, it uses the full -fopenmp rather than getting the link flag from omp_sequential_invocation. But even then it does not use the OpenMP of the user-requested host compiler, at least in my testing. fopenmp always presumes libomp unless I modify it to -fopenmp=libgomp, which the HIP backend then accepts. Thus the idea to pin the OpenMP flags to match the host compiler.

Okay - I can see the use case of maybe wanting to use clang with `libgomp`. But this is a truly new feature as `syclcc` has always just used `-fopenmp` so I don't think anything has changed here with this PR.
As far as I'm aware @sbalint98 has started some work to expose the compiler cxx flags and link lines for all backends (similarly to e.g. `CUDA_LINK_LINE` that we have already). Then the user could change this to `-fopenmp=libgomp` if gomp should be used. Do you think that there needs to be something in addition to that?

I'm not sure if just using the OpenMP library of the host compiler is the right thing to do. The user might have different requirements when only targeting CPU (which would then invoke the host compiler) vs when also compiling for GPU. For example, a user might want to use Intel C++ and Intel OpenMP for the best CPU performance when targeting CPUs exclusively, but when also compiling for GPUs, good compatibility with clang might be more important.

> Cmake is finding my GCC OpenMP so it's not failing, but not identifying the lack of LLVM OpenMP.

hm.. yeah maybe we should require LLVM OpenMP specifically in cmake when a GPU backend is enabled.

psath: > > So I can see that if OpenMP is selected as one of the targets, it uses the full -fopenmp rather than getting the link flag from omp_sequential_invocation. But even then it does not use the OpenMP of the user-requested host compiler, at least in my testing. fopenmp always presumes libomp unless I modify it to -fopenmp=libgomp, which the HIP backend then accepts. Thus the idea to pin the OpenMP flags to match the host compiler.
> 
> Okay - I can see the use case of maybe wanting to use clang with `libgomp`. But this is a truly new feature as `syclcc` has always just used `-fopenmp` so I don't think anything has changed here with this PR.
> As far as I'm aware @sbalint98 has started some work to expose the compiler cxx flags and link lines for all backends (similarly to e.g. `CUDA_LINK_LINE` that we have already). Then the user could change this to `-fopenmp=libgomp` if gomp should be used. Do you think that there needs to be something in addition to that?
> 
> I'm not sure if just using the OpenMP library of the host compiler is the right thing to do. The user might have different requirements when only targeting CPU (which would then invoke the host compiler) vs when also compiling for GPU. For example, a user might want to use Intel C++ and Intel OpenMP for the best CPU performance when targeting CPUs exclusively, but when also compiling for GPUs, good compatibility with clang might be more important.

Before merging this PR, I was able to build for the OpenMP and ROCm backend with no modification of flags without having LLVM OpenMP. I haven't done any diagnosing to figure out what has changed in terms of the underlying clang invocations though.

Exposing the compile and link flags makes good sense I think. More flexibility lets the end user do deployment-specific workarounds in weird cases like mine. Either way, compatibility with at least clang, gcc, and icc would definitely help the flexibility of our deployment. (We even have some optional components built for PGI that fallback on OpenMP if OpenACC isn't available, so ensuring consistency of the OpenMP implementation  across separately-compiled components is a high priority.)

illuhad: > Before merging this PR, I was able to build for the OpenMP and ROCm backend with no modification of flags without having LLVM OpenMP. I haven't done any diagnosing to figure out what has changed in terms of the underlying clang invocations though.

What this PR changes is that kernels are now always also compiled for CPU, even if only targeting GPU explicitly (in order to meet the specification's requirement to have the host device always available and working). Previously, CPU was only available if (and only if) you explicitly target CPU.

When testing without this PR, you have presumably been using `g++` as CPU compiler, in which case `-fopenmp` just uses libgomp. But if you try using clang as host compiler, you will get the same behavior as we are seeing now because it also just compiles with `-fopenmp`. So, compiling for CPU with clang but using libgomp is something that has never worked before. You might just not have noticed it because only now do we also need the CPU backend when compiling for GPUs, which always requires compilation with clang, so you cannot get around clang-based compilation for CPU anymore.

bin/syclcc-clang
illuhad: I can reproduce the issue on my Ubuntu machine here, so I switched to `-rpath=`, but I'm pretty sure that it worked on my Arch Linux machine (but the issue might not trigger depending on argument order, so maybe I just didn't notice).

@fodinabor I don't understand why we would need to add quotes. Is this something that is maybe specific to some linkers?

fodinabor: oh sorry.. didn't update.. should be fine this way.. just had to add quotes on the terminal to indicate where the parameter ends, that is not necessary when invoking a process using a parameter array, so should be fine!

illuhad: I see, thanks!

fodinabor:
Just as a note, clang 11.0.1 and 12 will have the issue fixed that prevented OpenMP and CUDA backends to be used simultaneously. Therefore (if we keep OMP as backend) it might make sense to introduce some clang version checks and allow this the combination for those versions.

Both llvm branches master and release/11.x have the required fix included since tonight.

illuhad:
Thanks for letting me know :) I'm not sure if we really need the clang version check, as `syclcc` doesn't prevent you from running those combinations. If you use `--hipsycl-targets=omp;cuda:sm_60` it will just do as it is told. Until now this would just fail because of the clang bug.

fodinabor:
would be okay for me in general, but currently there is this [check](https://github.com/illuhad/hipSYCL/blob/develop/bin/syclcc-clang#L664), which throws, so I thought syclcc would error out in this case?

illuhad:
Oh, I forgot about this. We can just remove that check. I don't think we need a clang version check. If this configuration is in principle supported it is clang's responsibility to get it right and not ours. If the user explicitly requests OpenMP and CUDA we should just give it to him :)

bin/syclcc-clang
sbalint98: linking Issue #437 here

PR #338, opened 2020-10-13
Fix several memory access bugs detecting using address sanitizer

illuhad:
* typo in unit tests causing invalid access to `item` element
* illegal access to `context` object in USM operations after it has vanished from stack caused by reference to temporary object

### DIFF
--- a/tests/sycl/item.cpp
+++ b/tests/sycl/item.cpp
@@ -70,7 +70,7 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(item_api, _dimensions, test_dimensions::type) {
             auto id3 = item.get_id();
             if(d >= 1) id3[0] = item.get_id(0);
             if(d >= 2) id3[1] = item.get_id(1);
-            if(d == 3) id3[3] = item.get_id(2);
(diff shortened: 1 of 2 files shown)


tests/sycl/item.cpp
```diff
@@ -70,7 +70,7 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(item_api, _dimensions, test_dimensions::type) {
             auto id3 = item.get_id();
             if(d >= 1) id3[0] = item.get_id(0);
             if(d >= 2) id3[1] = item.get_id(1);
-            if(d == 3) id3[3] = item.get_id(2);
```
psalz, line 73: Huh, how did these tests pass previously..?

illuhad: Now that is a very good question, and one that I also don't have an answer for..

psalz: Okay well, in this case it worked because the id is initialized with the result of `item::get_id()`, which was a questionable choice in the first place... Upon further investigation it seems like there is more copy-pasta issues here, as this entire block was intended to test `operator[]`.

PR #345, opened 2020-10-19
Mark item::operator[] const, fix tests

psalz:
As revealed by #338 there were some issues with the item tests. While fixing the tests to actually use `operator[]`, I noticed that it wasn't marked as const, so I fixed that as well!

### DIFF
--- a/include/hipSYCL/sycl/libkernel/item.hpp
+++ b/include/hipSYCL/sycl/libkernel/item.hpp
@@ -73,7 +73,7 @@ struct item_base
     return this->global_id[dimension];
   }
 
-  HIPSYCL_KERNEL_TARGET size_t operator[](int dimension)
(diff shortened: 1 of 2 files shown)


psalz:
By the way, I had some issues getting the unit tests to run on my machine with Boost 1.68: First I got some error about an ambiguous `operator<<` used somewhere within the `BOOST_TEST` macro expanded within usm.cpp, and then it wouldn't link because of `undefined symbol main`, which I fixed by adding `#define BOOST_TEST_DYN_LINK` back into various headers before including `<boost/test/unit_test.hpp>`. Not sure what is going on here.

illuhad:
Thank you! We could exchange `BOOST_TEST` with `BOOST_CHECK` or so if this causes some trouble.. And we probably should add `BOOST_TEST_DYN_LINK` back in. I think boost 1.68 is still wide spread, so I think it's worth to support it.

PR #351, opened 2020-10-21
Fix missing data transfers for buffers initialized from const pointers

illuhad:
When initializing a buffer from a const pointer, the `data_range` is constructed and an empty host allocation is created. This allocation is then filled with a `memcpy`. However, the runtime was not informed that the data inside the buffer has been modified, which causes it to not emit data transfers when the buffer is accessed on device.

This PR fixes the issue by manually marking the modified data range as current.

@9prady9 this should resolve our problem - with this the example code prints the expected output for me.


9prady9:
@illuhad Thank you. Yeah, I understood the mechanics after having a look at the buffer_impl class.

9prady9 (approved):

PR #355, opened 2020-10-24
fix cmake link libraries for runtime

DieGoldeneEnte:
This fixes a problem, when compiling code with hipSYCL (build with HIP support), in my case the tests, resulting in undefined references (hipDriverGetVersion, hipFree, hipGetDevice, ...) when compiling for CPU.
This could already be fixed by running cmake twice when building hipSYCL.

Problem:
The `src/runtime/CMakeLists.txt` file referenced `ROCM_LIBS` but was included before it was defined.
Solution:
This commit moves the include (`add_subdirectory(src)`) below the definition.


illuhad:
aaahhh, that was it! * bashes head on the table* :)

Thank you!

PR #357, opened 2020-10-27
Ensure vector types have correct alignment

### LINKED ISSUE #346: CUDA: Alignment mismatch between emulated and native vector types causes hard to debug errors
It seems that CUDA's `::float2` has an alignment of 8, while hipSYCL's `emulated_vector`, which is used on the host, has an alignment of 4. Obviously that's not good, and causes some very weird looking and hard to debug problems - see below. I'll mostly just dump the repro case here as I think it's quite self-explanatory:

```cpp
#include <CL/sycl.hpp>

namespace sycl = cl::sycl;

namespace foosycl {

struct emulated_float2 { float x; float y; };

// Fix:
// struct alignas(8) emulated_float2 { float x; float y; };
//
// #ifdef SYCL_DEVICE_ONLY
// static_assert(alignof(emulated_float2) == alignof(::float2));
// #endif

struct float2 {
  HIPSYCL_UNIVERSAL_TARGET
  float2(const float& x, const float& y) { data.x = x; data.y = y; }
  HIPSYCL_UNIVERSAL_TARGET float& x() { return data.x; }
  HIPSYCL_UNIVERSAL_TARGET float x() const { return data.x; }
  HIPSYCL_UNIVERSAL_TARGET float& y() { return data.y; }
  HIPSYCL_UNIVERSAL_TARGET float y() const { return data.y; }

#ifdef SYCL_DEVICE_ONLY
  ::float2 data;
#else
  emulated_float2 data;
#endif
};

}

    sycl::queue queue;

    //using float2_t = sycl::float2;
    using float2_t = foosycl::float2;

    float2_t vec = {42.f, 1337.f};

    printf("vec on host: %f,%f\n", vec.x(), vec.y());

        cgh.single_task<class buggy>([=, a = 9.f]() {
            printf("a in buggy kernel: %f\n", a);
            printf("vec in buggy kernel: %f,%f\n", vec.x(), vec.y());
        });
    });

        cgh.single_task<class worky>([=]() {
            printf("vec in worky kernel: %f,%f\n", vec.x(), vec.y());

            printf("sizeof(::float2) = %lu\n", sizeof(::float2));
            printf("sizeof(emulated_float2) = %lu\n", sizeof(foosycl::emulated_float2));
            printf("alignof(::float2) = %lu\n", alignof(::float2));
            printf("alignof(emulated_float2) = %lu\n", alignof(foosycl::emulated_float2));
        });

    queue.wait_and_throw();

    return EXIT_FAILURE;
}
```

I'm using `foosycl::float2` and `foosycl::emulated_float2` here, but these are just the result of reducing about 5000 lines of hipSYCL headers down to their essence. You can switch `float2_t` to `sycl::float2` to see that this error is reproducible with the original type as well.

As suggested in the code, a potential fix could be to set the alignment explicitly (this may need adjustment per backend), and to assert that the alignments (as well as the size) match during device compilation.

psalz:
This adds static assertions to check that emulated vector data types on the host fulfill the same alignment requirements as their device-side counterparts.

Additionally, a mechanism for specifying per-type, backend-specific alignment requirements is introduced. This is currently only used for CUDA, as I don't have access to a ROCm platform. I'm also not sure whether these empirically obtained values are the same across all shader model versions. However in any case, a static assertion will be raised should the current configuration be wrong and/or incomplete.

Fixes #346.


include/hipSYCL/sycl/libkernel/detail/vec.hpp
```diff
@@ -78,20 +78,82 @@ template<> struct logical_vector_op_result<detail::dp_float>
...
+template <> struct required_alignment<long, 4>
+{ constexpr static size_t value = 16; };
+template <> struct required_alignment<ulong, 4>
+{ constexpr static size_t value = 16; };
+template <> struct required_alignment<longlong, 4>
+{ constexpr static size_t value = 16; };
+template <> struct required_alignment<ulonglong, 4>
+{ constexpr static size_t value = 16; };
+template <> struct required_alignment<double, 4>
+{ constexpr static size_t value = 16; };
```
illuhad: Why do we need to hardcode the alignments? Couldn't we just set them to `alignof(NativeT)`, with `NativeT` being native CUDA/HIP vector types? Then it would automatically do the right thing on ROCm as  well.

include/hipSYCL/sycl/libkernel/detail/vec.hpp
```diff
@@ -78,20 +78,82 @@ template<> struct logical_vector_op_result<detail::dp_float>
...
+#if defined(HIPSYCL_PLATFORM_CUDA)
+  constexpr static size_t value = alignof(T) * Dim;
+#else
+  // Use default alignment otherwise
+  constexpr static size_t value = 0;
+#endif
+};
+
+// TODO: This should probably be moved to a backend-specific location.
+#if defined(HIPSYCL_PLATFORM_CUDA)
```
illuhad: Wouldn't it suffice to only enable this in the host pass of CUDA/HIP backends? e.g. `defined(HIPSYCL_PLATFORM_CPU) && defined(HIPSYCL_PLATFORM_ROCM/CUDA)`.

include/hipSYCL/sycl/libkernel/detail/vec.hpp
```diff
@@ -78,20 +78,82 @@ template<> struct logical_vector_op_result<detail::dp_float>
...
-struct emulated_vector<T,3> { T x; T y; T z; };
+struct alignas(required_alignment_v<T, 3>) emulated_vector<T,3> { T x; T y; T z; };
 
 template <class T>
-struct emulated_vector<T,4> { T x; T y; T z; T w; };
+struct alignas(required_alignment_v<T, 4>) emulated_vector<T,4> { T x; T y; T z; T w; };
+
+// This is a hack for the host device where float1,2,3,4 etc don't exist...
+// TODO: Reimplement vec<> properly
+#define HIPSYCL_DEFINE_EMULATE_VECTOR_TYPES(T)                                 \
```
illuhad: Looks like a missing d, e.g. `EMULATE*D*_VECTOR_TYPES`? Typo?

EDIT: Oopsie, I see it's already like this in the original code.

include/hipSYCL/sycl/libkernel/detail/vec.hpp
psalz: Uhh - to be honest I didn't really think too much about how the hipSYCL compilation model works for the new runtime. From the way the vector implementation was previously structured (with branching on `SYCL_DEVICE_ONLY`), I assumed that there was some phase of compilation where the native types wouldn't be available -- hence I also couldn't do something like `alignof(NativeT)`. But thinking about it now, that probably doesn't make any sense for CUDA/HIP, as they also support the native vector types in host code.

I realize the `emulated_vector` types are required for backends that don't have native vector types (i.e., OpenMP), but now my question is: Why use the emulated types at all when targeting CUDA/HIP?

illuhad: Great point - it seems we missed the best solution :) You are right, we now need to distinguish two cases:
1) CPU backend embedded in host pass of CUDA/HIP
2) Standalone CPU backend

For 2), the alignment won't matter (but it might make sense to align for vectorization), and we don't need to match alignments on other backends. For 1) we should be able to use the device types also on host, which should also solve the alignment issue.

psalz: Okay, so it should be sufficient to check for `#if defined(HIPSYCL_PLATFORM_ROCM) || defined(HIPSYCL_PLATFORM_CUDA)` for the first case, and the `#else` branch would be the second one, right?

By the way, it might still be worth considering having consistent alignment requirements across all backends. Otherwise dumping a buffer to file and reading it back in (on another backend) might lead to surprises. Does the 2020 spec say anything about this? Anyway, I guess that's out of scope for this PR.

illuhad: > Okay, so it should be sufficient to check for #if defined(HIPSYCL_PLATFORM_ROCM) || defined(HIPSYCL_PLATFORM_CUDA) for the first case, and the #else branch would be the second one, right?

Yes. It might be a good idea to add a macro to test this more easily to `libkernel/backend.hpp` so that we only need to change that one place in case we add more backends in the future.

> By the way, it might still be worth considering having consistent alignment requirements across all backends. Otherwise dumping a buffer to file and reading it back in (on another backend) might lead to surprises. Does the 2020 spec say anything about this?

Good question. I checked and it seems SYCL 2020 provisional specifies `vec` alignments explicitly in 4.16.2.6. Which would mean that we cannot just rely on CUDA/HIP types to do the right thing. For a fully compliant implementation, we may have to use our own types for all backends.

psalz:
Alright, I'd say we might as well improve spec compliance, if given the chance. I've changed the approach to the following: If the compiler supports native vector data types (detected using the new macro `HIPSYCL_LIBKERNEL_COMPILER_SUPPORTS_DEVICE`), hipSYCL will attempt to use those, unless they don't fulfill the alignment requirements dictated by the spec. In case they don't have the correct alignment, or we are using a pure host compiler, the emulated types are used instead.

I've added some compile time tests to show that the native types are in fact used in most cases. Again I'm not sure if these tests will hold across different backends (i.e., HIP) or shader model versions. But at least we'll be made aware then.

illuhad:
Can you please resolve conflicts and rebase on current `develop`?

psalz:
Done!

illuhad:
Thanks!

@hipsycl-ci-bot test on GPU please

Edit: try again

psalz:
Oops, `#ifdef` instead of `#if`!

psalz:
From CI:
```
 /home/runner/work/hipSYCL/hipSYCL/tests/sycl/vec.cpp:68:1: error: static_assert failed due to requirement '!std::is_same_v<HIP_vector_type<long, 4>, HIP_vector_type<long, 4>>'
EXPECT_12NATIVE_34EMULATED(long);
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/hipSYCL/hipSYCL/tests/sycl/vec.cpp:57:3: note: expanded from macro 'EXPECT_12NATIVE_34EMULATED'
```

Interesting. Looks like HIP has different alignments on some types.

Edit: This doesn't mean anything is wrong with my implementation btw, only that for HIP the set of types which need to be emulated for SYCL is different from that for CUDA (which is what I'm testing there). In this case it's actually positive as well, because it means that hipSYCL can use the native `long4` type for HIP. But the question is how this test should be adjusted then. Should I make a distinction between CUDA and HIP in the tests? Do we even care which types are emulated and which are native?

illuhad:
I think mid to long-term I would like to reimplement `vec` to just use a single `struct alignas(...) data { T [N]; };` uniformly for all backends, and not rely on CUDA/HIP types anymore (assuming there's not performance penalty). This would massively simplify the implementation in a whole lot of places, and would improve spec conformance for vectors with more than 4 components.

I am leaning towards we don't care which ones are emulated and which are native. I'm not even sure if it would make a performance difference as the native types are effectively also just aligned structs (well, they also have the `__device_builtin__` attribute on CUDA, but I have no idea what that does).

psalz:
> I am leaning towards we don't care which ones are emulated and which are native.

Alright, I've removed the test.

>  I'm not even sure if it would make a performance difference as the native types are effectively also just aligned structs (well, they also have the `__device_builtin__` attribute on CUDA, but I have no idea what that does).

Well PTX [has support](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#vectors) for vector types, so I guess the question is whether the LLVM codegen actually realizes that the custom struct represents a vector. I can imagine this having implications for the generated device assembly.

fodinabor:
not sure I'm missing anything, but doesn't seem that Clang uses the vector instructions at all.. so no difference here: https://cuda.godbolt.org/z/x3aqM6

illuhad:
@fodinabor Thanks, I had also just started to playing around in godbolt, but you beat me ;)
So it seems that in any case if we want to have nvcc vector behavior we need to put in inline ptx. Which seems like a good optimization to me.

psalz:
Alright, so what's the plan - will this be merged? Even though using native vector types may not have any benefits right now, this may change with future Clang versions. More importantly though, I'd like to finally get #346 fixed.

illuhad:
Yes, the plan is still to merge this for hipSYCL 0.9 as a short-term solution to fix #346. The question about the future direction of `vec` is a larger issue and I don't want to delay the release because of that.

I still believe that in the long term, reimplementing `vec` without native types (but potentially specialized `load()/store()` functions that use inline assembly) is the right thing - we are talking about thousands of lines of code that could be removed with that change.

illuhad:
Thank you!

PR #374, opened 2020-11-20
[SYCL 2020] Add initial support for reductions

illuhad:
This adds initial support for SYCL 2020 reductions in the form of `parallel_for(..., reduction(...), kernel)`

Supported features:
* All backends
* Variadic reductions, e.g. arbitrary amounts of simultaneous reductions
* Exceeds the spec by also providing support for reductions with hierarchical and scoped parallelism
* Queue shortcuts also accept reductions
* Caveat: Only scalar reductions for now (no multi-dimensional reductions)
* Caveat: Local reduction algorithm on GPU is somewhat simplistic - @DieGoldeneEnte is working on optimized local reductions in the context of group algorithms that can we use here once they are available
* Caveat: Lots of additional optimizations to explore in the future, e.g. when multiple reductions are carried out
* Caveat: Precise interface of `reduction()` is subject to change with SYCL 2020 final, in particular with respect to providing identity values


illuhad:
@hipsycl-ci-bot test to verify if there are regressions in hierarchical parallel for.

Edit: Run number 6

PR #376, opened 2020-11-20
Don't raise error when no CUDA/HIP devices are present

illuhad:
Don't raise an error if `cuda/hipGetDeviceCount` fails with `cuda/hipErrorNoDevice`.

In this case, the device list should simply be empty. Additionally, raising errors when the runtime is starting up can lead to deadlocks since the errors will try to register themselves with the runtime (which is not fully there yet at this point).
This issue should however probably be solved at a more fundamental level in the future.

### DIFF
--- a/src/runtime/hip/hip_hardware_manager.cpp
+++ b/src/runtime/hip/hip_hardware_manager.cpp
@@ -40,7 +40,10 @@ hip_hardware_manager::hip_hardware_manager(hardware_platform hw_platform)
   int num_devices = 0;
 
   auto err = hipGetDeviceCount(&num_devices);
-  if (err != hipSuccess) {
+  if(err == hipErrorNoDevice)
+    num_devices = 0;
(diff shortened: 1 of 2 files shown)


src/runtime/hip/hip_hardware_manager.cpp
```diff
@@ -40,7 +40,10 @@ hip_hardware_manager::hip_hardware_manager(hardware_platform hw_platform)
   int num_devices = 0;
 
   auto err = hipGetDeviceCount(&num_devices);
-  if (err != hipSuccess) {
+  if(err == hipErrorNoDevice)
+    num_devices = 0;
+  
+  if (err != hipSuccess && err != hipErrorNoDevice) {
```
fodinabor, line 46: `else if`?

illuhad:
Will be superseded by PR coming up shortly :)

PR #377, opened 2020-11-20
Handle variadic templates as kernel names

### LINKED ISSUE #375: Kernel name with variadic parameter pack is not accepted
I am getting the following error: `Not a valid kernel name: 'AnonKern<'.', '.', '<'>'`, but I am not sure what the requirements on the type are.
It seems like some class template instantiations are allowed, while others aren't.

Thanks :)

bmanga:
This recursively handles TemplateArguments for when an argument is a parameter pack.
Closes #375

### DIFF
--- a/include/hipSYCL/compiler/Frontend.hpp
+++ b/include/hipSYCL/compiler/Frontend.hpp
@@ -132,6 +132,56 @@ class CompleteCallSet : public clang::RecursiveASTVisitor<CompleteCallSet> {
     FunctionSet visitedDecls;
 };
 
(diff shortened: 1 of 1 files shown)


illuhad:
Thank you! I will have a closer look at this on monday :)

keryell:
Do you handle name conflict? Such as a playful programmer using `struct _nullptr {}` as a name and so, like both `::` and `__` in different names to create collision?

bmanga:
@keryell That's a good point. I created an alternative PR #378 which just uses clang's mangler.

illuhad:
Closing as the alternative solution #378 has been merged. We can reopen if it turns out that #378 causes issues.

PR #303, opened 2020-09-14
Remove name mangling in clang plugin to improve compatibility with newer clang versions

illuhad:
Removes name mangling attempts in clang plugin that can cause trouble with certain clang installations.
This is (hopefully) a hotfix for #302.
Since this prevents the old IR pass logic from working, the old IR pass should be removed if this is merged.

Resolves #302

Issue #302, opened 2020-09-14
ICE - from Clang/ItaniumMangle


reguly:
I compiled the latest develop branch with clang-11 (llvmorg-11.0.0-rc2), as well as with 10.0, both with CUDA 10.1, but in both cases, I am getting an ICE out from the ItaniumMangle module as below. I don't even know why Itanium is coming up, I only set llvm targets to X86 and NVPTX... I can compile other C++ and CUDA code okay. Any ideas how to work around this?

> syclcc-clang -O3 --hipsycl-gpu-arch=sm_60 test.cpp
> Can't mangle a deduction guide name!
> UNREACHABLE executed at /home/ireguly/projects/llvm-project/clang/lib/AST/ItaniumMangle.cpp:1509!
> PLEASE submit a bug report to https://bugs.llvm.org/ and include the crash backtrace, preprocessed source, and associated run script.
> Stack dump:
[...]

PR #378, opened 2020-11-20
Mangle type kernel names with clang's mangler

### LINKED ISSUE #375: Kernel name with variadic parameter pack is not accepted

bmanga:
This is an alternative approach to #377 (handling variadic templates in kernel name generation).
If compatibility with previously created kernel names is not a concern, I suspect this would be a much more robust approach.

### DIFF
--- a/include/hipSYCL/compiler/Frontend.hpp
+++ b/include/hipSYCL/compiler/Frontend.hpp
@@ -40,6 +40,7 @@
 #include "clang/Frontend/FrontendPluginRegistry.h"
 #include "clang/AST/AST.h"
 #include "clang/AST/Attr.h"
+#include "clang/AST/Mangle.h"
 #include "clang/AST/ASTConsumer.h"
 #include "clang/AST/RecursiveASTVisitor.h"
 #include "clang/AST/Mangle.h"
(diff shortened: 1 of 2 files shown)


illuhad:
clang is quite picky about the circumstances in which it allows name mangling calls. We had to merge in #303 which removes calls to the name mangler, since previously existing usages of the Itanium mangler caused ICEs such as #302. I'm therefore hesitant to add name mangling back in, unless we are certain that it is safe and works for all clang versions in all distributions...

bmanga:
@illuhad  Oh I see. However, the old code was using `mangleName`. It seems that `mangleTypeName` is more appropriate according to the doc: 
> Generates a unique string for an externally visible type for use with TBAA or type uniquing

BTW I built this with LLVM 11 (ubuntu) with no issue. I can verify also with 9 and 10 if you'd like.

illuhad:
Yes, I also think that, since only types are involved and not arbitrary function declarations as previously, it's less likely to cause issues due to a narrower use case. But even with the previous issue, I had trouble reproducing and mainly noticed because of users complaining ;-) I've never found out why the issue occured for some setups and didn't for others...

I also agree overall that letting clang handle things is a good strategy instead of effectively reinventing name mangling ourselves (after all relying on existing toolchains as much as possible is the philosophy behind hipSYCL).

Maybe it might be a good idea to merge this and see what happens, in which case we could revert to your alternative solution. After all, we are on the `develop` branch where things are allowed to break occasionally.

I'm currently working on a new CI pipeline. I would like to test this first with the new pipeline. Clang 10 support is important as clang 11 introduces some other issues (e.g. linking with device stubs of unnamed kernel lambdas), so if you could test clang 10 it would be great!

bmanga:
@illuhad That sounds good to me! I have pushed a cleaned-up version of the changes, and I will report back when I test with Clang10 :)

illuhad:
Can you add a test case for a variadic kernel name to the device compilation tests and rebase on current `develop`? Then we can push this through CI and have it run with all sort of different clang versions :-)

bmanga:
Done, let's hope for the best :D

bmanga:
@illuhad I think something is wrong with the CI pipeline

illuhad:
@bmanga It seems there are currently issues with the LLVM apt repositories. But don't worry, the github hosted tests are not the important ones :)

Our self-hosted testing infrastructure has just tested your PR with clang 10,11,12, successfully compiled all tests for CPU, CUDA and ROCm and executed the unit tests for CPU and CUDA without error :)

https://github.com/hipSYCL/hipSYCL-ci-bot/actions/runs/385751030
Unfortunately, you won't be able to see the results. We have to keep the repository private because using self-hosted github runners with public repositories has severe security issues.

illuhad (approved):

illuhad:
Thank you very much!

Issue #381, opened 2020-11-23
Linker error on clang 11+ for unnamed kernel lambdas and function object kernels


illuhad:
In clang 11 and newer, the way that clang generates kernel invocation stubs on CUDA and HIP was changed. This now causes code using unnamed kernel lambdas to fail with a linker error. clang 10 works fine.

The following code (which follows the same principles as hipSYCL internals) reproduces the issue:
```
#include <SYCL/sycl.hpp>

template<class kernel_name, class F>
__sycl_kernel void invoker(F f){
    f();
}


int main(){
  std::function<void()> bound_invoker = [](){
    invoker<class _unnamed_kernel>([=](){});
  };
  bound_invoker();
}
```

It seems to be related to the fact that we wrap the kernel launch in a function object - without the function object it works fine. Also named kernels work fine, so it seems that clang's default name mangling that we rely on doesn't do the right thing here.

illuhad:
It seems function objects as kernels are also affected, i.e. all cases where we don't explicitly rename the kernel.

illuhad:
It seems that starting with clang 11, we also need to rerun codegen during the host pass - if this is the entire problem in that case it should be easy to fix since we already do the same for the device pass.

PR #382, opened 2020-11-23
Move to GitHub workflows for testing

illuhad:
* Removes legacy Travis CI/CD and docker images
* Adds GitHub workflow
   * Builds hipSYCL against ROCm, multiple CUDA versions and for Ubuntu 18.04 and Ubuntu 20.04
   * Currently uses only clang 10; I had to disable testing with clang 11 and 12 because they are affected by #381 and hence cannot compile the unit tests at the moment.
   * Builds the unit tests for CPU, ROCm and CUDA
   * Runs unit tests on CPU
* Fixes runtime not properly starting up when compiled with CUDA/ROCm support, but no devices are available
* Fixes a couple of `BOOST_DYN_LINK` issues with the unit tests
* Adds option to the unit tests to disable tests with optional kernel naming

The workflow can take quite long to compile; in particular installing boost takes some time. This is necessary because Ubuntu 18.04 boost package is too old for our unit tests. Not sure how important testing on 18.04 is, if we can drop it we can make the workflow substantially faster.

As a next step I'd like to investigate custom runners to execute tests on GPUs.


psalz:
You could consider moving the hipSYCL build / test steps into custom Docker container actions, then you wouldn't need the whole prerequisite setup each time!

illuhad:
@psalz Thanks, I had briefly thought about that. This will be a great direction for future improvement.

PR #384, opened 2020-11-24
Added support for boost installations other than the default location

sbalint98:
This PR should allow the user to use boost installations that are in a custom location easily.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,8 +20,8 @@ FORCE)
   endif()
 endif()
 
+set(BOOST_ROOT ${BOOST_PATH})
 find_package(Boost COMPONENTS context fiber REQUIRED)
-
 # Check for CUDA/ROCm and clang
 find_package(CUDA QUIET)
 # We currently search for hipcc to check for ROCm installation
@@ -142,6 +142,7 @@ set(SYCLCC_CONFIG_FILE "{
   \"default-gpu-arch\"  : \"${DEFAULT_GPU_ARCH}\",
   \"default-cpu-cxx\"   : \"${CMAKE_CXX_COMPILER}\",
   \"default-rocm-path\" : \"${ROCM_PATH}\",
+  \"default-boost-path\" : \"${BOOST_PATH}\",
   \"default-use-bootstrap-mode\" : \"false\",
   \"default-is-dryrun\" : \"false\",
   \"default-clang-include-path\" : \"${CLANG_INCLUDE_PATH}\",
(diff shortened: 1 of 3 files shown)


illuhad:
Wouldn't it be better to add generalized `cxx_flags` and `link_line` arguments for the CPU backend? Otherwise we have the same issue for OpenMP, or other dependencies.

sbalint98:
Well, boost is a dependency of both the `omp` and the `omp_sequential` backend, if I understood correctly. And since the sequential backend is temporary it might be unnecessary to expose its dependencies.  And by the same token, would you also remove the `default-cuda-path` and the `default-rocm-path` vars?

illuhad:
boost as a hard dependency is probably going to remain even after `omp_sequential` is removed since we need `boost.fiber` for replacement backends e.g. based on TBB as well.

I don't think we can remove existing arguments such the cuda path or rocm path, but we might want to change how they work internally: I would like to have them set corresponding substitution variables which then are used by default cxx flags and link lines.

sbalint98:
> boost as a hard dependency is probably going to remain even after omp_sequential is removed since we need boost.fiber for replacement backends e.g. based on TBB as well.

I wanted to open a separate PR with the exposure of the cxx flags and the link lines for all backends, but probably it might be better to just close this PR and move these changes in the other one. 

> I don't think we can remove existing arguments such the cuda path or rocm path, but we might want to change how they work internally: I would like to have them set corresponding substitution variables which then are used by default cxx flags and link lines.

I am not sure if I fully understand this. In syclcc the `default-cuda/rocm-path` specified in syclcc.json are currently used to set substitutions variables for the `HIPSYCL_CUDA/ROCM_PATH` `HIPSYCL_CUDA/ROCM_LIB_PATH` vars that are substituted from the cxx and link lines for cuda and rocm. How would you like to change this behaviour?

illuhad:
> I wanted to open a separate PR with the exposure of the cxx flags and the link lines for all backends, but probably it might be better to just close this PR and move these changes in the other one.

okay :)

>  I am not sure if I fully understand this. In syclcc the default-cuda/rocm-path specified in syclcc.json are currently used to set substitutions variables for the HIPSYCL_CUDA/ROCM_PATH HIPSYCL_CUDA/ROCM_LIB_PATH vars that are substituted from the cxx and link lines for cuda and rocm. How would you like to change this behaviour?

Only in a subtle way. I think the idea would be to not refer to them directly, but only potentially when the cxx line or link line is evaluated and their substitution variables are referenced there. I think there are still a couple of places that use them directly inside `syclcc`.

PR #391, opened 2020-11-27
[SYCL2020] Group functions

DieGoldeneEnte:
This PR adds a naive implementation for group_functions to hipSYCL. This includes implementations of
* group_broadcast
* group_barrier
* group_{any,all,none]_of
* group_reduce
* group_{ex,in}clusive_scan

on CPU and NVIDIA/AMD GPUs, as well as tests for these functions. I will provide optimized versions in later PRs.
It also includes (group_)functions using two pointers (beginning/end), but these are not in the specification and are not meant to be used yet (as such they reside in the `detail` namespace).

At the moment all tests pass (except some problems with the pointer-based functions on CPU which sometimes fail, I am still investigating). [(private results)](https://github.com/hipSYCL/hipSYCL-ci-bot/actions/runs/387445691)

I would love to get some feedback. If you find some template-parameters or formatting you don't like, there is a chance I missed them in one of my cleanup/refactoring attempts. Just tell me so I can fix it. Some small changes like splitting the tests into multiple files for faster compilation might be added here,

### DIFF
--- a/include/hipSYCL/sycl/libkernel/cuda/group_functions.hpp
+++ b/include/hipSYCL/sycl/libkernel/cuda/group_functions.hpp
@@ -1,0 +1,242 @@
+/*
+ * This file is part of hipSYCL, a SYCL implementation based on CUDA/HIP
+ *
+ * Copyright (c) 2018 Aksel Alpay
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifdef SYCL_DEVICE_ONLY
+#ifdef HIPSYCL_PLATFORM_CUDA
+
+#ifndef HIPSYCL_LIBKERNEL_CUDA_GROUP_FUNCTIONS_HPP
+#define HIPSYCL_LIBKERNEL_CUDA_GROUP_FUNCTIONS_HPP
+
+#include "../backend.hpp"
+#include "../id.hpp"
+#include "../sub_group.hpp"
+#include "../vec.hpp"
+#include <type_traits>
+
+namespace hipsycl {
+namespace sycl {
+
+namespace detail {
+
+constexpr unsigned int AllMask = 0xFFFFFFFF;
+
+template<typename T,
+         typename std::enable_if_t<std::is_arithmetic_v<T>, int> = 0>
+__device__
+T shuffle_impl(T x, int id) {
+  return __shfl_sync(detail::AllMask, x, id);
+}
+
+template<>
+__device__
+char shuffle_impl(char x, int id) {
+  return static_cast<char>(shuffle_impl(static_cast<int>(x), id));
+}
+
+template<typename T, int N>
+__device__
+sycl::vec<T, N> shuffle_impl(sycl::vec<T, N> x, int id) {
+  sycl::vec<T, N> ret{};
+
+  if constexpr (1 <= N)
+    ret.s0() = shuffle_impl(x.s0(), id);
+  if constexpr (2 <= N)
+    ret.s1() = shuffle_impl(x.s1(), id);
+  if constexpr (3 <= N)
+    ret.s2() = shuffle_impl(x.s2(), id);
+  if constexpr (4 <= N)
+    ret.s3() = shuffle_impl(x.s3(), id);
+  if constexpr (8 <= N) {
+    ret.s4() = shuffle_impl(x.s4(), id);
+    ret.s5() = shuffle_impl(x.s5(), id);
+    ret.s6() = shuffle_impl(x.s6(), id);
+    ret.s7() = shuffle_impl(x.s7(), id);
+  }
+  if constexpr (16 <= N) {
+    ret.s8() = shuffle_impl(x.s8(), id);
+    ret.s9() = shuffle_impl(x.s9(), id);
+    ret.sA() = shuffle_impl(x.sA(), id);
+    ret.sB() = shuffle_impl(x.sB(), id);
+    ret.sC() = shuffle_impl(x.sC(), id);
+    ret.sD() = shuffle_impl(x.sD(), id);
+    ret.sE() = shuffle_impl(x.sE(), id);
+    ret.sF() = shuffle_impl(x.sF(), id);
+  }
+
+  return ret;
+}
+
+} // namespace detail
+
+// broadcast
+template<typename T>
+HIPSYCL_KERNEL_TARGET
+T group_broadcast(sub_group g, T x,
+                  typename sub_group::linear_id_type local_linear_id = 0) {
+  return detail::shuffle_impl(x, local_linear_id);
+}
+
+template<typename T, int N>
+HIPSYCL_KERNEL_TARGET
+sycl::vec<T, N> group_broadcast(sub_group g, sycl::vec<T, N> x,
+                                typename sub_group::linear_id_type local_linear_id = 0) {
+  return detail::shuffle_impl(x, local_linear_id);
+}
+
+
+// barrier
+template<typename Group>
+HIPSYCL_KERNEL_TARGET
+inline void group_barrier(Group g, memory_scope fence_scope = Group::fence_scope) {
+  if (fence_scope == memory_scope::device) {
+    __threadfence_system();
+  }
+  __syncthreads();
+}
+
+template<>
+HIPSYCL_KERNEL_TARGET
+inline void group_barrier(sub_group g, memory_scope fence_scope) {
+  if (fence_scope == memory_scope::device) {
+    __threadfence_system();
+  } else if (fence_scope == memory_scope::work_group) {
+    __threadfence_block();
+  }
+  __syncwarp(); // not necessarily needed, but might improve performance
+}
+
+
+// any_of
+template<>
+HIPSYCL_KERNEL_TARGET
+inline bool group_any_of(sub_group g, bool pred) {
+  return __any_sync(detail::AllMask, pred);
+}
+
+
+// all_of
+template<>
+HIPSYCL_KERNEL_TARGET
+inline bool group_all_of(sub_group g, bool pred) {
+  return __all_sync(detail::AllMask, pred);
+}
+
+
+// none_of
+template<>
+HIPSYCL_KERNEL_TARGET
+inline bool group_none_of(sub_group g, bool pred) {
+  return !__any_sync(detail::AllMask, pred);
+}
+
+
+// reduce
+template<typename T, typename BinaryOperation>
+HIPSYCL_KERNEL_TARGET
+T group_reduce(sub_group g, T x, BinaryOperation binary_op) {
+  auto lid = g.get_local_linear_id();
+
+  size_t lrange          = 1;
+  auto group_local_range = g.get_local_range();
+  for (int i = 0; i < g.dimensions; ++i)
+    lrange *= group_local_range[i];
+
+  group_barrier(g);
+
+  auto local_x = x;
+
+  for (size_t i = lrange / 2; i > 0; i /= 2) {
+    auto other_x = detail::shuffle_impl(local_x, lid + i);
+    if (lid < i)
+      local_x = binary_op(local_x, other_x);
+  }
+  return detail::shuffle_impl(local_x, 0);
+}
+
+
+// exclusive_scan
+template<typename V, typename T, typename BinaryOperation>
+HIPSYCL_KERNEL_TARGET
+T group_exclusive_scan(sub_group g, V x, T init, BinaryOperation binary_op) {
+  auto lid      = g.get_local_linear_id();
+  size_t lrange = g.get_local_linear_range();
+
+  auto local_x = x;
+
+  for (size_t i = 1; i < lrange; i *= 2) {
+    size_t next_id = lid - i;
+    if (i > lid)
+      next_id = 0;
+
+    auto other_x = detail::shuffle_impl(local_x, next_id);
+    if (i <= lid && lid < lrange)
+      local_x = binary_op(local_x, other_x);
+  }
+
+  size_t next_id = lid - 1;
+  if (g.leader())
+    next_id = 0;
+
+  auto return_value = detail::shuffle_impl(local_x, lid - 1);
+
+  if (g.leader())
+    return init;
+
+  return binary_op(return_value, init);
+}
+
+
+// inclusive_scan
+template<typename T, typename BinaryOperation>
+HIPSYCL_KERNEL_TARGET
+T group_inclusive_scan(sub_group g, T x, BinaryOperation binary_op) {
+  auto lid      = g.get_local_linear_id();
+  size_t lrange = g.get_local_linear_range();
+
+  auto local_x = x;
+
+  for (size_t i = 1; i < lrange; i *= 2) {
+    size_t next_id = lid - i;
+    if (i > lid)
+      next_id = 0;
+
+    auto other_x = detail::shuffle_impl(local_x, next_id);
+    if (i <= lid && lid < lrange)
+      local_x = binary_op(local_x, other_x);
+  }
+
+  return local_x;
+}
+
+
+} // namespace sycl
+} // namespace hipsycl
+
+#endif //HIPSYCL_LIBKERNEL_CUDA_GROUP_FUNCTIONS_HPP
+
+#endif //HIPSYCL_PLATFORM_CUDA
+#endif //SYCL_DEVICE_ONLY
--- a/include/hipSYCL/sycl/libkernel/generic/hiplike/group_functions.hpp
+++ b/include/hipSYCL/sycl/libkernel/generic/hiplike/group_functions.hpp
@@ -1,0 +1,593 @@
+/*
+ * This file is part of hipSYCL, a SYCL implementation based on CUDA/HIP
+ *
+ * Copyright (c) 2018 Aksel Alpay
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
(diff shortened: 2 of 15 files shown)


include/hipSYCL/sycl/libkernel/cuda/group_functions.hpp
```diff
@@ -0,0 +1,231 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifdef SYCL_DEVICE_ONLY
+#ifdef HIPSYCL_PLATFORM_CUDA
+
+#ifndef HIPSYCL_LIBKERNEL_CUDA_GROUP_FUNCTIONS_HPP
+#define HIPSYCL_LIBKERNEL_CUDA_GROUP_FUNCTIONS_HPP
+
```
illuhad, line 33: Should include `sub_group` header?

include/hipSYCL/sycl/libkernel/cuda/group_functions.hpp
```diff
@@ -0,0 +1,231 @@
...
+// broadcast
+template <typename T>
+HIPSYCL_KERNEL_TARGET
+T group_broadcast(sub_group g, T x, typename sub_group::linear_id_type local_linear_id = 0) {
+  return detail::_shfl(x, local_linear_id);
+}
+
+template <typename T, int N>
+HIPSYCL_KERNEL_TARGET
+sycl::vec<T, N> group_broadcast(sub_group g, sycl::vec<T, N> x, typename sub_group::linear_id_type local_linear_id = 0) {
```
illuhad: Line break? Maybe use something like clang-format

include/hipSYCL/sycl/libkernel/cuda/group_functions.hpp
```diff
@@ -0,0 +1,231 @@
...
+#define HIPSYCL_LIBKERNEL_CUDA_GROUP_FUNCTIONS_HPP
+
+namespace hipsycl {
+namespace sycl {
+
+namespace detail {
+
+template <typename T, typename std::enable_if<std::is_arithmetic<T>::value, T>::type * = nullptr>
+__device__
+T _shfl(T x, int id) {
```
illuhad: We don't have a convention for `_identifier`. There are some `__identifier()` for compiler-glue stuff or low-level macros such as `__hipsycl_lid_x` etc. How about just using `shufl_impl`?

include/hipSYCL/sycl/libkernel/hip/group_functions.hpp
```diff
@@ -0,0 +1,246 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifdef SYCL_DEVICE_ONLY
+#ifdef HIPSYCL_PLATFORM_HIP
+
+#ifndef HIPSYCL_LIBKERNEL_HIP_GROUP_FUNCTIONS_HPP
+#define HIPSYCL_LIBKERNEL_HIP_GROUP_FUNCTIONS_HPP
+
```
illuhad, line 33: Include `sub_group` header?

include/hipSYCL/sycl/libkernel/group_functions.hpp
```diff
@@ -0,0 +1,98 @@
...
+bool group_none_of(Group g, T x, Predicate pred) {
+  return group_none_of(g, pred(x));
+}
+
+// reduce
+template <typename Group, typename V, typename T, typename BinaryOperation>
+HIPSYCL_KERNEL_TARGET
+T group_reduce(Group g, V x, T init, BinaryOperation binary_op) {
+  T reduction = group_reduce(g, T{x}, binary_op);
+  return binary_op(reduction, init);;
```
illuhad: double semicolon ;)

include/hipSYCL/sycl/libkernel/host/group_functions.hpp
```diff
@@ -0,0 +1,514 @@
...
+    scratch[0] = x;
+  }
+
+  group_barrier(g);
+
+  T tmp = scratch[0];
+
+  group_barrier(g);
+
+  detail::free_shared_ptr(g, scratch);
```
illuhad: We are in SYCL. We call it local memory, except when referring explicitly to HIP/CUDA and their `__shared__` memory :-)

include/hipSYCL/sycl/libkernel/group_functions.hpp
```diff
@@ -0,0 +1,98 @@
...
+}
+
+
+} // namespace sycl
+} // namespace hipsycl
+
+#include "generic/hiplike/group_functions.hpp"
+#include "cuda/group_functions.hpp"
+#include "hip/group_functions.hpp"
+#include "host/group_functions.hpp"
```
illuhad: Better to have includes at the beginning of a file. Is there a hard reason why you need them at the end?
Maybe it would also be clearer if you `#ifdef backend` the include here. You already `ifdef` inside the headers, but doing it here in addition might make it clearer what happens?

include/hipSYCL/sycl/libkernel/nd_item.hpp
```diff
@@ -372,6 +374,7 @@ struct nd_item
   const range<dimensions> _local_range;
   const range<dimensions> _num_groups;
   const id<dimensions> _global_id;
+  void *_shared_ptr;
```
illuhad: `_local_ptr` ?

tests/sycl/group_functions.cpp
```diff
@@ -0,0 +1,1800 @@
...
+#include <math.h>
+#include <type_traits>
+
+#include <string>
+#include <sstream>
+
+using namespace cl;
+
+#ifdef TESTS_GROUPFUNCTION_FULL
+using test_types = boost::mpl::list<char, int, unsigned int,long long, float, double, sycl::vec<int, 1>, sycl::vec<int, 2>, sycl::vec<int, 3>, sycl::vec<int, 4>, sycl::vec<int, 8>,  sycl::vec<short, 16>, sycl::vec<long, 3>, sycl::vec<unsigned int, 3>>;
```
illuhad: Line breaks, e.g. use clang-format

tests/sycl/group_functions.cpp
```diff
@@ -0,0 +1,1800 @@
...
+
+namespace detail {
+
+template<typename T>
+using elementType = typename std::remove_reference<decltype(T{}.s0())>::type;
+
+template<typename T, int N>
+std::string T2str(sycl::vec<T, N> v)
+{
+    std::stringstream ss{};
```
illuhad: We use 2 spaces as indentation (`clang-format` can do it automatically for you  :-) )

tests/sycl/group_functions.cpp
```diff
@@ -0,0 +1,1800 @@
...
+
+}
+
+
+BOOST_FIXTURE_TEST_SUITE(group_functions_tests, reset_device_fixture)
+
+template<int N, int M, typename T>
+class test_kernel;
+
+template<int TEST_ID, typename T, typename DataGenerator, typename FunctionUnderTest, typename TestFunction>
```
illuhad: Why is `TEST_ID` capitalized?

tests/sycl/group_functions.cpp
```diff
@@ -0,0 +1,1800 @@
...
+
+    {
+        const auto functionUnderTest  = [](auto acc, size_t global_linear_id, sycl::sub_group sg, auto g, T local_value){
+          acc[global_linear_id] = sycl::group_all_of(g, static_cast<bool>(local_value));
+        };
+        const auto testFunction       = [](const std::vector<T> &vIn, const std::vector<T> &vOrig){
+            detail::check_binary_reduce<T, __LINE__>(vIn, local_size, global_size, std::vector<bool>{false, false, false, true}, "all_of");
+        };
+
+        test_nd_group_function_1d<__LINE__, T>(local_size, global_size,
```
illuhad: It's an interesting idea to use `__LINE__` as kernel name. I'm torn between thinking that it's genius or evil C-macro hacking ;)

illuhad (commented):
Some initial feedback :)

illuhad:
Let's try @fodinabor 's awesome new feature:

DieGoldeneEnte:
Do you know how I can stop clang-format from putting `HIPSYCL_KERNEL_TARGET` on the same line as the function?
I used it before, but since I need to reformat it by hand I only used it form the initial refactor :)

illuhad:
Let's try @fodinabor 's awesome new feature, take two:

@hipsycl-ci-bot test this again please!

EDIT: Retrigger!

include/hipSYCL/sycl/libkernel/group_functions.hpp
DieGoldeneEnte: It was needed in an earlier version, I'll move it to the top and add the `#ifdef`.

tests/sycl/group_functions.cpp
DieGoldeneEnte: missed it when renaming the other parameters, also changed it to `CallingLine`, since that's what I'm using as 'id' anyway.

tests/sycl/group_functions.cpp
DieGoldeneEnte: Since I call the kernel in the function I need some way to get unique names, the type alone is not enough. If you have another idea I'll glady use it. Once we use c++20 this can be replaced with [std::source_location::current()](https://en.cppreference.com/w/cpp/utility/source_location/current).

illuhad: The traditional way is to just name it: `test_nd_group_function_1d<class my_test_case_name_or_a_description_of_it, T>(...)`. But I see that this can be cumbersome when there are many test cases.

include/hipSYCL/sycl/libkernel/cuda/group_functions.hpp
DieGoldeneEnte: Do you really want `shufl_impl` or instead of `shfl_impl`/`shuffle_impl`? :)

tests/sycl/group_functions.cpp
DieGoldeneEnte: I use line to make the tests as copy&past ready as possible. As an added bonus I get the line number if something goes wrong and I have to read mangled names :>

include/hipSYCL/sycl/libkernel/cuda/group_functions.hpp
illuhad: :laughing: You're absolutely right, let's use `shuffle_impl`.

DieGoldeneEnte:
Let's see if I am mighty enough for this (and manage to guess all the necessary conditions)

@hipsycl-ci-bot test this again please!

DieGoldeneEnte:
I managed to create a clang-format config, that somewhat resembles existing code. If you want to get HIPSYCL_KERNEL_TARGET, \_\_device\_\_, ... on it's own line you can use:
```
StatementMacros:
  - HIPSYCL_KERNEL_TARGET
  - HIPSYCL_UNIVERSAL_TARGET
  - __device__
```

illuhad:
Sorry, but since your PR is already the test bed, you will have to endure some spam ;)

@hipsycl-ci-bot test with actual GPUs, please :)

EDIT: Attempt 3

DieGoldeneEnte:
Added a barrier tests and fixed the race-condition (which was in the tests...). No the groups don't override the input of the others. I also split the tests leading to a reduction of ~1min on my machine.

@hipsycl-ci-bot test it with CUDA for me <3

include/hipSYCL/sycl/libkernel/cuda/group_functions.hpp
```diff
@@ -0,0 +1,240 @@
...
+#include "../vec.hpp"
+#include <type_traits>
+
+namespace hipsycl {
+namespace sycl {
+
+namespace detail {
+
+template<typename T,
+         typename std::enable_if<std::is_arithmetic<T>::value, T>::type * = nullptr>
```
fodinabor: this is very C++11 style... `std::enable_if_t<std::is_arithmetic_v<T>, int> = 0>` could be used with C++17 :)

include/hipSYCL/sycl/libkernel/cuda/group_functions.hpp
```diff
@@ -0,0 +1,240 @@
...
+__device__
+T shuffle_impl(T x, int id) {
+  return __shfl_sync(0xFFFFFFFF, x, id);
```
fodinabor: maybe use a `constexpr std::uint32_t AllMask` or similar for the mask?

include/hipSYCL/sycl/libkernel/cuda/group_functions.hpp
```diff
@@ -0,0 +1,240 @@
...
+HIPSYCL_KERNEL_TARGET
+inline bool group_none_of(sub_group g, bool pred) {
+  return !__any_sync(0xFFFFFFFF, pred);
+}
+
+
+// reduce
+template<typename T, typename BinaryOperation>
+HIPSYCL_KERNEL_TARGET
+T group_reduce(sub_group g, T x, BinaryOperation binary_op) {
```
fodinabor, line 160: couldn't these _higher_ level implementations be shared between at least hiplike backends? looks like quite some code duplication .. or do you plan to possibly use some more intrinsics (like cuda's very new `__reduce_add_sync`)

include/hipSYCL/sycl/libkernel/generic/hiplike/group_functions.hpp
```diff
@@ -0,0 +1,591 @@
...
+    v.sC() = scratch[12];
+    v.sD() = scratch[13];
+    v.sE() = scratch[14];
+    v.sF() = scratch[15];
+  }
+}
+
+// reduce implementation
+template<typename Group, typename T, typename BinaryOperation,
+         typename std::enable_if<!std::is_same<Group, sub_group>::value, Group>::type * = nullptr>
```
fodinabor: as above with `std::is_same_v` :)

include/hipSYCL/sycl/libkernel/cuda/group_functions.hpp
```diff
@@ -0,0 +1,240 @@
...
+      local_x = binary_op(local_x, other_x);
+  }
+
+  size_t next_id = lid - 1;
+  if (lid == 0)
+    next_id = 0;
+
+  auto return_value = detail::shuffle_impl(local_x, lid - 1);
+
+  if (lid == 0)
```
fodinabor: just from looking at the spec, using `g.leader()` might be beautiful here.. but shouldn't change any semantics and in theory, should be well optimizable.. so guess it just comes down to whether one likes the common pattern or the readable one

include/hipSYCL/sycl/libkernel/generic/hiplike/group_functions.hpp
```diff
@@ -0,0 +1,591 @@
...
+  size_t group_range      = g.get_local_range().size();
+  auto elements_per_group = (last - first + group_range - 1) / group_range;
+  T *start_ptr            = first + elements_per_group * g.get_local_linear_id();
+  T *end_prt              = start_ptr + elements_per_group;
+
+  if (end_prt > last)
+    end_prt = last;
+
+  if (end_prt > start_ptr) {
+    auto local = binary_op(*start_ptr, *(start_ptr + 1));
```
fodinabor, line 294: what if `end_ptr == start_ptr + 1`? wouldn't `*(start_ptr + 1)` be an invalid access as the end_ptr is not in the range?
also typo in `end_prt` :)

include/hipSYCL/sycl/libkernel/hip/group_functions.hpp
```diff
@@ -0,0 +1,255 @@
...
+namespace detail {
+
+template<typename T,
+         typename std::enable_if<std::is_arithmetic<T>::value, T>::type * = nullptr>
+__device__
+T shuffle_impl(T x, int id) {
+  if constexpr (std::is_signed<T>::value) {
+    return __shfl(x, id);
+  } else {
+    return static_cast<T>(__shfl(static_cast<typename std::make_signed<T>::type>(x), id));
```
fodinabor: `static_cast<std::make_signed_t<T>>`

include/hipSYCL/sycl/libkernel/hip/group_functions.hpp
```diff
@@ -0,0 +1,255 @@
...
+namespace hipsycl {
+namespace sycl {
+
+namespace detail {
+
+template<typename T,
+         typename std::enable_if<std::is_arithmetic<T>::value, T>::type * = nullptr>
+__device__
+T shuffle_impl(T x, int id) {
+  if constexpr (std::is_signed<T>::value) {
```
fodinabor: `is_signed_v`

fodinabor (commented):
Love it! :)
Mostly some readability notes that I just mentioned once and could be applied generally and one small uncertainty about cerrectness :)

include/hipSYCL/sycl/libkernel/cuda/group_functions.hpp
DieGoldeneEnte: good Idea!

include/hipSYCL/sycl/libkernel/cuda/group_functions.hpp
DieGoldeneEnte: I could try to merge them into hiplike, I'll take a look. I originally split them up, because the shuffle-instruction is different. Now we have `shuffle_impl`, so I could give this a try.

include/hipSYCL/sycl/libkernel/cuda/group_functions.hpp
DieGoldeneEnte: I'll have a look again. I know it isn't yet implemented for groups (I could/should just do it if necessary), so I opted for `lid==0`.

include/hipSYCL/sycl/libkernel/generic/hiplike/group_functions.hpp
DieGoldeneEnte: correct I didn't think of this case in my last rewrite :/

DieGoldeneEnte:
@fodinabor 
Using the c++17 template-magic is a good point, didn't think about them at all ':)

include/hipSYCL/sycl/libkernel/cuda/group_functions.hpp
DieGoldeneEnte: I'll keep them separated for now, since it is likely I will separate them again when optimizing. I'll keep it in mind though in case the optimized code can be implemented in a hiplike fashion :)

DieGoldeneEnte:
@illuhad the first of the new commits adds leader() to sycl::group. I could put it in a separate PR, but then I have to rebase this PR to use it. I think it should be OK to hide it in here. In case you want to squash this, I can squash manually and keep that PR separate, as well as make sure the commit message is useful (it will likely be a long list of one-liners that don't make much sense when squashing automatically).

DieGoldeneEnte:
@hipsycl-ci-bot test please

Edit: Once more please

tests/sycl/group_functions/group_functions.hpp
```diff
@@ -53,7 +53,7 @@ using test_types = boost::mpl::list<char, unsigned int, float, double, sycl::vec
 namespace detail {
 
 template<typename T>
-using elementType = typename std::remove_reference<decltype(T{}.s0())>::type;
+using elementType = typename std::remove_reference_t<decltype(T{}.s0())>;
```
fodinabor: no need for `typename` before `_t` variants, it is only required for dependent names aka if you have a `::` after a type name (ignore those after namespace names... :)), see: https://en.cppreference.com/w/cpp/keyword/typename

DieGoldeneEnte: good catch :+1:

illuhad:
@DieGoldeneEnte that's fine. If you want to squash manually feel free, otherwise I would just squash automatically.

DieGoldeneEnte:
@illuhad Ok, I'll do that, just tell me when you want to merge it. It might make sense if the group functions are not merged under the title "removed redundant typename" or similar :D

illuhad:
please squash to prepare for merging :)

DieGoldeneEnte:
Done :)

Only lost a newline while moving the group-commit to the front.
```
diff --git a/include/hipSYCL/sycl/libkernel/group.hpp b/include/hipSYCL/sycl/libkernel/group.hpp
index c26e1d8..b553522 100644
--- a/include/hipSYCL/sycl/libkernel/group.hpp
+++ b/include/hipSYCL/sycl/libkernel/group.hpp
@@ -478,7 +478,6 @@ private:
   const host_barrier_type* _group_barrier;
   const id_type _local_id;
   void *_local_memory_ptr;
-
 public:
 #endif
```

DieGoldeneEnte:
@hipsycl-ci-bot test please, llvm doesn't want to be installed on the default actions

Edit: again please

DieGoldeneEnte:
there seems to be a problem with CUDA-broadcasts, it happens in less than every 100th run...
I'll take a look what the cause could be.

Edit: I could only get the test to fail 2 out of 500 runs...

illuhad:
If you are using host accessors or rely on buffer writebacks for validation in your tests, then this might also be cause by a bugged task graph if you don't have PR #395 merged yet.

DieGoldeneEnte:
That might be the reason, I could only reproduce it once. I think the problem occurred under heavy CPU-load, but that is just a suspicion.

DieGoldeneEnte:
@hipsycl-ci-bot test this :snake: 

Edit: again after rebase

DieGoldeneEnte:
FunctionUnderTest ->  TestedFunction
TestFunction -> ValidationFunction
TCompare -> compare_type
TInit -> initialize_type
T2str -> type_to_string
and a bunch of camel-case to snake-case conversion :)

Edit: removed emojis, it looked too much like some PR for a js-library...

DieGoldeneEnte:
@illuhad I think this is ready for merging now, pretty sure I got everything.

illuhad:
In it goes, thank you so much for this awesome contribution :)

PR #392, opened 2020-11-28
Expose the cxx flags and link lines for all backends

sbalint98:
This pr exposed the cxx flags and the link lines for all backends. This is a WIP PR since I am not sure if we should expose all the flags that are added in syclcc in syclcc.json. Currently, the flags necessary for hipSYCL to work with a spack base installation are exposed.

PR #393, opened 2020-11-29
use the rocm cmake integration

sbalint98:
This PR changes the cmake scripts to use the rocm cmake integration fi available. If hip si not found using find_package, we fall back to finding rocm by looking for hipcc

This PR is a draft because of the following issues:
* during compilation the following two types of warnings are displayed:
```
/tmp/hipsycl-installer-sbalint/include/hipSYCL/runtime/hip/hip_target.hpp:43:9: warning: '__HIP_PLATFORM_HCC__' macro redefined [-Wmacro-redefined]                        
#define __HIP_PLATFORM_HCC__
        ^
<command line>:5:9: note: previous definition is here
#define __HIP_PLATFORM_HCC__ 1
        ^
1 warning generated when compiling for gfx908.
In file included from /tmp/hipsycl-installer-sbalint/src/runtime/hip/hip_event.cpp:28:
In file included from /tmp/hipsycl-installer-sbalint/include/hipSYCL/runtime/hip/hip_event.hpp:32:
```
and
```
[ 62%] Building CXX object src/runtime/CMakeFiles/hipSYCL-rt.dir/hw_model/memcpy.cpp.o                                                                                      
clang-10: warning: -lclang_rt.builtins-x86_64: 'linker' input unused [-Wunused-command-line-argument]                                                                       
clang-10: warning: argument unused during compilation: '-amdgpu-function-calls=false' [-Wunused-command-line-argument]                                                      
clang-10: warning: argument unused during compilation: '-L/opt/hipSYCL/llvm/llvm/lib/clang/10.0.1/include/../lib/linux' [-Wunused-command-line-argument]                    
clang-10: warning: -lclang_rt.builtins-x86_64: 'linker' input unused [-Wunused-command-line-argument]                                                                       
clang-10: warning: argument unused during compilation: '-amdgpu-function-calls=false' [-Wunused-command-line-argument]                                                      
clang-10: warning: argument unused during compilation: '-L/opt/hipSYCL/llvm/llvm/lib/clang/10.0.1/include/../lib/linux' [-Wunused-command-line-argument]                    
```
I believe it would be okay to suppress these, but I am not sure. 
* Although compilation works for rocm targets, I was not able to check for any runtime problems. I don't expect any problems, but better to be sure.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,12 +24,19 @@ find_package(Boost COMPONENTS context fiber REQUIRED)
 
 # Check for CUDA/ROCm and clang
 find_package(CUDA QUIET)
-# We currently search for hipcc to check for ROCm installation
-find_program(HIPCC_COMPILER NAMES hipcc HINTS ${ROCM_PATH})
-set(ROCM_PATH /opt/rocm CACHE PATH "Path to ROCm installation")
+find_package(HIP QUIET HINTS ${ROCM_PATH} ${ROCM_PATH}/lib/cmake)
 
-if(HIPCC_COMPILER MATCHES "-NOTFOUND")
-  set(ROCM_FOUND false)
+# In case HIP is not found via the cmake pkgs we fall back to the legacy rocm discovery:
+if (NOT HIP_FOUND)
+  message(STATUS "Could not find hip cmake integration falling back to finding hipcc")
+  find_program(HIPCC_COMPILER NAMES hipcc HINTS ${ROCM_PATH})
+  set(ROCM_PATH /opt/rocm CACHE PATH "Path to ROCm installation")
+  if(HIPCC_COMPILER MATCHES "-NOTFOUND")
+    message(STATUS "Could not find rocm install with looking for hpcc. No rocm")
+    set(ROCM_FOUND false)
+  else()
+    set(ROCM_FOUND true)
+  endif()
 else()
   set(ROCM_FOUND true)
 endif()
--- a/include/hipSYCL/runtime/hip/hip_target.hpp
+++ b/include/hipSYCL/runtime/hip/hip_target.hpp
@@ -40,7 +40,9 @@
 #define __HIP_PLATFORM_NVCC__
 #include <hip/hip_runtime.h>
 #elif defined(HIPSYCL_RT_HIP_TARGET_ROCM)
+#ifndef __HIP_PLATFORM_HCC__
 #define __HIP_PLATFORM_HCC__
+#endif
 #include <hip/hip_runtime.h>
 #elif defined(HIPSYCL_RT_HIP_TARGET_HIPCPU)
 #include "hipCPU/hip/hip_runtime.h"
--- a/src/runtime/CMakeLists.txt
+++ b/src/runtime/CMakeLists.txt
@@ -1,4 +1,5 @@
 
+set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
 
 set(HIPSYCL_SOURCES
   application.cpp
@@ -74,8 +75,14 @@ endif()
 
 if(WITH_ROCM_BACKEND) 
   target_compile_definitions(hipSYCL-rt PRIVATE HIPSYCL_RT_ENABLE_HIP_BACKEND=1 HIPSYCL_RT_HIP_TARGET_ROCM=1)
-  target_include_directories(hipSYCL-rt PRIVATE ${ROCM_PATH}/include)
-  target_link_libraries(hipSYCL-rt PRIVATE ${ROCM_LIBS})
+  if(NOT HIP_FOUND)
+    target_include_directories(hipSYCL-rt PRIVATE ${ROCM_PATH}/include)
+    target_link_libraries(hipSYCL-rt PRIVATE ${ROCM_LIBS})
+  else()
+    # Supress warnings because wrongly set CXX arguments
+    target_compile_options(hipSYCL-rt PRIVATE -Wno-unused-command-line-argument)
+    target_link_libraries(hipSYCL-rt PRIVATE hip::host)
+  endif()
 endif()
 
 if(WITH_CPU_BACKEND)


fodinabor:
The first one seems to be generated, by CMake now adding `-D__HIP_PLATFORM_HCC__=1` to the CXX_FLAGS.
That should easily be supressable by adding a `#ifndef __HIP_PLATFORM_HCC` around the definition here:
https://github.com/illuhad/hipSYCL/blob/develop/include/hipSYCL/runtime/hip/hip_target.hpp#L43

For the second bunch of warnings, it might be interesting to look at the full command-line using `ninja -v` but I don't really know too much about HIP compilation...

illuhad:
Thank you! I agree with @fodinabor on the first one - we should only define this macro if it's not already defined.

For the second one, could it be that the ROCm cmake integration now tries to compile the runtime library as HIP code or something like that? As @fodinabor has said, we should investigate the flags that it generates.

sbalint98:
This is the output of `VERBOSE=1 make` would it be easier to read, and understand if had used ninja? I have never used ninja before.
```
cd /tmp/hipsycl-installer-sbalint/build/src/runtime && /opt/hipSYCL/llvm/llvm/bin/clang++ -DHIPSYCL_DEBUG_LEVEL=1 -DHIPSYCL_RT_ENABLE_HIP_BACKEND=1 -DHIPSYCL_RT_ENABLE_OMP_BACKEND=1 -DHIPSYCL_RT_HIP_TARGET_ROCM=1 -D__HIP_PLATFORM_HCC__=1 -D__HIP_ROCclr__=1 -DhipSYCL_rt_EXPORTS -I/tmp/hipsycl-installer-sbalint/include -I/tmp/hipsycl-installer-sbalint -isystem /opt/hipSYCL/rocm/hip/include -isystem /opt/hipSYCL/llvm/llvm/lib/clang/10.0.1/include/.. -isystem /opt/hipSYCL/rocm/hsa-rocr-dev/include -O3 -DNDEBUG -fPIC -mllvm -amdgpu-early-inline-all=true -amdgpu-function-calls=false -xhip --hip-device-lib-path=/opt/hipSYCL/rocm/rocm-device-libs/lib --cuda-gpu-arch=gfx900 --cuda-gpu-arch=gfx906 --cuda-gpu-arch=gfx908 -L/opt/hipSYCL/llvm/llvm/lib/clang/10.0.1/include/../lib/linux -lclang_rt.builtins-x86_64 -fopenmp=libomp -std=c++17 -o CMakeFiles/hipSYCL-rt.dir/device_id.cpp.o -c /tmp/hipsycl-installer-sbalint/src/runtime/device_id.cpp
clang-10: warning: -lclang_rt.builtins-x86_64: 'linker' input unused [-Wunused-command-line-argument]
clang-10: warning: argument unused during compilation: '-amdgpu-function-calls=false' [-Wunused-command-line-argument]
clang-10: warning: argument unused during compilation: '-L/opt/hipSYCL/llvm/llvm/lib/clang/10.0.1/include/../lib/linux' [-Wunused-command-line-argument]```
```
Adding the `#ifndef` around the definition worked :)

fodinabor:
no, the output would be the same.. I thought from your output, you'd be using ninja.. :D
I think there is a `-mllvm` missing in front of `-amdgpu-function-calls=false`
and the linker args are probably ignored as `-c` is passed and no linking is done.. so those are wrongly set as `CXX_FLAGS` instead of `LD_FLAGS` ... don't know where to fix those, though :D

illuhad:
> `-xhip --hip-device-lib-path=/opt/hipSYCL/rocm/rocm-device-libs/lib --cuda-gpu-arch=gfx900 --cuda-gpu-arch=gfx906 --cuda-gpu-arch=gfx908`

Yeah, so it's trying to compile for device...This is not necessary when building hipSYCL itself as `hipSYCL-rt` can be built with any C++ compiler and does not contain device code. So, the additional device code pass would only increase compilation time and should be avoided.

Is there a way to tell the ROCm cmake integration not to use device flags? Or only use it for detection whether ROCm is installed?

I think however that this compiler command line is quite valuable as it contains logic that we have for now in `syclcc`. It could be a great idea to use the command line that the ROCm cmake integration returns to generate the `syclcc` ROCm CXX and link lines once your PR #392 is in. This could help shift some of the responsibility to correctly build HIP applications from us to ROCm.

sbalint98:
Oh well yes, there are two teargets imported `hip::device` and `hip::host`. I used `hip::device` but as you said that is not necessary. using `hip::host` looks to be a step in the right direction:

I am not sure if this checks if rocm is installed properly. But it seems logical, to me since in order to have the targets imported all checks should pass regardless if the target is used or not, right?

> I think however that this compiler command line is quite valuable as it contains logic that we have for now in syclcc. It could be a great idea to use the command line that the ROCm cmake integration returns to generate the syclcc ROCm CXX and link lines once your PR #392 is in. This could help shift some of the responsibility to correctly build HIP applications from us to ROCm.

Yes, that definitely sounds like a great idea!

```
[ 24%] Building CXX object src/runtime/CMakeFiles/hipSYCL-rt.dir/operations.cpp.o
cd /tmp/hipsycl-installer-sbalint/build/src/runtime && /opt/hipSYCL/llvm/llvm/bin/clang++ -DHIPSYCL_DEBUG_LEVEL=1 -DHIPSYCL_RT_ENABLE_HIP_BACKEND=1 -DHIPSYCL_RT_ENABLE_OMP_BACKEND=1 -DHIPSYCL_RT_HIP_TARGET_ROCM=1 -D__HIP_PLATFORM_HCC__=1 -D__HIP_ROCclr__=1 -DhipSYCL_rt_EXPORTS -I/tmp/hipsycl-installer-sbalint/include -I/tmp/hipsycl-installer-sbalint -isystem /opt/hipSYCL/rocm/hip/include -isystem /opt/hipSYCL/rocm/hsa-rocr-dev/include -O3 -DNDEBUG -fPIC -L/opt/hipSYCL/llvm/llvm/lib/clang/10.0.1/include/../lib/linux -lclang_rt.builtins-x86_64 -fopenmp=libomp -std=c++17 -o CMakeFiles/hipSYCL-rt.dir/operations.cpp.o -c /tmp/hipsycl-installer-sbalint/src/runtime/operations.cpp
clang-10: warning: -lclang_rt.builtins-x86_64: 'linker' input unused [-Wunused-command-line-argument]
clang-10: warning: argument unused during compilation: '-L/opt/hipSYCL/llvm/llvm/lib/clang/10.0.1/include/../lib/linux' [-Wunused-command-line-argument]
```

I am not sure how to fix the linker args though. That seems to be a bug in the cmake integration?

illuhad:
Ok, great. It could be a bug, or maybe there are ROCm distributions where there are actually libraries in that path. I think at this point we can add `-Wno-unused-command-line-argument` when building `hipSYCL-rt` with ROCm support to suppress the warning.

sbalint98:
Thank you for the help @fodinabor and @illuhad , it compiles without warnings now :)

CMakeLists.txt
```diff
@@ -24,12 +24,19 @@ find_package(Boost COMPONENTS context fiber REQUIRED)
...
-
-if(HIPCC_COMPILER MATCHES "-NOTFOUND")
-  set(ROCM_FOUND false)
+find_package(HIP QUIET)
+
+# In case HIP is not found via the cmake pkgs we fall back to the legacy rocm discovery:
+if (NOT HIP_FOUND)
+  message(STATUS "Could not find hip cmake integration falling back to finding hipcc")
+  find_program(HIPCC_COMPILER NAMES hipcc HINTS ${ROCM_PATH})
+  set(ROCM_PATH /opt/rocm CACHE PATH "Path to ROCm installation")
```
illuhad, line 33: `ROCM_PATH` is used later for the `syclcc` configuration, so I think it should be set in any case. Also, if set, I don't think it should be ignored if HIP was found.
Is it possible to use it as a hint for the `find_package()` call?

sbalint98: So far the `ROCM_PATH` var was associated with the root directory of the ROCM directory right? My understanding is that `HINTS` can be specified for `find_package` but it should point to the directory containing the hip rocm configuration files, which would be something like: `ROCM_PATH/lib/cmake/` Therefore I don't think that using the `ROCM_PATH` as a hint a good idea, So I think the best is to have both `ROCM_PATH` and `ROCM_PATH`/lib/cmake added to the HINTS, but that is not a perfect solution. At least in my scripts for the build from a spack install I use the `CMAKE_PREFIX_PATH` to specify all the directories necessary for finding HIP and its dependnecies

illuhad: > ROCM_PATH and ROCM_PATH/lib/cmake added to the HINTS, but that is not a perfect solution

What would be the issue? If the hint doesn't work, the user can still override, no?

sbalint98: Nothing actually :D

illuhad:
Thank you!

PR #394, opened 2020-11-30
Add --rocm-path argument when compiling for HIP

illuhad:
Compile with `--rocm-path` when targeting ROCm.
It seems that newer clang versions need this, e.g. the clang 11 as shipped on Arch Linux.

EDIT: clang 10 supports neither `--rocm-path` nor the alternative `-nogpuinc`. This puts us in a difficult spot as clang 11 seems to require those arguments.

@sbalint98 This may be relevant for your work on syclcc.

### DIFF
--- a/bin/syclcc-clang
+++ b/bin/syclcc-clang
@@ -556,6 +556,10 @@ class hip_invocation:
     flags = [
         "-x", "hip",
         "--hip-device-lib-path=" + os.path.join(self._rocm_path, "lib"),
+        #--rocm-path is not yet supported by clang 10
+        #"--rocm-path=" + self._rocm_path, use -nogpuinc instead
+        # as we already handle includes
+        "-nogpuinc",
         "-D__HIPSYCL_ENABLE_HIP_TARGET__",
         "-D__HIPSYCL_CLANG__",
         # clang erroneously sets feature detection flags for 


illuhad:
@hipsycl-ci-bot test please :-)

EDIT: Rerun

illuhad:
hm.. it seems clang 10 doesn't support this argument, but clang 11 needs it..

DieGoldeneEnte:
You could try `--hip-device-lib-path` clang 10 knows it and it is an alternative for clang 12 (didn't check 11)
```

--rocm-device-lib-path=<arg>, --hip-device-lib-path=<arg>

ROCm device library path. Alternative to rocm-path.
```

illuhad:
We already use `--hip-device-lib-path`, but it seems this is no longer sufficient with clang 11, as it also wants to find the HIP runtime:
```
clang-11: error: cannot find HIP runtime. Provide its path via --rocm-path, or pass -nogpuinc to build without HIP runtime.
```
We could try this `-nogpuinc` since `syclcc` already takes care of include paths..

illuhad:
`-nogpuinc` also doesn't work on clang 10, unfortunately

DieGoldeneEnte:
I remenber I tried the same ~2 month ago and decided clang-10 is enough for me (but I wasn't sure if it isn't my setup) :D
I didn't find an easy fix at that time, but it wasn't a priority.

illuhad:
The fix will hopefully be PR #392, which would allow users to remove these arguments when compiling with clang 10 :)

illuhad:
Closing as the way to address this now is with the mechanism from PR #392

Issue #369, opened 2020-11-12
hipSYCL/sycl/buffer.hpp:185 "Writing back to external locations (not passed to " "buffer at construction) is unimplemented"' failed.


psath:
This only happens sporadically (Out of 100 runs last night, 14 gave an assertion failure, 2 just segfaulted.) All of my kernels that either do an explicit `cgh.copy` afterwards seem to be fine. This is the only one that relies on any destructor-based copyback.

ROCm 3.8, Relatively new Clang/libomp from Github. PRs #303 #337 #362 merged in. `--hipsycl-target=omp;hip:gfx803,gfx900`

This unit test is writing back to a scalar variable to double-check the accessor offset.  (In the wider context of a work-partitioning framework). The code basically looks like this. (I can work to extract a MWE tomorrow, if needed.)

```
sycl::buffer<int, 1> A = ... //(passed in by shared_ptr. heap-allocated, associated with heap-allocated host pointer, both with lifetimes >> the test function's scope. This data gets in and out of the kernels fine with separate cgh.copy calls
int64_t length = ...//passed in by value
int64_t offset = ...//passed in by value

sycl::queue q;

int scalar = 0
{ //data scope for scalar_buf
  sycl:buffer<int> scalar_buf(&scalar, sycl::range<1>{1});
  q.submit([&](sycl::handler &cgh) {
    auto scalar_acc = scalar_acc.get_access<sycl::access::mode::discard_write>(cgh);
    auto A_acc = A.get_access<sycl::access::mode::read>(cgh, sycl::range<1>(length), sycl::id<1>(offset));
    cgh.single_task<class offset_check>([=](){
      scalar_acc[0] = A_acc[0];
    });
  });
} // <-- Crashes here, during buffer destruction. scalar_buf is the only one that should be destroyed here.
```

It breaks regardless of whether scalar is stack-allocated or heap- and regardless of whether i use a `single_task` or `parallel_for`

I've managed to catch an instance of the assertion failure in GDB. The `writeback_ptr` is the correct address of scalar.
`get_host_device()` returns
```
$8 = {_backend = {id = (hipsycl::rt::backend_id::hip | hipsycl::rt::backend_id::omp | unknown: 316),
    sw_platform = hipsycl::rt::api_platform::cuda, hw_platform = (unknown: 368)}, _device_id = 0}
```
But attempting to check the other half of the conditional triggers another assertion failure
```
p data->get_memory(get_host_device())
[...]

PR #395, opened 2020-12-02
Add explicit requirements to the data users of own buffer

illuhad:
In order to construct the DAG and calculate dependencies between DAG nodes, we need to track by which DAG nodes a `data_region` (i.e. data content of a buffer) is already used.

For this purpose, the `dag_builder` will iterate over all requirements of a node and add the node to the `data_user_tracker` objects of the buffers that its requirements refer to.

This alone is however insufficient for explicit requirements, as used as underlying mechanism e.g. for host accessors or buffer writebacks. In that case, the DAG node *itself* is a requirement.
As such, we also need to add the requirement DAG node to the data users of its *own* buffer that it formulates a requirement for.

This was previously missing, and is fixed by this PR.

This missing step can lead to missing synchronization and incorrectly constructed DAGs, potentially leading to race conditions, corruption and a whole lot of undefined behavior.

* This PR may solve issue #369 
* This PR solves at least the issues reported in #380 where a crash occurs at the application end around buffer destruction (e.g. `nbody-sycl`). It might also resolve some of the other reported issues.

### DIFF
--- a/src/runtime/dag_builder.cpp
+++ b/src/runtime/dag_builder.cpp
@@ -47,6 +47,25 @@ namespace rt {
 
 namespace {
 
+// Add this node to the data users of the memory region of the specified
+// requirement
+void add_to_data_users(dag_node_ptr node, memory_requirement *mem_req) {
+  assert(mem_req);
+  if (mem_req->is_buffer_requirement()) {
+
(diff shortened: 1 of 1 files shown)


illuhad:
@hipsycl-ci-bot test this and do your duty, you most beautiful bot!

PR #396, opened 2020-12-03
Disable clang 12 in CI due to unrelated clang ICE

illuhad:
It seems clang 12 was recently updated in the apt.llvm.org repos, and it now segfaults when compiling our unit tests. Since nothing from hipSYCL appears in the stack trace, I assume that it is unrelated and simply caused by instability of clang 12 (which is still unreleased and in development, so entirely possible).

This PR disables clang 12 in CI.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        clang_version: [10, 11, 12]
+        clang_version: [10, 11]
         rocm_version: [3.8]
         os: [ubuntu-20.04, ubuntu-18.04]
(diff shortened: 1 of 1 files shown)


fodinabor:
also you might want to limit running pull request actions and push actions for your branches, this looks promising (courtesy to piersy in https://github.community/t/how-to-trigger-an-action-on-push-or-pull-request-but-not-both/16662/10): 
```
  push:
  pull_request:
    branches:
      # Branches from forks have the form 'user:branch-name' so we only run
      # this job on pull_request events for branches that look like fork
      # branches. Without this we would end up running this job twice for non
      # forked PRs, once for the push and then once for opening the PR.
    - '**:**'
```

illuhad:
@fodinabor performance optimization can come later ;)

PR #398, opened 2020-12-03
Use std::promise for omp_node_event instead of busy wait

illuhad:
The old implementation of `omp_node_event` was using a spin lock. For CPU backends, this is not a good strategy as it can lead to reduced performance for simultaneously running kernels. This PR uses `std::promise/future` instead as the next best thing in order to mitigate such performance issues.


illuhad:
@hipsycl-ci-bot test as well please to be sure :)

PR #399, opened 2020-12-04
Fix multi-gpu device handling

illuhad:
There was a bug preventing hipSYCL from working correctly with multiple GPUs. This PR fixes this:
* Fix missing assignment of updated active device in CUDA/HIP device manager
* Fix missing device activation before constructing an event

With this PR, I can run the multigpu nstream from PRK on two CUDA GPUs. Result validation fails with 4 GPUs, but I believe this is because PRK's calculation of the offsets for the decomposed buffers assigned to individual GPUs seems incorrect to me - ping @jeffhammond

CC @psath - I believe this might fix your multi-GPU issue, but unfortunately I only have a single AMD GPU for testing and can only do multi-GPU testing on NVIDIA.

### DIFF
--- a/src/runtime/cuda/cuda_device_manager.cpp
+++ b/src/runtime/cuda/cuda_device_manager.cpp
@@ -27,6 +27,7 @@
 
 #include <cuda_runtime_api.h>
 
+#include "hipSYCL/common/debug.hpp"
 #include "hipSYCL/runtime/cuda/cuda_device_manager.hpp"
 #include "hipSYCL/runtime/error.hpp"
 
@@ -48,6 +49,10 @@ cuda_device_manager::cuda_device_manager() {
 void cuda_device_manager::activate_device(int device_id)
 {
   if (_device != device_id) {
+
+    HIPSYCL_DEBUG_INFO << "cuda_device_manager: Switchting to device "
+                       << device_id << std::endl;
+
     auto err = cudaSetDevice(device_id);
 
     if (err != cudaSuccess){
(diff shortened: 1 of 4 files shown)


illuhad:
@hipsycl-ci-bot test this for me please :)

psath:
> CC @psath - I believe this might fix your multi-GPU issue, but unfortunately I only have a single AMD GPU for testing and can only do multi-GPU testing on NVIDIA.

Thanks!, I'll give it a test after morning meetings.

jeffhammond:
I never ran that code on more than 2 devices because I wrote it on a NUC and the only two devices I had were CPU and GPU.  However, I don't see the bug.  Can you provide some details on https://github.com/ParRes/Kernels/issues/524 with what you see?  I can run it on a DGX-A100 8 GPU box but don't know when I'll have time for that.

illuhad:
@jeffhammond Yes, it works just fine with two devices. I have provided some more information.

psath:
> > CC @psath - I believe this might fix your multi-GPU issue, but unfortunately I only have a single AMD GPU for testing and can only do multi-GPU testing on NVIDIA.
> 
> Thanks!, I'll give it a test after morning meetings.

No dice, still getting segfaults in `libamdhip64.so` when I try to do an explicit copy command group to my 2nd AMD GPU, that I don't see when running HIP directly. I'll do some more diagnostics and open a separate issue for it.

illuhad:
@psath Might be that there are additional issues in the buffer-accessor implementation. The application I have tested with uses USM, so no buffers.

PR #401, opened 2020-12-07
Remove explicit copy tests with host accessors that are not covered by the spec

### LINKED ISSUE #385: Spurious unit test failures with explicit copies
For some time we have seen spurious unit test failures for the explicit copy tests involving host accessors. I did some investigating and it turns out there's a massive race condition between the explicit copy and the host accessor that is used to access the data afterwards for validation. However, it might not be hipSYCL's fault:

In those tests, the explicit copy operates on two accessors, at least one of which is a host accessor. The problem is that, since host accessors are not associated with a command group handler, a SYCL implementation cannot know that the accessor is actually used by the copy operation [1]. Because of this, of course it cannot create a dependency between the explicit copy and the host access afterwards since it doesn't know that the accessor is used by the copy. As a result there are no synchronization guarantees whatsoever between the explicit copy and the host access.

I'm uncertain now whether the intent of the SYCL specification is to allow this at all as it doesn't work with the usual rules for accessor dependencies.

@psalz @fknorr Since you added those tests, do you depend on support for this? There are various approaches we could take:
1. remove those tests entirely, interpreting the SYCL specification as not allowing host accessors in explicit copy invocations
2. keep the tests, but acknowledge that additional manual synchronization is required by inserting `queue::wait()` calls before the host access to make sure the memcpy has completed
3. Add additional dependency semantics for explicit copies: Check the `copy()` arguments, and when an argument is a host accessor, manually add it to the operation's requirements. It is not yet entirely clear how this can best be implemented as host accessors work somewhat differently than the device accessors that the `handler`'s requirement handling is designed for.

[1] At least by relying on the regular mechanisms for that. See the proposed solutions.

EDIT: The specification also requires that host accessors block other operations on their buffer for the duration of their existence. This is not yet implemented in hipSYCL - if it were, I suspect that using host accessors in explicit copy operation might cause additional problems such as deadlocks.

illuhad:
Removes explicit copy tests using host accessors that are not covered by the SYCL specification for good. Those tests had already been disabled for some time (see #385)

Closes #385


psalz:
It would probably also be a good idea to add static assertions into `handler::copy` that prevent it from being instantiated with host accessors at all.

illuhad:
hm...I'm not yet sure if we want to do that. We could still expose using host accessors in explicit copy functions as an extension. After all it works if you have explicit synchronization afterwards.

PR #403, opened 2020-12-08
Updated repos for spack support and add testing

sbalint98:
This PR adds the self-built boost installation to the base package. From now on, we use spack to build the dependencies for hipSYCL using spack. This change intends to make the packages more flexible, up-to-date and more maintainable.

* `hipSYCL-nightly` has hipSYCL from the develop branch 
* `hipSYCL-base` has clang-11 and boost 1.73
* `hipSYCL-rocm` has rocm 3.8

Please note:
the `hipSYCL-nightly` package now has dependencies on both rocm and Cuda. If the installation of all backends is not intended there are hipSYCL packages available with the following backends enabled: omp, omp-cuda, omp-rocm. They are available through the `hipSYCL-<backend-combination>` packages.

The provisional packages are available at http://repo.urz.uni-heidelberg.de/sycl/test_spack_build/

Furthermore, a test step is added before packaging.

This PR is a draft because:
* The following warning is displayed when compiling for hip:
```
/tmp/hipSYCL-test/tests/sycl/extensions.cpp:253:7: warning: ignoring return value of function declared with 'nodiscard' attribute [-Wunused-result]
      hipMemcpyAsync(target_ptr, native_mem, test_size * sizeof(int),
      ^~~~~~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
* Runtime testing on hip device is needed


illuhad:
Thanks! The HIP compilation warning is expected - it's because we don't look at the `hipError_t` when invoking `hipMemcpyAsync` in that particular test. It's not related to your change.

>  If the installation of all backends is not intended there are hipSYCL packages available with the following backends enabled: omp, omp-cuda, omp-rocm. 

* If we add additional cpu backends, would it make sense to have them all available in a cpu package? e.g. if we also introduced a TBB backend, would it make sense to have a single hipSYCL-cpu-nightly instead of separate omp and tbb packages?

* What is the difference between `hipSYCL-rocm-nightly` and `hipSYCL-omp-rocm-nightly`? I think `hipSYCL-rocm-nightly` is only the ROCm stack, right? Maybe we can rename it to have a clearer distinction? Otherwise it might be confused with the packages for hipSYCL itself.

* I wonder if it would make sense to name the package that has support for everything `hipSYCL-full` or something like that?


Anyone reading this, I'm also interested in whether you prefer `hipSYCL-*-nightly` or `hipSYCL-nightly-*`.

devops/repos/build_nightly.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+source /etc/profile
+cd /home/sbalint/hipSYCL-test-spack/devops/repos/
```
illuhad: Please don't hardcode directories :)

devops/repos/create_pkgs.sh
```diff
@@ -10,15 +10,15 @@ cd $HIPSYCL_PKG_SCRIPT_DIR_ABS/packaging
 
 stage_dir=$HIPSYCL_PKG_DEVOPS_DIR/$HIPSYCL_PKG_UBUNTU_PKG_DIR
 echo "Starting Ubuntu 18.04 pkg build"
-singularity exec $HIPSYCL_PKG_CONTAINER_DIR/hipsycl-ubuntu-18.04.sif sh make-ubuntu-pkg.sh
+singularity exec $HIPSYCL_PKG_CONTAINER_DIR/hipsycl-ubuntu-18.04 sh make-ubuntu-pkg.sh
```
illuhad: By the way.. Do you have any opinion regarding Ubuntu 20.04 and whether/for how long we should still support 18.04?

devops/repos/test_installation.sh
```diff
@@ -0,0 +1,41 @@
...
+
+#ubuntu 18.04 for has insufficient cmake version for building the tests
+#TODO:Install recent cmake in ubuntu 18.04
+for distro in archlinux-rolling centos-7 #ubuntu-18.04 
+do
+echo "Build and comp test for omp backend for $distro"
+SINGULARITYENV_HIPSYCL_PLATFORM=omp singularity exec $HIPSYCL_PKG_CONTAINER_DIR/hipsycl-$distro   cmake \
+    -DCMAKE_PREFIX_PATH=/opt/hipSYCL/boost/boost -S /tmp/hipSYCL-test/tests \
+    -B /tmp/build/
+SINGULARITYENV_HIPSYCL_PLATFORM=omp singularity exec -H /tmp/build $HIPSYCL_PKG_CONTAINER_DIR/hipsycl-$distro \
```
illuhad: Maybe use the new `HIPSYCL_TARGETS` mechanism instead of old `HIPSYCL_PLATFORM`/`HIPSYCL_GPU_ARCH`?

install/scripts/base-archlinux-rolling.def
```diff
@@ -2,23 +2,16 @@ BootStrap: docker
...
-cp ./install-cuda.sh ${SINGULARITY_ROOTFS}/install-cuda.sh 
 cp ./install-hipsycl.sh ${SINGULARITY_ROOTFS}/install-hipsycl.sh
-cp ./install-llvm.sh ${SINGULARITY_ROOTFS}/install-llvm.sh
-cp ./add-repo-arch.sh ${SINGULARITY_ROOTFS}/add-repo-arch.sh
+cp ./install-cuda.sh ${SINGULARITY_ROOTFS}/install-cuda.sh
+cp ./spack-install-hipsycl.sh  ${SINGULARITY_ROOTFS}/spack-install-hipsycl.sh 
+cp ./spack-install-llvm.sh ${SINGULARITY_ROOTFS}/spack-install-llvm.sh 
+cp ./spack-install-boost.sh ${SINGULARITY_ROOTFS}/spack-install-boost.sh 
+cp ./spack-install-rocm.sh ${SINGULARITY_ROOTFS}/spack-install-rocm.sh 
+cp ./spack-syclcc.json ${SINGULARITY_ROOTFS}/syclcc.json
```
illuhad: Why do we need to copy `syclcc.json` in? Shouldn't it be possible to let cmake generate it and configure cmake so that it generates it as we require?

install/scripts/base-centos-7.def
```diff
@@ -1,31 +1,30 @@
...
 cp ./install-hipsycl.sh ${SINGULARITY_ROOTFS}/install-hipsycl.sh
-cp ./install-llvm.sh ${SINGULARITY_ROOTFS}/install-llvm.sh
-cp ./add-repo-centos.sh ${SINGULARITY_ROOTFS}/add-repo-centos.sh
+cp ./install-cuda.sh ${SINGULARITY_ROOTFS}/install-cuda.sh
+cp ./install-base-spack.sh ${SINGULARITY_ROOTFS}/install-base-spack.sh
+cp ./spack-install-hipsycl-centos.sh ${SINGULARITY_ROOTFS}/spack-install-hipsycl.sh 
+cp ./spack-install-llvm.sh ${SINGULARITY_ROOTFS}/spack-install-llvm.sh 
+cp ./spack-install-boost.sh ${SINGULARITY_ROOTFS}/spack-install-boost.sh 
+cp ./spack-install-rocm.sh ${SINGULARITY_ROOTFS}/spack-install-rocm.sh 
+cp ./spack-syclcc.json ${SINGULARITY_ROOTFS}/syclcc.json
```
illuhad: same here

install/scripts/install-hipsycl.sh
```diff
@@ -4,14 +4,14 @@ HIPSYCL_PKG_LLVM_VERSION_MINOR=${HIPSYCL_PKG_LLVM_VERSION_MINOR:-0}
 HIPSYCL_PKG_LLVM_VERSION_PATCH=${HIPSYCL_PKG_LLVM_VERSION_PATCH:-1}
 HIPSYCL_PKG_LLVM_REPO_BRANCH=${HIPSYCL_PKG_LLVM_REPO_BRANCH:-release/${HIPSYCL_PKG_LLVM_VERSION_MAJOR}.x}
 
-export HIPSYCL_INSTALL_PREFIX=${HIPSYCL_INSTALL_PREFIX:-/opt/hipSYCL}
+export HIPSYCL_INSTALL_PREFIX=${HIPSYCL_INSTALL_PREFIX:-/home/sbalint/hipSYCL-installation}
```
illuhad: Can we set the default maybe to something like `$HOME/hipSYCL` so that it also works for other users out of the box? We occasionally have users that use these scripts :-)

install/scripts/install-hipsycl.sh
```diff
@@ -4,14 +4,14 @@ HIPSYCL_PKG_LLVM_VERSION_MINOR=${HIPSYCL_PKG_LLVM_VERSION_MINOR:-0}
...
 
 set -e
-HIPSYCL_BUILD_DIR=${HIPSYCL_BUILD_DIR:-/tmp/hipSYCL-installer}
+HIPSYCL_BUILD_DIR=${HIPSYCL_BUILD_DIR:-/tmp/hipsycl-installer-sbalint}
```
illuhad: Also maybe use `$USER` here?

install/scripts/install-hipsycl.sh
```diff
@@ -35,15 +35,15 @@ mkdir -p $HIPSYCL_BUILD_DIR/build
...
 -DWITH_CUDA_BACKEND=$HIPSYCL_WITH_CUDA \
 -DWITH_ROCM_BACKEND=$HIPSYCL_WITH_ROCM \
--DLLVM_DIR=$HIPSYCL_INSTALL_PREFIX/llvm/lib/cmake/llvm \
--DROCM_PATH=$HIPSYCL_INSTALL_PREFIX/rocm \
--DCUDA_TOOLKIT_ROOT_DIR=$HIPSYCL_INSTALL_PREFIX/cuda \
--DCLANG_EXECUTABLE_PATH=$HIPSYCL_INSTALL_PREFIX/llvm/bin/clang++ \
+-DLLVM_DIR=/opt/hipSYCL/llvm/lib/cmake/llvm \
+-DROCM_PATH=/opt/hipSYCL/rocm \
+-DCUDA_TOOLKIT_ROOT_DIR=/opt/hipSYCL//cuda \
+-DCLANG_EXECUTABLE_PATH=/opt/hipSYCL//llvm/bin/clang++ \
```
illuhad: Avoid double /

install/scripts/install-spack-hipsycl.sh
```diff
@@ -0,0 +1,61 @@
...
+echo "Cloning hipSYCL"
+git clone --recurse-submodules -b $HIPSYCL_REPO_BRANCH https://github.com/$HIPSYCL_REPO_USER/hipSYCL $HIPSYCL_BUILD_DIR
+
+fi
+
+mkdir -p $HIPSYCL_BUILD_DIR/build
+cd $HIPSYCL_BUILD_DIR/build
+
+# We need the llvm module to be loaded in order to be able to find the openmp rt
+export SPACK_ROOT=/root/spack
```
illuhad: Is it possible to formulate it such that it works by default also for non-privileged users who wish to install on bare metal (i.e. not building singularity image)?

install/scripts/install-spack-hipsycl.sh
```diff
@@ -0,0 +1,61 @@
...
+-DWITH_CUDA_BACKEND=$HIPSYCL_WITH_CUDA \
+-DWITH_ROCM_BACKEND=$HIPSYCL_WITH_ROCM \
+-DLLVM_DIR=/opt/hipSYCL/llvm/llvm/lib/cmake/llvm \
+-DROCM_PATH=/opt/hipSYCL/rocm \
+-DBOOST_PATH=/opt/hipSYCL/boost/boost \
+-DCUDA_TOOLKIT_ROOT_DIR=/opt/hipSYCL/cuda \
+-DCLANG_EXECUTABLE_PATH=/opt/hipSYCL/llvm/llvm/bin/clang++ \
+-DCLANG_INCLUDE_PATH=$LLVM_INCLUDE_PATH \
+-DCMAKE_INSTALL_PREFIX=$HIPSYCL_INSTALL_PREFIX \
+-DCMAKE_PREFIX_PATH="$rocm_path/comgr/lib/cmake;$rocm_path/rocm-device-libs/lib/cmake;$rocm_path/hsa-rocr-dev/lib/cmake;$rocm_path/hsa-rocr-dev/;$rocm_path/hip/lib/cmake" \
```
illuhad: Does the ROCm cmake integration provide a better way, or is it intended that we need to manually look for all components?

fodinabor:
hipSYCL-*-nightly 👍🏼

install/scripts/install-spack-hipsycl.sh
sbalint98: Well, I am not sure if there is one.  The HIP package internally calls find_package for `comgr` `hsa-rocr-dev` and `rocm-device-libs`. I have to look but I think there is no option to pass HINTS to these so I think this might be the only way. Probably what makes this so ugly is that the spack layout is quite unusual, and we have to pass every component separately.

install/scripts/base-archlinux-rolling.def
sbalint98: I tried to do that but my cmake knowledge has proved to be insufficient for this task. The problem is that `find_package(HIP)` doesn't set the variables that we need but only imports the `hip::host` and `hip::device` targets. I tried to retrieve the paths using `get_properties` but the dependencies also only use other imported targets as dependencies not actual paths. I would be very happy to see if there is a clean solution for that :)

devops/repos/create_pkgs.sh
sbalint98: Not really, but with this update to the packages it should be fairly simple to add support for 20.04, (and for other oses as well)

devops/repos/build_nightly.sh
sbalint98: Oops :D

install/scripts/install-hipsycl.sh
sbalint98: I didn't intend to leave this in here sorry, I will remove it

install/scripts/install-spack-hipsycl.sh
sbalint98: I don't see right now how this could work well. I think these spack-install scripts are not intended to be used by a regular user. We have to do some hacking (changing where spack installs the packages) to have the different dependencies in separate folders so that they can be packaged. A user would probably be happy if all the dependencies are installed in the same directory. However, I think that a separate script would be handier for that purpose. Maybe these scripts should be moved to `devops/repo`. 

Ultimately I think all of this should be part of the CI workflows, but for that, we probably need to update jarvis, since the singularity docs recommend 3.18 kernel version for --fake-root. (jarvis has 3.10)

devops/repos/test_installation.sh
sbalint98: I'll have a look

illuhad:
When testing with the packages I noticed that `syclcc` doesn't actually generate any binaries. Further investigation showed that clang doesn't start correctly:
```
$ /opt/hipSYCL/llvm/llvm/bin/clang++
Ungültiger Maschinenbefehl (Speicherabzug geschrieben)
```
i.e. invalid machine instruction. Is it possible that you have compiled for a relatively new hardware architecture? Have you compiled with `-march` for jarvis, i.e. skylake? We need to make sure that poor people with older machines can still execute the software ;)

sbalint98:
> i.e. invalid machine instruction. Is it possible that you have compiled for a relatively new hardware architecture? Have you compiled with -march for jarvis, i.e. skylake? We need to make sure that poor people with older machines can still execute the software ;)

I thought we are only packaging software for the skylake master-race ;) Seriously, sorry for that. Spack compiles for the host arch by default. This should be fixed now.

> If we add additional CPU backends, would it make sense to have them all available in a CPU package? e.g. if we also introduced a TBB backend, would it make sense to have a single hipSYCL-CPU-nightly instead of separate omp and tbb packages?

I have never used TBB before, but I assume getting TBB installed on a system is more involved than the same for omp. Shouldn't it be better the user to be able to choose from to either use TBB or omp? Would it be possible to use both?

> What is the difference between hipSYCL-rocm-nightly and hipSYCL-omp-rocm-nightly? I think hipSYCL-rocm-nightly is only the ROCm stack, right? Maybe we can rename it to have a clearer distinction? Otherwise, it might be confused with the packages for hipSYCL itself.

Yes, exactly the `hipSYCL-rocm-nighly` is only the rocm stack. I was trying to keep some coherency with the already existing packages. But I guess we can break that. can you suggest a name for the rocm stack packages? maybe `hipSYCL-rocm-base` ?
 
> I wonder if it would make sense to name the package that has support for everything hipSYCL-full or something like that?

Here as well I was trying to keep consistent with the old packages, but sure we can do that :)

sbalint98:
I had a look at spack filesystem views today, but since it requires to keep the original spack installation tree intact, even if hard links are used, I don't think it is any good for us.

illuhad:
Can you give a quick summary of what the current state is? Is it ready for another review?

> I thought we are only packaging software for the skylake master-race ;) Seriously, sorry for that. Spack compiles for the host arch by default. This should be fixed now.

Thanks!

> I have never used TBB before, but I assume getting TBB installed on a system is more involved than the same for omp. Shouldn't it be better the user to be able to choose from to either use TBB or omp? Would it be possible to use both?

TBB is available as package pretty much everywhere, and as C++17 parallel STL implementations build on top of TBB afaik, it's going to be even more widespread in the future. Yes, it would absolutely be possible to use TBB and OpenMP simultaneously.

>  Yes, exactly the hipSYCL-rocm-nighly is only the rocm stack. I was trying to keep some coherency with the already existing packages. But I guess we can break that. can you suggest a name for the rocm stack packages? maybe hipSYCL-rocm-base ?

Ah, I see. Consistency with existing naming schemes is good, but I think not top priority when consistency leads to confusion with the new naming scheme. `hipSYCL-rocm-base` sounds fine to me.

> Here as well I was trying to keep consistent with the old packages, but sure we can do that :)

Okay, it might make sense to keep `hipSYCL` then. Maybe we can have `hipSYCL-full` as well and of the two be a meta-package that aliases the other one or something like that?

sbalint98:
I set this back to draft because I wanted to restructure this to be used in a GitHub Actions workflow. However, it seems that it is not possible from inside a singularity container to create, and open a shell to a second container; therefore, this is not a viable path currently. We would either have to move the GitHub runner into a VM or run them on bare metal. The restructuring should be done sooner or later anyway, but I wouldn't start that unless I can actually test that these scripts actually work well with actions. 

Furthermore, there resulting base package became quite large. I will need to have a look at how to reduce its size. Currently, it is around ~5GB. I haven't touched the defaults of spack, therefore I hope if I disable building `gold`, `internal unwind`, `lldb`, and `polly` we can save some space. 

What would you think about the following naming scheme:
The base packages would be: 
* `hipSYCL-base<-nightly>`
* `hipSYCL-base-rocm<-nightly>`
And for the hipSYCL packages I suggest:
* `hipSYCL-omp`
* `hipSYCL-omp-cuda`
* `hipSYCL-omp-rocm`
*  `hipSYCL-omp-rocm-cuda`
And I would introduce two meta-packages in order to keep consistent with the previous packages:
* `hipSYCL-full` -> `hipSYCL-omp-rocm-cuda`
* `hipSYCL` -> `hipSYCL-omp-rocm-cuda`

illuhad:
> I set this back to draft because I wanted to restructure this to be used in a GitHub Actions workflow. However, it seems that it is not possible from inside a singularity container to create, and open a shell to a second container; therefore, this is not a viable path currently

Is it also not possible with the singularity `--fakeroot` option?

> We would either have to move the GitHub runner into a VM or run them on bare metal.

VM will be very complicated as we would need to forward the GPUs into the VM using PCIe passthrough or some special drivers for virtualization. Most likely the GPUs then would also be inaccessible from the host for manual testing scenarios. I'd like to stay away from that if possible. I'd rather have it run on bare metal, although we would need to reevaluate security in that case.

> Furthermore, there resulting base package became quite large. I will need to have a look at how to reduce its size. Currently, it is around ~5GB. I haven't touched the defaults of spack, therefore I hope if I disable building gold, internal unwind, lldb, and polly we can save some space.

Yeah, this is quite large. I don't think gold, lldb etc would consume so much memory. Where is all of this coming from?

>  hipSYCL-base<-nightly>
    hipSYCL-base-rocm<-nightly>

Okay.

>  And for the hipSYCL packages I suggest:
    hipSYCL-omp
    hipSYCL-omp-cuda
    hipSYCL-omp-rocm
    hipSYCL-omp-rocm-cuda
    And I would introduce two meta-packages in order to keep consistent with the previous packages:
    hipSYCL-full -> hipSYCL-omp-rocm-cuda
    hipSYCL -> hipSYCL-omp-rocm-cuda

Okay. We might have to keep in mind that we might get Level Zero/SPIR-V/Intel support in the future, so I'm a bit worried that this might lead to combinatoric explosion. Is there any meaningful way to combine support for some targets? Could it make sense to only have CUDA as optional dependency since we cannot redistribute it for legal reasons?

EDIT: Wouldn't we also need `hipSYCL-omp-*` with `nightly` suffix?

sbalint98:
> Is it also not possible with the singularity --fakeroot option?

In the singularity admin [docs](https://sylabs.io/guides/3.7/admin-guide/installation.html#installation-on-linux) It says that it is not recommended and requires the outer container to run with full privileges. However, on this [github issue](https://github.com/hpcng/singularity/issues/4037) it is claimed that it can be done. However, I need to look into more into the actual process of setting this up. 

> VM will be very complicated as we would need to forward the GPUs into the VM using PCIe passthrough or some special drivers for virtualization. Most likely the GPUs then would also be inaccessible from the host for manual testing scenarios. I'd like to stay away from that if possible. I'd rather have it run on bare metal, although we would need to reevaluate security in that case.

I haven't thought about that. This sounds like something that we really want to avoid. What I had in mind was to have dedicated runners for building the packages, they could be located in a VM since they don't need GPU access in order to build the packages. For testing, we could have more runners with the supported operating systems available. Although I must admit that maintaining these runners might be a burden 

Regarding the size of the packages:

I used to the following command to install llvm:
`spack install llvm@$llvm_version cuda=False libcxx=False target=x86_64`

| Distro       | base-pkg | rocm-pkg | build-size-base       | build-size-rocm | install-size-base | install-size-rocm | old-rocm    |  old-base   |
| ------------ | -------- | -------- | --------------------- | --------------- | ----------------- | ----------------- | --- | --- |
| ubuntu-18.04 | 1.4G     | 807M     | 7.2G (after gc: 6.9G) | 4.4G  after gc  |                   |                   |     |     |
| centos-7     | 1.5G     | 820M     | 7.2G (after gc: 6.9G) | 4.0G  after gc  | 6.4GB             | 3.9GB             |   1.6G  | 3.2G    |
| Archlinux    | 1.8G     | 917M     | 7.3G (after gc: 6.9G) | 4.0G after gc   |                   |                   |     |     |



llvm is installed in the `/opt/hipSYCL/llvm/llvm` folder, above the build size is the size of the `/opt/hipSYCL/llvm` folder which contains besides llvm other software that spack installed as llvm dependencies. The size of the actual llvm installation is around 6.3GB. Therefore, spack has installed 700MB most likely superfluous software. This can be reduced if we tell spack to use the version installed on the system; in that case, we will have to introduce additional package dependencies.

Inside the actual llvm installation (`/opt/hipSYCL/llvm/llvm`)
* lib 1.7GB
* bin 2.1GB
* libexec 2.5GB

Using our install llvm script (llvm-9)
* lib 580M
* libexec 28K
* bin 918M

In the old handwritten version in libexec we only had `c++-analyer`.

In spack, one has the following options when building llvm:
```
Variants:                                                                                 
    Name [Default]          Allowed values          Description                           
    ====================    ====================    ====================================  
                                                                                          
    all_targets [off]       on, off                 Build all supported targets, default  
                                                    targets <current                      
                                                    arch>,NVPTX,AMDGPU,CppBackend         
    build_type [Release]    Debug, Release,         CMake build type                      
                            RelWithDebInfo,                                               
                            MinSizeRel
    clang [on]              on, off                 Build the LLVM C/C++/Objective-C     
                                                    compiler frontend                    
    code_signing [off]      on, off                 Enable code-signing on macOS         
    compiler-rt [on]        on, off                 Build LLVM compiler runtime,         
                                                    including sanitizers                 
    cuda [off]              on, off                 Build with CUDA                      
    cuda_arch [none]        none, 12, 80, 30,       CUDA architecture                    
                            86, 10, 11, 13, 60,                                          
                            62, 52, 35, 32, 75,                                          
                            72, 61, 37, 70, 21,                                          
                            20, 50, 53
    gold [on]               on, off                 Add support for LTO with the gold    
                                                    linker plugin                        
    internal_unwind [on]    on, off                 Build the libcxxabi libunwind        
    ipo [off]               on, off                 CMake interprocedural optimization   
    libcxx [on]             on, off                 Build the LLVM C++ standard library  
    lld [on]                on, off                 Build the LLVM linker                
    lldb [on]               on, off                 Build the LLVM debugger              
    mlir [off]              on, off                 Build with MLIR support              
    omp_debug [off]         on, off                 Include debugging code in OpenMP     
                                                    runtime libraries                    
    omp_tsan [off]          on, off                 Build with OpenMP capable thread     
                                                    sanitizer                            
    polly [on]              on, off                 Build the LLVM polyhedral            
                                                    optimization plugin, only builds for 
                                                    3.7.0+                               
    python [off]            on, off                 Install python bindings              
    shared_libs [off]       on, off                 Build all components as shared       
                                                    libraries, faster, less memory to    
                                                    build, less stable                   
    split_dwarf [off]       on, off                 Build with split dwarf information   

```
When excluding things that we definitely don't need we get: 6.5GB  (5.7GB with grabage collect)
`spack install llvm@11.0.0 cuda=False libcxx=False polly=False lldb=False lld=False internal_unwind=False gold=False target=x86_64`

with MinSizeRel: we get: 5.2GB (5.0GB with garbage collect)
` spack install llvm@11.0.0 cuda=False libcxx=False polly=False lldb=False lld=False internal_unwind=False gold=False target=x86_64 build_type=MinSizeRel `

Here is the directory structure of the [old base](https://github.com/illuhad/hipSYCL/files/5883812/llvm-9-script.txt) package (llvm-9) and the [new one](https://github.com/illuhad/hipSYCL/files/5883813/llvm-11-spack-minsized.txt) (llvm-11.0.0).

I find it very strange that the compressed packages are so much smaller than the actual install size...

> Okay. We might have to keep in mind that we might get Level Zero/SPIR-V/Intel support in the future, so I'm a bit worried that this might lead to combinatoric explosion. Is there any meaningful way to combine support for some targets? Could it make sense to only have CUDA as optional dependency since we cannot redistribute it for legal reasons?

Sorry, I don't think I understand what you mean by optional dependencies. Do you recommend that we have a hipSYCL-omp-rocm-etc... package and one where additionally we have Cuda? I should measure the build time for the hipSYCL package, but I have the feeling that it does not have a huge impact on the performance of the packaging pipeline, especially if we run them in parallel. However, it might not make too much sense to have every combination. It might make sense to exclude certain less probable combinations.

> EDIT: Wouldn't we also need hipSYCL-omp-* with nightly suffix?

Sorry I forgot to add that. Yes, I imagine all listed package to have a nightly equivalent.

sbalint98:
It turns out that, unsurprisingly disabling flang reduces the resulting build with around ~800MB.
However, spack still builds `clang-toosl-extra` which I think we don't need. Hower the spack package does not provide a way to disable this since it is automatically added if we build clang:
```
        if "+clang" in spec:
            projects.append("clang")
            projects.append("clang-tools-extra")
            projects.append("openmp")
```
[here](https://github.com/spack/spack/blob/de99535f9afce5f516374ea9a8d706384d03324b/var/spack/repos/builtin/packages/llvm/package.py#L472)

I wonder if we should open a PR or an issue in order to make this optional, I am not sure how likely it is to get something like this merged. 

If I comment this line out the llvm build is around 2.8GB(+700MB dependencies built by spack) if the `clang-tools-extra` is also built we get around 3.8GB(+700MB).

I copy the cmake line here, do you see anything else we could get rid of?

```
==> [2021-02-07-16:17:56.252818] 'cmake' '-G' 'Unix Makefiles' '-DCMAKE_INSTALL_PREFIX:STRING=/root/spack/opt/spack/linux-arch-x86_64/gcc-10.2.0/llvm-11.0.0-o3q3yuf2h7q3macm3x5noqgxxahugihd' '-DCMAKE_BUILD_TYPE:STRING=MinSizeRel' '-DCMAKE_INTERPROCEDURAL_OPTIMIZATION:BOOL=OFF' '-DCMAKE_VERBOSE_MAKEFILE:BOOL=ON' '-DCMAKE_INSTALL_RPATH_USE_LINK_PATH:BOOL=OFF' '-DCMAKE_INSTALL_RPATH:STRING=/root/spack/opt/spack/linux-arch-x86_64/gcc-10.2.0/llvm-11.0.0-o3q3yuf2h7q3macm3x5noqgxxahugihd/lib;/root/spack/opt/spack/linux-arch-x86_64/gcc-10.2.0/llvm-11.0.0-o3q3yuf2h7q3macm3x5noqgxxahugihd/lib64;/root/spack/opt/spack/linux-arch-x86_64/gcc-10.2.0/hwloc-2.4.0-wk2rx6i6abtiwnhwtd6ki7oifzdgtnbe/lib;/root/spack/opt/spack/linux-arch-x86_64/gcc-10.2.0/libpciaccess-0.16-shslczgtfa2aqe2tp37jt2fsgu3wf5ek/lib;/root/spack/opt/spack/linux-arch-x86_64/gcc-10.2.0/libxml2-2.9.10-a6jpv3uxubjh2gsrspl7rcd3mxibecg5/lib;/root/spack/opt/spack/linux-arch-x86_64/gcc-10.2.0/libiconv-1.16-kllnz4udjhrtzhmedn4som3rfkwzpjcj/lib;/root/spack/opt/spack/linux-arch-x86_64/gcc-10.2.0/xz-5.2.5-mob3iuqccwpnfzzznji375kbtxkv656l/lib;/root/spack/opt/spack/linux-arch-x86_64/gcc-10.2.0/zlib-1.2.11-mpalo66lmzmgyc5h4juo6o4eyhr4venz/lib;/root/spack/opt/spack/linux-arch-x86_64/gcc-10.2.0/ncurses-6.2-owvwlo2wnntq6672ne3ct7bqmfcyjwik/lib;/root/spack/opt/spack/linux-arch-x86_64/gcc-10.2.0/z3-4.8.9-ushfxyqzwygzvq6q4zsnczth3kxsohwl/lib' '-DCMAKE_PREFIX_PATH:STRING=/root/spack/opt/spack/linux-arch-x86_64/gcc-10.2.0/cmake-3.19.2-kj3i7aam7tnqkm2nof6awxobw3yeb44i;/root/spack/opt/spack/linux-arch-x86_64/gcc-10.2.0/python-3.8.7-uoorpfubul232wtay6ds2pnjj7witicz;/root/spack/opt/spack/linux-arch-x86_64/gcc-10.2.0/pkgconf-1.7.3-vmakq3qayhhc6csswvxujrv7fopwmjzs;/root/spack/opt/spack/linux-arch-x86_64/gcc-10.2.0/z3-4.8.9-ushfxyqzwygzvq6q4zsnczth3kxsohwl;/root/spack/opt/spack/linux-arch-x86_64/gcc-10.2.0/perl-data-dumper-2.173-rj3tosd4cxewfoer4ntab6s6csn6mt5m;/root/spack/opt/spack/linux-arch-x86_64/gcc-10.2.0/hwloc-2.4.0-wk2rx6i6abtiwnhwtd6ki7oifzdgtnbe;/root/spack/opt/spack/linux-arch-x86_64/gcc-10.2.0/ncurses-6.2-owvwlo2wnntq6672ne3ct7bqmfcyjwik' '-DLLVM_REQUIRES_RTTI:BOOL=ON' '-DLLVM_ENABLE_RTTI:BOOL=ON' '-DLLVM_ENABLE_EH:BOOL=ON' '-DCLANG_DEFAULT_OPENMP_RUNTIME:STRING=libomp' '-DPYTHON_EXECUTABLE:PATH=/root/spack/opt/spack/linux-arch-x86_64/gcc-10.2.0/python-3.8.7-uoorpfubul232wtay6ds2pnjj7witicz/bin/python3.8' '-DLIBOMP_USE_HWLOC:BOOL=ON' '-DLIBOMP_HWLOC_INSTALL_DIR=/root/spack/opt/spack/linux-arch-x86_64/gcc-10.2.0/hwloc-2.4.0-wk2rx6i6abtiwnhwtd6ki7oifzdgtnbe' '-DPython3_EXECUTABLE=/root/spack/opt/spack/linux-arch-x86_64/gcc-10.2.0/python-3.8.7-uoorpfubul232wtay6ds2pnjj7witicz/bin/python3.8' '-DCUDA_TOOLKIT_ROOT_DIR:PATH=IGNORE' '-DCUDA_SDK_ROOT_DIR:PATH=IGNORE' '-DCUDA_NVCC_EXECUTABLE:FILEPATH=IGNORE' '-DLIBOMPTARGET_DEP_CUDA_DRIVER_LIBRARIES:STRING=IGNORE' '-DLLVM_TARGETS_TO_BUILD:STRING=NVPTX;AMDGPU;X86' '-DGCC_INSTALL_PREFIX=/usr' '-DCMAKE_BUILD_WITH_INSTALL_RPATH=1' '-DLLVM_ENABLE_PROJECTS:STRING=clang;clang-tools-extra;openmp;compiler-rt' '/tmp/root/spack-stage/spack-stage-llvm-11.0.0-o3q3yuf2h7q3macm3x5noqgxxahugihd/spack-src/llvm'

```

illuhad (approved):

PR #404, opened 2020-12-09
Fix linker flags order for GNU

illuhad:
GNU ld/gcc needs to have the linked libraries after the source files for which they are needed. This was not taken into account, so using gcc as host compiler was broken. This PR fixes this.

Additionally fixes some minor typos in the cmake message output.


illuhad:
@hipsycl-ci-bot test on GPU please

PR #406, opened 2020-12-09
fixed overflow in group-function tests

DieGoldeneEnte:
The changes in `group_functions.hpp` are mostly quality of life changes.
The diffs for the reduce/scan tests go crazy, but I only extracted the tests using multiplication and put them into their own test-case and replacing the data generator with one, that is 1 almost everywhere. Also since the multiplication-tests doesn't overflow anymore I don't need to disable float anymore :tada:


illuhad:
@hipSYCL-ci-bot test please

illuhad:
Thank you!

PR #407, opened 2020-12-10
fix memory leak in cpu-group_functions

DieGoldeneEnte:
Replaced a void pointer with the correct Type and make the function that frees the memory aware of the correct size of the memory region.

Before the memory was only freed (if at all because of void*-pointer) if the datatype was larger than `void*`.

### DIFF
--- a/include/hipSYCL/sycl/libkernel/host/group_functions.hpp
+++ b/include/hipSYCL/sycl/libkernel/host/group_functions.hpp
@@ -59,12 +59,12 @@ T *get_local_memory_ptr(Group g, size_t number_elements = 1) {
 }
 
 template<typename T, typename Group>
-void free_local_memory_ptr(Group g, T *local_memory_ptr) {
-  if (sizeof(T) > sizeof(void *)) {
+void free_local_memory_ptr(Group g, T *local_memory_ptr, size_t number_elements = 1) {
+  if (sizeof(T) * number_elements > sizeof(void *)) {
     group_barrier(g);
 
     if (g.leader())
-      delete[] static_cast<void *>(local_memory_ptr);
+      delete[] local_memory_ptr;
   }
 }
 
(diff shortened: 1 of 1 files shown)


include/hipSYCL/sycl/libkernel/host/group_functions.hpp
```diff
@@ -59,12 +59,12 @@ T *get_local_memory_ptr(Group g, size_t number_elements = 1) {
 }
 
 template<typename T, typename Group>
-void free_local_memory_ptr(Group g, T *local_memory_ptr) {
-  if (sizeof(T) > sizeof(void *)) {
+void free_local_memory_ptr(Group g, T *local_memory_ptr, size_t number_elements = 1) {
+  if (sizeof(T) * number_elements > sizeof(void *)) {
```
illuhad, line 63: The way this works still seems like a hack to me ;) But okay, I guess everything will get simpler once we just use the stack for that...

DieGoldeneEnte: Yes using the stack is the next point on my TODO-list.
But the fix is not a hack, it is just the same condition I use when allocating the memory :)

illuhad: My interpretation is that this check is needed because for small required amounts of memory you use the memory of the void pointer itself. That *is* a hack ;)

DieGoldeneEnte: Can't really argue against that statement, but it is faster than allocating memory (almost 40% in the one benchmark run I did).

fodinabor: It's not a _hack_ - it's a vaild optimization!
You want stack memory, then don't complain, when it's used :P

illuhad: It's a hacky optimization that's only necessary because we don't use stack memory for all local allocations yet ;)

PR #409, opened 2020-12-10
Specify CMake package version (set to 0.9.0)

psalz:
This adds versioning to the CMake package, so one can do e.g. `find_package(hipSYCL 0.9.0 REQUIRED)`. Would be cool if we could get this into the 0.9.0 release! :-)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,6 @@
 cmake_minimum_required (VERSION 3.5)
-project(hipSYCL)
+set(hipSYCL_VERSION 0.9.0)
+project(hipSYCL VERSION ${hipSYCL_VERSION})
 
 set(HIPSYCL_DEVICE_COMPILER ${PROJECT_SOURCE_DIR}/bin/syclcc-clang)
 set(HIPSYCL_SOURCE_DIR ${PROJECT_SOURCE_DIR})
@@ -214,6 +215,12 @@ install(FILES ${SYCLCC_CONFIG_FILE_PATH} DESTINATION ${SYCLCC_CONFIG_FILE_INSTAL
 include(CMakePackageConfigHelpers)
 include(GNUInstallDirs)
 
+write_basic_package_version_file(
+  "${PROJECT_BINARY_DIR}/hipsycl-config-version.cmake"
+  VERSION ${hipSYCL_VERSION}
+  COMPATIBILITY SameMajorVersion
+)
+
 # Set relative paths for install root in the following variables so that
 # configure_package_config_file will generate paths relative whatever is
 # the future install root
@@ -231,6 +238,7 @@ configure_package_config_file(
 )
 install(FILES
   ${PROJECT_BINARY_DIR}/hipsycl-config.cmake
+  ${PROJECT_BINARY_DIR}/hipsycl-config-version.cmake
   DESTINATION ${HIPSYCL_INSTALL_CMAKE_DIR}
 )
 install(EXPORT install_exports


illuhad:
Ah, good catch, thank you! Didn't think about cmake. 

We should start thinking about unifying the version description from `version.hpp` with the one from cmake. And think about the process after the release - should the cmake package version just stick to 0.9.0 until the next official release? Or should the cmake package version be more of an identification of "major flavors" of hipSYCL, e.g. "0.9.x" so that we don't have to update it so frequently?

psalz:
> We should start thinking about unifying the version description from `version.hpp` with the one from cmake

Yes. One way of doing that would be to have the `version.hpp` file be generated by CMake automatically. That would make the CMake file the single source of truth for the version (and it could even automatically add git hashes, and worktree-dirty flags).

> And think about the process after the release - should the cmake package version just stick to 0.9.0 until the next official release? Or should the cmake package version be more of an identification of "major flavors" of hipSYCL, e.g. "0.9.x" so that we don't have to update it so frequently?

I don't know what your plans regarding the frequency of releases are in the future, but I always prefer libraries that are quick to update versions and adhere to Semantic Versioning over slow release cycles that put too much significance into the numbers themselves (such as the elusive "1.0" release). I thus also think the CMake version should be kept up-to-date (which, with the approach mentioned above, would be easy). After all, there is great self-documenting value in being able to just say that a project depends on version 0.10.2 of hipSYCL because of some bug fix, instead of having to note that somewhere in a README.

In this PR I set the compatibility mode of versions to `SameMajorVersion`, which means that a `find_package(hipSYCL 0.9.1)` would match e.g. `0.9.2` or `0.10.0`, but not `0.9.0` or `1.0.0`. This roughly reflects Semantic Versioning (although not fully, as SemVer allows breaking changes in minor releases before 1.0.0).

fodinabor:
I'd also love to see symbol versioning, i.e. having an `inline namespace v0.9.0` that could also be configured from the CMake version.
This would prevent symbol clashes in the case an application is using one or more libraries using hipSYCL which were built against other versions of it (which might be API incompatible, therefore a simple recompilation would not fix it..).

I know that is anyway not an ideal case, as the runtime and DAG would be replicated.. but if just using hipSYCL internally in a library it should still _work_ with a different version..

illuhad:
> Yes. One way of doing that would be to have the version.hpp file be generated by CMake automatically. That would make the CMake file the single source of truth for the version (and it could even automatically add git hashes, and worktree-dirty flags).

Agreed.

> I don't know what your plans regarding the frequency of releases are in the future, but I always prefer libraries that are quick to update versions and adhere to Semantic Versioning over slow release cycles that put too much significance into the numbers themselves (such as the elusive "1.0" release). I thus also think the CMake version should be kept up-to-date (which, with the approach mentioned above, would be easy). After all, there is great self-documenting value in being able to just say that a project depends on version 0.10.2 of hipSYCL because of some bug fix, instead of having to note that somewhere in a README.

Yes, I think I tend to agree. As an alternative I wonder if it could make sense to just switch to something like `year.month` for the releases and then just release every month.

> This would prevent symbol clashes in the case an application is using one or more libraries using hipSYCL which were built against other versions of it (which might be API incompatible, therefore a simple recompilation would not fix it..).
> I know that is anyway not an ideal case, as the runtime and DAG would be replicated.. but if just using hipSYCL internally in a library it should still work with a different version..

I don't think we can or should support that. There would be massive resource contention on GPUs and CPUs and completely break any scheduling we can do if we have two hipSYCL versions active simultaneously. I think things *should* break in that case because the alternative is silently introducing potentially massive performance issues.

psalz:
> As an alternative I wonder if it could make sense to just switch to something like `year.month` for the releases and then just release every month.

Well, I guess it depends on what guarantees you want to make. With SemVer, the package version implicitly tells me whether it should be safe to upgrade (minor or patch releases), allowing me to potentially benefit from performance or functional bug fixes without having to put any additional work into it. With a `year.month` scheme on the other hand I'll always have to read the changelog. Probably not a huge deal, but still.

illuhad:
> With SemVer, the package version implicitly tells me whether it should be safe to upgrade (minor or patch releases)

Given that we are following a specification for the API, unless you are using hipSYCL-specific extensions, is this information not already contained in the SYCL version we are aiming at? (and we already have branches for `sycl/2020` and `sycl/1.2.1` if you want to target specific SYCL versions).
The advantage I see with `year.month` is that it would probably be straight-forward to automate.

psalz:
> Given that we are following a specification for the API, unless you are using hipSYCL-specific extensions, is this information not already contained in the SYCL version we are aiming at?

Only if you assume that hipSYCL always implements (a subset of) the spec in a fully compliant way, and the spec is free of ambiguities (which has obviously not been the case thus far, for any SYCL implementation). Any clarification in a spec revision could thus also lead to breaking changes (warranting a major release in SemVer), depending on how the spec was interpreted previously (as a recent example the H2D explicit copy operations come to mind).

illuhad:
> Only if you assume that hipSYCL always implements (a subset of) the spec in a fully compliant way, and the spec is free of ambiguities (which has obviously not been the case thus far, for any SYCL implementation). Any clarification in a spec revision could thus also lead to breaking changes (warranting a major release in SemVer), depending on how the spec was interpreted previously (as a recent example the H2D explicit copy operations come to mind).

hm.. I'm not sure we can track clarifications of the spec on such a fine-grained level. I don't really want to have to think about whether every little change that is done to align better with the spec could break some minor details of (incorrect) SYCL code. For example the `vec` reimplementation - does it behave differently than the old implementation in some unusual edge cases? Who knows, it's a lot of code that was replaced and the old one certainly had its quirks.. I'm not sure we can realistically make such guarantees.

Do you mean copy operations with host accessors? I'm not sure if this *ever* really worked. I think even with the old runtime it might have just been UB with us not realizing it.

PR #415, opened 2020-12-14
Reimplement vec<> class

illuhad:
Reimplement and simplify vec class to not rely on CUDA/HIP vector types which currently don't yield any performance benefits with clang anyway (note that, if there are performance benefits in the future we could still always specialize the `vec_storage` backend to use them for cases where it makes sense).

Key stats of this change:

* -3000 lines of code (!)
* +50% readability
* +100% maintainability
* +20% C++17 (old implementation was based on C++14 and hence did not use fold expressions)
* +/- 0% performance (expected)
* +10% spec conformance, e.g. alignments.
* +5% SYCL 2020 (implements SYCL 2020 additions to the vec interface)


illuhad:
@hipSYCL-ci-bot test this please, my favorite bot!

PR #416, opened 2020-12-15
Set default log level to print warnings

illuhad:
Set default log level such that warnings are printed by default. hipSYCL emits warnings when it thinks there's something going wrong, or it thinks that there might be a performance issue. It makes sense to bring this kind of information to the user's attention by default.


illuhad:
@hipSYCL-ci-bot test

PR #418, opened 2020-12-16
Add command group properties, hipSYCL_retarget and hipSYCL_prefer_group_size

illuhad:
This adds a new extension: Properties for individual command groups, i.e. allows passing a `property_list` to `queue::submit()` calls.

`submit()` accepts the two new properties as SYCL extension:
* `sycl::property::command_group::hipSYCL_prefer_group_size` to give a hint to hipSYCL which group size should be selected for basic parallel for kernels (CC @jeffhammond)
* `sycl::property::command_group::hipSYCL_retarget` to execute the command group on an arbitrary different device than the one the queue is bound to (without introducing additional overhead)

Detailed documentation (copied from the documentation added with this PR):

### Command group properties

hipSYCL supports attaching special command group properties to individual command groups. This is done by passing a property list to the queue's `submit` member function:

```cpp
template <typename T>
event queue::submit(const property_list& prop_list, T cgf)
```

A list of supported command group properties follows:

##### API reference

```cpp
namespace sycl::property::command_group {

template<int Dim>
struct hipSYCL_prefer_group_size {
  hipSYCL_prefer_group_size(range<Dim> group_size);
};

}
```

##### Description

If this property is added to a command group property list, it instructs the backend to prefer a particular work group size for kernels for models where a SYCL implementation has the freedom to decide on a work group size.

In the current implementation, this property only affects the selected local size for basic parallel for on HIP and CUDA backends.

*Note:* The property only affects kernel launches of the same dimension. If you want to set the group size for 2D kernels, you need to attach a `hipSYCL_prefer_group_size<2>` property.

##### API reference

```cpp
namespace sycl::property::command_group {

template<int Dim>
struct hipSYCL_retarget {
  hipSYCL_retarget(const device& dev);
};

}
```

##### Description

If this property is added to a command group property list, it instructs the hipSYCL runtime to execute the submitted operation on a different device than the one the queue was bound to. This can be useful as a more convenient mechanism to dispatch to multiple devices compared to creating multiple queues.

Using this property does *not* introduce additional overheads compared to using multiple queues. In particular, it does *not* silently lead to the creation of additional backend execution resources such as CUDA streams.

In order to understand this, it is important to realize that because of the design of the hipSYCL runtime, a queue is decoupled from backend objects. Instead, the hipSYCL runtime internally manages a pool of backend execution resources such as CUDA streams, and automatically distributes work across those resources.
In this design, a queue is nothing more than an interface to hipSYCL runtime functionality. This allows us to efficiently retarget operations submitted to a queue arbitrarily.

Compared to using multiple queues bound to different devices, using a single queue and submitting with the `hipSYCL_retarget` property comes with some minor semantic differences:

* A single `queue::wait()` call guarantees that all operations submitted to the queue, no matter to which device they were retargeted, have completed. With multiple queues on the other hand, multiple `wait()` calls are necessary which can add some overhead.
* If the queue is an in-order queue, the in-order property is *preserved even if the operations are retargeted to run on different devices*. This can be a highly convenient way to formulate in-order USM algorithms that require processing steps on different devices.


illuhad:
@hipSYCL-ci-bot test this for me again

PR #420, opened 2020-12-17
work around exp10f on MacOS

jeffhammond:
This resolves the final remaining issue in math.hpp that was identified in https://github.com/illuhad/hipSYCL/issues/228.

Feel free to implement a different way, but `exp10` is missing on MacOS.

Signed-off-by: Hammond, Jeff R <jeff.r.hammond@intel.com>


illuhad:
Thanks, that looks fine to me. Will merge once CI finishes.

Issue #422, opened 2020-12-17
How do I specify boost location?


VictorEijkhout:
It looks like cmake is picking up the wrong boost installation on my system. 
What "-D" variable do I set to specify the boost include path?

illuhad:
I think it's `-DBOOST_PATH`, but haven't actually used it so far. @sbalint98 might know more.

fodinabor:
Usually it is `-DBOOST_ROOT=<path>` and you might also have to add `-DBoost_NO_SYSTEM_PATHS=On`
also see: [docs](https://cmake.org/cmake/help/latest/module/FindBoost.html)

illuhad:
I think for hipSYCL specifically `BOOST_PATH` might work as well because of this:
https://github.com/illuhad/hipSYCL/blob/db122df23b5bb5616e9c516e7bee5ddae1ac3af3/CMakeLists.txt#L24

VictorEijkhout:
That worked. Thanks much.

V.

jeffhammond:
I am so tired of debugging CMake.  What is the right input when I download a release tarball of Boost and unpack it in `$HOME`?

# Attempt 1
```
[jhammond@corona82:build]$ rm -rf /g/g12/jhammond/hipSYCL-0.9.0/build/* ; cmake .. -DBoost_NO_SYSTEM_PATHS=On -DBOOST_ROOT=$HOME/boost_1_75_0/libs
-- The C compiler identification is GNU 4.9.3
-- The CXX compiler identification is GNU 4.9.3
-- Check for working C compiler: /usr/tce/bin/cc
-- Check for working C compiler: /usr/tce/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/tce/bin/c++
-- Check for working CXX compiler: /usr/tce/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Error at /usr/tce/packages/cmake/cmake-3.14.5/share/cmake/Modules/FindBoost.cmake:2165 (message):
  Unable to find the requested Boost libraries.

  Unable to find the Boost header files.  Please set BOOST_ROOT to the root
  directory containing Boost or BOOST_INCLUDEDIR to the directory containing
[...]

PR #423, opened 2020-12-17
Rquire BOOST_PATH if BOOST is not in system dir

sbalint98:
Make sure that BOOST_ROOT is never overwritten if it is set. This has probably contributed to the issue #422. Alternatively, we could remove BOOST_PATH from CMake entirely.


illuhad:
hm.. I suspect it might be easier to just use `BOOST_ROOT` rather than introducing a complex logic just so we can call the argument `BOOST_PATH`. Is there a specific reason why we have `BOOST_PATH`?

sbalint98:
`BOOST_ROOT` is an environment variable while `BOOST_PATH` is a CMake variable. The consideration was to have all hipSYCL related options as CMake variables since this is what the script was using. Furthermore `BOOST_PATH` is used multiple times for setting the cxx and link flags. Therefore `BOOST_PATH` currently must be defined. (Therefore this PR is also incorrect in its current form) I don't think this behaviour is preferable I think rather variables set by CMake should be used. 

However, when we are using recent boost I could not find a straightforward way to retrieve the `INCLUDE_DIR` and `LIB_DIR` variables that we need for setting the cxx and link line defaults in sylcc.json . In the documentation of [findBoost](https://cmake.org/cmake/help/latest/module/FindBoost.html): 

> If Boost was built using the boost-cmake project or from Boost 1.70.0 on it provides a package configuration file for use with find_package’s config mode. This module looks for the package configuration file called BoostConfig.cmake or boost-config.cmake and stores the result in CACHE entry “Boost_DIR”. If found, the package configuration file is loaded and this module returns with no further action. See the documentation of the Boost CMake package configuration for details on what it provides.

And in the BoostConfig.cmake the use of find package is described like this:

> A typical use might be
>
> find_package(Boost 1.70 REQUIRED COMPONENTS filesystem regex PATHS C:/Boost)
>
> On success, the above invocation would define the targets Boost::headers,
> Boost::filesystem and Boost::regex. Boost::headers represents all
> header-only libraries. An alias, Boost::boost, for Boost::headers is
> provided for compatibility.

This means that we don't have the `Boost_INCLUDE_DIR` and `Boost_LIBRARY_DIR_RELEASE` variables defined which would be very useful for us when setting the cxx and link defaults. There might be an option to retrieve these from the imported targets using `get_properties` but this solution also seemed quite hacky to me, and also might have the same problem as the [rocm cmake integration.](https://github.com/illuhad/hipSYCL/pull/403#discussion_r539590212)

There is an option to force cmake to use the old way of finding boost:

> Set Boost_NO_BOOST_CMAKE to ON, to disable the search for boost-cmake.

But it failed to properly find the very recent boost installed by spack, and forcing cmake to work in legacy mode does not seem to be the right direction to me.

Therefore I think the best solution might be to require the `-DBOOST_PATH` variable when boost is not installed into a default location.

fodinabor:
You indeed can specify `-DBOOST_ROOT=<path>` (just did that several times the last few days :)), which acts just as the `BOOST_ROOT` environment variable would be used.

With that, I personally would prefer just using `BOOST_ROOT` all over the place, as this then is in line with the default behavior of FindBoost. You could of course also `set(BOOST_PATH ${BOOST_ROOT})` for further internal usage... but not sure there's any benefit to this.

sbalint98:
Oh well, okay. Thanks for letting me know! I agree then `BOOST-ROOT` makes way more sense.

sbalint98:
@fodinabor It seems like that for cmake (version 3.18.2 at least ) it is not equivalent. If BOOST_ROOT is specified as a env variable cmake uses the cmake config of boost, if it is specified as a cmake variable it falls back to the module mode :D 
 
This is the code in FindBoost.cmake that makes the difference in my opinion: 
```
#-------------------------------------------------------------------------------
# Before we go searching, check whether a boost cmake package is available, unless
# the user specifically asked NOT to search for one.
#
# If Boost_DIR is set, this behaves as any find_package call would. If not,
# it looks at BOOST_ROOT and BOOSTROOT to find Boost.
#
if (NOT Boost_NO_BOOST_CMAKE)
  # If Boost_DIR is not set, look for BOOSTROOT and BOOST_ROOT as alternatives,
  # since these are more conventional for Boost.
  if ("$ENV{Boost_DIR}" STREQUAL "")
    if (NOT "$ENV{BOOST_ROOT}" STREQUAL "")
      set(ENV{Boost_DIR} $ENV{BOOST_ROOT})
    elseif (NOT "$ENV{BOOSTROOT}" STREQUAL "")
      set(ENV{Boost_DIR} $ENV{BOOSTROOT})
    endif()
  endif()

  # Do the same find_package call but look specifically for the CMake version.
  # Note that args are passed in the Boost_FIND_xxxxx variables, so there is no
  # need to delegate them to this find_package call.
  find_package(Boost QUIET NO_MODULE)
  if (DEFINED Boost_DIR)
    mark_as_advanced(Boost_DIR)
  endif ()

```
Here it only copies to `Boost_DIR` if `BOOST_ROOT` is an environmental variable. Cna you confirm this or am I missing something?

This is important because if `find_package(boost)` runs in module mode for the boost built via spack I get the following warning:

```
CMake Warning at /usr/share/cmake-3.18/Modules/FindBoost.cmake:1187 (message):
  New Boost version may have incorrect or missing dependencies and imported
  targets
  /usr/share/cmake-3.18/Modules/FindBoost.cmake:1311 (_Boost_COMPONENT_DEPENDENCIES)
  /usr/share/cmake-3.18/Modules/FindBoost.cmake:1919 (_Boost_MISSING_DEPENDENCIES)
  CMakeLists.txt:24 (find_package)


CMake Warning at /usr/share/cmake-3.18/Modules/FindBoost.cmake:1187 (message):
  New Boost version may have incorrect or missing dependencies and imported
  targets
  /usr/share/cmake-3.18/Modules/FindBoost.cmake:1311 (_Boost_COMPONENT_DEPENDENCIES)
  /usr/share/cmake-3.18/Modules/FindBoost.cmake:1919 (_Boost_MISSING_DEPENDENCIES)
  CMakeLists.txt:24 (find_package)


-- Found Boost: /opt/hipSYCL/boost/boost/include (found version "1.74.0") found components: context fiber
```
EDIT:
I think just adding `set(ENV{BOOST_ROOT} ${BOOST_ROOT})`
 before find_package(Boost COMPONENTS context fiber ) should probably be the best solution...

fodinabor:
I guess you should only do that if `-DBOOST_ROOT` is set, as otherwise, you overwrite the environment variable that some people might have set with an empty string.

That really is awful there... much further down in the file, in the manual searching part, they use the actual variables.. so I guess for consistent handling, it could actually be the best way, how you just did it, except that the environment variable should not be overwritten in case `-DBOOST_ROOT` is empty.

sbalint98:
Good point thanks!

illuhad:
Thanks!

illuhad:
@facuMH context is not header only, in fact, the page you linked says it ;) 

> Most Boost libraries are header-only: they consist entirely of header files containing templates and inline functions, and require  no separately-compiled library binaries or special treatment when linking.

> The only Boost libraries that *must* be built separately are:
>    Boost.Chrono
>    Boost.Context
>    ...

PR #429, opened 2020-12-26
update for new bitcode path in llvm-12

louiszhuang:


illuhad:
Thank you. We are aware of this issue. It's more of a ROCm issue than a clang issue since AMD decided to change their directory layout with ROCm 3.9. So, ROCm 3.8 should work fine with clang 12. Note that newer clang/ROCm combinations may require additional arguments such as `--rocm-path`.

Unfortunately we cannot merge this at the moment because it would break earlier ROCm installations by default, and we are not yet prepared to do this - at least until our packages are updated to also ship ROCm 3.9+, in which case we could drop support for older ROCm versions.
As you can see, it breaks CI (clang 10/11 with ROCm 3.8).

The correct way to target newer ROCm versions at the moment is to use `cmake -DROCM_CXX_FLAGS=... ` to customize the arguments passed to clang. We should however probably update the documentation to describe the current situation better.

Note that we do not routinely test with clang 12 in CI anyway because the stability of pre-release clang versions is not always good enough to run hipSYCL.

louiszhuang:
hipSYCL implements SYCL without recompiling llvm and/or ROCm. However ROCm 4 is the only offical package from vendor repository with own branch of llvm-12. It would be great to update document to warn users for the change of location so that they can save a day.

illuhad:
> hipSYCL implements SYCL without recompiling llvm and/or ROCm. However ROCm 4 is the only offical package from vendor repository with own branch of llvm-12

This is incorrect, AMD provides packages for older ROCm versions as well. See e.g. http://repo.radeon.com/rocm/apt/3.8/ for ROCm 3.8. As noted in the hipSYCL readme, it is generally neither necessary nor recommended to use AMD's LLVM distribution.

But yes, we should probably update the Readme to clarify.

PR #430, opened 2020-12-29
syclcc export correct include paths in 'syclcc -v -E -x c /dev/null'

louiszhuang:
vscode cpptool uses 'syclcc -v -E -x c /dev/null' to fetch include directories. The change fixes the issue in wrapper.

### DIFF
--- a/bin/syclcc-clang
+++ b/bin/syclcc-clang
@@ -706,13 +706,12 @@ class compiler:
      self._backends.append(omp_sequential_invocation(config))
 
   @property
-  def common_cxx_flags(self):
-    args = [
-      "-I" + os.path.join(self._hipsycl_path, "include/"),
-      "-D__HIPSYCL__"
-    ] + self._common_compiler_args
+  def common_cxx_include_path(self):
+    return ["-I" + os.path.join(self._hipsycl_path, "include/")]
 
-    return args
+  @property
+  def common_cxx_flags(self):
+    return self.common_cxx_include_path + ["-D__HIPSYCL__"] + self._common_compiler_args
 
   @property
   def common_linker_flags(self):
@@ -741,6 +740,8 @@ class compiler:
     args = []
     if self._requires_compilation:
       args += cxx_flags
+    else:
+      args += self.common_cxx_include_path
 
     args += self._user_args
 
--- a/bin/syclcc-clang
+++ b/bin/syclcc-clang
@@ -740,7 +740,8 @@ class compiler:
     args = []
     if self._requires_compilation:
       args += cxx_flags
-    else:
+
+    if not self._requires_compilation and not self._requires_linking:
       args += self.common_cxx_include_path
 
     args += self._user_args
--- a/bin/syclcc-clang
+++ b/bin/syclcc-clang
@@ -530,18 +530,26 @@ class syclcc_config:
           return True
     return False
   
-  def contains_linking_stage(self):
+  def is_preprocessing(self):
     for arg in self.forwarded_compiler_arguments:
       if (arg == "-E" or
           arg == "-fsyntax-only" or
-          arg == "-S" or
-          arg == "-c"):
-        return False
-    return True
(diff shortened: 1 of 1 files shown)


illuhad:
Thank you!

I think the root cause of this is that `syclcc` does not treat the invocation as a compilation. The logic to determine whether a `syclcc` invocation is a compilation at the moment is pretty basic: It only checks if there is a source file argument that ends in `.cpp` or other common C++ file endings.

So I wonder, wouldn't it be a much cleaner solution to just change this so that if `-x c` (and `-x c++`) arguments are present it also causes it to treat the invocation like a compiler invocation?

This would also cause it to correctly emit the include paths of backend dependencies (e.g. CUDA or ROCm) which I think your change does not do.

I think this would only require changing `is_pure_linking_stage()` here:
https://github.com/illuhad/hipSYCL/blob/77db25b6d077d3fb1fb2079fd085e629689f46e0/bin/syclcc-clang#L542

PS: If you want vscode integration, it might be a good idea to rely on the hipSYCL cmake integration which should automatically export this kind of stuff.

louiszhuang:
nice suggestion - however "-std=c++17" in comm_cxx_flags is incompatible with "-x c". 
In general, we should treat invocation with 4 possible modes - preprocess (-E, -S etc), compilation (-c), compilation + linkage, or pure linkage. 
'syclcc -v -E -x c /dev/null' is preprocess mode - common include should be passed but no other flags.

the issue exploded by vscode but not vscode specific, so we should solve the root cause in syclcc...

bin/syclcc-clang
```diff
@@ -742,6 +749,9 @@ class compiler:
     if self._requires_compilation:
       args += cxx_flags
 
+    if not self._requires_compilation and not self._requires_linking:
```
illuhad: `elif`? And maybe add comment that this is for pure preprocessing cases?

bin/syclcc-clang
```diff
@@ -530,23 +530,31 @@ class syclcc_config:
...
-  def contains_linking_stage(self):
+  def is_preprocessing(self):
     for arg in self.forwarded_compiler_arguments:
       if (arg == "-E" or
           arg == "-fsyntax-only" or
-          arg == "-S" or
-          arg == "-c"):
-        return False
-    return True
+          arg == "-S"):
```
illuhad: I think `-S` also runs the compiler. Would it make more sense to move it to `is_compiling_only`?

illuhad (commented):
I think this is a good change. But I'm still a bit unsure about the specific use case. You are right that `-x c` clashes with `-std=c++17`, but then SYCL is inherently tied to C++ and I don't think a  SYCL implementation has to (or even should?) support `-x c`. So I wonder if this is also in part a tooling issue inside vscode. Of course there's no reason why it shouldn't work with `-x c++` so we should at least get this to work.

> 'syclcc -v -E -x c /dev/null' is preprocess mode - common include should be passed but no other flags.

Wouldn't we also need include paths for backends in general? And preprocessor definitions that backends might set? E.g. if somebody does `syclcc -v -E --hipsycl-targets="hip:gfx906" -x c /dev/null`? This could be accomplished by splitting the `cxx_flags` obtained from backends into `preprocessor_flags` and `cxx_flags`.

bin/syclcc-clang
louiszhuang: `-S` should be compiler though not sure what's expected behaviour for single source multitargets.

bin/syclcc-clang
louiszhuang: backend flags are interesting topic - we should have them for completeness.

louiszhuang: it is fine to pass ALL flags to clang for preprocessing or compiling. Just not to add `-std=c++17` when `-x c` is provided. I agree hipsycl should be c++ only, but wrapper should not be aggressive on c++ standard when -x c is present... What do you think?

PR #433, opened 2021-01-02
allow groups < 128 and > 256 on AMD-GPUs

DieGoldeneEnte:
This extends the clang-plugin to create the `amdgpu_flat_work_group_size` to allow groups between 64 and 1024 work items on AMD-GPUs.
Previously when using larger/smaller groups could result in wrong results or even crashes.

I chose to set the valid range to [64, 1024]. Maybe the limits could be exposed via environment variables, for now they are hardcoded.

It compiles (on my system), but maybe we want additional includes **?**  I didn't add for example `llvm/ADT/APInt.h`, because it wasn't present in [lib/Sema/SemaDeclCXX.cpp](https://github.com/llvm/llvm-project/blob/main/clang/lib/Sema/SemaDeclCXX.cpp) which I referenced.

Currently I only add the attribute to marked kernels. It might also be needed for other functions or add a check to only apply on AMD-GPU passes **?**

### DIFF
--- a/include/hipSYCL/compiler/Frontend.hpp
+++ b/include/hipSYCL/compiler/Frontend.hpp
@@ -205,7 +205,7 @@ public:
   bool VisitFunctionDecl(clang::FunctionDecl *f) {
     if(!f)
       return true;
-    
+
     this->processFunctionDecl(f);
 
     return true;
@@ -318,8 +318,21 @@ public:
           CustomAttributes::SyclKernel.isAttachedTo(F)) {
 
         auto* NewAttr = clang::CUDAGlobalAttr::CreateImplicit(Instance.getASTContext());
-        
+
         F->addAttr(NewAttr);
+
+        // create amdgpu_flat_work_group_size attribute to allow sub_groups outside of [128, 256]
+        // first we need to create Expressions containing the workgroup-sizes
+        auto sizeType = Instance.getASTContext().getSizeType();
+        auto minFlatWorkgroupSize = llvm::APInt(Instance.getASTContext().getTypeSize(sizeType), 64);
+        auto maxFlatWorkgroupSize = llvm::APInt(Instance.getASTContext().getTypeSize(sizeType), 1024);
+        auto* minFlatWorkgroupExpr = clang::IntegerLiteral::Create(Instance.getASTContext(), minFlatWorkgroupSize, sizeType, clang::SourceLocation{});
+        auto* maxFlatWorkgroupExpr = clang::IntegerLiteral::Create(Instance.getASTContext(), maxFlatWorkgroupSize, sizeType, clang::SourceLocation{});
+
+        // to finally create the attribute itself
+        auto * NewAMDGPUAttr = clang::AMDGPUFlatWorkGroupSizeAttr::CreateImplicit(Instance.getASTContext(), minFlatWorkgroupExpr , maxFlatWorkgroupExpr);
+
+        F->addAttr(NewAMDGPUAttr);
       }
     }
 


DieGoldeneEnte:
@hipSYCL-ci-bot test please to make sure I didn't make hipSYCL AMD-only :)

illuhad:
Awesome!!! 

I think however that we should not enable this unconditionally; even for AMD passes there might be a reason why the default only allows group sizes of up to 256. I suspect that this restriction allows the LLVM backend to perform additional optimizations, so we shouldn't just throw this away by default. When compiling for CUDA I think the attribute will be just ignored, but it might be cleaner to not set it in this case. There are various options to accomplish setting it only conditionally:

* Environment variable: Simple, but has the inherent problem that we cannot distinguish between multiple kernels. E.g. a user might only want to set this attribute only for some kernels where group sizes are `> 256`.
* Custom kernel attribute/constexpr `property_list`. Introduce some SYCL extension that we can expose which allows the user to attach to kernels whether this attribute should be used. Such a mechanism might also be useful for other things.
* Automatically add it when compiling for AMD and using group size > 256. This might be the best option, but it requires some more thought. It would involve multi-versioning kernels when compiling for AMD (compiling one version with this attribute and one without), and then selecting at runtime which kernel to use - this could be done e.g. with an additional template argument for kernel invocation functions.

If you want to get this merged quickly, we could start with the environment variable and then look into the more elaborate 
approaches in the future.

> Currently I only add the attribute to marked kernels. It might also be needed for other functions

I would not expect this to be necessary. AFAIK the HIP use case is to use it with `__global__` functions.

DieGoldeneEnte:
I thought about making it configurable, but this is the easiest way to allow bigger groups which solves my problem. 
Since the user gets undefined results for the same code on AMD-GPUs but not on NVIDIA-GPUs when using bigger groups I think it makes sense to have the larger groups sizes by default and give the user a way to change it.

Maybe it is possible to warn/throw an error if an incorrect configuration is used (maybe in `hiplike_kernel_launcher`)? Until the attribute is set a simple check for flat work_group size `<128 && > 256` would be sufficient and potentially prevent unnecessary confusion.

I don't need to get this merged soon, since I can apply it locally when needed. We can just use this PR as a reminder/hot-fix until we have a better solution for it.

illuhad:
> Since the user gets undefined results for the same code on AMD-GPUs but not on NVIDIA-GPUs when using bigger groups I think it makes sense to have the larger groups sizes by default and give the user a way to change it.

Well... the SYCL spec does not guarantee that a device has to support a particular work group size. We might just as well say that AMD GPUs, at least by default, only support 128 and 256 as work group sizes. This would allow us to run with a performance-optimized configuration by default. In any case, we should raise an error if this is violated instead of silently executing and hoping things work out.

> Maybe it is possible to warn/throw an error if an incorrect configuration is used (maybe in hiplike_kernel_launcher)? Until the attribute is set a simple check for flat work_group size <128 && > 256 would be sufficient and potentially prevent unnecessary confusion.

Yes, exactly :) Preferably error handling is done inside `handler` because we can use synchronous errors there. But we would still need some handling inside the kernel launcher in case the kernel launcher is invoked by something else.

DieGoldeneEnte:
"In the previous ROCm release, this value was 256. In the ROCm v4.1 release, it is changed to 1024."
They have changed the default maximum groupsize in ROCm 4.1 (as far as I can see it is only usable with AMD's llvm, because of currently unmerged changes to atomic floating point instructions).

illuhad:
We should implement this functionality using the interface from #543 :)

PR #434, opened 2021-01-11
HIP,CUDA: Don't emit warning if hip/cudaErrorNoDevice encountered during startup

illuhad:
Currently, if hipSYCL is compiled with HIP/CUDA support but no devices are detected, at startup a warning is emitted.

This removes this warning since not having HIP/CUDA devices is a valid configuration - just because the hipSYCL runtime is built with HIP/CUDA support, this does not necessarily mean that there's something wrong if no such devices are present in the target machine.

### DIFF
--- a/src/runtime/cuda/cuda_hardware_manager.cpp
+++ b/src/runtime/cuda/cuda_hardware_manager.cpp
@@ -44,10 +44,12 @@ cuda_hardware_manager::cuda_hardware_manager(hardware_platform hw_platform)
   if (err != cudaSuccess) {
     num_devices = 0;
 
-    print_warning(
-        __hipsycl_here(),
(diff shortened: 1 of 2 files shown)


fodinabor:
Might make sense to keep it for the info level or so, though?

illuhad:
@fodinabor hm.. not sure. What diagnostic value would we gain by printing information about something that is expected? What might be more interesting is to print the complete topology of all discovered devices during rt startup - this would then include information about devices from other backends than just CUDA/HIP.

fodinabor:
Well.. in case something is misconfigured, it might not be expected ;-)
But sure, taking a more holistic approach by printing the complete topology would be a good added value.

illuhad:
@fodinabor Implemented in #436 :)

fodinabor:
great, then go ahead :)

PR #435, opened 2021-01-11
[SYCL 2020][USM] Add support for host prefetches, add prefetch_host extension

illuhad:
# Support for host prefetches using host queue

This adds support for prefetching shared USM allocations on the host device. For this, the user needs to construct a second queue running on the host device with a context that is shared with the device queue:

```c++
sycl::queue q = ...; //construct device queue
void* shared_mem = sycl::malloc_shared(...);
// prefetch to device
q.prefetch(shared_mem, size);
q.parallel_for(...);
q.wait();
{
    sycl::queue host_queue{q.get_context(), sycl::host_selector{}};
    // prefetch back to host
    host_queue.prefetch(shared_mem, size);
    host_queue.wait();
}
```

To make this use case work, the host device is now always added to device contexts. This should not have any practical implications apart from avoiding error checks that would trigger otherwise because we are operating on a device that is outside the device context.

# prefetch_host extension

Since this is a bit cumbersome (and potentially suboptimal in terms of performance because of the additional `wait()` calls), this PR also adds the `prefetch_host()` extension which allows prefetching back to host directly from the device queue:

```c++
  sycl::queue q{sycl::property_list{sycl::property::queue::in_order{}}};
  int *shared_mem = sycl::malloc_shared(size, q);

  q.parallel_for<class usm_prefetch_host_test_kernel>();
  q.prefetch_host(shared_mem, size);
```


illuhad:
@hipSYCL-ci-bot test please

PR #445, opened 2021-01-26
Make cuda|hip_device_manager thread_local

### LINKED ISSUE #444: Multi-GPU Regression (CUDA Error 400)
**Bug summary**
There appears to be a regression in 9cd4aff67 that causes CUDA Error 400 (`cudaErrorInvalidResourceHandle`) when running hipSYCL on a system with multiple GPUs, whenever the last GPU is selected.

**To Reproduce**
Simply run this on any system with more than one NVIDIA GPU, and make sure to select the last one (this is for 4 GPUs, but I would presume it happens for other setups as well):
```c++
	sycl::queue queue(sycl::platform::get_platforms()[0].get_devices()[3]); // Pick last device
	queue.submit([&](sycl::handler& cgh) { cgh.single_task<class Bug>([=] {}); });
	queue.wait_and_throw();
	return EXIT_SUCCESS;
}
```

Running this with `CUDA_VISIBLE_DEVICES=0` (or any other single device id) fixes the problem.

**Describe your setup**
hipSYCL @ 91258d2
4x NVIDIA RTX 2070 running on CUDA 10.2, driver version 450.80.02.

**Additional context**
I haven't had time to thoroughly investigate the cause for this, but I can say that removing [the check](https://github.com/illuhad/hipSYCL/blob/f2904b3e68f1825feb8bfcda2393b5abd9c86c7e/src/runtime/cuda/cuda_device_manager.cpp#L51) inside `cuda_device_manager::activate_device` whether the given device is already active fixes the issue. (By the way, this also fixes any issues I've had with the unit tests @illuhad). From looking at the debug log of hipSYCL, I think what is happening is that each device is activated once during device discovery, with the last one then remaining the "active" one until any actual device interaction is made. Once the kernel is submitted, the device should be activated again, however since it already is active, the call to `cudaSetDevice` is skipped. I suspect that this check may not be adequate as CUDA tracks streams on a per-thread basis, so tracking a single active device id across all threads wont' suffice in all situations (unless `cuda_device_manager` were to be made `thread_local`).

psalz:
CUDA/HIP keep track of the active device on a per-thread basis. While the `cuda|hip_device_manager`  classes were originally never intended to be used from multiple threads, they in fact are (for example within the `cuda|hip_queue` constructors). This can lead to subtle bugs when running hipSYCL on a machine with multiple CUDA/HIP devices.

To remedy this, the `cuda|hip_device_manager` singletons are now stored as `thread_local`, meaning there is a unique instance for each new thread.

Fixes #444.

---

Side note: I may have gone slightly overboard with the unit tests, but the functionality is actually pretty cool: When running the tests using `./rt_tests --log-level=test_suite` on my machine I now get this:

image

And when running with less than 2 GPUs (i.e. `CUDA_VISIBLE_DEVICES=0 ./rt_tests --log-level=test_suite`) it will say:

image

### DIFF
--- a/include/hipSYCL/runtime/cuda/cuda_device_manager.hpp
+++ b/include/hipSYCL/runtime/cuda/cuda_device_manager.hpp
@@ -1,7 +1,7 @@
 /*
  * This file is part of hipSYCL, a SYCL implementation based on CUDA/HIP
  *
- * Copyright (c) 2019 Aksel Alpay
+ * Copyright (c) 2019-2021 Aksel Alpay and contributors
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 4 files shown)


psalz:
(Jeez that was sloppy, as soon as I view the code in the GitHub UI I see tons of copy-pasta errors 😅)

illuhad:
@hipSYCL-ci-bot test please

illuhad:
Thank you!

PR #447, opened 2021-01-26
Add missing sycl::queue constructors

al42and:
Hello!

Both SYCL-1.2.1 and SYCL 2020 (provisional) require two additional
constructors for the `sycl::queue` class, that directly accept device handle instead of device_selector:

```cpp
explicit queue(const context &syclContext, const device &syclDevice,
    const property_list &propList = {});

explicit queue(const context &syclContext, const device &syclDevice,
    const async_handler &asyncHandler,
    const property_list &propList = {});
```

They are currently missing in hipSYCL but seem rather trivial to add.

### DIFF
--- a/include/hipSYCL/sycl/queue.hpp
+++ b/include/hipSYCL/sycl/queue.hpp
@@ -167,36 +167,45 @@ public:
   explicit queue(const context &syclContext,
                  const device_selector &deviceSelector,
                  const property_list &propList = {})
-      : detail::property_carrying_object{propList}, _ctx{syclContext} {
-
-    _handler = _ctx._impl->handler;
-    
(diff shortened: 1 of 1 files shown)


illuhad:
Thank you! :) This is a great addition.

I wonder, would it maybe make sense to map the existing variants with the device selector on these new constructors? E.g. something like this:

```c++

explicit queue(const context& ctx, const device_selector& sel, ...)
: queue{ctx, sel.select_device(), ...} {}
```

@hipSYCL-ci-bot test again

al42and:
@illuhad 

Thanks, that's a very good point! Amended my code.

illuhad:
Lovely, thank you! :)

PR #448, opened 2021-01-28
Specify alignment when allocating buffer memory

fodinabor:
This is required for types with required alignments, otherwise bad things might happen.
As we do currently not propagate the alignment requirements of types, I just used the element size as alignment, capped at `sizeof(double) * 16` (for `sycl::vec<double, 16>` without having to couple the rt there to libkernel)
`aligned_alloc` has implementation defined requirements regarding the alignment.
As posix & Windows require the alignment to be a power of two, I borrowed LLVMs `PowerOf2Ceil` to make sure this requirement is met.

### DIFF
--- a/src/runtime/omp/omp_allocator.cpp
+++ b/src/runtime/omp/omp_allocator.cpp
@@ -29,6 +29,7 @@
 #include "hipSYCL/runtime/device_id.hpp"
 #include "hipSYCL/runtime/error.hpp"
 #include "hipSYCL/runtime/omp/omp_allocator.hpp"
+#include "hipSYCL/runtime/util.hpp"
 
 namespace hipsycl {
 namespace rt {
@@ -42,6 +43,7 @@ void *omp_allocator::allocate(size_t min_alignment, size_t size_bytes) {
   
   if(size_bytes % min_alignment != 0)
     return nullptr;
+  min_alignment = power_of_2_ceil(min_alignment);
 
   // ToDo: Mac OS CI has a problem with std::aligned_alloc
   // but it's unclear if it's a Mac, or libc++, or toolchain issue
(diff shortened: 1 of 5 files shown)


src/runtime/omp/omp_allocator.cpp
```diff
@@ -33,6 +33,29 @@
 namespace hipsycl {
 namespace rt {
 
+namespace {
+/// borrowed from LLVM
+/// Returns the next power of two (in 64-bits) that is strictly greater than A.
+/// Returns zero on overflow.
+uint64_t NextPowerOf2(uint64_t A) {
```
illuhad: `#include <cstdint>`?

src/runtime/omp/omp_allocator.cpp
```diff
@@ -33,6 +33,29 @@
...
+  A |= (A >> 32);
+  return A + 1;
+}
+
+/// Returns the power of two which is greater than or equal to the given value.
+/// Essentially, it is a ceil operation across the domain of powers of two.
+uint64_t PowerOf2Ceil(uint64_t A) {
+  if (!A)
+    return 0;
+  return NextPowerOf2(A - 1);
```
illuhad: Since this is such a small section of code, would it make sense to adapt it to the style of the surrounding code? Or maybe move it to the runtime's `util.hpp` as it might be useful in other places as well?

illuhad (commented):
Thank you! This is a great PR :)

src/runtime/omp/omp_allocator.cpp
fodinabor: in `util.hpp` already present :P

fodinabor:
Really looking forward to having @hipSYCL-ci-bot test this! :)

illuhad:
Thanks!

PR #449, opened 2021-01-28
Add subgroup-related fields to sycl::info::device

al42and:
Hello!

It seems that while the `sycl::sub_group` class is implemented fully, the related device properties (per SYCL 2020 spec) are
missing.

This PR adds the following descriptors to `sycl::info::device`:

- `max_num_sub_groups` (`uint32_t`): Set to maxThreadsPerBlock / warpSize.
- `sub_group_independent_forward_progress` (`bool`). Set to `true` for NVIDIA Volta devices, `false` otherwise. Depending on the compilation settings, warp-synchronicity can be ensured even on Volta, but the spec says "device supports," so we always return true for CC>=7.0.
- `sub_group_sizes` (`vector<size_t>`). Since neither NVIDIA, nor AMD give us any choice about the warp size, a vector of one element is always returned: {1} for CPU, {warpSize} for GPU.

For CPU device, there is a minor inconsistency: `max_num_sub_groups` is set to the upper value of size_t, 2<sup>64</sup>-1.
Since the sub_group on CPU always has the size of 1, `max_num_sub_groups` should also be 2<sup>64</sup>-1. However, it is uint32 per SYCL 2020 provisional specification and thus is capped at 2<sup>32</sup>-1.


src/runtime/cuda/cuda_hardware_manager.cpp
```diff
@@ -172,6 +172,15 @@ cuda_hardware_context::get_property(device_uint_property prop) const {
   case device_uint_property::max_group_size:
     return _properties->maxThreadsPerBlock;
     break;
+  case device_uint_property::max_num_sub_groups:
+    return _properties->maxThreadsPerBlock / _properties->warpSize;
+    break;
+  case device_uint_property::sub_group_independent_forward_progress:
+    return (_properties->major >= 7) ? 1 : 0; // True on Volta or newer
```
illuhad, line 179: The spec says:
> Returns true if the device supports independent forward progress of sub-groups with respect to other sub-groups in the same work-group.

So this is about forward progress of *different* subgroups/warps with respect to each other. My understanding is that this was also true before Volta. I think what has changed with Volta is forward-progress of individual threads within a single warp.

src/runtime/hip/hip_hardware_manager.cpp
```diff
@@ -183,6 +183,19 @@ hip_hardware_context::get_property(device_uint_property prop) const {
   case device_uint_property::max_group_size:
     return _properties.maxThreadsPerBlock;
     break;
+  case device_uint_property::max_num_sub_groups:
+    return _properties.maxThreadsPerBlock / _properties.warpSize;
+    break;
+  case device_uint_property::sub_group_independent_forward_progress:
+#ifdef HIPSYCL_RT_HIP_TARGET_CUDA
+    return (_properties.major >= 7) ? 1 : 0; // True on Volta or newer
```
illuhad, line 191: I don't think we need to concern ourselves with a potential CUDA path in the HIP backend. There are other places where we assume AMD-specific behavior here. Only a madman would attempt to use hipSYCL's HIP backend to target CUDA instead of just using CUDA ;)

include/hipSYCL/sycl/info/device.hpp
```diff
@@ -176,6 +179,10 @@ HIPSYCL_PARAM_TRAIT_RETURN_VALUE(device, device::max_work_item_dimensions, detai
 HIPSYCL_PARAM_TRAIT_RETURN_VALUE(device, device::max_work_item_sizes, id<3>);
 HIPSYCL_PARAM_TRAIT_RETURN_VALUE(device, device::max_work_group_size, size_t);
 
+HIPSYCL_PARAM_TRAIT_RETURN_VALUE(device, device::max_num_sub_groups, detail::u_int);
+HIPSYCL_PARAM_TRAIT_RETURN_VALUE(device, device::sub_group_independent_forward_progress, bool);
+HIPSYCL_PARAM_TRAIT_RETURN_VALUE(device, device::sub_group_sizes, vector_class<size_t>);
```
illuhad, line 184: `vector_class` is deprecated in SYCL 2020, just use std::vector.

include/hipSYCL/runtime/hardware.hpp
```diff
@@ -52,6 +52,9 @@ enum class device_uint_property {
   max_global_size1,
   max_global_size2,
   max_group_size,
+  max_num_sub_groups,
+  sub_group_independent_forward_progress,
+  sub_group_size,
```
illuhad, line 57: It might be better to not force this into the uint property API as we might have to return multiple values for this one. This is especially true for the new Level Zero/SPIR-V backend.
Add a dedicated virtual function for that?

tests/sycl/sub_group.cpp
```diff
@@ -108,6 +108,13 @@ BOOST_AUTO_TEST_CASE(sub_group) {
       }
     }
   }
+
+  const s::device dev = q.get_device();
+  const std::vector<size_t> supported_subgroup_sizes = dev.get_info<cl::sycl::info::device::sub_group_sizes>();
+  BOOST_CHECK(supported_subgroup_sizes.size() == 1);
+  BOOST_CHECK(supported_subgroup_sizes[0] == subgroup_size);
```
illuhad, line 115: The assumptions behind this part of the tests have changed recently because we can now target multiple backends simultaneously. In that case we run into trouble with the assumption that we can just determine the expected subgroup size based on those macros. I don't think we can actually test this apart from testing that it compiles.

illuhad (commented):
Sorry for the delay, I wanted to wait for SYCL 2020 final first because that part of the device info API was reworked. But I still think we can merge this, as we will have to change the property interface in the SYCL headers anyway.

illuhad:
Thanks, based on your PR I have restructured a bit how the information is exposed inside the runtime in PR #491. This PR effectively supersedes this one.

PR #451, opened 2021-01-29
[Win] Part 0: Link to Windows usage Wiki

fodinabor:
This is the first PR of a series, targeting basic and super experimental Windows support for OpenMP and CUDA backends.
Code PRs to follow :)

This also includes the required LLVM patch (based on 11.1.0 rc1), which is not yet merged.

### DIFF
--- a/doc/installing.md
+++ b/doc/installing.md
@@ -29,12 +29,11 @@ In order to successfully build and install hipSYCL, the following major requirem
   * hipSYCL requires clang for CUDA compilation. clang usually produces CUDA programs with very competitive performance compared to nvcc. For more information on compiling CUDA with clang, see [here](http://llvm.org/docs/CompileCudaWithLLVM.html).
   * Read more about [compatible clang and CUDA versions for the CUDA backend](install-cuda.md).
 * *For the ROCm backend*: 
-  * A recent **ROCm** must be installed. 
+  * A recent **ROCm**, in particular HIP, must be installed.
   * hipSYCL requires clang for ROCm compilation. While AMD's clang forks can in principle be used, regular clang is usually easier to set up. Read more about [compatible clang versions for the ROCm backend](install-rocm.md).
 * *For the CPU backend*: Any C++ compiler with **OpenMP** support should do.
 * python 3 (for the `syclcc` and `syclcc-clang` compiler wrappers)
 * `cmake`
-* hipSYCL also depends on HIP, but already comes bundled with a version of the HIP headers for the CUDA backend. The ROCm backend will use the HIP installation that is installed as part of ROCm.
 * the Boost C++ libraries (in particular `boost.fiber`, `boost.context` and for the unit tests `boost.test`)
 
 If hipSYCL does not automatically configure the build for the desired clang/LLVM installation, the following cmake variables can be used to point hipSYCL to the right one:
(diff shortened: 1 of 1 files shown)


doc/installing.md
```diff
@@ -34,7 +34,7 @@ In order to successfully build and install hipSYCL, the following major requirem
 * *For the CPU backend*: Any C++ compiler with **OpenMP** support should do.
 * python 3 (for the `syclcc` and `syclcc-clang` compiler wrappers)
 * `cmake`
-* hipSYCL also depends on HIP, but already comes bundled with a version of the HIP headers for the CUDA backend. The ROCm backend will use the HIP installation that is installed as part of ROCm.
+* HIP (the ROCm backend will use the HIP installation that is installed as part of ROCm).
```
illuhad: This should probably be moved to *For the ROCm backend* anyway.. Recent hipSYCL only needs HIP when targeting AMD.

doc/llvmPatchWindows.diff
```diff
@@ -0,0 +1,407 @@
+ clang/include/clang/AST/DeclCXX.h                | 13 ++++++++++
```
illuhad: Where does this patch come from? Is it also available somewhere else where we could just link to? I'm not sure that shipping clang/llvm patches that will inevitably break at some point and need maintenance is a good strategy.

doc/installing-windows.md
```diff
@@ -0,0 +1,79 @@
+# Using hipSYCL on Windows
```
illuhad: As this is more of an experimental guide rather than "official documentation", I wonder if the project Wiki might be a more appropriate place for it as it might allow for quicker editing and adapting to changes? We could still link to it in the main readme.

doc/llvmPatchWindows.diff
fodinabor: It's basically https://reviews.llvm.org/D69322 just being manually rebased on 11.1.0.. but yeah, I agree that it's not the best strategy.. so for now, will just set a link to the review in the Wiki?

doc/installing.md
fodinabor: Could be just removed then, as HIP is part of ROCm, so only ROCm is required?

illuhad: Hm.. It's a component of ROCm. I'm not sure we can assume every ROCm installation is a *full* installation. There are some packages in some distributions that only install ROCm's OpenCL for example. So we could list as requirement something like *ROCm, in particular HIP*

doc/installing.md
```diff
@@ -29,12 +29,11 @@ In order to successfully build and install hipSYCL, the following major requirem
   * hipSYCL requires clang for CUDA compilation. clang usually produces CUDA programs with very competitive performance compared to nvcc. For more information on compiling CUDA with clang, see [here](http://llvm.org/docs/CompileCudaWithLLVM.html).
   * Read more about [compatible clang and CUDA versions for the CUDA backend](install-cuda.md).
 * *For the ROCm backend*: 
-  * A recent **ROCm** must be installed. 
+  * A recent **ROCm** must be installed.
```
illuhad:
```suggestion
  * A recent **ROCm**, in particular HIP, must be installed.
```

illuhad (approved):

PR #452, opened 2021-01-29
[Win] Part 1: Make build system Windows aware

fodinabor:
This is the first code PR of a series, targeting basic and super experimental Windows support for OpenMP and CUDA backends.

What this PR does: In CMake, syclcc and friends are many Unix specifics, thus I extended this with the Windows specifics.
Some weird things have to be done, like using `clang.exe` instead of `clang++.exe` in syclcc, as plugins can only link against `clang.exe`.  
For boost, full library paths are used which are obtained from `Boost_{LIB_LIBRARY_{BUILD_TYPE}`.  
On Windows, one links against the `.lib` regardless, of having built a shared or static library (they just contain very different information), the `.dll` then must be located in the `%PATH%` or reside the executable. As one usually put's the `bin` path into `%PATH%`, the `.dll` are stored there, while `.lib`s (archives) must be stored in the `lib` folder.  
Nothing is changed about the interface for the user.

Windows also needs `python` before the scripts, thus the launcher adds this, before the invocation.
As `python` needs to be a separate element in the call to `subprocess.run`, it is separated from the rest by a character, which is forbidden in paths.


CMakeLists.txt
```diff
@@ -153,7 +165,11 @@ endif()
 set(DEFAULT_GPU_ARCH "" CACHE STRING "Optional: Default GPU architecture to compile for when targeting GPUs (e.g.: sm_60 or gfx900)")
 set(ROCM_LIBS "-L${ROCM_PATH}/lib -L${ROCM_PATH}/hip/lib -lamdhip64" CACHE STRING "Necessary libraries for ROCm")
 if(NOT ROCM_LINK_LINE)
-  set(ROCM_LINK_LINE "-Wl,-rpath=$HIPSYCL_ROCM_PATH/lib -Wl,-rpath=$HIPSYCL_ROCM_PATH/hip/lib ${ROCM_LIBS}")
+  if(WIN32)
+    set(ROCM_LINK_LINE "${ROCM_LIBS}")
+  else()
+    set(ROCM_LINK_LINE "-Wl,-rpath=$HIPSYCL_ROCM_PATH/lib -Wl,-rpath=$HIPSYCL_ROCM_PATH/hip/lib ${ROCM_LIBS}")
```
illuhad: Is this really an issue with ROCm not being supported on Windows?

CMakeLists.txt
```diff
@@ -162,7 +178,11 @@ if(NOT ROCM_CXX_FLAGS)
   set(ROCM_CXX_FLAGS "-isystem ${CLANG_INCLUDE_PATH} -U__FLOAT128__ -U__SIZEOF_FLOAT128__ -I$HIPSYCL_ROCM_PATH/include -I$HIPSYCL_ROCM_PATH/include --hip-device-lib-path=$HIPSYCL_ROCM_PATH/lib ")
 endif()
 if(NOT CUDA_LINK_LINE) 
-  set(CUDA_LINK_LINE "-Wl,-rpath=$HIPSYCL_CUDA_LIB_PATH -L$HIPSYCL_CUDA_LIB_PATH -lcudart" CACHE STRING "Arguments passed to compiler to link CUDA libraries to SYCL applications")
+  if(WIN32)
+    set(CUDA_LINK_LINE "-L$HIPSYCL_CUDA_LIB_PATH -lcudart" CACHE STRING "Arguments passed to compiler to link CUDA libraries to SYCL applications")
+  else()
+    set(CUDA_LINK_LINE "-Wl,-rpath=$HIPSYCL_CUDA_LIB_PATH -L$HIPSYCL_CUDA_LIB_PATH -lcudart" CACHE STRING "Arguments passed to compiler to link CUDA libraries to SYCL applications")
```
illuhad: I wonder if it might be cleaner to maintain `DEFAULT_CUDA_LINK_LINE` and `DEFAULT_WIN32_CUDA_LINK_LINE` that we can then just assign to `CUDA_LINK_LINE`?

CMakeLists.txt
```diff
@@ -217,10 +253,15 @@ endif()
...
-install(DIRECTORY include/hipSYCL DESTINATION include/ FILES_MATCHING PATTERN "*.hpp")
+install(DIRECTORY include/hipSYCL DESTINATION include/ FILES_MATCHING PATTERN "*.hpp" PATTERN include/hipSYCL/compiler EXCLUDE)
 
+if(NOT WIN32)
 install(PROGRAMS bin/syclcc DESTINATION bin)
+else()
+install(PROGRAMS bin/syclcc-clang DESTINATION bin RENAME syclcc)
+endif()
 install(PROGRAMS bin/syclcc-clang DESTINATION bin)
+
```
illuhad, line 273: Can you comment on what is going on here with the `syclcc`'s?

illuhad (commented):
There is a lot of if/else branching. I wonder if we could somehow reduce some of that to make it look a little cleaner and more robust...

CMakeLists.txt
fodinabor: `syclcc` is a symlink to `syclcc-clang`, which is not too well supported on Windows.. CMake doesn't like copying it:
```
CMake Error at cmake_install.cmake:57 (file):
  file INSTALL cannot read symlink
  "D:/Programmierung/SYCL/hipSYCLWin/bin/syclcc" to duplicate at
  "D:/Programmierung/SYCL/hipSYCL_install/bin/syclcc": No such file or
  directory.
```

thus, I made CMake just copy `syclcc-clang` to `syclcc` in the install directory.

fodinabor:
rebased on `develop` and checked with explicit multipass.. as far as I could tell, everything was fine, except for some small merge conflicts ...

PR #453, opened 2021-01-29
[Win] Part 2: Make frontend plugin Windows compatible.

fodinabor:
This reverts to always using Itanium mangling, but on Windows anyways relies on this [LLVM patch](https://reviews.llvm.org/D69322) to get consistent lambda numbering.
Thus, the new API, introduced in that patch, is always used on Windows, to get the correct behavior.

Also included is a hack to fix some compiler crashes, caused by nullptrs being handed into `TraverseDecl`.

### DIFF
--- a/include/hipSYCL/compiler/Frontend.hpp
+++ b/include/hipSYCL/compiler/Frontend.hpp
@@ -37,6 +37,7 @@
 #include "clang/AST/Decl.h"
 #include "clang/AST/DeclCXX.h"
 #include "clang/Basic/LLVM.h"
+#include "clang/Basic/TargetInfo.h"
 #include "clang/Frontend/FrontendPluginRegistry.h"
 #include "clang/AST/AST.h"
 #include "clang/AST/Attr.h"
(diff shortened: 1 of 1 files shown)


include/hipSYCL/compiler/Frontend.hpp
```diff
@@ -167,8 +172,14 @@ class FrontendASTVisitor : public clang::RecursiveASTVisitor<FrontendASTVisitor>
 public:
   FrontendASTVisitor(clang::CompilerInstance &instance)
       : Instance{instance}
-      , KernelNameMangler(instance.getASTContext().createMangleContext())
-  {}
+      , KernelNameMangler(clang::ItaniumMangleContext::create(instance.getASTContext(), instance.getASTContext().getDiagnostics()))
```
illuhad: Maybe add a line break?

illuhad:
@hipSYCL-ci-bot test with various clangs :)

illuhad (approved):

psalz:
So Clang plugins now work on Windows huh? Interesting... 🤔

fodinabor:
It's possible with clang.exe (not clang++ or clang-cl).. But as *all* symbols are exported, it might happen at some point that there are too many (don't remember the limit, but still a bit to go) and everything will break.. So won't be officially supported at any point in time I guess...

PR #454, opened 2021-01-29
[Win] Part 3: Make the runtime Windows compatible

fodinabor:
Windows, too does not have `exp10`, has `[u]int64_t` as `[unsigned] long long`, does not support `std::aligned_alloc` and thus requires special handling for host allocations.

CUDA also does not support memory prefetch or advices on Windows, so a warning is emitted instead.


include/hipSYCL/sycl/libkernel/vec.hpp
```diff
@@ -181,6 +181,14 @@ template<> struct logical_vector_op_result<uint64_t>
 template<> struct logical_vector_op_result<int64_t>
 { using type = int64_t; };
 
+#ifdef _WIN32
+template<> struct logical_vector_op_result<unsigned long>
+{ using type = int32_t; };
+
+template<> struct logical_vector_op_result<long>
+{ using type = int32_t; };
```
illuhad, line 189: So, the issue here is that on Windows `(unsigned) long != (u)int32_t` so that we need extra specializations? How are those types defined on Windows? It might also be worth checking with the SYCL spec here.

src/runtime/cuda/cuda_allocator.cpp
```diff
@@ -167,6 +167,11 @@ result cuda_allocator::mem_advise(const void *addr, std::size_t num_bytes,
   }
 
   return make_success();
+#else
+  HIPSYCL_DEBUG_WARNING << "cuda_allocator: Ignoring mem_advise() hint"
+                        << std::endl;
+  return make_success();
```
illuhad: Combine with previous `make_success()` and move to end of function?

src/runtime/cuda/cuda_queue.cpp
```diff
@@ -253,6 +253,11 @@ result cuda_queue::submit_prefetch(const prefetch_operation& op) {
                                  error_code{"CUDA", err}});
   }
   return make_success();
+#else
+  HIPSYCL_DEBUG_WARNING << "cuda_queue: Ignoring prefetch() hint"
+                        << std::endl;
+  return make_success();
```
illuhad: Same here?

src/runtime/omp/omp_allocator.cpp
```diff
@@ -52,6 +53,16 @@ void *omp_allocator::allocate(size_t min_alignment, size_t size_bytes) {
...
 #endif
+#else
+  if(min_alignment >= 1 && size_bytes % min_alignment != 0)
+    return nullptr;
+
+  if(min_alignment < alignof(void*))
+    min_alignment = alignof(void*); 
+  min_alignment = power_of_2_ceil(min_alignment);
+
+  return _aligned_malloc(size_bytes, min_alignment);
```
illuhad, line 59: Maybe move to its own function that is called here to simplify `allocate()`?

include/hipSYCL/sycl/libkernel/vec.hpp
fodinabor: from stdint.h:
```c++
typedef signed char        int8_t;
typedef short              int16_t;
typedef int                int32_t;
typedef long long          int64_t;
typedef unsigned char      uint8_t;
typedef unsigned short     uint16_t;
typedef unsigned int       uint32_t;
typedef unsigned long long uint64_t;
```
And `[unsigned] long` is afaik just an alias for `[unsigned] int` but still handled as a distinct type..

From Table 4.108 (the member definition of `sycl::vec`) I mostly read, that the width of the input values should be equal to the width of the output, which hereby is satisfied.

Elsewhere (D.3) it acknowledges the differences and for interop, requires to use `cl_int` and so forth.
It also copied the definition of the types from the C++ standard into `5.5`. Where this discrepancy is allowed..

Otherwise the SYCL spec is little bit confused it self 🤔
>4.16.2.2 Aliases
The SYCL programming API provides all permutations of the type alias:
`using <type><elems> = vec<<storage-type>, <elems>>`
where `<elems>` is 2, 3, 4, 8 and 16, and pairings of `<type>` and `<storage-type>` for integral types are char and int8_t, uchar and uint8_t, short and int16_t, ushort and uint16_t, int and int32_t, uint and uint32_t, long and int64_t, ulong and uint64_t, and for floating point types are both half, float and double. 
**For example uint4 is the alias to `vec<uint64_t, 4>` and float16 is the alias to `vec<float, 16>`.**

illuhad: Okay. I think the quote in 4.16.2.2 is a typo in the spec - this is clearly not the intent ;)

illuhad (approved):

PR #455, opened 2021-01-29
[Win] Part 4: Make Boost.Test and Windows happy

fodinabor:
We don't want `min/max` macros or similar weird stuff from Windows headers, thus this adds some defines to remove them.

Boost.Test (in the version 1.70.0 I managed to get running at all (🙄)) did not use tolerance based comparison, thus using some casts and explicit `double`s to get to tolerance based comparisons.

### DIFF
--- a/tests/sycl/math.cpp
+++ b/tests/sycl/math.cpp
@@ -29,6 +29,8 @@
 
 #include <boost/mpl/joint_view.hpp>
 
+#include <cmath>
+
 BOOST_FIXTURE_TEST_SUITE(math_tests, reset_device_fixture)
 
 // list of types classified as "genfloat" in the SYCL standard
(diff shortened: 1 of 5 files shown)


tests/sycl/math.cpp
```diff
@@ -318,14 +320,14 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(math_genfloat_binary, T,
...
-      BOOST_TEST(comp(acc[i++], c) == hypot(comp(acc[0], c), comp(acc[1], c)), tolerance);
-      BOOST_TEST(comp(acc[i++], c) == pow(comp(acc[0], c), comp(acc[1], c)), tolerance);
+      BOOST_TEST(comp(acc[i++], c) == std::atan2<double>(comp(acc[0], c), comp(acc[1], c)), tolerance);
+      BOOST_TEST(comp(acc[i++], c) == std::copysign<double>(comp(acc[0], c), comp(acc[1], c)), tolerance);
+      BOOST_TEST(comp(acc[i++], c) == std::fdim<double>(comp(acc[0], c), comp(acc[1], c)), tolerance);
+      BOOST_TEST(comp(acc[i++], c) == std::fmin<double>(comp(acc[0], c), comp(acc[1], c)), tolerance);
+      BOOST_TEST(comp(acc[i++], c) == std::fmax<double>(comp(acc[0], c), comp(acc[1], c)), tolerance);
+      BOOST_TEST(comp(acc[i++], c) == std::fmod<double>(comp(acc[0], c), comp(acc[1], c)), tolerance);
+      BOOST_TEST(comp(acc[i++], c) == std::hypot<double>(comp(acc[0], c), comp(acc[1], c)), tolerance);
+      BOOST_TEST(comp(acc[i++], c) == std::pow<double>(comp(acc[0], c), comp(acc[1], c)), tolerance);
```
illuhad, line 330: Why do we need the template arguments? I'm not sure we are guaranteed that these are templated functions, I think they might also be overloads.

fodinabor: we could also make that a static_cast, but for some reason, I had to aid Boost here, to use tolerance comparisons..
And as of now (until maybe we got Concepts) the 4th overload has to be a template, if I'm not mistaken: https://en.cppreference.com/w/cpp/numeric/math/atan2

illuhad (approved):

PR #459, opened 2021-02-03
Add support for explicit multipass to target multiple device backends (e.g. CUDA+HIP) simultaneously

illuhad:
While it was possible to use the OpenMP backend simultaneously with either CUDA or HIP backend, it was not yet possible to create a single binary that can run on both CUDA and HIP devices. This PR adds support for explicit multipass mode which allows to target CUDA, HIP (and OpenMP) simultaneously (e.g. using an argument like `--hipsycl-targets="omp;cuda:sm_61;hip:gfx906"`). Explicit multipass requires clang >= 11.

Here is the new documentation -- integrated multipass is what we had previously.

---
hipSYCL distinguishes two modes of compilation:
* *integrated multipass*, where host and device compilation passes are handled by clang's CUDA and HIP drivers. This mode allows for the most interoperability with backends because backend-specific language extensions are also available in the host pass. For example, kernels can be launched using the `<<<>>>` syntax. However, limitations in clang's compilation drivers also affect hipSYCL in this mode. In particular, CUDA and HIP cannot be targeted simultaneously because host code cannot support language extensions from *both* at the same time.
* *explicit multipass*, where host and device compilation passes are handled by hipSYCL's `syclcc` compiler driver. Here, `syclcc` invokes backend-specific device passes, which then result in a compiled kernel image for each target device. `syclcc` then embeds the compiled kernel image into the host binary. At runtime, the kernel is extracted from the image and invoked using runtime functions instead of language extensions. Therefore, in the host pass only C++ support is required. As a consequence, explicit multipass compilation for one backend can be combined with arbitrary other backends simultaneously, allowing hipSYCL to target multiple device backends at the same time.

Not all backends support all modes:

| Backend | Integrated multipass? | Explicit multipass? | Comment |
|------------------|-------------------|------------------|------------------|
| OpenMP | N/A | N/A | Does not require multipass compilation |
| CUDA   | Yes | Yes |  |
| HIP   | Yes | No |  |
---

When compiling with `--hipsycl-explicit-multipass`, all backends supporting explicit multipass run in this mode. Currently, this only affects the CUDA backend. In HIP it could be implemented as well, but it's not terribly important because we can already target HIP and CUDA simultaneously if only CUDA supports explicit multipass.

The implementation in the CUDA backend is as follows:
* `syclcc` does `--cuda-device-only` clang invocations to generate PTX for each target;
* PTX is bundled into a header file;
* the header file is included during the host pass;
* At runtime, the PTX kernels are loaded into the CUDA driver API and kernels launched using driver API functionality.

Explicit multipass requires clang >= 11 because we need `__builtin_unique_stable_name()` which originated from DPC++, but has been upstreamed in clang 11. When compiling with clang >= 11, kernel naming inside the clang plugin is always active.

**TODO**
* CI fails because due to the way we install the CUDA toolkit we don't have libcuda.so (the driver API) available, which is now a dependency.
* We need to play around with it some more

**Caveats**
* In explicit multipass mode, it is currently not possible to compile and link multiple source files in a single invocation, e.g. `syclcc -o test test1.cpp test2.cpp`. This is not a big restriction since build systems used in production (e.g. cmake) don't do that anyway.

**Note:**
* This changes the meaning of the `HIPSYCL_PLATFORM_*` macros somewhat: They are defined when, in the current compilation pass, backend-specific language extensions (e.g. CUDA syntax) are available. In explicit multipass mode, we now have the case where we target a backend without having those language extensions available in the host pass. This means that e.g. in CUDA explicit multipass, `HIPSYCL_PLATFORM_CUDA` will not be defined in the host pass. The macros `__HIPSYCL_ENABLE_*_TARGET__` can be used to query if a particular backend is targeted.
* Independently of this PR, the hipSYCL runtime cannot yet emit data transfers between two device backends. This means, if you try to run on a system that has both CUDA and HIP GPUs and your accessors trigger a data copy between a CUDA and a HIP GPU, things will cease to work. This is because data transfers are not directly possible between the CUDA and HIP backends - you would first need to copy from source to host, and then from host to target. As a workaround, you can emit the required copies manually.
* `libcuda.so` (CUDA driver API) is now a dependency


CMakeLists.txt
```diff
@@ -201,7 +201,9 @@ set(SYCLCC_CONFIG_FILE "{
   \"default-rocm-link-line\" : \"${ROCM_LINK_LINE}\",
   \"default-rocm-cxx-flags\" : \"${ROCM_CXX_FLAGS}\",
   \"default-cuda-link-line\" : \"${CUDA_LINK_LINE}\",
-  \"default-cuda-cxx-flags\" : \"${CUDA_CXX_FLAGS}\"
+  \"default-cuda-cxx-flags\" : \"${CUDA_CXX_FLAGS}\",
+  \"default-is-explicit-multipass\" : \"false\",
```
fodinabor, line 208: Would be great to have this as a CMake option, I guess?

bin/syclcc-clang
```diff
@@ -555,6 +583,95 @@ def run_or_print(command, print_only):
     print(' '.join(command))
     return 0
 
+class cuda_multipass_invocation:
+  def __init__(self, config):
+    self._config = config
+    self._temp_dir = None
```
fodinabor: any reason to init this when setting it in `set_temp_dir` immediately?

bin/syclcc-clang
```diff
@@ -555,6 +583,95 @@ def run_or_print(command, print_only):
...
+    self._integration_header = os.path.join(
+      self._temp_dir,"__hipsycl_embedded_cuda_kernels.hpp")
+
+  def get_requested_targets(self):
+    return self._config.targets["cuda"]
+
+  def get_device_compiler(self):
+    return self._config.clang_path
+
+  def get_compiler_preference(self):
```
fodinabor, line 604: interesting... from this not being used so far I guess this is preps for plans to support HIP Clang or DPC++ with this? Nice :)

bin/syclcc-clang
```diff
@@ -555,6 +583,95 @@ def run_or_print(command, print_only):
...
+        ptx_content.append(f.read())
+
+    result = """ 
+#ifndef HIPSYCL_CUDA_INTEGRATION_HEADER_HPP
+#define HIPSYCL_CUDA_INTEGRATION_HEADER_HPP
+
+static std::array __hipsycl_cuda_bundle {
+"""
+    for target, ptx in zip(targets, ptx_content):
+      result += "__hipsycl_cuda_embedded_object {{ std::string {{\"{0}\"}}, std::string {{R\"__hipsycl_embed({1})__hipsycl_embed\"}} }},\n".format(target, ptx)
```
fodinabor: or in a single long line without having to manually remove the end :) (and I'm sure your editor will tell you how this can be formatted in shorter lines ;-))
```python
result += ",\n".join(["__hipsycl_cuda_embedded_object {{ std::string {{\"{0}\"}}, std::string {{R\"__hipsycl_embed({1})__hipsycl_embed\"}} }}".format(target, ptx) for target, ptx in zip(targets, ptx_content)])
result += """
};
"""
```

bin/syclcc-clang
```diff
@@ -683,28 +800,61 @@ class compiler:
...
       # get access to things like host tasks
      self._backends.append(omp_sequential_invocation(config))
 
+    self._source_files = config.source_file_arguments
+    self._multipass_user_args = self._construct_multipass_user_args()
+
+  def _construct_multipass_user_args(self):
+    args = list(self._user_args)
+
+    # For multipass, we nyeed to remove -c and -o arguments
```
fodinabor: typo `need`

include/hipSYCL/glue/generic/hiplike/hiplike_kernel_launcher.hpp
```diff
@@ -587,8 +662,9 @@ class hiplike_kernel_launcher : public rt::backend_kernel_launcher
 
           if constexpr (type == rt::kernel_type::basic_parallel_for) {
 
-            __hipsycl_launch_kernel(
-                hiplike_dispatch::parallel_for_kernel<KernelName>,
+            __hipsycl_invoke_kernel(
+                hiplike_dispatch::parallel_for_kernel<KernelName>, KernelName,
```
fodinabor, line 666: weird formatting (move `KernelName` down)..

include/hipSYCL/glue/generic/hiplike/hiplike_kernel_launcher.hpp
```diff
@@ -627,8 +705,9 @@ class hiplike_kernel_launcher : public rt::backend_kernel_launcher
             for (int i = 0; i < Dim; ++i)
               assert(global_range[i] % effective_local_range[i] == 0);
 
-            __hipsycl_launch_kernel(
-                hiplike_dispatch::parallel_region<KernelName>,
+            __hipsycl_invoke_kernel(
+                hiplike_dispatch::parallel_region<KernelName>, KernelName,
```
fodinabor, line 709: formatting again :)
(it's a good sign when that's the only thing to complain about :))

include/hipSYCL/runtime/cuda/cuda_module.hpp
```diff
@@ -0,0 +1,96 @@
...
+#include "hipSYCL/runtime/error.hpp"
+#include "hipSYCL/runtime/device_id.hpp"
+#include "hipSYCL/glue/generic/module.hpp"
+
+struct CUmod_st;
+
+namespace hipsycl {
+namespace rt {
+
+using cuda_module_id_t = unsigned long long;
```
fodinabor, line 43: any reason for not using `std::uint64_t` in these occasions?

src/runtime/cuda/cuda_module.cpp
```diff
@@ -0,0 +1,211 @@
...
+cuda_module::cuda_module(cuda_module_id_t module_id, const std::string &target,
+                         const std::string &code_content)
+    : _id{module_id}, _target{target}, _content{code_content} {
+
+  std::istringstream code_stream(code_content);
+  std::string line;
+
+  while (std::getline(code_stream, line)) {
+
+    std::string kernel_identifier = ".visible .entry";
```
fodinabor: I don't usually comment this, but for some reason, this triggered me, that this is not `const` :)

src/runtime/cuda/cuda_module.cpp
```diff
@@ -0,0 +1,211 @@
...
+    }
+  }
+}
+
+const std::vector<std::string> &
+cuda_module::get_kernel_names() const {
+  return _kernel_names;
+}
+
+std::string cuda_module::get_content() const {
```
fodinabor, line 90: any reason why this does not return a const-ref as well?

src/runtime/cuda/cuda_module.cpp
```diff
@@ -0,0 +1,211 @@
...
+cuda_module_manager::obtain_module(cuda_module_id_t id,
+                                   const std::string &target,
+                                   const std::string &content) {
+  for (const cuda_module &mod : _modules) {
+    if (mod.get_id() == id && mod.get_target() == target) {
+      return mod;
+    }
+  }
+
+  _modules.push_back(cuda_module(id, target, content));
```
fodinabor: `emplace_back` ?

fodinabor (changes requested):
Love it 🎉 
Just some nits..

bin/syclcc-clang
illuhad: I see no reason at the moment to work with HIP-clang or other clang forks. Main reason is to satisfy the compiler invocation concept that the other classes also implement (maybe should turn it into a base class at some point).

include/hipSYCL/runtime/cuda/cuda_module.hpp
illuhad: The idea was to have as little dependencies as possible in the generated header file for robustness reasons (apart from the `std::array` and `std::string` which are needed for obvious reasons). The definition here then just aligns with what `syclcc` generates.

src/runtime/cuda/cuda_module.cpp
illuhad: IIRC this is only called when we actually need to compile the PTX/construct the CUDA module anyway. In that situation, copying an std::string does not impact performance. In that context I felt that risking accidentally returning a reference to a temporary if the content is maybe post-processed in the future is not worth it.

CMakeLists.txt
illuhad: I think as long as this is a new thing that we need to get experience with we should disable it by default.
`syclcc` will automatically enable explicit multipass mode anyway if you compile for both CUDA and HIP :)

illuhad:
We may need to change the `#ifdefs` since `__builtin_unique_stable_name` was merged for clang 11, but reverted afterwards and is currently being revised. It might not be in clang 12 yet.

bin/syclcc-clang
illuhad: Yeaah, this is what happens when you write too much C++ code - sometimes you forget how easy things are in Python ;)

include/hipSYCL/glue/generic/hiplike/hiplike_kernel_launcher.hpp
illuhad: Soo.. I checked all `__hipsycl_invoke_kernel()` calls again, and it seems that this is what `clang-format` thinks is optimal. Apparently it just tries to put all arguments into one line until it would exceed the maximum line width, then just inserts a line break and continues with the next line.
I can change this manually, but it will probably be changed back the next time somebody auto-formats that code.

src/runtime/cuda/cuda_module.cpp
illuhad: Fair point, done ;)

src/runtime/cuda/cuda_module.cpp
illuhad: Okay.

bin/syclcc-clang
illuhad: Done

bin/syclcc-clang
illuhad: Probably just historic reasons ;) Removing it.

CMakeLists.txt
fodinabor: Not sure I agree that the new-ness is a reason for not having it as an option, but yeah, the automatic fall back to explicit is fine, I guess :)

include/hipSYCL/runtime/cuda/cuda_module.hpp
fodinabor: okay.. that's a valid reason :) and `unsigned long long` is a good type, then (as it's guaranteed to be 64bit :))

src/runtime/cuda/cuda_module.cpp
fodinabor: okay.

fodinabor (approved):

illuhad:
@hipsycl-ci-bot test again.

PR #462, opened 2021-02-07
[SYCL2020] optimized group_functions

DieGoldeneEnte:
optimized group_functions and fixed some bugs. Notable changes:
- subgroups are used to lessen local-memory-requirements
- static memory is used for CPU local-memory
- better memory accesses
- group-sizes that are not a multiple of the warpSize work now (not always the case before)
- shuffle-instruction abstraction added

Problems:
- [ ] inclusive_scan_ptr can use so many registers it doesn't run with 1024-threads and large sycl::vec (usually >=8 elements)
- [x] some renaming/refactoring is still incoming

### DIFF
--- a/include/hipSYCL/compiler/Frontend.hpp
+++ b/include/hipSYCL/compiler/Frontend.hpp
@@ -205,7 +205,7 @@ public:
   bool VisitFunctionDecl(clang::FunctionDecl *f) {
     if(!f)
       return true;
-    
+
     this->processFunctionDecl(f);
 
     return true;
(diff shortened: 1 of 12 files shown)


DieGoldeneEnte:
@hipsycl-ci-bot test please check the new refactoring didn't break cuda (x2)

DieGoldeneEnte:
I removed the dpp code again, since clang already generates dpp-instructions. This also explains the inconsistent behavior  during measurements.

Edit: I ran the tests locally, since hipsycl-ci-bot doesn't execute them yet and the change is purely hip I let the ci-bot relax :)

illuhad (approved):

PR #465, opened 2021-02-10
Transition to plugin architecture for runtime backends

illuhad:
Previously, all backends for the runtime were compiled into a single library, `libhipSYCL-rt`. This has turned out to be problematic for two reasons:
1. All dependencies from all backends are contained in `linhipSYCL-rt`, and as such all backend dependencies must be available when symbols are resolved when linking with `syclcc`. This can be cumbersome for dependencies like `libcuda.so`, which is only available on systems where the NVIDIA kernel driver is installed
2. It makes it very difficult to provide binary packages for hipSYCL:
   1. Either the package maintainer provides a hipSYCL distribution that has all backends enabled such that it works with as much user hardware as possible, at the cost of having hard dependencies on very large dependencies (e.g. CUDA/ROCm) which the user might not need;
   2. Or only provide a hipSYCL distribution for specific backends to avoid large dependencies at the cost of excluding users who have different hardware;
   3. Or provide multiple hipSYCL packages that have various combinations of backends enabled. This is difficult to maintain and doesn't scale as the number of hipSYCL backends increases due to the combinatoric explosion.

This PR addresses these issues by restructuring the runtime such that backends are individual plugin libraries. When the core hipSYCL runtime starts, it scans `$HIPSYCL_INSTALL_PREFIX/lib/hipSYCL` for runtime backends and dynamically loads them using `dlopen()`/`dlsym()`. Runtime backends are now built as individual libraries `librt-backend-*`.

This decouples backend dependencies from the core `libhipSYCL-rt` that is linked by `syclcc`. Unfortunately we cannot purge `ROCM_LINK_LINE` and `CUDA_LINK_LINE` entirely since the CUDA/HIP runtime libraries still need to be linked to SYCL applications to provide kernel invocation mechanisms when using integrated (non-explicit) multipass compilation.

As a consequence of this new approach, it is now possible to extend an existing hipSYCL installation (e.g. from a package) after the initial installation with support for additional backends by simply dropping additional backend libraries into the `lib/hipSYCL` directory.

Another important consequence of this change is that it is **no longer possible to directly call functions defined by runtime backends from kernel launchers or client code since the symbols are no longer defined in `libhipSYCL-rt`**. My impression is that this modularizes the design more and forces us to think more about generic interfaces, which might be a good thing in the long run.

This has required a couple of additional changes, such as refactoring explicit multipass to have a backend-independent interface (that for now is only implemented for CUDA).
I also had to temporarily disable the `device_manager` runtime tests, as that functionality is now no longer available to client code. I think the right way to approach such tests in the future is to develop tests directly against the individual backend libraries. But it is still unclear how to best structure this and how the build system could look like, especially if we don't want to have the `-lcuda` difficulties again. (cc @psalz)

**The recommended way of packaging hipSYCL changes as follows with this PR**:
* Provide a core hipSYCL package containing `libhipSYCL-rt`, `syclcc` and the include headers
* Provide one package per backend containing the appropriate `librt-backend-*` library.
* The core package *must* have a dependency on the package for the OpenMP backend since hipSYCL always needs a CPU backend to function.

Pinging package maintainers: @acxz @nazavode @sbalint98

### DIFF
--- a/include/hipSYCL/runtime/backend_loader.hpp
+++ b/include/hipSYCL/runtime/backend_loader.hpp
@@ -1,7 +1,7 @@
 /*
  * This file is part of hipSYCL, a SYCL implementation based on CUDA/HIP
  *
- * Copyright (c) 2018 Aksel Alpay
+ * Copyright (c) 2021 Aksel Alpay
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -25,30 +25,52 @@
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#ifndef HIPSYCL_BACKEND_LOADER_HPP
+#define HIPSYCL_BACKEND_LOADER_HPP
 
-#ifndef HIPSYCL_VERSION_HPP
-#define HIPSYCL_VERSION_HPP
 
 #include <string>
+#include <vector>
+#include <utility>
 
-#include "hipSYCL/common/config.hpp"
+namespace hipsycl::rt {
+class backend;
+}
+
+
+#define HIPSYCL_PLUGIN_API_EXPORT extern "C"
+
+
+HIPSYCL_PLUGIN_API_EXPORT
+hipsycl::rt::backend *hipsycl_backend_plugin_create();
+
+HIPSYCL_PLUGIN_API_EXPORT
+const char* hipsycl_backend_plugin_get_name();
+
 
 namespace hipsycl {
-namespace sycl {
-namespace detail {
+namespace rt {
 
-static std::string version_string()
-{
-  std::string hipsycl_version = std::to_string(HIPSYCL_VERSION_MAJOR)
-      + "." + std::to_string(HIPSYCL_VERSION_MINOR)
-      + "." + std::to_string(HIPSYCL_VERSION_PATCH)
-      + "-" + std::string(HIPSYCL_VERSION_TYPE);
+class backend_loader {
+public:
+  ~backend_loader();
 
-  return "hipSYCL " + hipsycl_version;
-}
+  void query_backends();
+  
+  std::size_t get_num_backends() const;
+  std::string get_backend_name(std::size_t index) const;
+  bool has_backend(const std::string &name) const;
 
-}
+  backend *create(std::size_t index);
+  backend *create(const std::string &name);
+
+private:
+  using handle_t = void*;
+  std::vector<std::pair<std::string, handle_t>> _handles;
+};
+
 }
 }
 
 #endif
+
--- a/include/hipSYCL/runtime/backend_loader.hpp
+++ b/include/hipSYCL/runtime/backend_loader.hpp
@@ -61,8 +61,8 @@ public:
   std::string get_backend_name(std::size_t index) const;
   bool has_backend(const std::string &name) const;
 
-  backend *create(std::size_t index);
-  backend *create(const std::string &name);
+  backend *create(std::size_t index) const;
+  backend *create(const std::string &name) const;
 
 private:
   using handle_t = void*;
--- a/src/runtime/CMakeLists.txt
+++ b/src/runtime/CMakeLists.txt
@@ -57,7 +57,7 @@ if(WITH_CUDA_BACKEND)
   target_link_libraries(rt-backend-cuda PRIVATE hipSYCL-rt ${CUDA_LIBRARIES} ${CUDA_DRIVER_LIBRARY})
 
   install(TARGETS rt-backend-cuda
-        EXPORT install_exports
+        #EXPORT install_exports
         LIBRARY DESTINATION lib/hipSYCL
         ARCHIVE DESTINATION lib/static)
 endif()
@@ -83,7 +83,7 @@ if(WITH_ROCM_BACKEND)
   endif()
 
   install(TARGETS rt-backend-hip
-        EXPORT install_exports
+        #EXPORT install_exports
         LIBRARY DESTINATION lib/hipSYCL
         ARCHIVE DESTINATION lib/static)
 endif()
@@ -109,7 +109,7 @@ if(WITH_CPU_BACKEND)
     target_link_libraries(rt-backend-omp PRIVATE hipSYCL-rt  OpenMP::OpenMP_CXX)
 
     install(TARGETS rt-backend-omp
-        EXPORT install_exports
+        #EXPORT install_exports
         LIBRARY DESTINATION lib/hipSYCL
         ARCHIVE DESTINATION lib/static)
 endif()
--- a/src/runtime/CMakeLists.txt
+++ b/src/runtime/CMakeLists.txt
@@ -57,7 +57,6 @@ if(WITH_CUDA_BACKEND)
   target_link_libraries(rt-backend-cuda PRIVATE hipSYCL-rt ${CUDA_LIBRARIES} ${CUDA_DRIVER_LIBRARY})
 
   install(TARGETS rt-backend-cuda
-        #EXPORT install_exports
         LIBRARY DESTINATION lib/hipSYCL
         ARCHIVE DESTINATION lib/static)
 endif()
@@ -83,7 +82,6 @@ if(WITH_ROCM_BACKEND)
   endif()
 
   install(TARGETS rt-backend-hip
-        #EXPORT install_exports
         LIBRARY DESTINATION lib/hipSYCL
         ARCHIVE DESTINATION lib/static)
 endif()
@@ -109,7 +107,6 @@ if(WITH_CPU_BACKEND)
     target_link_libraries(rt-backend-omp PRIVATE hipSYCL-rt  OpenMP::OpenMP_CXX)
 
     install(TARGETS rt-backend-omp
-        #EXPORT install_exports
         LIBRARY DESTINATION lib/hipSYCL
         ARCHIVE DESTINATION lib/static)
 endif()
--- a/src/runtime/CMakeLists.txt
+++ b/src/runtime/CMakeLists.txt
@@ -1,11 +1,13 @@
 
 set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
 
-set(HIPSYCL_SOURCES
+
+add_library(hipSYCL-rt SHARED
   application.cpp
   runtime.cpp
   error.cpp
   backend.cpp
+  backend_loader.cpp
   hints.cpp
   device_id.cpp
   operations.cpp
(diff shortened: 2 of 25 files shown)


src/runtime/backend_loader.cpp
```diff
@@ -0,0 +1,178 @@
...
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+
+#include "hipSYCL/runtime/backend_loader.hpp"
+#include "hipSYCL/common/debug.hpp"
+#include "hipSYCL/common/config.hpp"
+
+#include <cassert>
+#include <filesystem>
```
sbalint98, line 34: Maybe use `#include <boost/filesystem.hpp>` instead since we already have a dependency on boost and that way we can support oses that do not have recent enough standard library.

illuhad: `std::filesystem` is part of C++17, which we already require anyway. We don't have a dependency on `boost::filesystem` at the moment. Is there an OS where `std::filesystem` is a problem?

sbalint98: Sorry for some reason I thought we require filesystem as well. On Ubuntu 18.04 the standard library is too old. The gcc version is gcc7.5, but filesystem only properly supported in 8.1. Probably the same is true for the centos7. 
https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html#status.iso.2017

src/runtime/backend_loader.cpp
```diff
@@ -0,0 +1,178 @@
...
+  return create_backend_func();
+}
+
+}
+
+namespace hipsycl {
+namespace rt {
+
+void backend_loader::query_backends() {
+  std::string install_prefix = HIPSYCL_INSTALL_PREFIX;
```
fodinabor, line 105: feels super unflexible and completely un-redistributable, I'd say.. there should be at least a fallback to look relative to the binaries' directory.

Also, all these strings and so on could be `const`.. helps making intent clear (also being nice to the compiler :))

src/runtime/backend_loader.cpp
illuhad: Is there a way to update the toolchain? E.g. newer `devtoolset` packages on CentOS? We might want to use other C++17 library features in the future, so giving up on C++17 is not an attractive strategy ;)
Is this even a problem for packaging with spack? Won't it provide its own headers?

include/hipSYCL/runtime/backend_loader.hpp
```diff
@@ -0,0 +1,76 @@
...
+#include <string>
+#include <vector>
+#include <utility>
+
+namespace hipsycl::rt {
+class backend;
+}
+
+
+#define HIPSYCL_PLUGIN_API_EXPORT extern "C"
```
fodinabor, line 41: Great, we can misuse this for adding `__declspec(dllexport)` then for Windows ;-)
See: https://docs.microsoft.com/en-us/cpp/build/exporting-from-a-dll-using-declspec-dllexport?view=msvc-160 and the more manual alternative of specifying the exports with their ordinals (resulting in ensured binary compatibility) https://docs.microsoft.com/en-us/cpp/build/exporting-from-a-dll-using-def-files?view=msvc-160

include/hipSYCL/runtime/backend_loader.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+
+namespace hipsycl {
+namespace rt {
+
+class backend_loader {
+public:
+  ~backend_loader();
+
+  void query_backends();
```
fodinabor, line 58: any reason for a dedicated function instead of immediately doing this in the Ctor?

src/runtime/CMakeLists.txt
```diff
@@ -21,85 +23,103 @@ set(HIPSYCL_SOURCES
...
+    $<INSTALL_INTERFACE:include>
+  PRIVATE
+    ${HIPSYCL_SOURCE_DIR}
+    ${PROJECT_BINARY_DIR}/include
+)
+
+install(TARGETS hipSYCL-rt
+        EXPORT install_exports
+        LIBRARY DESTINATION lib
+        ARCHIVE DESTINATION lib/static)
```
fodinabor: don't really get the benefit of moving the archive into a separate folder (I know, it's not new... but I still don't get it)

src/runtime/CMakeLists.txt
```diff
@@ -21,85 +23,103 @@ set(HIPSYCL_SOURCES
...
-    if(CMAKE_VERSION VERSION_LESS "3.16")
-      message(FATAL_ERROR "CMake 3.16.0+ is required for macOS OpenMP support!")
-    endif()
-    target_include_directories(hipSYCL-rt PRIVATE ${OpenMP_CXX_INCLUDE_DIRS})
-  endif()
-
-  target_link_libraries(hipSYCL-rt PRIVATE OpenMP::OpenMP_CXX)
 
-  target_compile_definitions(hipSYCL-rt PRIVATE HIPSYCL_RT_ENABLE_OMP_BACKEND=1)
+if(WITH_CPU_BACKEND)
```
fodinabor: forgot something here?

src/runtime/backend_loader.cpp
```diff
@@ -0,0 +1,178 @@
...
+#include <cassert>
+#include <filesystem>
+#include <dlfcn.h>
+
+namespace {
+
+void close_plugin(void *handle) {
+  int err = dlclose(handle);
+
+  if (err != 0) {
```
fodinabor: Just, preference, but I personally really like limiting the scope for error codes like this, to prevent name spilling:
`if(int err = dlclose(handle))`

src/runtime/backend_loader.cpp
```diff
@@ -0,0 +1,178 @@
...
+    HIPSYCL_DEBUG_ERROR << "backend_loader: dlclose() failed" << std::endl;
+  }
+}
+
+bool load_plugin(const std::string &filename, void *&handle_out,
+                 std::string &backend_name_out) {
+  void *handle = dlopen(filename.c_str(), RTLD_NOW);
+
+  if (!handle) {
```
fodinabor: Again (but here little less of just preference and more of a really useful thing), if you invert the `if` you can do: 
`if(void *handle = ..)` and the handle will always be valid in the scope, it is present in..
helps when thinking about, whether something is valid deep inside a branch :)

src/runtime/backend_loader.cpp
```diff
@@ -0,0 +1,178 @@
...
+    if (entry.is_regular_file()) {
+      auto p = entry.path();
+      if (p.extension().string() == shared_lib_extension) {
+        std::string backend_name;
+        void *handle;
+        if (load_plugin(p.string(), handle, backend_name)) {
+          HIPSYCL_DEBUG_INFO << "Successfully opened plugin: " << p
+                             << " for backend '" << backend_name << "'"
+                             << std::endl;
+          _handles.emplace_back(std::make_pair(backend_name, handle));
```
fodinabor, line 124: with `emplace_back`, you should be able to omit the `std::make_pair` :)
`_handles.emplace_back(backend_name, handle);`

src/runtime/backend_loader.cpp
```diff
@@ -0,0 +1,178 @@
...
+      return true;
+    }
+  }
+
+  // In case of failure, close handle to avoid resource leak
+  close_plugin(handle);
+  return false;
+}
+
+hipsycl::rt::backend *create_backend(void *plugin_handle) {
```
fodinabor, line 80: planning on caching anything here or so? else, all the create / close functions can be `const`
(and for caches, it's common to allow `mutable`)

src/runtime/backend_loader.cpp
```diff
@@ -0,0 +1,178 @@
...
+}
+
+backend *backend_loader::create(std::size_t index) {
+  assert(index < _handles.size());
+  
+  return create_backend(_handles[index].second);
+}
+
+backend *backend_loader::create(const std::string &name) {
+  if (!has_backend(name))
```
fodinabor: this is checked in the for loop as well.. no reason to go through the backends twice

fodinabor (changes requested):
Should be portable to Windows, but I really think we need some smarter way of getting the folder to load the plugins from..

src/runtime/backend_loader.cpp
illuhad: Not sure what you mean. It's a free function in an anonymous namespace, not a member function.

fodinabor: whoopsie.. commented on the wrong function then obviously :D
but still holds true for the `backend_loader::create` functions..?

src/runtime/backend_loader.cpp
illuhad: First priority is making sure that current packaging strategies with fixed install paths work as expected. Redistributing is a different story and might also require changes to the way we use `-rpath`, but I see the use case. Not sure if it's a good idea to just randomly start loading shared libraries from paths that are not necessarily trusted if something doesn't work.

The better approach is probably to introduce a field for the `settings` infrastructure that can be set via environment variable to override this. But I'm not sure if this might not better be a PR in itself.

src/runtime/CMakeLists.txt
illuhad: I don't actually know where this decision originally came from.. I just copy-pasted the existing pattern ;)

include/hipSYCL/runtime/backend_loader.hpp
illuhad: That was why I did it this way ;)

include/hipSYCL/runtime/backend_loader.hpp
illuhad: Since the backend_loader is a member of the runtime class, the constructor is called when static constructors are processed. My thinking was that I wanted to have evil low-level stuff which might go wrong like `dlsym()` at a well defined point where it's a bit easier to get to debugging-wise.

src/runtime/backend_loader.cpp
sbalint98: I don't think spack builds any standard library by default... On centos surely devtoolset can be used, on ubuntu, I have to check. I think the universal solution would be to build and ship our own standard library, we just have to enable `libcxx` when building llvm. (and link against it when building hipSYCL)

include/hipSYCL/runtime/backend_loader.hpp
fodinabor: I understand the reasoning, but not sure, the current version helps here at all? currently that function is called only inside `backend_manager`'s ctor, which actually is called when `runtime` is initialized.. Thus I don't really see a difference?

illuhad: Yes, the benefit at the moment is limited. It might still be easier to add diagnostic code around the call to `query_backend()`, or introduce additional logic such as instructing the backend_loader to only load certain backends etc.

src/runtime/backend_loader.cpp
illuhad: ~Well, the `backend_loader` manages the handles returned by dlopen, therefore state of the class changes. I don't consider it a cache, more of a RAII manager object for the handles.~
EDIT: Nevermind, was thinking about the wrong function ;) I guess we can make this here const.

src/runtime/CMakeLists.txt
illuhad: I have removed the separate installation directory.

src/runtime/CMakeLists.txt
illuhad: whopsie - forgot to delete that old `if()` ;) Removed.

src/runtime/backend_loader.cpp
illuhad: Good idea, I have applied this.

src/runtime/backend_loader.cpp
illuhad: hmm.. I don't really like the variadic template usage of `emplace_back` as it hides the relation of the arguments.

src/runtime/backend_loader.cpp
illuhad: Done.

src/runtime/backend_loader.cpp
illuhad: Done :)

src/runtime/backend_loader.cpp
illuhad: I have applied this here, but in general this is a common pattern that we also use e.g. for CUDA error codes. If you have a lot of subsequent calls that return error codes it is a bit more convenient to just define in the beginning and then just assign, but your suggestion makes sense here.

fodinabor (approved):
LGTM now.

illuhad:
@hipsycl-ci-bot test this tasty treat of text again

PR #466, opened 2021-02-11
Add requirement to data users even if the same operation is already a user

### LINKED ISSUE #464: Regression with hipSYCL >= 0.8.2 for hand-rolled tree reduction
**Bug summary**
We found a bug with hipSYCL > 0.8.2 where our existing hand-rolled tree reduction in [CloverLeaf](https://github.com/UoB-HPC/cloverleaf_sycl) produced the wrong answer.
This was not the case for versions <= 0.8.2 or DPC++ and ComputeCpp.

**To Reproduce**/**Expected behavior**

To facilitate testing with other SYCL implementations, I've create a [separate repo](https://github.com/UoB-HPC/reduction_sycl).

**Describe your setup**
Programs are compiled using CMake, this bug has only been verified on the CPU.

illuhad:
Fixes #464 

We need to add an operation to data users even if the same operation is already listed as a user, because the access mode might be different. This in turn can affect conflict detection in the `dag_builder`. In certain scenarios, this can lead to missing dependencies (and hence missing synchronization) in the DAG.

This does not introduce additional synchronization due to potential duplicate DAG edges, because `dag_node::add_requirement()` still only inserts a requirement to a node if is a new one.

### DIFF
--- a/src/runtime/dag_builder.cpp
+++ b/src/runtime/dag_builder.cpp
@@ -57,11 +57,17 @@ void add_to_data_users(dag_node_ptr node, memory_requirement *mem_req) {
                            ->get_data_region()
                            ->get_users();
 
-    if (!data_users.has_user(node)) {
-      data_users.add_user(
-          node, mem_req->get_access_mode(), mem_req->get_access_target(),
-          mem_req->get_access_offset3d(), mem_req->get_access_range3d());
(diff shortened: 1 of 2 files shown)


illuhad:
@hipsycl-ci-bot test again please.

PR #471, opened 2021-02-13
Don't export the backend libraries

sbalint98:
Don't export the backend libraries in order to make package creation more straightforward.


src/runtime/CMakeLists.txt
```diff
@@ -57,7 +57,7 @@ if(WITH_CUDA_BACKEND)
   target_link_libraries(rt-backend-cuda PRIVATE hipSYCL-rt ${CUDA_LIBRARIES} ${CUDA_DRIVER_LIBRARY})
 
   install(TARGETS rt-backend-cuda
-        EXPORT install_exports
+        #EXPORT install_exports
```
fodinabor: Do you see any reason why one might want to keep this? Or could those lines just be removed?

illuhad: I have removed the lines. I don't think they are required.

PR #474, opened 2021-02-18
Support new ROCm versions by default

illuhad:
Support newer ROCm versions (>= 4.0) by default:
* Enable HIP managed memory API for `hipMemPrefetchAsync()` and `hipMemAdvise()` by default
* Use new `--rocm-path` and `--rocm-device-lib-path` flags (from clang >= 11)
* Adapt default ROCm device library path to the location used by ROCm since 3.9 (`$PREFIX/amdgcn/bitcode`)

With this PR, hipSYCL should be able to compile out-of-the-box against clang 11 and ROCm 4.0.

Because we now use clang 11 specific flags in the ROCm path, I had to disable clang 10 in CI.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -8,8 +8,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        clang_version: [10, 11]
-        rocm_version: [3.8]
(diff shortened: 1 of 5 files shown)


illuhad:
@hipSYCL-ci-bot test new ROCm again

PR #476, opened 2021-02-21
[Win][CI] Add a Windows CI using pre-built LLVM 11*.

fodinabor:
Just as with the Linux CI, CUDA and CPU tests are built, but only the CPU tests are actually executed.
* LLVM 11.1.0 was built with the required patch from https://reviews.llvm.org/D69322
It's using the action cache, as fetching of those >400MB would increase the overall build time heavily.

To fix an issue with Boost.Intrusive during device pass, the collective_execution_engine is ifdef'd away during device pass.


fodinabor:
@hipsycl-ci-bot test the fiber change, please

illuhad (approved):

PR #485, opened 2021-03-09
Only set policy CMP0074 if cmake >= 3.12 in root cMakeLists.txt

sbalint98:
This policy has been added to cmake 3.12 first, therefore it should only be set in case cmake is more recent than 3.12.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,7 @@
 cmake_minimum_required (VERSION 3.5)
-cmake_policy(SET CMP0074 NEW) # Don't complain about using BOOST_ROOT...
+if(NOT CMAKE_VERSION VERSION_LESS 3.12)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1,5 +1,7 @@
 cmake_minimum_required (VERSION 3.5)
-cmake_policy(SET CMP0074 NEW) # Don't complain about using BOOST_ROOT...
+if(CMAKE_VERSION VERSION_EQUAL 3.12 OR CMAKE_VERSION VERSION_GREATER 3.12)
```
illuhad: `NOT LESS` might be a bit simpler? I assume we are not using `GREATER_EQUAL` because it was introduced in cmake later than 3.5?

sbalint98: Yes it is a feature added in cmake 3.7.

illuhad (approved):

illuhad:
@hipSYCL-ci-bot test

PR #486, opened 2021-03-10
[SYCL 2020] Install SYCL/sycl.hpp also to sycl/sycl.hpp

illuhad:
SYCL 2020 final has changed the header location to `sycl/sycl.hpp`. For compatibility, we still provide `SYCL/sycl.hpp`.

On Windows, don't install `sycl/sycl.hpp` as filesystems are case-insensitive (? cc @fodinabor)


fodinabor (approved):
Yeah, wouldn't make too much sense on Windows :)
APFS (which might be used on MacOS) is iirc also case-insensitive, but in general copying both, will probably just result in the same file being overwritten by itself... so shouldn't matter :D

illuhad:
Alright, let's merge and see if Mac people will start complaining ;)

PR #487, opened 2021-03-10
Change signature of unary vec operators to accept const ref

illuhad:
Change signature of unary `vec` operators to accept const ref argument. Non-const ref is arguably a bug in the spec.

cc @mujjingun


daniel-geon-park:
FYI, one of the spec contributors has confirmed it's a mistake.
image

illuhad:
@mujjingun  I'm also one of the spec contributors, but great to see that Tom and I share the same opinion ;)

daniel-geon-park:
Ah nice! I should have guessed so.

PR #490, opened 2021-03-16
[SYCL 2020] Add accessor simplifications/deduction guides, host_accessor

illuhad:
(cc @tomdeakin)

Initial implementation of
* Accessor simplifications, deduction guides and `no_init` property
* host_accessor (partial implementation)
* Redesigned `property_list` which was required by the new accessor interface.

Allows code patterns such as
```c++
q.submit([&](sycl::handler& cgh){
  // read-write default for non-const buffer type
  sycl::accessor acc{buff, cgh};
  // Using deduction guides and no_init property
  sycl::accessor acc2{buff, cgh, sycl::read_only, sycl::no_init};
});
sycl::host_accessor hacc{buff, sycl::read_only};
```

*Note*: This PR does not yet contain additional changes/additions to the members of accessors, such as the iterator interface.

The only thing I'm not sure about is the pattern:
```
sycl::buffer<int> buff {...};
// construct read-only accessor
sycl::accessor<const int> acc{buff};
```
I think originally the intent was to allow this pattern in order to make read-write/read-only expressible only with accessor const-ness, but I don't see how the interface in the spec allows for that, and the wording in the spec appears pretty vague to me.
The spec hardcodes that the buffer arguments to the `accessor<T>` constructors is of type `buffer<T>`. 

We would need additional `accessor(buffer<std::remove_const_t<T>>&)` constructors to make constructing `accessor<const T>` from `buffer<T>` work, which the spec does not seem to anticipate (which would also double the number of constructors because we don't have enough yet ;-) )

### DIFF
--- a/include/hipSYCL/sycl/access.hpp
+++ b/include/hipSYCL/sycl/access.hpp
@@ -32,20 +32,13 @@
 
 namespace hipsycl {
 namespace sycl {
-namespace access {
-
-enum class mode {
-  read = 1024,
-  write,
-  read_write,
-  discard_write,
-  discard_read_write,
-  atomic
-};
 
 enum class target {
-  global_buffer = 2014,
-  constant_buffer,
+  device,
+  host_task,
+  global_buffer = device,
+  // Deprecated targets
+  constant_buffer = 2,
   local,
   image,
   host_buffer,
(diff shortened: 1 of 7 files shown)


illuhad:
@hipSYCL-ci-bot test again

PR #499, opened 2021-03-19
Redesign initialization mechanism for captured accessors using binary tagging

illuhad:
This PR redesigns pretty much from scratch how hipSYCL handles turning accessors into pointers. This is of course super-critical for functionality and correctness, so I appreciate reviews :-)

## Problem setup

Because (hip)SYCL allocates lazily, in general we cannot know when an accessor is constructed what the actual device pointer is going to be. That information is only available once the accessor has been bound to a command group, and the scheduler has processed it and assigned a device for execution.
This means we need to somehow put the information about the device pointer into the kernel lambda *after* it has captured the accessors. But because hipSYCL supports library-only CPU backends, in general we cannot introspect the contents of kernel lambdas at all. As a consequence, we need some workaround to modify the content of the lambda without knowing what the structure of the lambda object is.

## Current solution

The current approach is based on accessors that contain `glue::deferred_pointer<T>` objects. `deferred_pointer<T>` is a pointer wrapper with a twist: In its copy constructor it checks a predefined memory location for the actual device pointer. When the runtime has finished scheduling, it will put the device pointer in this location.
Afterwards, the kernel launchers only need to make sure that the kernel lambda is copied at least once. This then triggers the copy constructors of the captured `deferred_pointer` objects, which will then see the device pointer and copy it into the accessor.

## Limitations of current approach

This approach has worked reasonably well, but actually we have cheated slightly: The `handler::require()` method in hipSYCL has the signature `handler::require(accessor&)` while per spec it should be `handler::require(accessor)`, i.e. take the argument by value. 
While I used to think that this was a mostly unimportant detail, it turns out there are libraries out there that massively rely on the by-value behavior.

With the current approach, the by-value signature *cannot* work: In order to generate the deferred pointer not only for the copy passed into the `require()` argument, but also for the original accessor that is later captured, some shared state between accessors is required.
However, this does not really work in practice beyond trivial cases because of race conditions. Consider e.g.
```
sycl::queue queue_on_device1;
sycl::queue queue_on_device2;

sycl::accessor myplaceholder{...};

queue_on_device1.submit([&](handler& cgh){
  cgh.require(myplaceholder);
  cgh.parallel_for(...);
});

queue_on_device2.submit([&](handler& cgh){
  cgh.require(myplaceholder);
  cgh.parallel_for(...);
});
```
Both `require()` calls need to modify the state of `myplaceholder` to initialize the deferred pointers and bind them to the requirement objects of the hipSYCL runtime. But because of shared state, this then also impacts other captured copies of the accessor, potentially even from other kernel submissions. This can lead to kernels ending up with the wrong pointers inside their accessors, in the worst case even pointers for different devices.

## New solution

The approach taken in this PR is entirely different: We replace `deferred_pointer` with `embedded_pointer`. `embedded_pointer` generates and stores a 128-bit random (hopefully unique) id on construction. When accessors are copied, they copy the same unique id.
The requirement objects from the hipSYCL runtime that are bound to the accessors also know the unique id. Because we know all requirements for a kernel, we know the unique ids for all embedded pointers that might be used inside a kernel.
After scheduling, the kernel lambda object is scanned for the binary pattern of the unique id of each requirement. If found, it is overwritten with the actual device pointer.
Effectively, the unique id acts as a binary identification tag for the accessor inside the kernel lambda blob.

This is a lot of low-level binary hacking, but passes all unit tests. Of course there is the pathological case of user data captured by the kernel by chance being equal to an accessor's unique id. In practice I believe this risk is probably negligible.

This approach effectively allows us to detect accessors inside lambda objects. As such, we can now emit warnings if users register accessors with the runtime, but then don't use them inside kernels. 

@hipSYCL-ci-bot test please


illuhad:
@hipSYCL-ci-bot test please.

keryell:
What a hack! :-)
Otherwise a possible implementation would be a PIMPL implementation of the accessors and `require` can update the PIMPL-ed part.
I have the feeling that your can control the race-conditions by having `require` and others to check for some atomic protection inside each PIMPL-ed part.

fknorr:
I feel like there should still be a way to do this with copy constructor abuse.

```c++
thread_local handler *submit_cgh;

void queue::submit(CGF cgf) {
    handler cgh;
    submit_cgh = &cgh;
    cgf(cgh);
    submit_cgh = nullptr;
}

class accessor {
    accessor() {
        static std::atomic<uint64_t> next_id;
        _id = next_id++;
        _ptr = nullptr;
    }

    accessor(const accessor &) {
        if (!_ptr && submit_cgh) {
            _ptr = submit_cgh.resolve_device_pointer(_id);
        }
    }
};
```

illuhad:
@keryell Something pimpl-like was the first thing I attempted. It didn't go well ;) In principle you can solve this by synchronizing such that all kernels using the same accessors are processed sequentially. But this is not only bad for performance, it also breaks down when the runtime decides to batch up multiple command groups and process them together. In this case you can get deadlocks in the user code because scheduling doesn't progress until more nodes have been collected.

@fknorr The command group handler is not intended to outlive the submission call, but that's probably a detail. Again, I think this assumes that there is only a single command group in flight.

fknorr:
Ah yes, `handler` is probably not the right type to attach this `resolve_device_pointer` functionality to. Let's call it `device_pointer_resolver` instead. The `device_pointer_resolver` would be created by the backend executor with knowledge of all accessors required by the command group it wants to launch, and passed through to the final accessor copy ctor via a thread local. As long as the executor has some way of updating this thread local between "final kernel copy" operations, concurrency should not be a concern.

illuhad:
@fknorr Okay, I think I understand it now, thanks. That could work. We would need some of the infrastructure from this PR (such as having the `dag_node_ptr` available in the kernel launcher, and the requirements storing accessor ids). 
I'm not actually sure which approach is better, but I'm leaning towards binary tagging.

Pro `thread_local`+copy constructor:
* Similar behavior as currently
* No low-level wizardry, just well-defined C++
* Needs ~8 bytes less than binary tagging approach (but accessors already need to store offset, range, buffer shape, etc so wouldn't matter much)
* No theoretical risk of incorrect binary identification (but in practice I believe this risk is basically non-existent)
* Construction of new accessors might be slightly faster (binary tagging generates random number and timestamp, but probably something simpler like timestamp+global atomic counter would also work. But if your code is bound by accessor construction time, you are doing something wrong anyway)

Pro binary tagging:
* No hidden implicit behavior with copy constructors; might actually be easier to understand. This implicit behavior has actually cost me some debugging time several times in the past, and is somewhat error-prone
* Implementation is actually structurally simpler than the old deferred pointer design with copy constructor
* Potential performance advantage for CPU backend inside kernels, because accessor copies can now just be memcopied because copy constructor is trivial. This is important because many OpenMP implementations seem to copy the kernel lambda by value into each loop chunk, so this can actually contribute to fork-join time
* Allows us to actually know which accessors are used inside kernels; this PR already introduces diagnostics for unused accessors, in the future we might even add logic to elide runtime dependencies from unused dependencies.

keryell:
I do not understand which dead-lock you are talking about. I assume that you will use `std::scoped_lock` or equivalent inside the scheduler à la https://github.com/triSYCL/triSYCL/blob/3b8879eefe8525f5168a52fccc9fe9fe338a884d/include/CL/sycl/command_group/detail/task.hpp#L221

By PIMPL I think that you can hide a lot under the hood, even

illuhad:
@keryell The basic process is as follows:

1. Accessor is constructed
2. Accessor is required and bound to a runtime requirement object. This initializes a mechanism that allows the runtime to inject the device pointer even after the accessor has been captured
3. `queue::submit()` returns, background runtime worker thread does schduling, lazy allocation etc and generates actual data pointer for the accessor
4. Kernel is submitted to backend and accessors initialize with the pointer obtained from the runtime
5. Kernel starts executing

If there is shared state, in order to guarantee that the mechanism to initialize accessor pointers works correctly, no other kernel submission processes can interfere from 2 to 4, as other kernels might use the same accessor which would then require modifications to the same shared state.
This leads to a deadlock as soon as the runtime decides to queue up kernel submissions and form batches instead of passing them to the scheduler right away.

In the initial example:
```
sycl::accessor myplaceholder{...};

queue_on_device1.submit([&](handler& cgh){
  // the accessor is locked here until the kernel starts executing
  cgh.require(myplaceholder);
  cgh.parallel_for(...);
});
// Assume runtime doesn't submit the first kernel yet
// because it wants to form a batch of multiple kernels,
// so it waits for more work...
queue_on_device2.submit([&](handler& cgh){
  // Requests the same lock as before and deadlocks because
  // the additional work that the runtime waits for cannot be submitted
  cgh.require(myplaceholder);
  cgh.parallel_for(...);
});
```

EDIT: This is probably not a problem if you only support a single device/backend (e.g. host) as the pointer would in typical implementations be the same anyway for all kernels using the same accessor.

PR #505, opened 2021-03-20
Add retrieving CPU name for OpenMP host device.

fodinabor:
This uses cpuid and thus does only work on x86 based CPUs.

Currently, we only provide the following information for the OpenMP backend:
```
Discovered devices from backend 'OpenMP': 
  device 0: 
    vendor: the hipSYCL project
    name: hipSYCL OpenMP host device
```

This PR extends this to include the processor's name on x86 platforms:
e.g.
```
Discovered devices from backend 'OpenMP': 
  device 0: 
    vendor: the hipSYCL project
    name: hipSYCL OpenMP host device (AMD FX(tm)-8350 Eight-Core Processor)
```
or e.g. `name: hipSYCL OpenMP host device (Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz)`

I was wondering whether it would also be a good idea to add the thread count from `omp_get_num_procs` and maybe shorten the `hipSYCL OpenMP...`

### DIFF
--- a/src/runtime/omp/omp_hardware_manager.cpp
+++ b/src/runtime/omp/omp_hardware_manager.cpp
@@ -26,8 +26,19 @@
  */
 
 
-#include <omp.h>
+#include <array>
+#include <cstring>
 #include <limits>
+#include <omp.h>
+#include <string>
+
+#if defined(__x86_64__) || defined(__i386__)
+#ifdef _WIN32
+  #include <intrin.h>
+#else
+  #include <cpuid.h>
+#endif
+#endif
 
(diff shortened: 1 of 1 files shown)


illuhad:
Hmm.. In general I think the term "host device" is the correct description because in general, we could be running on multiple CPUs in a multi-socket scenario. In an extreme case, those might even be different CPUs if we have a multi-socket system with different CPUs in it. An abstract description like "host device" therefore seems correct to me.

It might make sense to expose individual CPUs as subdevices of the complete host, but for this I suspect portable libraries like `hwloc` to collect topology information might be a better match?

PR #506, opened 2021-03-20
Allow selecting active backends via environment variable

fodinabor:
Usage is as similar to `HIPSYCL_TARGETS` as possible: e.g. `HIPSYCL_ACTIVE_BACKENDS="omp;cuda"` would activate OpenMP and CUDA. No device selection is implemented here, so don't use `:` ..
OpenMP will always be active, as we need it as a host device.

Also, only output warnings, if a backend library failed to load as that might be the case if no CUDA driver is installed, e.g.. An error is output and an exception is thrown if no CPU backend was loaded, as we always need one.

### DIFF
--- a/include/hipSYCL/runtime/settings.hpp
+++ b/include/hipSYCL/runtime/settings.hpp
@@ -45,7 +45,7 @@ enum class scheduler_type { direct };
 std::istream &operator>>(std::istream &istr, scheduler_type &out);
 std::istream &operator>>(std::istream &istr, std::vector<rt::backend_id> &out);
 
-enum class setting { debug_level, scheduler_type, active_backends };
+enum class setting { debug_level, scheduler_type, visibility_mask };
 
 template <setting S> struct setting_trait {};
 
@@ -57,7 +57,7 @@ template <setting S> struct setting_trait {};
 
 HIPSYCL_RT_MAKE_SETTING_TRAIT(setting::debug_level, "debug_level", int)
 HIPSYCL_RT_MAKE_SETTING_TRAIT(setting::scheduler_type, "rt_scheduler", scheduler_type)
-HIPSYCL_RT_MAKE_SETTING_TRAIT(setting::active_backends, "active_backends", std::vector<rt::backend_id>)
+HIPSYCL_RT_MAKE_SETTING_TRAIT(setting::visibility_mask, "visibility_mask", std::vector<rt::backend_id>)
 
 class settings
 {
@@ -77,8 +77,8 @@ public:
       return _debug_level.value();
     } else if constexpr (S == setting::scheduler_type) {
       return _scheduler_type.value();
-    } else if constexpr (S == setting::active_backends) {
-      return _active_backends.value();
+    } else if constexpr (S == setting::visibility_mask) {
+      return _visibility_mask.value();
     }
     return typename setting_trait<S>::type{};
   }
@@ -97,8 +97,8 @@ public:
     _scheduler_type =
         get_environment_variable_or_default<setting::scheduler_type>(
             scheduler_type::direct);
-    _active_backends =
-        get_environment_variable_or_default<setting::active_backends>(
+    _visibility_mask =
+        get_environment_variable_or_default<setting::visibility_mask>(
             std::vector<rt::backend_id>{});
   }
 
@@ -139,7 +139,7 @@ private:
 
   std::optional<int> _debug_level;
   std::optional<scheduler_type> _scheduler_type;
-  std::optional<std::vector<rt::backend_id>> _active_backends;
+  std::optional<std::vector<rt::backend_id>> _visibility_mask;
 };
 
 }
--- a/include/hipSYCL/runtime/settings.hpp
+++ b/include/hipSYCL/runtime/settings.hpp
@@ -28,6 +28,7 @@
 #ifndef HIPSYCL_RT_SETTINGS_HPP
 #define HIPSYCL_RT_SETTINGS_HPP
 
+#include "hipSYCL/runtime/device_id.hpp"
 #include <ios>
 #include <optional>
 #include <string>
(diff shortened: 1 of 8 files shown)


src/runtime/settings.cpp
```diff
@@ -0,0 +1,72 @@
...
+      while(std::getline(istream, name, ';')) {
+        if(name.empty())
+          continue;
+        std::transform(name.cbegin(), name.cend(), name.begin(), ::tolower);
+
+        if (name == "cuda" || name == "nvidia") {
+          out.push_back(rt::backend_id::cuda);
+        } else if (name == "rocm" || name == "amd" || name == "hip") {
+          out.push_back(rt::backend_id::hip);
+        } else if(name == "host" || "cpu" || "hipcpu" || "omp") {
```
illuhad: I think we can just use "cuda", "hip", "omp". The other synonyms are deprecated and also not accepted by the new `--hipsycl-targets` argument.

include/hipSYCL/runtime/settings.hpp
```diff
@@ -28,29 +28,24 @@
...
-    out = scheduler_type::direct;
-  else
-    istr.setstate(std::ios_base::failbit);
-  return istr;
-}
+std::istream &operator>>(std::istream &istr, scheduler_type &out);
+std::istream &operator>>(std::istream &istr, std::vector<rt::backend_id> &out);
 
-enum class setting { debug_level, scheduler_type };
+enum class setting { debug_level, scheduler_type, active_backends };
```
illuhad: I would suggest to use the term "visibility" in analogy to `CUDA_VISIBLE_DEVICES` and `HIP_VISIBLE_DEVICES`?

It would probably also be a good idea to structure it such that it could be extended to remove individual devices in the future, like `HIPSYCL_VISIBILITY_MASK` or `HIPSYCL_VISIBLE_DEVICES` or something like that. I think DPC++ has `SYCL_DEVICE_FILTER` (or similar) which could also serve as a template here

src/runtime/settings.cpp
```diff
@@ -0,0 +1,72 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "hipSYCL/common/debug.hpp"
+
+namespace hipsycl {
+  namespace rt {
+
+    std::istream &operator>>(std::istream &istr, scheduler_type &out) {
```
illuhad: We don't use indentation for namespaces

src/runtime/settings.cpp
```diff
@@ -0,0 +1,72 @@
...
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "hipSYCL/common/debug.hpp"
```
illuhad, line 28: Include `settings.hpp`?

src/runtime/backend.cpp
```diff
@@ -73,6 +74,15 @@ backend_manager::backend_manager()
...
     }
   });
+
+  if(std::none_of(_backends.cbegin(), _backends.cend(), 
+                  [](const std::unique_ptr<backend>& b){
+                    return b->get_hardware_platform() == hardware_platform::cpu;
+                    }))
+  {
+    HIPSYCL_DEBUG_ERROR << "No CPU backend has been loaded. Things will break." << std::endl;
+    throw sycl::runtime_error{"No CPU backend has been loaded."};
```
illuhad: We should `std::terminate()` imo. This is not a situation where we want the user to have the option to catch the exception and continue as the entire setup is broken at this point.

illuhad (commented):
This is a great feature! :)

include/hipSYCL/runtime/settings.hpp
fodinabor: Yeah, makes sense to use something more standard. I initially thought about the extension path would be to go to a `--hipsycl-targets` compatible syntax for device selection, except that it uses device ids instead of architectures of course, like `cuda:0,1;hip:4` or similar. But for now, thought it'd be nice to just be able to select backends (and devices could be _easily_ selected via the known backend variables, but then again, you could use those to basically disable the backend as well :D).

As of [DPC++ docs](https://github.com/intel/llvm/blob/sycl/sycl/doc/EnvironmentVariables.md) `SYCL_DEVICE_FILTER` uses `backend:device_type:device_num,backend:device_type:device_num,..` as syntax, which feels slightly overkill for hipSYCL as we currently only have one device_type per backend.. but OpenMP offload and HIP for FPGAs might become a thing, right?
What annoys me slightly with this is, that you'd have to specify the full triple multiple times to select multiple devices of the same backend, e.g. `cuda:gpu:0,cuda:gpu:1` and that they use `,` for triple splitting which is unintuitive given the `hipsycl-targets` syntax.. 

So we could of course support `SYCL_DEVICE_FILTER` for compatibility.. but have `HIPSYCL_VISIBLE_DEVICES` use the `hipsycl-targets` syntax. What are your thoughts, which one we should actually support?
And I guess I should probably try to revamp the PR to include preliminary support for device selection as well.

src/runtime/backend.cpp
fodinabor: This always assumes that SYCL is the only reason for the application to run... what if an application just tries to load hipSYCL and if something fails, would back off to another backend at runtime..?
I know that is pretty hypothetical and might get messy, but I'm not really a fan for a _library_ to decide about whether the actual application should terminate...

illuhad: Except it's not just a library, it's also a compiler stack that is used to compile the application. Since you cannot compile code with two compilers, I actually think a SYCL implementation is well within its rights to assume special privileges :) If other compiler runtimes (think libstdc++) detect a problem they probably also terminate.

Additionally it's not uncommon for HPC stacks to terminate in case of problems that make recovering impossible, e.g. MPI implementations also do this.

If you have borked your setup so much that the hipSYCL installation is broken there's something fundamentally wrong with your build environment.

include/hipSYCL/runtime/settings.hpp
fodinabor: Decided to use `HIPSYCL_VISIBILITY_MASK` with `hipsycl-targets` like syntax for now, with the expectation to implement compatibility options, when we can properly support device selection.

illuhad (approved):

illuhad:
Thanks!

PR #507, opened 2021-03-21
Fix ambiguity of atomic accessor operator[].

fodinabor:
As reported in https://github.com/illuhad/hipSYCL/issues/380#issuecomment-803572968 [1], the `operator[]` is ambiguous, if using an atomic accessor.

### DIFF
--- a/include/hipSYCL/sycl/libkernel/accessor.hpp
+++ b/include/hipSYCL/sycl/libkernel/accessor.hpp
@@ -168,18 +168,34 @@ public:
     return create_next_proxy(index);
   }
 
-
   template <int D = dimensions,
             int C = current_dimension,
-            std::enable_if_t<can_invoke_access(C, D), bool> = true>
+            access_mode M = accessmode,
+            std::enable_if_t<can_invoke_access(C, D) && (M != access_mode::atomic), bool> = true>
   HIPSYCL_UNIVERSAL_TARGET
   reference operator[](size_t index) const {
     return invoke_value_access(index);
   }
 
+  template <int D = dimensions,
+            int C = current_dimension,
+            access_mode M = accessmode,
+            std::enable_if_t<can_invoke_access(C, D) && (M == access_mode::atomic), bool> = true>
+  [[deprecated("Atomic accessors are deprecated as of SYCL 2020")]] HIPSYCL_UNIVERSAL_TARGET
+  auto operator[](size_t index) const {
+    return invoke_atomic_value_access(index);
+  }
+
 private:
   HIPSYCL_UNIVERSAL_TARGET
   reference invoke_value_access(size_t index) const {
+    // Set the last index
+    _access_id[dimensions - 1] = index;
+    return (*_original_accessor)[_access_id];
+  }
+
+  HIPSYCL_UNIVERSAL_TARGET
+  auto invoke_atomic_value_access(size_t index) const {
     // Set the last index
     _access_id[dimensions - 1] = index;
     return (*_original_accessor)[_access_id];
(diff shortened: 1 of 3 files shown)


illuhad:
Actually we implement the interface exactly as described by the SYCL 2020 spec :)
https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#_interface_for_buffer_command_accessors

It seems that there is now an oversight in the spec w.r.t. to the conditions in which to enable `operator[]`, presumably because `atomic` is completely deprecated...

tests/sycl/accessor.cpp
```diff
@@ -161,6 +161,17 @@ BOOST_AUTO_TEST_CASE(accessor_api) {
...
+  queue.submit([&](s::handler& cgh) {
+    run_test([&](auto buf, auto... args) {
+      return buf.template get_access<s::access::mode::atomic>(cgh, args...);
+    });
+    // mostly compilation test
+    auto atomicAcc = buf_a.template get_access<s::access::mode::atomic>(cgh);
+    cgh.single_task<class accessor_api_atomic_device_accessors>([atomicAcc](){
+      atomicAcc[0].exchange(0);
+    });
+  });
```
illuhad, line 182: Philosophical question: Do we actually need to/Should we test deprecated features?

include/hipSYCL/sycl/libkernel/accessor.hpp
```diff
@@ -537,15 +537,17 @@ class accessor : public detail::accessor_base<std::remove_const_t<dataT>> {
...
   HIPSYCL_UNIVERSAL_TARGET
   reference operator[](id<dimensions> index) const
   {
     return (this->_ptr.get())[detail::linear_id<dimensions>::get(index, _buffer_range)];
   }
 
   template<int D = dimensions,
-            std::enable_if_t<(D == 1), bool> = true>
+            access::mode M = accessmode,
+            std::enable_if_t<(D == 1) && (M != access::mode::atomic), bool> = true>
```
illuhad, line 566: Wouldn't local accessor be affected as well?

fodinabor: only if they had atomic support in the first case (which they [deprecatedly should have as of SYCL 2020](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#_accessor_specialization_with_targetlocal))

tests/sycl/accessor.cpp
fodinabor: I think as long as we did not remove them, we should test it, as we still provide it and the user will expect it to work as long as it's there.

fodinabor (commented):
The spec seems to be undecided how to deal with this..
```cpp
/// note the missing accessMode check
  /* Available only when: (dimensions > 0) */
  reference operator[](id<dimensions> index) const;

/// note the missing accessMode check
  /* Available only when: (dimensions > 1) */
  __unspecified__ &operator[](size_t index) const;

/// note the accessMode check
  /* Available only when: (accessMode != access_mode::atomic && dimensions == 1) */
  reference operator[](size_t index) const;

  /* Deprecated
  Available only when: (accessMode == access_mode::atomic && dimensions ==  0) */
  operator cl::sycl::atomic<dataT, access::address_space::global_space> () const;

  /* Deprecated
  Available only when: (accessMode == access_mode::atomic && dimensions == 1) */
  cl::sycl::atomic<dataT, access::address_space::global_space> operator[](
    id<dimensions> index) const;
```

Just had another look.. seems we're missing atomic support for multi-dimensional accessors in `subscript_proxy` in general..

Thus.. do we plan on keeping atomics around? If so, I will add the missing parts for local and subscript proxy

fodinabor:
@hipsycl-ci-bot test this now as well, please.

illuhad:
@fodinabor Yes, I think when rewriting the spec we didn't really think about if/how things should play with deprecated stuff in this particular case...

I can get behind continuing to support for now what we already have, but not sure if it's a good idea to complicate `subscript_proxy` just for some deprecated feature...

EDIT: Okay, I see you already did the `subscript_proxy`. Nevermind then, so be it ;)

illuhad (approved):
And I was so happy that I could remove all those checks when I changed the interfaces to SYCL 2020 definitions... Can't wait for atomic accessors to go away for good :)

PR #508, opened 2021-03-21
Add missing logical_vector_op_result<char> specialization.

fodinabor:
Which is not a typedef of `int8_t` for some reason (https://godbolt.org/z/fGEqcr).
Reported by @zjin-lcf in https://github.com/illuhad/hipSYCL/issues/380#issuecomment-803572968 [4]

### DIFF
--- a/include/hipSYCL/sycl/libkernel/vec.hpp
+++ b/include/hipSYCL/sycl/libkernel/vec.hpp
@@ -151,6 +151,9 @@ template<class T>
 struct logical_vector_op_result
 {};
 
+template<> struct logical_vector_op_result<char>
+{ using type = int8_t; };
+
 template<> struct logical_vector_op_result<int8_t>
 { using type = int8_t; };
 


illuhad:
Is unsigned char also affected?

fodinabor:
Nope, see the godbolt link. `uint8_t` is an alias for `unsigned char` but `char`  is not an alias for `int8_t`, as `char` has no defined signage.. `signed char` again is an alias for `int8_t`

illuhad (approved):

illuhad:
Ah, yeah, I remember the `char` weirdness in C++...

keryell:
> Ah, yeah, I remember the `char` weirdness in C++...

You mean in C?

illuhad:
I meant `char`, `signed char` and `unsigned char` being three distinct types, which AFAIK is also the case in C++, or was that changed?
https://stackoverflow.com/questions/16503373/difference-between-char-and-signed-char-in-c

keryell:
What I mean is that most of the weirdness of C++ is because it is kindly compatible with C. :-)

PR #510, opened 2021-03-23
[HIP] Use new HIP launch API and define __HIP_ROCclr__

illuhad:
* Use new HIP push call configuration kernel launch API
* Compile with `__HIP_ROCclr__`

Aligns `syclcc` compilation with current `hipcc` compilation.

Draft because first needs AMD runtime testing

### DIFF
--- a/include/hipSYCL/glue/generic/hiplike/clang.hpp
+++ b/include/hipSYCL/glue/generic/hiplike/clang.hpp
@@ -48,11 +48,7 @@ extern "C" unsigned __cudaPushCallConfiguration(dim3 gridDim, dim3 blockDim,
                                                 size_t sharedMem,
                                                 void *stream);
 
-extern "C" unsigned __cudaPopCallConfiguration();
-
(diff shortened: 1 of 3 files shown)


include/hipSYCL/glue/generic/hiplike/clang.hpp
```diff
@@ -54,6 +54,10 @@ extern "C" unsigned __cudaPopCallConfiguration();
 
 hipError_t hipConfigureCall(dim3 gridDim, dim3 blockDim, size_t sharedMem, hipStream_t stream);
```
fodinabor: This could be removed then, I guess?
And while at it, I think `__cudaPopCallConfiguration` isn't used anywhere anymore as well..?

illuhad: Done ;)

illuhad:
@hipSYCL-ci-bot test please

PR #514, opened 2021-03-27
Implement queue::wait() using node group properties, optimize wait() in dag_submitted_ops

illuhad:
* Introduce `node_group` execution hint in runtime
* Implement `queue::wait()` by generating a unique node group id for each queue, at submit attaching `node_group(id)` execution hint to nodes, and then only waiting for nodes with this node group id in `wait()`
* Optimize `dag_submitted_ops::wait()` by not locking until all waits have completed, instead only lock until a copy of the current node list could be created.
* Fix minor bugs in `execution_hints`

**This can boost performance considerably when working with multiple queues** because `queue::wait()` now waits selectively for nodes and also doesn't block runtime operations.

### DIFF
--- a/include/hipSYCL/runtime/dag_submitted_ops.hpp
+++ b/include/hipSYCL/runtime/dag_submitted_ops.hpp
@@ -44,7 +44,7 @@ public:
   void update_with_submission(dag_node_ptr single_node);
   
   void wait_for_all();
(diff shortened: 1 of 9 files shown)


illuhad:
@hipSYCL-ci-bot test please.

PR #517, opened 2021-03-29
Add buffer construction on top of USM pointers, add explicit buffer policies

### LINKED ISSUE #276: [Extension] Explicit buffer policies
Based on user feedback, I believe we should try to allow users to better specify intent when constructing a `buffer`. At the moment, `buffer` can have a lot of varying behaviors depending on (seemingly) subtle differences in initialization. This makes it error-prone to use and difficult to understand, especially for people who are new to SYCL. Problems I have noticed include
* users having wrong assumptions about writeback behavior
* users having wrong assumptions whether it acts as a "view" for a supplied host pointer or just copies the content to an internal buffer
* It's unclear for most users what the most performant way of initialization is
* Synchronization requirement in the destructor makes it potentially inefficient to destroy multiple buffers at the end of one scope since each one will have to do separate synchronization
* etc

I think we should give users a method to explicitly specify their desired behavior and use cases. The rationale in SYCL behind not having multiple types of `buffer` classes to distinguish different behaviors is to make it easy to manage buffers (e.g. store many buffers in a vector). But we can still have this advantage if we
* introduce buffer properties that explicitly require a certain behavior
* add classes derived from `buffer` with clear names e.g. `view` that just set the behavior policy properties of the base class

At the moment I'm thinking of the following properties:
* ` use_external_storage(true)` and ` use_external_storage(false)` -- if false, copies any input pointers to internal storage instead of operating directly on them.
* `writes_back(true)` and `writes_back(false)` -- if true, a writeback operation will be submitted in the destructor. Note: `writes_back(true)` does not guarantee that a write back will be *executed*, only that it will be *submitted*. If it turns out that a write back is unnecessary because data is already up-to-date, the runtime would still optimize the operation away.
* `destructor_waits(true)` and `destructor_waits(false)` -- if true, the destructor waits until all work on that buffer has completed. Note: In future hipSYCL with the new runtime, the runtime manages the actual data allocations of a `buffer`. Any operations submitted to the runtime working with this `buffer`'s allocations will still complete successfully even if the user-facing `buffer` object is gone.

This would allow us to define the following buffer types:
| buffer type| `use_external_storage`| `writes_back` | `destructor_waits` | Comments |
| -------------- | --------- | ---------- | --------- | -------- |
| (makes no sense (?)) | false| true | true | writing back to internal storage makes no sense(*) |
| (makes no sense (?)) | false | true| false | writing back to internal storage makes no sense(*) |
| `sync_buffer` | false | false | true | Similar to current `buffer<T>{sycl::range(...)}`|
| `async_buffer` | false | false| false | Useful for performance optimization |
| `sync_view` | true | true | true | Similar to current `buffer<T>{T*, sycl::range(...)}`|
| `async_view` | true | true | false | Needs mechanism to provide event handle such that user can wait for async writeback (?) |
| `sync_explicit_view` | true | false | true | Requires explicit construction of host accessor to move data back to host |
| `async_explicit_view` | true | false | false | Requires explicit construction of host accessor to move data back to host |

As always, feedback welcome.

(*) It might make sense if an external writeback location is explicitly set later using `set_final_data()`

illuhad:
This adds two major extensions:
* Construct buffers on top of arbitrary USM pointers (in addition to being able to extract USM pointers from buffers which was already possible previously)
* Construct buffers using explicit buffer policies using convenient factory functions. Add support for buffers that do not block in the destructor. For example `auto b = sycl::make_async_writeback_view(ptr, range)` operates directly on the host pointer `ptr`, does not block in the destructor and performs a writeback.

Copy-pasted from documentation contained in this PR:

# hipSYCL buffer-USM interoperability

hipSYCL supports interoperability between `buffer` objects and USM pointers. All `buffer` memory allocations are USM pointers internally that can be extracted and be used with USM operations.
Similarly, `buffer` objects can be constructed on top of existing USM pointers.

hipSYCL follows its own [specification](runtime-spec.md) for the hipSYCL buffer-accessor model. Refer to this document to understand the memory management and allocation behavior of hipSYCL buffer objects. Using buffer-USM interoperability without a solid understanding of hipSYCL's buffer model is not recommended.

## Buffer introspection

[already merged and available in hipSYCL]

## Constructing buffer objects on top of USM pointers

hipSYCL can also construct buffers on top of existing USM pointers.

### API reference

```c++

namespace sycl {
namespace buffer_allocation {

/// A type based on buffer_allocation::descriptor<T> that stores additionally
/// information about the data state (outdated/current) of the data referenced
/// by the descriptor
template<class T>
using tracked_descriptor = __unspecified__;

enum class management_mode {
  owning,
  non_owning
};

static constexpr bool take_ownership = management_mode::owning;
static constexpr bool no_ownership = management_mode::non_owning;

/// Construct descriptor for a data region with initially outdated content.
/// When first accessed, the runtime will emit data transfers to update
/// date content.
tracked_descriptor<T> empty_view(T *ptr, device dev,
                                 management_mode m = no_ownership);

/// Construct descriptor for a data region with initially recent content.
/// When first accessed, the runtime will not emit data transfers to
/// update data content. Instead, the allocation may be used as a
/// potential data source to update other outdated allocations from.
tracked_descriptor<T> view(T *ptr, device dev, management_mode m = no_ownership);

}
}

template <typename T, int dimensions,
          typename AllocatorT>
class buffer {
public:
  /// Construct buffer on top of existing USM pointers with given range.
  /// Will not write back at destruction unless set_final_data() is used.
  /// Modifying the USM pointers externally during the duration of the buffer lifetime
  /// should be avoided unless the user has expert knowledge of hipSYCL's data
  /// state tracking mechanisms.
  buffer(
    const std::vector<buffer_allocation::tracked_descriptor<T>>& input_allocations,
    const range<dimensions>& r,
    const property_list& prop_list = {});
  
  buffer(
    const std::vector<buffer_allocation::tracked_descriptor<T>>& input_allocations,
    const range<dimensions>& r,
    AllocatorT allocator,
    const property_list& prop_list = {});
};

```

## Example code

```c++

sycl::queue q;
sycl::range size{1024};

int* alloc1 = sycl::malloc_shared<int>(size.size(), q);
int* alloc2 = sycl::malloc_shared<int>(size.size(), q);

{
  // Construct buffer on top of alloc1. Use empty_view() because alloc1
  // does not old live data.
  // Note: To give ownership over the pointer to the pointer, pass
  // the optional argument take_ownership to empty_view()
  sycl::buffer<int> b1{
      {sycl::buffer_allocation::empty_view(alloc1, q.get_device())}, size};

  // The buffer now uses the given allocation for all operations on
  // q.get_device()
  assert(b1.has_allocation(q.get_device()));
  assert(b1.get_pointer(q.get_device()) == alloc1);

  // Iterate over all allocations of the buffer. We did not grant
  // ownership of the pointer to the buffer. This is checked here.
  b1.for_each_allocation([&](const auto& alloc){
    if(alloc.ptr == alloc1){
      assert(!alloc.is_owned);
    }
  });

  // We can use the buffer just as usual
  q.submit([&](sycl::handler& cgh){
    sycl::accessor<int> acc{b1, cgh};

    cgh.parallel_for(size, [=](sycl::id<1> idx){
      acc[idx] = idx.get(0);
    });
  });
}

{
  // Construct another buffer - this time we use view instead of empty_view
  // because the pointer now holds up-to-date data
  sycl::buffer<int> b2{
      {sycl::buffer_allocation::view(alloc1, q.get_device())}, size};
  
  // Can again use the buffer just as usual
  q.submit([&](sycl::handler& cgh){
    sycl::accessor<int> acc{b2, cgh};

    cgh.parallel_for(size, [=](sycl::id<1> idx){
      alloc2[idx.get(0)] = acc[idx];
    });
  });

  sycl::host_accessor<int> hacc{b2};
  for(int i = 0; i < size.get(0); ++i){
    assert(hacc[i] == i);
  }  
}

sycl::free(alloc1, q);
sycl::free(alloc2, q);

```

# Explicit buffer policies

`sycl::buffer` objects can have a variety of different behaviors, sometimes depending on subtle differences in how they were constructed.
For example, whether a writeback occurs depends on the combination of constness of the buffer's pointer argument, constness of buffer type and potentially buffer properties.
This can potentially make it error-prone to use. Errors can either lead to undesired behavior affecting correctness (e.g. no writeback when it *was* expected) or performance issues (e.g. writeback when it *was not* expected).

hipSYCL therefore introduces explicit buffer policies that allow the programmer to explicitly specifiy the desired buffer behavior. Additionally, hipSYCL adds certain buffer behaviors not covered by the SYCL specification.

Buffer policies can be enabled by passing buffer policy property to the `property_list` of the `buffer` constructor.

## Available buffer policies

| Policy property | Behavior if set to true | Behavior if set to false |
| ------ | ----- | ----- |
| `property::buffer::hipSYCL_buffer_uses_external_storage(bool)` | Instruct the buffer to act as view and operate directly on provided input pointers. Pointer should not be used by the user during the lifetime of the buffer. | Only use supplied pointer as input argument and copy data to internal storage. Pointer can be used by used as desired after buffer construction. |
| `property::buffer::hipSYCL_buffer_writes_back(bool)` | Submit writeback to supplied user pointer at buffer destruction. If no data needs to be transferred, the writeback may be optimized away by the runtime. | Do not submit writeback in destructor. |
| `property::buffer::hipSYCL_buffer_destructor_blocks(bool)` | buffer destructor blocks until all tasks operating on the buffer have completed. | buffer destructor does not block. The user is responsible for making sure that all kernels and other operations working on the buffer are synchronized, e.g. using `queue::wait()` or `event::wait()`. It is allowed for the buffer to be destroyed before operations have completed because operations will retain references to the buffer's data storage. |

To describe buffers and their behavior better, hipSYCL adopts the following terminology:

| | Destructor blocks? | Writes back ? | Uses external storage? |
| ----- | ----- | ----- | ----- |
| yes | `sync_` | `_writeback_` | `view` |
| no  | `async_` | - | `buffer` |

For example, a `sync_view` blocks in the destructor, does not write back and operates directly on provided input pointers. An `async_writeback_view` does not block in the destructor, writes data back, and operates directly on provided input pointers.

These are logical data types, they are still expressed using regular `buffer<T>` objects, but they have dedicated factory functions to simplify construction.

## API reference

```c++
namespace sycl {
namespace property::buffer {

struct hipSYCL_buffer_uses_external_storage {
  hipSYCL_buffer_uses_external_storage(bool toggle);
};

struct hipSYCL_buffer_writes_back {
  hipSYCL_buffer_writes_back(bool toggle);
};

struct hipSYCL_buffer_destructor_blocks {
  hipSYCL_buffer_destructor_blocks(bool toggle);
};

} // namespace property buffer


/// Only uses internal storage, no writeback, blocking destructor
template <class T, int Dim>
buffer<T, Dim, buffer_allocator<std::remove_const_t<T>>>
make_sync_buffer(sycl::range<Dim> r);

/// Only uses internal storage, no writeback, blocking destructor
/// Data pointed to by ptr is copied to internal storage.
template <class T, int Dim>
buffer<T, Dim, buffer_allocator<std::remove_const_t<T>>>
make_sync_buffer(const T* ptr, sycl::range<Dim> r);

/// Only internal storage, no writeback, non-blocking destructor
template <class T, int Dim>
buffer<T, Dim, buffer_allocator<std::remove_const_t<T>>>
make_async_buffer(sycl::range<Dim> r);

/// Only internal storage, no writeback, non-blocking destructor
/// Data pointed to by ptr is copied to internal storage.
template <class T, int Dim>
buffer<T, Dim, buffer_allocator<std::remove_const_t<T>>>
make_async_buffer(const T* ptr, sycl::range<Dim> r);

/// Uses provided storage, writes back, blocking destructor
/// Directly operates on host_view_ptr.
template <class T, int Dim>
buffer<T, Dim, buffer_allocator<std::remove_const_t<T>>>
make_sync_writeback_view(T* host_view_ptr, sycl::range<Dim> r);

/// Uses provided storage, writes back, non-blocking destructor
/// Directly operates on host_view_ptr.
/// The queue can be used by the user to wait for the writeback to complete.
template <class T, int Dim>
buffer<T, Dim, buffer_allocator<std::remove_const_t<T>>>
make_async_writeback_view(T* host_view_ptr, sycl::range<Dim> r, const sycl::queue& q);

/// Uses provided storage, writes back, non-blocking destructor
/// Directly operates on host_view_ptr.
/// Note: Because the writeback is asynchronous, there is no way
/// for the user to wait for its completion. In most cases, users
/// should use make_async_writeback_view(T*, sycl::range<Dim>, const sycl::queue&)
/// instead.
template <class T, int Dim>
buffer<T, Dim, buffer_allocator<std::remove_const_t<T>>>
make_async_writeback_view(T* host_view_ptr, sycl::range<Dim> r);

/// Uses provided storage, does not write back, blocking destructor
/// Directly operates on host_view_ptr.
template <class T, int Dim>
buffer<T, Dim, buffer_allocator<std::remove_const_t<T>>>
make_sync_view(T* host_view_ptr, sycl::range<Dim> r);

/// Uses provided storage, does not write back, non-blocking destructor
/// Directly operates on host_view_ptr.
template <class T, int Dim>
buffer<T, Dim, buffer_allocator<std::remove_const_t<T>>>
make_async_view(T* host_view_ptr, sycl::range<Dim> r);

/// USM interoperability types

/// Uses provided storage, does not write back, blocking
template <class T, int Dim>
buffer<T, Dim, buffer_allocator<std::remove_const_t<T>>>
make_sync_view(
    const std::vector<buffer_allocation::tracked_descriptor<T>>
        &input_allocations,
    sycl::range<Dim> r);

/// Uses provided storage, does not write back, non-blocking
template <class T, int Dim>
buffer<T, Dim, buffer_allocator<std::remove_const_t<T>>>
make_async_view(
    const std::vector<buffer_allocation::tracked_descriptor<T>>
        &input_allocations,
    sycl::range<Dim> r);

/// same as make_sync_view(
//     std::vector<buffer_allocation::tracked_descriptor<T>>&, sycl::range<Dim>)
template <class T, int Dim>
buffer<T, Dim, buffer_allocator<std::remove_const_t<T>>>
make_sync_usm_view(const std::vector<buffer_allocation::tracked_descriptor<T>>
                       &input_allocations,
                   sycl::range<Dim> r);

/// same as make_async_view(
//     std::vector<buffer_allocation::tracked_descriptor<T>>&, sycl::range<Dim>)
template <class T, int Dim>
buffer<T, Dim, buffer_allocator<std::remove_const_t<T>>>
make_async_usm_view(const std::vector<buffer_allocation::tracked_descriptor<T>>
                       &input_allocations,
                    sycl::range<Dim> r);


}

```

## Example

```c++

int* data1 = ...;
int* data2 = ...;
std::size_t size = ...;

sycl::queue q;

// Because these are async writeback types, need to provide a queue
// so we can wait for the writeback
{
  auto v1 = sycl::make_async_writeback_view(data1, sycl::range{size}, q);
  auto v2 = sycl::make_async_writeback_view(data2, sycl::range{size}, q);

  q.submit([&](sycl::handler& cgh){
    sycl::accessor<int> a1 {v1, cgh};
    sycl::accessor<int> a2 {v2, cgh};

    cgh.parallel_for(size, [=](sycl::id<1> idx){
      a1[idx] += a2[idx];
    });
  });
} // Asynchronous writebacks triggered here

// At some point, wait for writebacks to complete
q.wait();

```

Resolves #276


illuhad:
@hipSYCL-ci-bot test please

PR #529, opened 2021-04-19
Implement basic vendor_id 

sbalint98:
The vendor id, as in dpc++, is assumed to be the PCI address vendor-prefix.
It is furthermore assumed that the HIP backend runs on AMD, and the CUDA backend runs on NVIDIA devices.

### DIFF
--- a/include/hipSYCL/runtime/hardware.hpp
+++ b/include/hipSYCL/runtime/hardware.hpp
@@ -95,7 +95,9 @@ enum class device_uint_property {
   local_mem_size,
 
   printf_buffer_size,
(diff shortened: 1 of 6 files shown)


fodinabor:
I think we should also handle the property in the OMP backend, as otherwise one runs into an `assert` .. not sure what to report though..

illuhad:
Yeah, we should also return something on CPU. It only needs to be (hopefully) unique for the backend, so maybe just return `numeric_limits<T>::max()` as this is unlikely to be used by any other vendor?

illuhad (approved):

PR #535, opened 2021-04-29
Fix issues on MacOS and enable running them on CI

### LINKED ISSUE #437: Error building hipSYCL applications on macOS / Apple Clang
A few minor issues in the CMake setup and `syclcc-clang` prevent hipSYCL applications from compiling on macOS with Apple Clang.

* The macOS linker expects a different syntax for `-rpath`: Instead of `-rpath=/lib`, it requires `-rpath /lib`. That syntax should work on other systems as well. Patches are required in `bin/syclcc-clang` and the root `CMakeLists.txt`
* Since OpenMP is enabled through `-Xclang -fopenmp` on Mac, the linker invocation is missing the dependency to `-lomp`. This flag can simply be added to `OMP_LINK_LINE` in CMakeLists.txt.
* `syclcc-clang` hardcodes link dependencies to `-lboost_context` and `-lboost_fiber`, but the libraries might have different names depending on the installation. For the installation in question, they have a `-mt` suffix (`-lboost_context-mt` and `-lboost_fiber-mt`).

The last bullet point is the reason I'm opening an issue rather than a pull request. CMake's `FindBoost` module knows how to discover these library file names, but there seems to be no obvious way to convert them into literal `-L`/`-l` flags that can be used by `syclcc-clang`. Since CMake already knows about the dependencies, the most viable option might be to move the Boost dependencies outside the compiler wrapper so that they can be specified explicitly in CMake or a manual syclcc invocation. 

Compiler info:
```
Apple clang version 12.0.0 (clang-1200.0.32.27)
Target: x86_64-apple-darwin20.1.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

Linker info:
```
@(#)PROGRAM:ld  PROJECT:ld64-609.7
BUILD 18:10:07 Oct 19 2020
configured to support archs: armv6 armv7 armv7s arm64 arm64e arm64_32 i386 x86_64 x86_64h armv6m armv7k armv7m armv7em
LTO support using: LLVM version 12.0.0, (clang-1200.0.32.27) (static support for 27, runtime is 27)
TAPI support using: Apple TAPI version 12.0.0 (tapi-1200.0.23.4)
```

cc @almightyvats

### LINKED ISSUE #534: macOS shared_lib_extension wrong
**Bug summary**
On macOS, the backend tries to load `librt-backend-omp.so` which does not exist. 

**Fix**
Edit line 196 in src/runtime/backend_loader.cpp
```
#ifndef _WIN32
#ifdef __APPLE__
  std::string shared_lib_extension = ".dylib";
  #else
  std::string shared_lib_extension = ".so";
  #endif
#else
  std::string shared_lib_extension = ".dll";
#endif
```
solves it!

fodinabor:
The backend loader now looks for `.dylib` on MacOS.
MacOS gets its own link line with OpenMP and Boost libs detected via CMake and fixes `rpath` argument.
The `unsigned long` issue with `sycl::vec` as on Windows has been taken care of.
Enables building and running tests on MacOS.

This fixes #437 and fixes #534.


fodinabor:

illuhad (approved):

PR #540, opened 2021-05-02
Some bugfixes and additions required to run most of the SYCL-BLAS tests

fodinabor:
This adds some missing builtins (including mul24 #380, although I wasn't willing to make up a concise test for this).

`init_from` only allowed an allocator of the same type.

`sycl::vec`'s load and store have been (naively) implemented.

`multi_ptr`'s assignment operators missed a return statement 😡

With this and the proposed changes in the SYCL-BLAS repo (+ build system adoption in my fork), one can now use SYCL-BLAS with hipSYCL (at least on GPU all unit tests pass, on CPU some tests fail due to illegal memory access in the kernels, which I'd expect to be SYCL-BLAS issues..). For the benchmarks to run, profiling is still missing.
The kernel configurations in my fork are also completely un-tuned..
See my fork for the build system: https://github.com/fodinabor/sycl-blas/tree/feature/hipsycl-build

PR #541, opened 2021-05-02
Fix missing return statements in multi_ptr assignment operators

illuhad:
Add missing return statements for multi_ptr assignment operators.


illuhad:
Superseded by #540

PR #545, opened 2021-05-06
Redesign builtin definition/reimplement builtins

illuhad:
* Unify builtin definition - we previously had separate files and definition mechanisms for math functions, common functions, integer functions etc. This PR unifies all in a single framework
* Builtins are now exposed in a single file `libkernel/builtins.hpp`. This file implements broadcasting across vector types and makes sure exactly the right overloads exists. It then invokes scalar `__hipsycl_<builtin>` functions provided by backends.
* Individual builtin definition files for each backend in `libkernel/<backend>/builtins.hpp`. This allows us to optimize builtins for each backend separately with more freedom, without having to worry about breaking other backends.
* Implements builtins using overloads instead of templates as much as possible, to get more consistent behavior w.r.t overload resolution compared to STL. We may have to also introduce template variants once we add `marray` support.
* Also implements some builtins that were not implemented previously (e.g. adds interface for native and half precision math functions, even though they still call the regular functions for the most part)


illuhad:
@hipSYCL-ci-bot test.

include/hipSYCL/sycl/libkernel/vec.hpp
```diff
@@ -154,44 +154,42 @@ struct logical_vector_op_result
...
 
-template<> struct logical_vector_op_result<uint32_t>
+template<> struct logical_vector_op_result<unsigned>
 { using type = int32_t; };
 
-template<> struct logical_vector_op_result<int32_t>
+template<> struct logical_vector_op_result<int>
+{ using type = int32_t; };
+
+template<> struct logical_vector_op_result<long>
```
fodinabor: this will be truncating on Linux, where long refers to 64bit... we could use `ULONG_WIDTH` macros to check this independently of the OSs..

include/hipSYCL/sycl/libkernel/spirv/builtins.hpp
```diff
@@ -0,0 +1,475 @@
...
+
+HIPSYCL_DEFINE_SPIRV_BUILTIN(log)
+HIPSYCL_DEFINE_SPIRV_BUILTIN(log2)
+HIPSYCL_DEFINE_SPIRV_BUILTIN(log10)
+HIPSYCL_DEFINE_SPIRV_BUILTIN(log1p)
+HIPSYCL_DEFINE_SPIRV_BUILTIN(logb)
+
+template<class T>
+HIPSYCL_BUILTIN T __hipsycl_mad(T x, T y, T z) noexcept {
+  return __hipsycl_fma(x, y, z);
```
fodinabor, line 147: spirv should have mad? https://github.com/intel/llvm/blob/54ede407edeb93b7e9334d1e725f48bf981b2965/sycl/include/CL/sycl/detail/builtins.hpp#L152

fodinabor (commented):
Pretty extensive work... nice :)

include/hipSYCL/sycl/libkernel/vec.hpp
illuhad: Ah, thanks, it should return int64_t on Linux. We can just use `sizeof(long)` :) In practice there won't really be truncation because this is only used for the return type of boolean operators, so will store 0 or 1.

include/hipSYCL/sycl/libkernel/spirv/builtins.hpp
illuhad: I think on SPIR-V we could probably map *all* of the SYCL builtins to SPIRV ones because the SYCL kernel library is based on the OpenCL kernel library. 
I have based the initial reimplementation of the SPIR-V builtins on what was previously mapped to SPIR-V builtins/what other backends map to builtins to get this PR going quicker. We can probably optimize a lot in the future :)

include/hipSYCL/sycl/libkernel/vec.hpp
illuhad: It should now automagically do the correct thing for any type, and do so with less code :)

PR #547, opened 2021-05-07
Add documentation for HIPSYCL_TARGETS.

fodinabor:
I think this was not documented at all outside of code.. so I added it to the environment variables and just mention that it can also be used as arguments directly..

### DIFF
--- a/doc/using-hipsycl.md
+++ b/doc/using-hipsycl.md
@@ -1,0 +1,168 @@
+# Using hipSYCL in projects
+It is recommended to use the CMake integration for larger projects.
+
+## Compiling with syclcc
+`syclcc` is the compilation driver used by hipSYCL to build the final compiler invocations.
+After installing hipSYCL, it can be used as a standalone tool to manually build single files or integrated in other build systems than CMake.
+Exemplary, building the nbody example while targeting CPU and CUDA backends could be done using `syclcc bruteforce_nbody.cpp -O3 --hipsycl-targets="omp;cuda:sm_61"`.
+
+Following is the full excerpt from `syclcc --help`. Note the options can also be set via environment variables or corresponding CMake options.
+```
+syclcc [hipSYCL compilation driver], Copyright (C) 2018-2020 Aksel Alpay
+Usage: syclcc <options>
+
+Options are:
+--hipsycl-platform=<value>
+  [can also be set with environment variable: HIPSYCL_PLATFORM=<value>]
+  [default value provided by field 'default-platform' in /install/path/etc/hipSYCL/syclcc.json.]
+  [current value: NOT SET]
+  (deprecated) The platform that hipSYCL should target. Valid values:
+    * cuda: Target NVIDIA CUDA GPUs
+    * rocm: Target AMD GPUs running on the ROCm platform
+    * cpu: Target only CPUs
+
+--hipsycl-clang=<value>
+  [can also be set with environment variable: HIPSYCL_CLANG=<value>]
+  [default value provided by field 'default-clang' in /install/path//etc/hipSYCL/syclcc.json.]
+  [current value: NOT SET]
+  The path to the clang executable that should be used for compilation
+    (Note: *must* be compatible with the clang version that the 
+     hipSYCL clang plugin was compiled against!)
+
+--hipsycl-cuda-path=<value>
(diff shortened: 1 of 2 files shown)


illuhad:
It is documented in `syclcc --help`. If we add this, would we also need to document all the other `syclcc` flags (all of which can also be set using environment variables)?

DieGoldeneEnte:
Maybe another approach would be to make the
> See syclcc --help for a comprehensive list of options.

part in "Installing and using hipSYCL" more visible. At the moment it's easy to miss especially when looking for cmake variables (at least I missed it multiple times and ended up reading the CMakeLists.txt file).
I think most of the options can also be directly passed to  cmake, this could also be highlighted.

fodinabor:
mhh.. it just really sucks to have to write in some docs out there: hey and btw. I'm not gonna explain to you how to use this, but please call this thing that you might not have in your path yet...
so maybe another approach would be to do it similar to what LLVM does with their docs: just have a place where we paste the output of `syclcc --help`?

DieGoldeneEnte:
That would improve the documentation, but only if it is updated (can this be automated?).

illuhad:
> another approach would be to do it similar to what LLVM does with their docs: just have a place where we paste the output of syclcc --help?

That sounds like a good strategy.

> That would improve the documentation, but only if it is updated (can this be automated?).

Anything is possible... we could generate `syclcc --help` in CI, and then somehow pipeline the output into some documentation, but I'm not sure if the benefit justifies the substantial engineering effort. Since we are not changing `syclcc` arguments that frequently, it might be doable to just update the documentation together with every change.

DieGoldeneEnte:
I think this looks good now (for some reason I can't approve :/ )

doc/using-hipsycl.md
```diff
@@ -0,0 +1,168 @@
+# Using hipSYCL in projects
+It is recommended to use the CMake integration for larger projects.
+
+## Compiling with syclcc
+`syclcc` is the compilation driver used by hipSYCL to build the final compiler invocations.
+After installing hipSYCL, it can be used as a standalone tool to manually build single files or integrated in other build systems than CMake.
```
illuhad:
```suggestion
After installing hipSYCL, it can be used as a standalone tool to manually build source files similarly to regular compilers, or it can be integrated in build systems other than CMake.
```
(It can also compile multiple files, not just single files - unless you are in explicit multipass. Also some style suggestions :) )

doc/using-hipsycl.md
```diff
@@ -0,0 +1,168 @@
...
+Exemplary, building the nbody example while targeting CPU and CUDA backends could be done using `syclcc bruteforce_nbody.cpp -O3 --hipsycl-targets="omp;cuda:sm_61"`.
```
illuhad:
```suggestion
For example, compiling a SYCL source `example.cpp` to an executable, while targeting CPU and CUDA backends, is possible using `syclcc -o example example.cpp -O3 --hipsycl-targets="omp;cuda:sm_61"`.
```
not sure if that example is really the best thing to point people to, as it is not routinely tested, and is still very SYCL 1.2.1-ish

doc/using-hipsycl.md
```diff
@@ -0,0 +1,168 @@
...
+
+Following is the full excerpt from `syclcc --help`. Note the options can also be set via environment variables or corresponding CMake options.
```
illuhad:
```suggestion
The full excerpt from `syclcc --help` follows below. Note the options can also be set via environment variables or corresponding CMake options. Default values can be set in `/hipsycl/install/path/etc/hipSYCL/syclcc.json`.
```
Maybe we should go more into detail with the cmake options in the future. I'm not sure all `syclcc` options are exposed.

doc/using-hipsycl.md
```diff
@@ -0,0 +1,168 @@
...
+  If set, do not delete temporary files created during compilation.
+
+
+Any other options will be forwarded to the compiler.
+
+Note: Command line arguments take precedence over environment variables.
+```
+
+## Using the CMake integration
+Setting a project using the hipSYCL CMake integration is quite straight forward.
```
illuhad:
```suggestion
Setting up a project using the hipSYCL CMake integration is quite straight forward.
```

illuhad (approved):
Thank you! :)

PR #549, opened 2021-05-08
Add templated device constructor

### LINKED ISSUE #548: Device can't be created using sycl::*_selector_v
According to the specification a device can be initialized using the one of `sycl::*_selector_v` constructs, this is currently not possible. According to the [specification](https://www.khronos.org/registry/SYCL/specs/sycl-2020/pdf/sycl-2020.pdf) page 56 this should work.

code:
```
#include <sycl/sycl.hpp>

int main() {
  sycl::device gpu{sycl::gpu_selector_v};
}
```

error:
```
/build/nbody_hipsim/src/test.cpp:4:16: error: no matching constructor for initialization of 'sycl::device'
  sycl::device gpu{sycl::gpu_selector_v};
               ^  ~~~~~~~~~~~~~~~~~~~~~~
/nix/store/f1jrjf6b8813dbjal6kj8qkphgn22596-hipsycl/bin/../include/sycl/../hipSYCL/sycl/device.hpp:69:7: note: candidate constructor (the implicit copy constructor) not viable: no known conversion from 'const hipsycl::sycl::gpu_selector' to 'const hipsycl::sycl::device' for 1st argument
class device {
      ^
/nix/store/f1jrjf6b8813dbjal6kj8qkphgn22596-hipsycl/bin/../include/sycl/../hipSYCL/sycl/device.hpp:69:7: note: candidate constructor (the implicit move constructor) not viable: no known conversion from 'const hipsycl::sycl::gpu_selector' to 'hipsycl::sycl::device' for 1st argument
/nix/store/f1jrjf6b8813dbjal6kj8qkphgn22596-hipsycl/bin/../include/sycl/../hipSYCL/sycl/device.hpp:75:3: note: candidate constructor not viable: no known conversion from 'const hipsycl::sycl::gpu_selector' to 'rt::device_id' for 1st argument
  device(rt::device_id id)
  ^
/nix/store/f1jrjf6b8813dbjal6kj8qkphgn22596-hipsycl/bin/../include/sycl/../hipSYCL/sycl/device_selector.hpp:231:16: note: candidate constructor not viable: no known conversion from 'const hipsycl::sycl::gpu_selector' to 'const hipsycl::sycl::device_selector' for 1st argument
inline device::device(const device_selector &deviceSelector) {
               ^
/nix/store/f1jrjf6b8813dbjal6kj8qkphgn22596-hipsycl/bin/../include/sycl/../hipSYCL/sycl/device.hpp:78:3: note: candidate constructor not viable: requires 0 arguments, but 1 was provided
  device()
  ^
1 error generated when compiling for gfx906.
```

The specification is in my opinion very loosely formulated, so I have no idea what should be changed.

sbalint98:
Add the missing templated device constructor, fixes #548

### DIFF
--- a/include/hipSYCL/sycl/device.hpp
+++ b/include/hipSYCL/sycl/device.hpp
@@ -79,7 +79,8 @@ public:
       : _device_id(detail::get_host_device()) {}
 
   // Implemented in device_selector.hpp
-  explicit device(const device_selector &deviceSelector);
+  template <class DeviceSelector>
+  explicit device(const DeviceSelector &deviceSelector);
 
   bool is_host() const 
   {
--- a/include/hipSYCL/sycl/device_selector.hpp
+++ b/include/hipSYCL/sycl/device_selector.hpp
@@ -228,8 +228,9 @@ auto aspect_selector() {
   return aspect_selector(aspectList...);
 }
 
-inline device::device(const device_selector &deviceSelector) {
-  this->_device_id = deviceSelector.select_device()._device_id;
+template <class DeviceSelector>
+inline device::device(const DeviceSelector &deviceSelector) {
+  this->_device_id = detail::select_device(deviceSelector)._device_id;
 }
 
 


fodinabor:
hmm.. shouldn't this be similar to https://github.com/illuhad/hipSYCL/blob/develop/include/hipSYCL/sycl/queue.hpp#L129 where the template argument is constrained with `detail::is_device_selector_v` and use `detail::select_device(selector)` to invoke the selection?

sbalint98:
I think in the case of a queue it needs to be constrained to prevent it from getting confused with async_handler. In this case, we don't have this problem, right?

EDIT: and `deviceSelector.select_device()` is just mapped to `detail::selcet_device(*this)` so isn't it the same thing?

fodinabor:
Yeah, makes sense, using `.select_device()` is still a bit SYCL 1.2.1-isch, SYCL 2020 allows to get a callable like a lambda as device selectors, so the selector would not have `.select_device` implemented: 
```c++
sycl::device([](const sycl::device& dev) -> int { 
if(dev.is_cpu()) return 1;
return -1;
});
```
And select device does the actual selection and just uses the call operator `selector()` to get the preference

illuhad:
> I think in the case of a queue it needs to be constrained to prevent it from getting confused with async_handler. In this case, we don't have this problem, right?

This is correct, the `enable_if` is only necessary for the queue.

> EDIT: and deviceSelector.select_device() is just mapped to detail::selcet_device(*this) so isn't it the same thing?

No, `selector.select_device()` only works for the old device selectors. You have to call `detail::select_device`, otherwise the new ones won't work.

fodinabor (approved):
LGTM :)

illuhad (approved):
Thanks! :)

PR #556, opened 2021-05-21
Include <limits> where necessary

illuhad:
This fixes one of several issues that all of a sudden popped up in the github CI pipeline.

### DIFF
--- a/src/runtime/cuda/cuda_hardware_manager.cpp
+++ b/src/runtime/cuda/cuda_hardware_manager.cpp
@@ -32,6 +32,7 @@
 #include <cuda_runtime_api.h>
 #include <exception>
 #include <cstdlib>
+#include <limits>
 
 namespace hipsycl {
 namespace rt {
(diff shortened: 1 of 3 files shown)


fodinabor (approved):
Making the CI really happy isn't this easy... I gave up yesterday when it started to complain that it doesn't find `-lomp`, while still not having fixed, that incompatible GCC 11.1 is used with Clang CUDA...

illuhad:
> Making the CI really happy isn't this easy... I gave up yesterday when it started to complain that it doesn't find -lomp, while still not having fixed, that incompatible GCC 11.1 is used with Clang CUDA...

Yeah.. I stumbled across the `-lomp` issue too and decided to fix the includes first ;) I think with clang CUDA we could probably come up with a workaround, at least for CI..

fodinabor:
:P
GCC's libstdc++ 11.2 will have that most likely fixed, until then we could try some weird stuff, like this: https://cuda.godbolt.org/z/e44qcaah5
In the essence, we would just redefine the offending `__glibcxx_assert` to only do what it did before version 11, before anything else is included:
```c++
#include <bits/c++config.h>
#undef __glibcxx_assert
# define __glibcxx_assert(_Condition)	  \
    __glibcxx_assert_2(_Condition);
```
(the corresponding LLVM bug for reference: https://bugs.llvm.org/show_bug.cgi?id=50383)

PR #559, opened 2021-05-24
Add workaround for libstdc++ 11 clang CUDA issue

illuhad:
This adds a workaround for https://bugs.llvm.org/show_bug.cgi?id=50383 - an incompatibility between clang CUDA/HIP and libstdc++ 11.1. This is not specific to hipSYCL; with libstdc++ 11.1.0 any code including `complex` fails compilation with clang CUDA or HIP.

This is already fixed in libstdc++ and clang upstream, but the problematic releases are already out there. In particular, our CI pipeline is affected, so currently this issue breaks our CI.

This PR adds a workaround by introducing our own std header wrappers with the same compatibility workarounds that the upstream clang patch uses. Currently we fix `complex` and `atomic`, for which this issue was reported. I don't know if other headers might be affected as well.

At the moment this workaround is always active. I'm not sure if there's harm in doing that and we should only conditionally enable it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -243,14 +243,14 @@ if(NOT ROCM_CXX_FLAGS)
   # __float128 even though it is not supported for CUDA / HIP,
   # see https://bugs.llvm.org/show_bug.cgi?id=47559.
 
-  set(ROCM_CXX_FLAGS "-isystem ${CLANG_INCLUDE_PATH} -U__FLOAT128__ -U__SIZEOF_FLOAT128__ -I$HIPSYCL_ROCM_PATH/include -I$HIPSYCL_ROCM_PATH/include --rocm-device-lib-path=$HIPSYCL_ROCM_PATH/amdgcn/bitcode --rocm-path=$HIPSYCL_ROCM_PATH -fhip-new-launch-api -D__HIP_ROCclr__" CACHE STRING "Arguments passed to compiler to compile SYCL applications with ROCm")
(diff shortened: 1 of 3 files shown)


fodinabor:
Ah.. you don't need that for atomic, afaik, it just fails If you include `complex` and another header (and in the example this was `atomic`) that uses `__glibcxx_assert`, because of the way `complex` is handled by clang cuda

illuhad:
@hipSYCL-ci-bot test also on other infrastructure

illuhad:
@fodinabor you are right, it seems `<atomic>` is not needed as it works on Linux without it. However it seems we are also getting this error Windows:
```c++
C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.28.29910\include\complex:565:16: error: call to 'signbit' is ambiguous
```
Any ideas?

fodinabor:
:/ seems like MS STL defines a `std::complex` specialization for `long double` but the clang CUDA `cmath` wrappers only define `signbit` for `float` and `double`.. 
one workaround could be to add something like the following to the std wrapper header...
```c++
#include <type_traits>

namespace std {
    bool signbit(long double v) { return signbit(static_cast<double>(v)); }
}
```
adapted from https://developercommunity.visualstudio.com/t/stdsignbit-misses-overloads-for-integer-types/923187 which is not our issue, though

illuhad:
@fodinabor thanks, it seems that did the trick.

PR #555, opened 2021-05-20
Add accessor variants, raw accessors to optimize register-pressure dominated kernels

illuhad:
This adds the hipSYCL accessor variant extension which introduces distinct accessor types depending on how they were constructed - in particular, depending on whether the accessor is a placeholder or ranged or unranged. The accessor implementation then only stores those fields that are strictly required for the specified use case. This can help performance for kernels where registers are spilling due to large accessor objects.

This PR also introduces *raw accessors*, which are a new, very lean accessor flavor that effectively only stores a pointer and has minimal overhead.

Accessor variants are expressed by extending the `access::placeholder` template parameter of the accessor class with additional values.

There are still some aspects that are unclear, e.g. how the new accessor variants interact with standard SYCL 2020 accessors  (where `access::placeholder` equals `true_t` or `false_t`). Currently such conversions are allowed for code compatibility purposes, but the argument can be made that this is dangerous since some conversions can lead to a loss of information.

Accessor variants can also be deduced automatically using SYCL accessor deduction guides. This allows automatic accessor size optimizations for existing SYCL code bases. Currently, this feature is enabled by default, but it is unclear if it should rather be disabled by default because it might in theory break existing SYCL code: Since different deduced accessor variants are implemented as distinct types via template parameters, there is an impact on conversion and compatibility between different variants.

Full documentation of this feature:

# hipSYCL accessor variants

## Terminology

In the following, we will use the following terminology:
* A *placeholder* accessor is an accessor that was constructed without providing a SYCL `handler` object. Placeholder need to be bound to a `handler` by calling `handler::require()` at a later point.
* A *ranged* accessor is an accessor that is constructed by specifying an access range consisting of a `sycl::range` and optionally a `sycl::id` describing the access offset. Ranged accessors may be used if only a subset of the buffer is accessed.
* An *unranged* accessor is an accessor that is not a ranged accessor.

## Motivation

SYCL accessors need to store several different kinds of information, depending on their use case. In general, this includes:
* A data pointer
* The n-dimensional shape of the allocation
* the access offset and access range
* Certain flags, e.g. whether the accessor is `no_init`
* Placeholder accessors generally need to store a `shared_ptr` or similar to the `buffer` object
* Additional implementation-specific information that the SYCL runtime needs to enable core accessor functionality.

[...]

PR #560, opened 2021-05-25
Add device update extension and fix update_host() cyclic DAG bug

illuhad:
Depends on #555 to avoid a lot of merge conflicts.

This PR adds the device `handler::update()` extension in analogy to `handler::update_host`.

Excerpt from documentation:

---
### `HIPSYCL_EXT_UPDATE_DEVICE`

An extension that adds `handler::update()` for device accessors in analogy to `update_host()`. While `update_host()` makes sure that the host allocation of the buffer is updated, `update()` updates the allocation on the device to which the operation is submitted. This can be used
* To preallocate memory if the buffer is uninitialized;
* To separate potential data transfers from kernel execution, e.g. for benchmarking;
* To control buffer data state when using buffer-USM interoperability(`HIPSYCL_EXT_BUFFER_USM_INTEROP`);
* To inform the runtime earlier of expected data usage in order to optimize data transfers or overlap of compute and data transfers

#### API Reference

```c++
namespace sycl {
class handler {
public:
  template <typename T, int dim, access::mode mode, access::target tgt,
          accessor_variant variant>
  void update(accessor<T, dim, mode, tgt, variant> acc);
};
}
```
---

Additionally, this PR fixes a bug that has potentially affected anything that build on top of explicit requirements, including `update_host()`, host accessors and buffer writebacks.
If an explicit requirement is submitted, the detection of conflicting requirements was potentially interleaved with the addition of nodes to data users. This could lead to cycles in the DAG, which is not good because nothing is prepared to handle a DCG :)


illuhad:
@hipSYCL-ci-bot test.

PR #561, opened 2021-05-26
add missing methods for `sycl::group`

DieGoldeneEnte:
Some methods are missing or have the wrong/old name. I added the missing ones (except `get_max_local_range`) and deprecated and renamed those with wrong name.
Some methods are removed in SYCL 2020, so they where deprecated.

I didn't implement  `get_max_local_range`, because I don't understand how it should work.
It is described as:
> Return a range representing the maximum number of work-items in any work-group in the nd_range.

But the work group size is defined by the maximum number of work items (on AMD/NVIDIA GPUs), so it doesn't feel right to return a range like {1024,1,1} or {16, 8, 8}. Maybe someone knows what the expected return value is?
In DPC++ it is only implemented for sub_groups as far as I could see.

### DIFF
--- a/include/hipSYCL/sycl/libkernel/group.hpp
+++ b/include/hipSYCL/sycl/libkernel/group.hpp
@@ -269,10 +269,17 @@ struct group
   }
 #endif
 
-  /*
-   * TODO: needs to be implemented
-  range<Dimensions> get_max_local_range() const;
-  */
+  range<Dimensions> get_max_local_range() const{
+    if constexpr (Dimensions == 1) {
+      return {1024};
+    } else if constexpr (Dimensions == 2) {
+      return {1024, 1024};
+    } else if constexpr (Dimensions == 3) {
+      return {1024, 1024, 1024};
+    } else {
+      static_assert(std::is_same_v<int, int>, "Only three dimensional ranges are supported!");
+    }
+  }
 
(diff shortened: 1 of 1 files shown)


illuhad:
> But the work group size is defined by the maximum number of work items (on AMD/NVIDIA GPUs), so it doesn't feel right to return a range like {1024,1,1} or {16, 8, 8}. Maybe someone knows what the expected return value is?

It's there to satisfy the new group concept. The lowest reasonable bound we can give on CUDA/HIP is {1024}, {1024,1024} and {1024, 1024, 1024}.

include/hipSYCL/sycl/libkernel/group.hpp
```diff
@@ -216,12 +244,43 @@ struct group
...
+    return detail::get_local_id<Dimensions>();
+  }
+
+  size_t get_local_id(int dimension) const
+  {
+    return detail::get_local_id<Dimensions>(dimension);
+  }
+#endif
+
+  range<Dimensions> get_max_local_range() const{
```
illuhad, line 275: I think those functions should be marked `HIPSYCL_KERNEL_TARGET` for consistency, no?

illuhad (approved):

illuhad:
Thanks!

PR #564, opened 2021-05-31
Add accessor variants aliases, forbid invalid conversions

illuhad:
* Describe `raw_accessor operator[]` behavior better in documentation
* Forbid conversion from standard accessor to `ranged` variants because of potentially incorrect semantics
* Add `ranged_accessor`, `unranged_accessor` etc aliases


illuhad:
@hipSYCL-ci-bot test

PR #565, opened 2021-05-31
Add queue::get_wait_list() extension

illuhad:
From documentation:

### `HIPSYCL_EXT_QUEUE_WAIT_LIST`

Adds a `queue::get_wait_list()` method that returns a vector of `sycl::event` in analogy to `event::get_wait_list()`, such that waiting for all returned events guarantees that all operations submitted to the queue have completed. This can be used to express asynchronous barrier-like semantics when passing the returned `vector` into `handler::depends_on()`.
If the queue is an in-order queue, the returned vector will contain at most one event.

Note that `queue::get_wait_list()` might not return an event for all submitted operations, e.g. completed operations or operations that are dependencies of others in the the dependency graphs may be optimized away in the returned set of events.

#### API Reference

```c++
namespace sycl {
class queue {
public:
  std::vector<event> get_wait_list() const;
};
}
```

### DIFF
--- a/doc/extensions.md
+++ b/doc/extensions.md
@@ -49,6 +49,24 @@ public:
 }
 ```
 
+### `HIPSYCL_EXT_QUEUE_WAIT_LIST`
+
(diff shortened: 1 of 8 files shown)


illuhad:
@hipSYCL-ci-bot test.

PR #566, opened 2021-05-31
Disable CUDA 10.0 in Windows CI

illuhad:
It seems it's no longer supported by Microsoft STL: https://github.com/microsoft/STL/pull/1544


illuhad:
Closing because it does not fix the issue - it seems Microsoft STL broke compilation with any CUDA version.

PR #570, opened 2021-06-02
Add a placeholder for half

sbalint98:
In order to support libraries using half without major changes to their interface, we need to define half in hipSYCL. 

This change allows reducing the number of `#ifdefs` necessary when making oneMKL work with hipSYCL. see:  https://github.com/hipSYCL/oneMKL/pull/6

### DIFF
--- a/include/hipSYCL/sycl/types.hpp
+++ b/include/hipSYCL/sycl/types.hpp
@@ -103,6 +103,7 @@ using sp_float = float;
 using dp_float = double;
 } //detail
 
+using half = detail::hp_float;
 } // sycl
 } // hipsycl
 


include/hipSYCL/sycl/types.hpp
```diff
@@ -80,6 +80,7 @@ using hash_class = std::hash<T>;
 using exception_ptr_class = std::exception_ptr;
 
 
+using half = short;
```
illuhad: Would it make sense to set this as an alias for `detail::hp_float` that is defined below?

sbalint98: mhmm Sure, :)

illuhad (approved):

PR #573, opened 2021-06-05
Fix no-init accessor property for non-placeholder accessor variants

illuhad:
Fix regression where no-init accessor property is not correctly processed for non-placeholder accessor variants when using `HIPSYCL_EXT_ACCESSOR_VARIANT_DEDUCTION`


illuhad:
@hipSYCL-ci-bot test

PR #574, opened 2021-06-07
[CI] Use Windows Server 2016 for CI

fodinabor:
As upstream MS STL broke Clang CUDA compilation, we use VS 2017 which is installed on the Windows Server 2016 runners.


illuhad (approved):
Awesome, thanks :)

PR #580, opened 2021-06-11
Critical: Fix bug in async_worker which might not correctly wait for operations

illuhad:
There is an ancient bug in `async_worker` which apparently did not cause any issues in testing for three years until PR #576 came along.

The worker thread of `async_worker` currently pops the next operation before executing it. If during execution of this operation `async_worker::wait()` is called, `wait()` no longer knows that there is still an operation being processed because it already has been popped from the queue, and hence cannot wait for its completion.

Since `async_worker` is used by the runtime to perform its background operations such as scheduling and submitting operations to the backend, this bug in particular causes `dag_manager::flush_sync()` calls to be unreliable in the sense that it cannot guarantee that all DAG operations that have been queued up have actually been flushed and submitted to a backend as it is supposed to.

At a higher level, this impacts all features that rely on `flush_sync()` functionality, such as buffer writebacks or host accessors. Without this fix, spurious errors, triggered debug assertions, or missing synchronization might occur.

This PR fixes this issue by only popping the operation after it has been executed.

@phiadaarr This might be related to your issue regarding buffer writebacks.

### DIFF
--- a/src/runtime/generic/async_worker.cpp
+++ b/src/runtime/generic/async_worker.cpp
@@ -58,6 +58,7 @@ void worker_thread::wait()
     // Wait until no operation is pending
     _condition_wait.wait(lock, [this]{return _enqueued_operations.empty();});
   }
+  assert(_enqueued_operations.empty());
 }
 
 
@@ -98,6 +99,7 @@ void worker_thread::work()
     // In any way, process the pending operations
 
     async_function operation = [](){};
+    bool has_operation = false;
 
     {
       std::lock_guard<std::mutex> lock(_mutex);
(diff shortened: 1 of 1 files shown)


illuhad:
@hipSYCL-ci-bot test

phiadaarr:
Thanks for informing me! Unfortunately this does not fix things for me. I will investigate the issue further and let you know of any insights.

phiadaarr:
At the moment it looks like it is reproducible accross different machines. If the buffer to which I write is of shape `(104800, n)` or smaller it gives the expected result. If it is of shape `(104900, n) ` or larger, the buffer is not written to at all. The issue seems to be independent of `n` (tried it for `n=1`, `10` and `100`). These numbers are reproducible on two different machines with different GPUs with different sizes of RAM of both the GPU and the CPU. And it is independent of whether I use `complex<single>` or `complex<double>`. But we'll investigate further and open a dedicated issue in the (unlikely) case that we believe something is wrong on hipSYCL's side.

@mreineck

PR #583, opened 2021-06-15
Add flush_sync() before querying node group in queue::get_wait_list()

illuhad:
For non-in-order queues, it is likely that we need to `flush_sync()` the dag before querying the node list so that we actually retrieve an up-to-date node list. This PR adds a call `flush_sync()` prior to querying the node group.


illuhad:
@hipsycl-ci-bot test

pszi1ard:
Verified that it fixes the issues observed on https://gitlab.com/gromacs/gromacs/-/merge_requests/1735

PR #584, opened 2021-06-27
[SYCL 2020] Add atomic_ref, refactor atomics, implement atomic builtins for all types and operations

illuhad:
* Add SYCL 2020 `atomic_ref`
* Refactor atomics in a fashion similar to how math builtins work: A common `atomic_builtins.hpp` header includes backend-specific headers which provide `__hipsycl_atomic_*` backend builtins. These can then be used by both `atomic` and `atomic_ref` classes.
* Add support for previously unsupported atomic operations, such as load/store on GPU, or floating point atomics on CPU by adding appropriate CAS emulations. With this PR, all atomic operations are supported for all types on all backends (except SPIR-V, which requires more work in a separate PR)
* Add test coverage


illuhad:
@hipSYCL-ci-bot test again.

PR #585, opened 2021-06-28
Avoid using BOOST_ROOT other than for finding Boost

mabraham:
Having found Boost, it is better to find a path relative to it via the
the variables set by the finding procedure. If so, builds that
e.g. found Boost via CMAKE_PREFIX_PATH can work.

Also fixed inverted logic for Apple build added in #535


mabraham:

illuhad:
Thanks!

@hipSYCL-ci-bot test

(@mabraham  currently only members of the hipSYCL organization can invoke that for security reasons ;) )

illuhad (approved):

PR #587, opened 2021-07-03
Make buffer::reinterpret more standard conforming.

fodinabor:
- return buffer with rebound allocator,
- add overload without range parameter,
- allow and handle differently sized target types,
- throw an exception if the byte size is not fully covered.

Handling differently sized types is a bit tricky as hipSYCL's page mechanism works on the original element's size. Thus, whenever the size of the element type of a ranged accessor is different than the size of the type, the buffer was originally created with, all pages of the buffer are migrated to prevent partial values.

### DIFF
--- a/doc/runtime-spec.md
+++ b/doc/runtime-spec.md
@@ -53,7 +53,7 @@ Using an accessor that is not of a read-only access mode on a device *d*  shall 
 
 Data transfers generated from accessors (see below) shall cause transferred pages to be marked as up-to-date on the target allocation.
 
-If a `buffer` is reinterpreted to a data type of different size than the original buffer element size, the implementation may assume a page range for accessors to the reinterpreted buffer that is larger than the page range as defined above.
+If a `buffer` is reinterpreted to a data type of different size than the original buffer element size or reshaped into a different range, the implementation may assume a page range for accessors to the reinterpreted buffer that is larger than the page range as defined above.
 
 #### Data transfers
 
--- a/doc/runtime-spec.md
+++ b/doc/runtime-spec.md
@@ -53,7 +53,7 @@ Using an accessor that is not of a read-only access mode on a device *d*  shall 
 
 Data transfers generated from accessors (see below) shall cause transferred pages to be marked as up-to-date on the target allocation.
 
-After reinterpreting a `buffer` with a differing type, the implementation may require the transfer of all pages, despite a smaller range being specified by an accessor.
+If a `buffer` is reinterpreted to a data type of different size than the original buffer element size, the implementation may assume a page range for accessors to the reinterpreted buffer that is larger than the page range as defined above.
 
 #### Data transfers
 
--- a/doc/runtime-spec.md
+++ b/doc/runtime-spec.md
@@ -49,9 +49,11 @@ For each allocation managed on each device, the `buffer` implementation shall tr
 
 If a page is fully contained within or overlaps with the accessed range of an accessor (taking into account the accessor's access offset and range), we use the terminology that the page is part of the accessor's *page range*.
 
-Using an accessor that is not of a read-only access mode on a device *d*  shall cause all pages within its page range to be marked as outdated on all allocations except for those on device *d*. This is because the implementations has to assume that data was modified on `d`.
+Using an accessor that is not of a read-only access mode on a device *d*  shall cause all pages within its page range to be marked as outdated on all allocations except for those on device *d*. This is because the implementation has to assume that data was modified on `d`.
 
 Data transfers generated from accessors (see below) shall cause transferred pages to be marked as up-to-date on the target allocation.
+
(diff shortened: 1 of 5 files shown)


fodinabor:
@hipsycl-ci-bot test this change again

tests/sycl/buffer.cpp
```diff
@@ -65,28 +65,79 @@ BOOST_AUTO_TEST_CASE(buffer_versioning) {
...
+
+  BOOST_REQUIRE(buf_a == buf_a);
+  BOOST_REQUIRE(buf_a != buf_b);
+  BOOST_REQUIRE(buf_a == buf_c);
+
+  // compile test, that reinterpreted buffer uses rebound allocator
+  s::buffer<std::uint16_t, 1, s::buffer_allocator<std::uint16_t>> buf_d = buf_a.reinterpret<std::uint16_t>(s::range<1>{32});
+  s::buffer<std::uint32_t, 1, s::buffer_allocator<std::uint32_t>> buf_e = buf_a.reinterpret<std::uint32_t, 1>();
+  s::buffer<std::uint16_t, 2, s::buffer_allocator<std::uint16_t>> buf_f = buf_a.reinterpret<std::uint16_t>(s::range<2>{4, 8});
+  s::buffer<std::uint32_t, 2, s::buffer_allocator<std::uint32_t>> buf_g = buf_a.reinterpret<std::uint32_t, 2>();
```
illuhad: Some line breaks?

include/hipSYCL/sycl/libkernel/accessor.hpp
```diff
@@ -1097,19 +1098,21 @@ class accessor
...
-                      "buffer are not yet supported");
-
     rt::dag_node_ptr node;
     {
       rt::dag_build_guard build{rt::application::dag()};
 
-      auto explicit_requirement =
-          rt::make_operation<rt::buffer_memory_requirement>(
+      std::unique_ptr<rt::buffer_memory_requirement> explicit_requirement;
+      if(sizeof(dataT) != data->get_num_elements())
```
illuhad:
```suggestion
      if(sizeof(dataT) != data->get_element_size())
```
I don't understand this - shouldn't it be as in the suggestion?

include/hipSYCL/sycl/libkernel/accessor.hpp
```diff
@@ -1097,19 +1098,21 @@ class accessor
...
+        explicit_requirement.reset(new rt::buffer_memory_requirement(
+              data, rt::id<3>{}, data->get_num_elements(),
```
illuhad: Since only offset and range are different, would it make sense to only find the right offset/range inside the `if` and construct the requirement outside, to share some more code between the code paths?

include/hipSYCL/sycl/handler.hpp
```diff
@@ -105,20 +106,22 @@ class handler {
...
+    std::unique_ptr<rt::buffer_memory_requirement> req;
+    if (element_size != sizeof(typename AccessorType::value_type)) {
+      req.reset(new rt::buffer_memory_requirement(
+        data.mem, rt::id<3>{}, data.mem->get_num_elements(), mode,
+        AccessorType::access_target));
+    } else {
+      req.reset(new rt::buffer_memory_requirement(
         data.mem, rt::make_id(data.offset), rt::make_range(data.range), mode,
-        AccessorType::access_target);
+        AccessorType::access_target));
```
illuhad: We have this logic now in multiple places. I wonder if it might be a good idea to have a single `get_effective_offset/range` or similar somewhere, so that we only need to change a single place if we want to optimize this case some more?

doc/runtime-spec.md
```diff
@@ -49,10 +49,12 @@ For each allocation managed on each device, the `buffer` implementation shall tr
 
 If a page is fully contained within or overlaps with the accessed range of an accessor (taking into account the accessor's access offset and range), we use the terminology that the page is part of the accessor's *page range*.
 
-Using an accessor that is not of a read-only access mode on a device *d*  shall cause all pages within its page range to be marked as outdated on all allocations except for those on device *d*. This is because the implementations has to assume that data was modified on `d`.
+Using an accessor that is not of a read-only access mode on a device *d*  shall cause all pages within its page range to be marked as outdated on all allocations except for those on device *d*. This is because the implementation has to assume that data was modified on `d`.
 
 Data transfers generated from accessors (see below) shall cause transferred pages to be marked as up-to-date on the target allocation.
 
+After reinterpreting a `buffer` with a differing type, the implementation may require the transfer of all pages, despite a smaller range being specified by an accessor.
```
illuhad:
```suggestion
If a `buffer` is reinterpreted to a data type of different size than the original buffer element size, the implementation may assume a page range for accessors to the reinterpreted buffer that is larger than the page range as defined above.
```

illuhad (approved):

PR #593, opened 2021-07-14
Add instrumentation and profiling API

### LINKED ISSUE #417: Implement kernel profiling
hipSYCL is currently missing an API for kernel profiling. I would like to implement this, but I think it warrants some discussion beforehand (and I also need some directions for navigating the runtime internals).

For reference, there was an attempt at implementing this for the old runtime #251, but that PR has since become obsolete.

## Problems with profiling in the SYCL spec

The spec mandates `event::get_profiling_info`, which is basically the same interface as [`clEventGetProfilingInfo`](https://www.khronos.org/registry/OpenCL/sdk/1.0/docs/man/xhtml/clGetEventProfilingInfo.html). This makes it easy to implement for an OpenCL backend, where a SYCL event corresponds to an OpenCL event exactly and OpenCL events already report submission, start and end times.

CUDA profiling works differently. To measure the device execution time of a command, one needs to call `cudaEventRecord` before and after the kernel launch, and obtain the duration via `cudaEventElapsedTime(endEvt, startEvt)`. hipSYCL already records a CUDA event for the kernel completion, and when queue profiling is enabled, will have to insert an additional event before kernel launch.

But even with multiple CUDA events, we cannot implement the spec API faithfully. We don't have access to timestamps, only durations, so we'd have to report `command_start` as constant-zero and `command_end` as the duration. That might be enough for simple kernel profiling, but losing the temporal ordering between different command groups might surprise users. Also, I see no way to measure `command_submit` in device time at all.

The spec is also not too precise on when those individual timings should be taken. It talks about profiling a command group, which can be understood as including times for implicit data transfers, but the OpenCL-like interface suggestst that only kernel times are measured.

## To be discussed

1. Should we add our own extension for kernel profiling that can actually be implemented correctly with CUDA? I'm thinking along the lines of an `event::get_execution_time()` or `event::get_profiling_info<info::execution_time>`.
2. If we do add an extension, should we partially implement the spec API subject to the limitations above? This would improve conformance, but a partially-correct implementation might be worse than none at all for users migrating from a different SYCL implementation.
3. Should we only record kernel execution (or explicit-copy) times, or should requirement / implicit transfer durations be included in the profile? Regardless of how we decide, the extension API (if we want to implment it) should be unambiguous in that regard.

## Directions for an implementation

Where should these profiling events be created? There are code comments mentioning that `backend_executor::create_event_{before,after}` is meant for profiling in the scheduler, but I'm not sure where to apply that. I think we should only add events for nodes that do return a `sycl::event`, there is no need for profiling functionality e.g. on implicit transfer operations.

illuhad:
This PR is based on PR #428, but introduces a redesign of the instrumentation and API, which allows for some simplifications. In particular, in this PR instrumentations can also be added by the background worker thread (e.g. the scheduler). This is done by collecting instrumentations until the scheduler has submitted the node.

Additionally, this PR moves to a design that relies more on existing idioms and functionality of the hipSYCL runtime.

All backends except for Level Zero are currently supported.

@fknorr feedback appreciated :)

TODO: Think about how we can access instrumentations of requirements and how we can handle multi-operation nodes in the future


include/hipSYCL/runtime/instrumentation.hpp
```diff
@@ -0,0 +1,193 @@
...
+class submission_timestamp : public instrumentation {
+public:
+  virtual profiler_clock::time_point get_time_point() const = 0;
+  virtual ~submission_timestamp() = default;
+
+  std::size_t get_ns_ticks() const {
+    return get_time_point().time_since_epoch().count();
+  }
+
+  double get_time_seconds() const {
```
fknorr: `get_time_seconds` is never used. For best type-safety I suggest we remove both `get_ns_ticks` and `get_time_seconds`, doing the `profiler_clock::time_point -> uint64_t` conversion only at the top-level `event::get_profiling_info`

include/hipSYCL/runtime/instrumentation.hpp
```diff
@@ -0,0 +1,193 @@
...
+    _is_registration_complete = true;
+    _registration_complete_signal.signal();
+  }
+
+private:
+  std::vector<std::pair<std::type_index, std::shared_ptr<instrumentation>>>
+      _instrs;
+
+  mutable signal_channel _registration_complete_signal;
+  mutable std::atomic<bool> _is_registration_complete = false;
```
fknorr: This flag looks redundant, we should be able to just call `_registration_complete_signal->has_signalled()`

src/runtime/omp/omp_instrumentation.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+void omp_execution_start_timestamp::wait() const { _signal.wait(); }
+
+profiler_clock::time_point
+omp_execution_start_timestamp::get_time_point() const {
+  return _time;
+}
+
+void omp_execution_start_timestamp::record_time() {
+  _time = profiler_clock::now();
```
fknorr: I've been thinking some time about whether the access to `_time` from `record_time()` and `get_time_point()` are thread-safe or not. I would expect `signal::signal()` to synchronize-with `signal::wait()` for practical purposes, but cppref does not say anything about such a relationship between `std::promise::set_value()` and `std::shared_future::wait()`. As long as nobody calls `get_time_point()` before wait, it *should* be safe.

Since this is a brittle interface as-is, it might be best to make `signal_channel` generic and transport a value (here, a `profiler_clock::time_point`) to avoid this question altogether.

fknorr (commented):
Thank you for taking the time to look into profiling again! Overall, I like this approach. I was not thinking about potential deadlocks around waiting for instrumentation results.

include/hipSYCL/runtime/instrumentation.hpp
illuhad: The idea was that we might need those conversions for future functionality that I want to work on, but maybe it makes more sense to have those conversions as free functions (which would also remove code duplication).

include/hipSYCL/runtime/instrumentation.hpp
illuhad: The idea here is to cache whether the registration is complete (a similar pattern exists in the dag node implementation), but I guess it could be argued that this is a case of premature optimization.

src/runtime/omp/omp_instrumentation.cpp
illuhad: Do I understand correctly that you say that we cannot actually rely on `signal_channel::wait()` to synchronize with `signal_channel::signal()`? The whole point of the `signal_channel` class was to allow for such patterns, if it doesn't do that then I think we should rather fix `signal_channel`. 
Especially since it's not a new class: It has existed before and is e.g. used to implement events in the OpenMP backend (this PR only moves it to a separate header file to make it usable by the instrumentation API). So, if it's broken existing code relying on it will be broken as well.

fknorr: I dug into the standard and *[futures.state]* does say that `shared_future::wait()` synchronizes-with `promise::set_value()`, so the use in `omp_node_event` is definitely file. The issue that remains IMO is that `omp_execution_*_timestamp` can be easily mis-used: Calling `record_time()` twice or calling `get_time_point()` before `wait()` will usually cause a data race because `_time` is not synchronized otherwise. The straight-forward solution should be to transport a value through `signal_channel` instead of the implicit `bool` it uses now.

include/hipSYCL/runtime/instrumentation.hpp
fknorr: I would expect futures to be implemented with atomics where possible anyway.

src/runtime/omp/omp_instrumentation.cpp
illuhad: Thanks for checking with the standard!

So the issue is that it could be misused. Calling `get_time_point()` before `wait()` cannot really happen because the `instrumentation_set` is deliberately designed to not grant direct access to instrumentations, and internally executing `wait()` before accessing results. This simplifies the implementation of the instrumentations, because such synchronization issues are now handled at a central place.
In this design, `instrumentation` objects should mainly be seen as a simple vehicle to store a result.

I have now made `record_time()` private such that it can only be accessed by the OpenMP instrumentation setup logic. It is after all more of an implementation detail of OpenMP instrumentation setup. There are now also asserts to check that it was not invoked before.

include/hipSYCL/runtime/cuda/cuda_instrumentation.hpp
```diff
@@ -0,0 +1,60 @@
...
+#include "hipSYCL/runtime/event.hpp"
+#include <chrono>
+
+namespace hipsycl {
+namespace rt {
+
+class cuda_event_time_delta {
+public:
+  profiler_clock::duration operator()(std::shared_ptr<dag_node_event> t0,
+                                      std::shared_ptr<dag_node_event> t1) const;
```
fknorr: These parameters don't need to transfer ownership, `const dag_node_event &` should be enough

include/hipSYCL/runtime/hip/hip_instrumentation.hpp
```diff
@@ -0,0 +1,60 @@
...
+#include "hipSYCL/runtime/event.hpp"
+#include <chrono>
+
+namespace hipsycl {
+namespace rt {
+
+class hip_event_time_delta {
+public:
+  profiler_clock::duration operator()(std::shared_ptr<dag_node_event> t0,
+                                      std::shared_ptr<dag_node_event> t1) const;
```
fknorr: `const dag_node_event &` should be enough here

src/runtime/cuda/cuda_instrumentation.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+#include <cassert>
+#include <memory>
+
+namespace hipsycl {
+namespace rt {
+
+profiler_clock::duration
+cuda_event_time_delta::operator()(std::shared_ptr<dag_node_event> t0,
+                                  std::shared_ptr<dag_node_event> t1) const {
```
fknorr: `const dag_node_event &` is sufficient

src/runtime/hip/hip_instrumentation.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include "hipSYCL/runtime/error.hpp"
+
+#include <cassert>
+#include <memory>
+
+namespace hipsycl {
+namespace rt {
+
+profiler_clock::duration
+hip_event_time_delta::operator()(std::shared_ptr<dag_node_event> t0,
```
fknorr: `const dag_node_event &` is sufficient

src/runtime/omp/omp_queue.cpp
```diff
@@ -69,8 +71,100 @@ bool is_contigous(id<3> offset, range<3> r, range<3> allocation_shape) {
...
 
+class instrumentation_task_guard;
+
+template <class BaseInstrumentation>
+class omp_task_timestamp : public BaseInstrumentation {
+public:
+  friend class instrumentation_task_guard;
+
+  virtual profiler_clock::time_point get_time_point() const override {
+    return _time;
```
fknorr, line 82: This can still be misused by calling `get_time_point()` before `wait()`, which will race on `_time`.

include/hipSYCL/runtime/dag_node.hpp
```diff
@@ -93,6 +106,16 @@ class dag_node
 
   void for_each_nonvirtual_requirement(std::function<void(dag_node_ptr)>
                                            handler) const;
+  /// Iterates across all operations that have actually been executed.
+  /// Precondition: is_submitted() returns true
+  template<class Handler>
+  void for_each_executed_operation(Handler h) {
```
fknorr, line 112: This function is added but not currently used, is that intentional?

src/runtime/omp/omp_queue.cpp
illuhad: Not really - the whole point of the instrumentation API is that `instrumentation_set` guarantees that no instrumentation objects will be handed out until `wait()` has returned. This simplifies the design of instrumentations because it can be assumed that they will only be accessed (after initial setup) once they are ready.

include/hipSYCL/runtime/dag_node.hpp
illuhad: It provides a mechanism to access instrumented requirements. It's not yet used because it's unclear what should be returned for user-facing queries in case a requirement was optimized away, but I have some other PRs coming up that use it for internal runtime functionality.

src/runtime/omp/omp_queue.cpp
fknorr: Ok, then I'll be happy with an `assert(_signal.has_signalled())` in there.

fknorr (approved):
Aside from my comments above, this looks good to me!

illuhad:
@fknorr thanks, changing the `shared_ptr`s to const ref as you suggested is on my to do list and will happen before merge :)

illuhad:
@hipSYCL-ci-bot test again please

PR #594, opened 2021-07-15
Add support for boost 1.70+

sbalint98:
In case Boost has been built with the boost-cmake project or from Boost 1.70.0 on it provides a package configuration file for use with find_package's config mode.  If the boost provided configuration is found, the package configuration file is loaded and the FindBoost module returns with no further action. [[1](https://cmake.org/cmake/help/latest/module/FindBoost.html#boost-cmake)]

In this case, the variables `Boost_FIBER_LIBRARY_DEBUG` and `Boost_FIBER_LIBRARY_RELEASE` won't be defined which results in crashing the cmake configuration. Unfortunately, the boost cmake configuration does not define any cmake variables which could be used for setting the `BOOST_LIBRARY_DIR` variable. 

Alternatively to this PR, we could set the `Boost_NO_BOOST_CMAKE` variable but this does not seem to be the right way to go since it will just force finding the boost libraries in the legacy way.


illuhad:
I recently came across an issue where we ended up with an empty `"-I"` as sequential cxx flag in syclcc.json, which messed up the entire compilation command line. Presumably the reason was because `$Boost_INCLUDE_DIR` was not set here:
https://github.com/illuhad/hipSYCL/blob/0520e1e5f156efa7784a9b9c8f6e857f467c7dd2/CMakeLists.txt#L265

Can this be related?

sbalint98:
I think based on [this](https://github.com/boostorg/boost_install/blob/develop/BoostConfig.cmake) file that the boost configuration doesn't set any variable with paths, it only exports targets. And since we don't set `Boost_INCLUDE_DIR` anywhere this could explain the issue you have seen.

fodinabor (commented):
CMake and Boost is just hell...
re @illuhad the BoostConfig seems to set `Boost_INCLUDE_DIRS`, not `Boost_INCLUDE_DIR`...

Also if `Boost_DIR` is set it could just be used to determine the lib dir.. 
The Boost CMake integration uses
```
get_filename_component(_BOOST_CMAKEDIR "${CMAKE_CURRENT_LIST_DIR}/../" REALPATH)
get_filename_component(_BOOST_LIBDIR "${_BOOST_CMAKEDIR}/../" ABSOLUTE)
```
where `CMAKE_CURRENT_LIST_DIR` should be replaceable by `Boost_DIR`, if set.

fodinabor:
Another possible approach would be to query the defined targets..
e.g. if `Boost_FIBER_LIBRARY_{DEBUG/RELEASE}` isn't set, query 
`get_target_property(FIBER_IMPORT_LIBRARY Boost::fiber IMPORTED_LOCATION_RELEASE)` and `get_target_property(FIBER_IMPORT_LIBRARY Boost::fiber IMPORTED_LOCATION_DEBUG)` and then `get_filename_component(Boost_LIBRARY_DIR "${FIBER_IMPORT_LIBRARY}" ABSOLUTE)`

Similarly the include directory can be found: `get_target_property(Boost_INCLUDE_DIR Boost::fiber INTERFACE_INCLUDE_DIRECTORIES)`

sbalint98:
@fodinabor Can we assume that `Boost_INCLUDE_DIR` and `FIBER_IMPORT_LIBRARY` will always be a single path, not a list?

fodinabor:
For now at least.. and can't really imagine that it changes, so yeah..

CMakeLists.txt
```diff
@@ -166,10 +166,19 @@ else()
   set(SYCLCC_CONFIG_FILE_INSTALL_DIR etc/hipSYCL)
 endif()
 
-if(Boost_FIBER_LIBRARY_DEBUG)
-  get_filename_component(Boost_LIBRARY_DIR ${Boost_FIBER_LIBRARY_DEBUG} DIRECTORY)
+if(NOT Boost_DIR)
```
fodinabor: maybe don't check `Boost_DIR` here but instead do `elseif(Boost_FIBER_LIBRARY_RELEASE`) ... else()` with the original if/else?

sbalint98: okay

fodinabor:
@hipsycl-ci-bot test this AS well

fodinabor (approved):
LGTM :)

illuhad (approved):

zjin-lcf:
Did you see the following error ?

-- Could NOT find Boost (missing: context fiber) (found version "1.72.0")
CMake Error at CMakeLists.txt:38 (message):
  Please provide the path to the root directory of the boost installation
  using the -DBOOST_ROOT option

PR #600, opened 2021-07-28
Feature: add missing copy functions to sycl::queue and sycl::handler

breyerml:
This pull request adds to following functions:
- `void handler::copy(const T *src, T *dest, size_t count)`: 
  the `src` and `dest` parameters are ordered with respect to [commandGroupHandler.h](https://github.com/KhronosGroup/SYCL-Docs/blob/SYCL-2020/master/adoc/headers/commandGroupHandler.h) (line 79)

- copy shortcuts for the `sycl::queue` class:
  - `event queue::copy(const T* src, T *dest, size_t count)`
  - `event queue::copy(const T* src, T *dest, size_t count, event depEvent)`
  - `event queue::copy(const T* src, T *dest, size_t count, const std::vector<event> &depEvents)`
  
  again the `src` and `dest` parameter order is the same as [queue.h](https://github.com/KhronosGroup/SYCL-Docs/blob/07775c05b7ef08b8d6c8dac52ca69639b8cce74c/adoc/headers/queue.h) (line 126 ff.)

- the "Explicit copy functions" in the `sycl::queue` class:
  - `event copy(accessor<T, dim, mode, tgt, variant> src, shared_ptr_class<T> dest)`
  - `event copy(shared_ptr_class<T> src, accessor<T, dim, mode, tgt, variant> dest)`
  - `event copy(accessor<T, dim, mode, tgt, variant> src, T *dest)`
  - `event copy(const T *src, accessor<T, dim, mode, tgt, variant> dest)`
  - `event copy(accessor<T, dim, srcMode, srcTgt, VariantSrc> src, accessor<T, dim, dstMode, destTgt, VariantDest> dest)`
  - `event update_host(accessor<T, dim, mode, tgt, variant> acc)`
  - `event fill(accessor<T, dim, mode, tgt, variant> dest, const T &src)`
  
  same as [queue.h](https://github.com/KhronosGroup/SYCL-Docs/blob/07775c05b7ef08b8d6c8dac52ca69639b8cce74c/adoc/headers/queue.h) (line 164 ff.) but with the template parameters changed to align with the same functions implemented in the `sycl::handler` class 

No functionality is newly implemented:
the new `copy` functions call the already implemented `memcpy` functions and the "Explicit copy functions" use the implementation provided in the `sycl::handler` class.

### DIFF
--- a/include/hipSYCL/sycl/handler.hpp
+++ b/include/hipSYCL/sycl/handler.hpp
@@ -558,8 +558,8 @@ public:
   
   template <typename T>
   void copy(const T* src, T* dest, std::size_t count) {
-    this->memcpy(reinterpret_cast<void*>(dest), 
-      reinterpret_cast<const void*>(src), count * sizeof(T));
+    this->memcpy(static_cast<void*>(dest),
+      static_cast<const void*>(src), count * sizeof(T));
   }
 
 
--- a/include/hipSYCL/sycl/handler.hpp
+++ b/include/hipSYCL/sycl/handler.hpp
@@ -558,7 +558,8 @@ public:
   
   template <typename T>
   void copy(const T* src, T* dest, std::size_t count) {
-    this->memcpy(reinterpret_cast<void*>(dest), reinterpret_cast<const void*>(src), count * sizeof(T));
+    this->memcpy(reinterpret_cast<void*>(dest), 
+      reinterpret_cast<const void*>(src), count * sizeof(T));
   }
 
 
--- a/include/hipSYCL/sycl/handler.hpp
+++ b/include/hipSYCL/sycl/handler.hpp
@@ -555,6 +555,11 @@ public:
 
     _command_group_nodes.push_back(node);
   }
+  
+  template <typename T>
+  void copy(const T* src, T* dest, std::size_t count) {
+    this->memcpy(reinterpret_cast<void*>(dest), reinterpret_cast<const void*>(src), count * sizeof(T));
+  }
 
 
   template <class T> void fill(void *ptr, const T &pattern, std::size_t count) {
--- a/include/hipSYCL/sycl/queue.hpp
+++ b/include/hipSYCL/sycl/queue.hpp
@@ -606,6 +606,26 @@ public:
       cgh.memcpy(dest, src, num_bytes);
     });
   }
+  
+  template <typename T>
+  event copy(const T* src, T* dest, std::size_t count) {
+    return this->memcpy(reinterpret_cast<void*>(dest), 
+      reinterpret_cast<const void*>(src), count * sizeof(T));
+  }
+  
+  template <typename T>
+  event copy(const T* src, T* dest, std::size_t count, 
+             event dependency) {
+    return this->memcpy(reinterpret_cast<void*>(dest), 
+      reinterpret_cast<const void*>(src), count * sizeof(T), dependency);
+  }
+  
+  template <typename T>
+  event copy(const T* src, T* dest, std::size_t count, 
+             const std::vector<event>& dependencies) {
+    return this->memcpy(reinterpret_cast<void*>(dest), 
+      reinterpret_cast<const void*>(src), count * sizeof(T), dependencies);
+  }
 
   event memset(void *ptr, int value, std::size_t num_bytes) {
     return this->submit([&](sycl::handler &cgh) {
--- a/include/hipSYCL/sycl/queue.hpp
+++ b/include/hipSYCL/sycl/queue.hpp
@@ -825,6 +825,15 @@ public:
       cgh.update_host(acc);
     });  
   }
+  
+  template <typename T, int dim, access::mode mode, access::target tgt,
+            access::placeholder isPlaceholder>
+  event update(accessor<T, dim, mode, tgt, isPlaceholder> acc) {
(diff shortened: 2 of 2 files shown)


include/hipSYCL/sycl/handler.hpp
```diff
@@ -555,6 +555,11 @@ class handler {
 
     _command_group_nodes.push_back(node);
   }
+  
+  template <typename T>
+  void copy(const T* src, T* dest, std::size_t count) {
+    this->memcpy(reinterpret_cast<void*>(dest), reinterpret_cast<const void*>(src), count * sizeof(T));
```
illuhad: Line break maybe?

include/hipSYCL/sycl/queue.hpp
```diff
@@ -606,6 +606,26 @@ class queue : public detail::property_carrying_object
       cgh.memcpy(dest, src, num_bytes);
     });
   }
+  
+  template <typename T>
+  event copy(const T* src, T* dest, std::size_t count) {
+    return this->memcpy(reinterpret_cast<void*>(dest), 
+      reinterpret_cast<const void*>(src), count * sizeof(T));
```
illuhad: I think a `static_cast` should be sufficient to convert to `void*`. No need to bring out the big guns :)

include/hipSYCL/sycl/queue.hpp
```diff
@@ -741,6 +761,71 @@ class queue : public detail::property_carrying_object
...
   }
 
+  
+  // Explicit copy functions
+  template <typename T, int dim, access::mode mode, access::target tgt,
+            accessor_variant variant>
+  event copy(accessor<T, dim, mode, tgt, variant> src,
+             shared_ptr_class<T> dest) {
+    return this->submit([&](sycl::handler &cgh) {
+      cgh.copy(src, dest);
```
illuhad, line 774: The spec is not really clear how those should behave, but my thinking is that this only makes sense with placeholder accessors, since the user doesn't have access to the command group handler to create non-placeholders.

In this case, you still need to call `require` on the accessor, otherwise it will not be connected to the command group and none of these functions will work :)

include/hipSYCL/sycl/queue.hpp
```diff
@@ -741,6 +761,71 @@ class queue : public detail::property_carrying_object
...
+  event copy(accessor<T, dim, srcMode, srcTgt, VariantSrc> src,
+             accessor<T, dim, dstMode, destTgt, VariantDest> dest) {
+    return this->submit([&](sycl::handler &cgh) {
+      cgh.copy(src, dest);
+    });  
+  }
+
+  template <typename T, int dim, access::mode mode, access::target tgt,
+            accessor_variant variant>
+  event update_host(accessor<T, dim, mode, tgt, variant> acc) {
```
illuhad: If we expose `update_host`, we should probably also expose hipSYCL's device `update()` extension (`handler::update()`)

include/hipSYCL/sycl/queue.hpp
```diff
@@ -741,6 +761,71 @@ class queue : public detail::property_carrying_object
     });
   }
 
+  
+  // Explicit copy functions
+  template <typename T, int dim, access::mode mode, access::target tgt,
```
illuhad: Can we directly move to the SYCL 2020 `access_mode` alias instead of `access::mode`?

illuhad (commented):
Thank you for this nice contribution :) I have some remarks :)

include/hipSYCL/sycl/handler.hpp
breyerml: Sure :)

include/hipSYCL/sycl/queue.hpp
breyerml: Fixed.

include/hipSYCL/sycl/queue.hpp
breyerml: I've changed all accessors to `access::placeholder` and added the missing `require()` calls.

include/hipSYCL/sycl/queue.hpp
breyerml: Done.

include/hipSYCL/sycl/queue.hpp
breyerml: Sure. I mainly used `access::mode` and `access::target` since the implementation in the `sycl::handler` class also uses them :).

include/hipSYCL/sycl/queue.hpp
illuhad: I think you can stick to using `accessor_variant`. In the current implementation it's equivalent to `access::placeholder` but also enables hipSYCL's accessor variant extension :)

illuhad:
@hipSYCL-ci-bot test this please! :)

include/hipSYCL/sycl/queue.hpp
illuhad: Yes, they are still used in some places, but I think for new code we should just use the new ones - we will have to switch everything at some point anyway :)

include/hipSYCL/sycl/queue.hpp
breyerml: Done.

illuhad (approved):

illuhad:
Thank you! :)

PR #602, opened 2021-07-28
Add atomic flag to signal channel

### LINKED ISSUE #601: Performance issue pre libstdc++-11 for signal_channel::has_signalled()
We use `signal_channel::has_signalled()` to check if certain events have completed. This method internally uses the pattern `std::future::wait_for(std::chrono::seconds(0))` to check the state of the future.

Pre libstdc++11, it seems that this is extremely inefficient since it waits for the event to complete even though the wait duration is 0.

sbalint98:
The change should make checking the signal channel more efficient on gcc versions below 11.
see issue #601 for details

### DIFF
--- a/include/hipSYCL/runtime/signal_channel.hpp
+++ b/include/hipSYCL/runtime/signal_channel.hpp
@@ -39,9 +39,11 @@ class signal_channel {
 public:
   signal_channel() {
     _shared_future = _promise.get_future().share();
+    _has_signalled_flag = false;
   }
 
   void signal() {
+    _has_signalled_flag = true;
     _promise.set_value(true);
   }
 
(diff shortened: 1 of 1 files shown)


include/hipSYCL/runtime/signal_channel.hpp
```diff
@@ -51,14 +53,13 @@ class signal_channel {
...
-    auto future = _shared_future;
-    return future.wait_for(std::chrono::seconds(0)) ==
-           std::future_status::ready;
+    return(has_signalled_flag.load());
   }
 
 private:
   std::promise<bool> _promise;
   std::shared_future<bool> _shared_future;
+  std::atomic<bool> has_signalled_flag;
```
illuhad: `_has_signalled_flag` for coding convention :)

include/hipSYCL/runtime/signal_channel.hpp
```diff
@@ -51,14 +53,13 @@ class signal_channel {
   }
 
   bool has_signalled() const {
-    auto future = _shared_future;
-    return future.wait_for(std::chrono::seconds(0)) ==
-           std::future_status::ready;
+    return(has_signalled_flag.load());
```
illuhad: `atomic<T>` has conversion operators and assignment operators for `T`, so you don't need to use explicit `load()` and `store()`.

PR #606, opened 2021-08-05
Support accessor CTAD without deduction tags

illuhad:
Previously, `sycl::accessor acc{buff, cgh}` would not work because there were no deduction guides. Additionally, `sycl::accessor acc{buff, cgh, sycl::no_init}` would not do what was expected since `no_init` would be interpreted as deduction tag.

The spec is not super clear, but I think we should allow `sycl::accessor acc{buff, cgh}` since people will be using it. And it makes sense to use it in conjunction with the hipSYCL accessor variant deduction extension.

This PR
* Allows accessor CTAD without deduction tags such as `sycl::accessor acc{buff, cgh}`. In this case, the access mode will be following the default template argument of the accessor class: `read_write` if `T` is non-const, `read` if T is `const`. This is implemented using new, additional deduction guides.
* Restricts the types that are accepted as valid deduction tags. This fixes e.g. `no_init` being misinterpreted as deduction tag.

### DIFF
--- a/include/hipSYCL/sycl/libkernel/accessor.hpp
+++ b/include/hipSYCL/sycl/libkernel/accessor.hpp
@@ -437,7 +437,18 @@ protected:
   glue::embedded_pointer<T> _ptr;
 };
 
+template<class T>
+inline constexpr auto default_access_tag(){
+  if constexpr(std::is_const_v<T>)
+    return sycl::read_only;
+  else
+    return sycl::read_write;
+}
 
+template<class T>
+inline constexpr sycl::access_mode default_access_mode() {
+  return std::is_const_v<T> ? access_mode::read : access_mode::read_write;
+}
 
 } // detail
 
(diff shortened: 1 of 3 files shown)


keryell (commented):
At least `mode_tag` instead of `tag` is a progress in answering https://github.com/KhronosGroup/SYCL-Docs/issues/174

illuhad:
@hipSYCL-ci-bot test.

include/hipSYCL/sycl/libkernel/accessor.hpp
```diff
@@ -1221,63 +1228,125 @@ using unranged_placeholder_accessor =
 // Accessor deduction guides
 #ifdef HIPSYCL_EXT_ACCESSOR_VARIANT_DEDUCTION
  #define HIPSYCL_ACCESSOR_VARIANT_SELECTOR(TagT, fallback, optimized) \
-  detail::deduce_accessor_variant<TagT>(optimized)
+  detail::deduce_accessor_variant(TagT, optimized)
```
fodinabor: `TagT` implies a type, bug as below you rather would want `tag`, as it's a value, I'd say.

fodinabor (approved):
One nit, otherwise LGTM :)

include/hipSYCL/sycl/libkernel/accessor.hpp
illuhad: Thanks, fixed :)

PR #607, opened 2021-08-07
rename group-algorithms to match SYCL 2020 rev.3

DieGoldeneEnte:
Thanks @fodinabor for noticing :)

This renames group_x functions to x(_over)_group and adds the old names as deprecated alias. The aliases are added in an additional file (`group_functions_alias.hpp`), since the functions are defined across multiple files and to not clutter the files containing the implementation.
The tests now also use the new names instead of the old ones.


fodinabor:
Thanks for the quick fix 😏
@hipsycl-ci-bot test real quick

illuhad:
Thank you!

We already have some aliases for the `joint_` functions in `group_functions.hpp`, which then call the old names. Do I see it correctly that with this
* the functions `_over_group` are implemented as defined in the SYCL 2020 spec, with aliases for backwards compatibility
* the `joint_` functions are implemented with their old names, with aliases for SYCL 2020 compatibility.
?

If so, would it make sense to adopt SYCL 2020 names universally and only provide aliases for backwards compatibility? I.e. apply this change to the `joint_` functions as well?

DieGoldeneEnte:
The joint_* functions were not in the preliminary specification and the implementation is in the detail namespace, as a result I don't think we need to have an alias for them. I could rename the implementation and move it out of the detail namespace.
The only change regarding the joint_* functions in this PR was to use them in the tests, instead of the implementation.

illuhad:
>  I could rename the implementation and move it out of the detail namespace.

This sounds reasonable to me!

DieGoldeneEnte:
@hipsycl-ci-bot test again please :)

The is_group type trait is in the group_functions.hpp file, while I don't think it really belongs there I couldn't find a better location ':D

I left the leader_* functions (used by scoped parallelism) in the detail namespace until they can be selected automatically.

illuhad (approved):
LGTM!

PR #610, opened 2021-08-10
Use R-Value references in buffer::get_access

### LINKED ISSUE #609: How to avoid dag_direct_scheduler warning "Detected a requirement ... that accesses uninitialized data"?
GROMACS is using hipSYCL to call rocFFT to do 3D FFTs, first forward, then backward. The ability to enqueue a custom operation has been very useful for that, so thanks!

However, regardless of the build configuration of either GROMACS or hipSYCL, I get a run-time warning:
`[hipSYCL Warning] dag_direct_scheduler: Detected a requirement that is neither of discard access mode (SYCL 1.2.1) nor noinit property (SYCL 2020) that accesses uninitialized data. Consider changing to discard/noinit. Optimizing potential data transfers away.`

The code triggering that warning is https://gitlab.com/gromacs/gromacs/-/merge_requests/1833/diffs#c95806302062c303359425a19287ab63d40b9a16_0_342, where `impl_->realGrid_` is filled with data and `impl_->complexGrid_` is allocated but not filled. If instead I declare like `auto complexGridAccessor = cl::sycl::accessor<float, 1, mode::write>(*impl_->complexGrid_, cgh, cl::sycl::property::no_init{});` the warning remains. If I fill the complex grid before this kernel is called, then the warning goes away, but as this buffer is strictly write-only, I should not need to do that.

How should we write the code to avoid generating such a warning?

One can suppress it by setting`HIPSYCL_DEBUG_WARNING` appropriately, but I think it would be better to fix in code somewhere.

I can produce a minimal reproducer if that would be useful.

illuhad:
Potential fix for variadic `buffer::get_access` issue reported in #609

### DIFF
--- a/include/hipSYCL/sycl/buffer.hpp
+++ b/include/hipSYCL/sycl/buffer.hpp
@@ -674,12 +674,12 @@ public:
   }
 
   template<typename... Args>
-  auto get_access(Args... args) {
+  auto get_access(Args&&... args) {
     return accessor{*this, args...};
   }
 
   template<typename... Args>
-  auto get_host_access(Args... args) {
+  auto get_host_access(Args&&... args) {
     return host_accessor{*this, args...};
   }
 
--- a/include/hipSYCL/sycl/buffer.hpp
+++ b/include/hipSYCL/sycl/buffer.hpp
@@ -37,6 +37,7 @@
 #include <mutex>
 #include <type_traits>
 #include <algorithm>
+#include <utility>
 
 #include "hipSYCL/common/debug.hpp"
 #include "hipSYCL/runtime/allocator.hpp"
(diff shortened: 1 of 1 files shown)


include/hipSYCL/sycl/buffer.hpp
```diff
@@ -674,12 +674,12 @@ class buffer : public detail::property_carrying_object
...
 
   template<typename... Args>
-  auto get_access(Args... args) {
+  auto get_access(Args&&... args) {
     return accessor{*this, args...};
   }
 
   template<typename... Args>
-  auto get_host_access(Args... args) {
+  auto get_host_access(Args&&... args) {
```
fodinabor, line 683: ain't this a forward reference -> you'd ideally do `std::forward<Args>(args)...`?

illuhad: You are right, using `std::forward` is better. Fixed :)

fodinabor (approved):
LGTM now :)

illuhad:
@hipSYCL-ci-bot test please

mabraham:
Works for me. Fix looks good also! Thanks!

PR #577, opened 2021-06-08
Update pruning IR pass.

fodinabor:
This cleans up the function pruning and moves the globals pruning into an accordingly named pass.
To support the new pass manager infrastructure which will be the default for Clang 13+.
The new PM can be activated in Clang 11/12 with -fexperimental-new-pass-manager.
The option -fpass-plugin= is already present in Clang 11.

PR #611, opened 2021-08-11
Don't build NewPM passes for LLVM < 11.

fodinabor:
As mentioned in #577 a quick fix, if compatibility with previous LLVM versions is broken, is to extend the preprocessor check, which is hereby done. I don't really expect the new PM to be used with LLVM < 11, so no one will miss the IR pass there and it saves me from installing yet another LLVM version.... 💀 

@psalz could you pull this change in and verify it indeed fixes the LLVM 10 build?


fodinabor:
@hipsycl-ci-bot test this with almost modern Clang please

psalz (approved):
> @psalz could you pull this change in and verify it indeed fixes the LLVM 10 build?

Yes, builds again. Thanks for the quick fix!

illuhad (approved):

PR #616, opened 2021-08-16
Callset might not be complete when marking local variables in hierarchical.

fodinabor:
CompleteCallSet only succeeds when the full AST is built.
As some functions might be predeclared and not yet defined when Visitor.VisitFunctionDecl is called, the CCS traversal can't add calls from that predeclared functions.
Thus CCS should only be used from within HandleTranslationUnitDecl.


illuhad:
@hipSYCL-ci-bot test

illuhad (approved):

PR #617, opened 2021-08-18
implement group_barrier default argument as overload

DieGoldeneEnte:
Use function-overloads instead of default argument for group_barrier.


DieGoldeneEnte:
@hipsycl-ci-bot test please :]

include/hipSYCL/sycl/libkernel/cuda/group_functions.hpp
```diff
@@ -52,12 +52,17 @@ T group_broadcast(sub_group g, T x,
 // barrier
 template<typename Group>
 HIPSYCL_KERNEL_TARGET
-inline void group_barrier(Group g, memory_scope fence_scope = Group::fence_scope) {
+inline void group_barrier(Group g, memory_scope fence_scope) {
```
illuhad: What do we need this template argument for? Couldn't we just overload with `sycl::group<int Dim>` instead of allowing pretty much anything as argument? After all, this barrier implementation assumes work groups, right?

include/hipSYCL/sycl/libkernel/cuda/group_functions.hpp
```diff
@@ -69,6 +74,11 @@ inline void group_barrier(sub_group g, memory_scope fence_scope) {
   }
   __syncwarp(); // not necessarily needed, but might improve performance
 }
+template<>
```
illuhad: Do we need the `template<>` here? Can we not make it a pure overload?

include/hipSYCL/sycl/libkernel/cuda/group_functions.hpp
DieGoldeneEnte: Originally it was to stay in line with the signature presented in the specification. We could adjust it, but then I would do it for all group functions, which have specializations for sub_groups to keep it consistent.

illuhad: I think this might be a step that we might have to take in light of the fact that we will soon get yet another work group type for scoped parallelism ;)

Do you want to do this change in this PR or in a subsequent one? I'm fine with either.

include/hipSYCL/sycl/libkernel/cuda/group_functions.hpp
DieGoldeneEnte: We could go pure overload. Again I would apply it to all sub_group specialized functions, as above.

include/hipSYCL/sycl/libkernel/cuda/group_functions.hpp
DieGoldeneEnte: I think a second PR would make it easier to check the diff for errors.

DieGoldeneEnte: Should I use `int Dim` in the template in combinaion with `sycl::group<DIM>` or, as it is currently done for the joint_* functions, `std::enable_if_t<is_group_v<std::decay_t<Group>>, bool> = true`?
Also the PR adjusting the templates makes sense to be merged after #607, so I can make the changes on the already renamed functions.

illuhad: Overload for `sycl::group` specifically, please. The second option is going to break when scoped parallelism 2.0 hits, which already comes with its own `group_barrier` overloads for the new `sp_group` type :)

> Also the PR adjusting the templates makes sense to be merged after #607, so I can make the changes on the already renamed functions.

Okay, so we'll pull in #607 and then #617.

include/hipSYCL/sycl/libkernel/host/group_functions.hpp
```diff
@@ -388,12 +388,22 @@ inline void group_barrier(Group g, memory_scope fence_scope = Group::fence_scope
     g.barrier();
```
fodinabor: btw. there's also system scope.. should at least do a `g.barrier()` there, too? Maybe add a `switch` without default, so this (hopefully) generates a warning if enum values are missing?
https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#_memory_scope

illuhad: hm, this looks fishy indeed - don't we need a barrier in all cases if the `Group` is a work group, regardless of the fence scope?
If the group is a work group, wouldn't you always expect work group barrier semantics?

DieGoldeneEnte: yes, we always need a barrier and a memory-fence for the device-scope, no idea how I messed that up :/

DieGoldeneEnte:
I rebased this commit on top of #607 and adjusted the template parameters for all files, so that they now use `group<Dim>`, `sub_group` or `Group` with an `enable_if_t<is_group_v...`.

I also adjusted `group_barrier` on host to always synchronize and call `detail::mem_fence()` for device-scope. This function is not implemented yet (does nothing on host), but I didn't want to add the implementation to this PR as it is already big enough.

@hipsycl-ci-bot test please

illuhad (approved):

illuhad:
Thanks!

PR #618, opened 2021-08-23
Minor improvements around sycl::backend

al42and:
1. Added backend macros, defining the available enum values. Per [SYCL 2020 spec, paragraph 4.1.1](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:backend-macros).
2. Fixes a conditional in backend name parsing.
3. Updated serializer for platform/API/backend to include all possible enum values.

### DIFF
--- a/include/hipSYCL/sycl/backend.hpp
+++ b/include/hipSYCL/sycl/backend.hpp
@@ -37,6 +37,12 @@ namespace sycl {
 
 using backend = hipsycl::rt::backend_id;
 
+// Per SYCL 2020, 4.1.1 Backend macros
+#define SYCL_BACKEND_CUDA
+#define SYCL_BACKEND_HIP
+#define SYCL_BACKEND_LEVEL_ZERO
+#define SYCL_BACKEND_OMP
+
 #if defined(HIPSYCL_PLATFORM_CPU) && defined(__HIPSYCL_ENABLE_OMPHOST_TARGET__)
  #define SYCL_EXT_HIPSYCL_BACKEND_OMPHOST
 #endif
--- a/src/runtime/serialization/serialization.cpp
+++ b/src/runtime/serialization/serialization.cpp
@@ -48,6 +48,9 @@ std::ostream &operator<<(std::ostream &out, const hardware_platform value) {
   case rt::hardware_platform::rocm:
     out << "ROCm";
     break;
+  case rt::hardware_platform::level_zero:
+    out << "Level Zero";
+    break;
   default:
     out << "<unknown>";
     break;
@@ -57,8 +60,14 @@ std::ostream &operator<<(std::ostream &out, const hardware_platform value) {
 
 std::ostream &operator<<(std::ostream &out, const api_platform value) {
   switch (value) {
+  case rt::api_platform::cuda:
+    out << "CUDA";
+    break;
   case rt::api_platform::hip:
     out << "HIP";
+    break;
+  case rt::api_platform::level_zero:
+    out << "Level Zero";
     break;
   case rt::api_platform::omp:
     out << "OpenMP";
@@ -77,6 +86,9 @@ std::ostream &operator<<(std::ostream &out, const backend_id value) {
     break;
   case rt::backend_id::cuda:
     out << "CUDA";
+    break;
+  case rt::backend_id::level_zero:
+    out << "Level Zero";
     break;
   case rt::backend_id::omp:
     out << "OpenMP";
--- a/src/runtime/settings.cpp
+++ b/src/runtime/settings.cpp
@@ -62,7 +62,7 @@ std::istream &operator>>(std::istream &istr, std::vector<rt::backend_id> &out) {
       out.push_back(rt::backend_id::hip);
     } else if (name == "ze") {
       out.push_back(rt::backend_id::level_zero);
-    } else if("omp") {
+    } else if (name == "omp") {
       // looking for this, even though we have to allow it always.
       out.push_back(rt::backend_id::omp);
     } else {


include/hipSYCL/sycl/backend.hpp
```diff
@@ -37,6 +37,12 @@ namespace sycl {
 
 using backend = hipsycl::rt::backend_id;
 
+// Per SYCL 2020, 4.1.1 Backend macros
+#define SYCL_BACKEND_CUDA
+#define SYCL_BACKEND_HIP
+#define SYCL_BACKEND_LEVEL_ZERO
+#define SYCL_BACKEND_OMP
```
illuhad, line 44: I think the intent is for this to already be handled by the backend macros below. Since none of our backends are specified by the Khronos group, the following section is relevant for backend feature test macros instead:

 https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#_adding_a_backend

i.e. the macros should in principle be called `SYCL_EXT_HIPSYCL_BACKEND_*`, as defined a couple lines further down :-) However, we also don't follow those extension guidelines from the spec very rigorously - I think adding `ext_` in front of pretty much all our interop and backends is very unattractive... From that perspective I can see that adding non-ext macros additionally makes sense for the sake of consistency.

It's a bit unclear when these macros should be defined.  In earlier versions of the spec it said that they should be available when the backend is "enabled" which is very ambiguous (when the implementation is compiled with runtime support for the backend? When the backend is being targeted so that it can actually execute kernels?)

Currently the macros are defined when the backend is targeted, although we could change that interpretation. I'm not sure if it would make sense to enable them unconditionally. If you want to check if you have, say `HIP` support, but hipSYCL was compiled without HIP and so cannot anything HIP-related, and the macro still says it's there, it's not very useful I think. Do you have a use case for which you would need those macros unconditionally?

al42and: Thanks for thoughtful feedback.

> Do you have a use case for which you would need those macros unconditionally?

My intention was checking the availability of a `sycl::backend::X` value.

Macros being defined conditionally is fine for that. The enum value is not very useful when the backend is not operational.

The name discrepancy, though, is a bit confusing, since: "As the identifiers defined in `enum class backend` are implementation-defined, a SYCL implementation must also define a preprocessor macro for each of these identifiers." ([4.1.1. Backend macros](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:backend-macros)). So, one would expect that the availability of `sycl::backend::cuda` value would be checkable with `SYCL_BACKEND_CUDA`. Not a blocker, of course.

> I think adding ext_ in front of pretty much all our interop and backends is very unattractive...

I can sympathize. A compromise might be making prefixed versions always available, and shorter versions available if a feature macro is defined. That said, I don't have a strong preference either way.

illuhad: > My intention was checking the availability of a sycl::backend::X value.
> Macros being defined conditionally is fine for that. The enum value is not very useful when the backend is not operational.

Got it, this makes sense.

> The name discrepancy, though, is a bit confusing, since: "As the identifiers defined in enum class backend are implementation-defined, a SYCL implementation must also define a preprocessor macro for each of these identifiers." (4.1.1. Backend macros). So, one would expect that the availability of sycl::backend::cuda value would be checkable with SYCL_BACKEND_CUDA. Not a blocker, of course.

Okay, I see the issue. I guess if we wanted to follow the spec that strictly, we should have named the enum values `ext_*` too. In light of this, I think we can proceed with what you have proposed.

Since we also have the `EXT_*` macros already, I think your new macros should be defined under the same conditions for the sake of consistency. Whether the current conditions are the correct ones is less clear. I think the proper condition would be "if the runtime has support", but we don't have a way of querying this at compile time since backend plugins for the runtime are loaded dynamically.

There is a lot of nuance to this especially when things like explicit multipass are involved, where the visible backends in host and device pass might be different.
The current macros are defined if either the particular backend is targeted using `--hipsycl-targets`, or if the compiler supports language extensions from a particular backend. I think the logic there is to make sure that if those macros are defined, the corresponding CUDA/HIP headers are included as well which might pull in code that cannot be compiled by a regular compiler.

So, if one compiles with `--hipsycl-targets="hip:gfx906"`, the feature test macro for CUDA would be undefined even though the setup would be perfectly capable of performing CUDA interop to call e.g. cuBLAS, assuming hipSYCL's CUDA plugin has been built...

I guess this is a more fundamental issue here that needs some more thought.. I don't want to delay this PR more than necessary, so I'd say, let's just move your macros to the `EXT_*` ones for now, such that they are defined under the same conditions and we are good to go :-)

al42and: > Okay, I see the issue. I guess if we wanted to follow the spec that strictly, we should have named the enum values ext_* too. In light of this, I think we can proceed with what you have proposed.

Another option is to have both, as IntelLLVM does: https://github.com/intel/llvm/blob/97f916e596fc62c00666aace9b211de87b1f7561/sycl/include/CL/sycl/backend_types.hpp#L21

Intel are deprecating "short" names, but, I personally don't see any harm in keeping them around.

> The current macros are defined if either the particular backend is targeted using --hipsycl-targets, or if the compiler supports language extensions from a particular backend. I think the logic there is to make sure that if those macros are defined, the corresponding CUDA/HIP headers are included as well which might pull in code that cannot be compiled by a regular compiler.

Can't `__HIPSYCL_ENABLE_*_TARGET__` be used for this?

> I don't want to delay this PR more than necessary, so I'd say, let's just move your macros to the EXT_* ones for now, such that they are defined under the same conditions and we are good to go :-)

That's not urgent, so I'd prefer waiting and arriving to a nice, stable solution than having the meaning of macros subtly change between versions :)

illuhad: > Can't __HIPSYCL_ENABLE_*_TARGET__ be used for this?

They are currently already in use for this purpose. There are limitations to this approach especially in an explicit multipass scenario where multiple device backends are targeted simultaneously.

Because of this, #683 goes in a different direction: The latest commits there introduce the change that backend interop is enabled unconditionally, no matter which backends are targeted or supported by the runtime. We can do this because we can just forward declare e.g. `CUStream_t*` which is the same as `cudaStream_t`. With this change, it is no longer necessary to include actual backend headers to enable the interop API.

Of course,the interop API will only succeed if you have a corresponding device available.

So, we could also simplify this PR to just unconditionally enable both `SYCL_EXT_BACKEND*` and `SYCL_BACKEND_*`.

PR #621, opened 2021-08-29
Make `syclcc` directly usable on Windows

fodinabor:
This mostly fixes using `syclcc` for combined compilation and linking on Windows. To keep MinGW happy, the `-l:` prefix is used instead of `-l`.

Apart from that, the CUDA driver lib search is now also performed if CUDA is found, the first time calling CMake and some QoL fixes in the nbody example.


illuhad (approved):
LGTM!

PR #623, opened 2021-08-31
Workaround CUDA / MS ABI mismatch on Windows.

fodinabor:
On Windows the MS ABI is used for host code, that also influences the outline of structs.
For some obscure reason multiple empty base classes lead to 8 byte padding in MS ABI.
CUDA device code on the otherhand uses Itanium ABI.
Itanium ABI doesn't randomly add padding for empty base classes.
As the size of the accessor variants thus differs on host and device, invalid accesses occur, as the device loads the pointer from the kernel lambda from the wrong position -> bumms.
The standard SYCL accessors are fine as they don't have any empty base classes.

To mitigate this, we add __declspec(empty_bases) as presented in https://devblogs.microsoft.com/cppblog/optimizing-the-layout-of-empty-base-classes-in-vs2015-update-2-3/ to get the Itanium compatible layout.

### DIFF
--- a/include/hipSYCL/sycl/libkernel/accessor.hpp
+++ b/include/hipSYCL/sycl/libkernel/accessor.hpp
@@ -458,11 +458,20 @@ struct no_init : public detail::property {};
 
 inline constexpr property::no_init no_init;
 
+// manually align layout with Itanium ABI, until MS ABI default changes
+// in a "future major version"
+// https://devblogs.microsoft.com/cppblog/optimizing-the-layout-of-empty-base-classes-in-vs2015-update-2-3/
+#if !defined(HIPSYCL_EMPTY_BASES) && defined(_WIN32)
+#define HIPSYCL_EMPTY_BASES __declspec(empty_bases)
+#else
+#define HIPSYCL_EMPTY_BASES
+#endif // HIPSYCL_EMPTY_BASES
+
 template <typename dataT, int dimensions = 1,
           access_mode accessmode = detail::default_access_mode<dataT>(),
           target accessTarget = target::device,
           accessor_variant AccessorVariant = accessor_variant::false_t>
-class accessor
+class HIPSYCL_EMPTY_BASES accessor
     : public detail::accessor_base<std::remove_const_t<dataT>>,
       public detail::accessor::conditional_buffer_pointer_storage<
           detail::accessor::stores_buffer_pointer(AccessorVariant)>,


include/hipSYCL/sycl/libkernel/accessor.hpp
```diff
@@ -111,6 +111,18 @@ constexpr accessor_variant deduce_accessor_variant(mode_tag<TagDescT> tag,
   }
 }
 
+#if defined(_WIN32) && defined(__clang__) && defined(__CUDA__)
```
illuhad: Might also add `__HIP__` there. There's no ROCm on Windows yet, but in principle it should be affected by the same issue.

CMakeLists.txt
```diff
@@ -233,6 +233,8 @@ if(WIN32)
   if(NOT SEQUENTIAL_LINK_LINE) 
     set(SEQUENTIAL_LINK_LINE ${DEFAULT_WIN32_SEQUENTIAL_LINK_LINE} CACHE STRING "Arguments passed to compiler to link host libraries to SYCL applications")
   endif()
+  # Clang on Windows adds invalid padding to accessor structs, workaround by using standard (inefficient) accessors
+  set(CUDA_EXTRA_CXX_FLAGS "-DHIPSYCL_STRICT_ACCESSOR_DEDUCTION")
```
illuhad: Maybe `CUDA_PLATFORM_SPECIFIC_CXX_FLAGS` or similar - I believe that's the intention?

include/hipSYCL/sycl/libkernel/accessor.hpp
```diff
@@ -111,6 +111,18 @@ constexpr accessor_variant deduce_accessor_variant(mode_tag<TagDescT> tag,
...
+#define WARN_ATTR(condition, text) __attribute__((diagnose_if(condition, text, "warning")))
+#else
+#define WARN_ATTR(condition, text)
+#endif
+
+constexpr bool warn_if_variant_on_windows(accessor_variant V) 
+WARN_ATTR(V != accessor_variant::true_t && V != accessor_variant::false_t,
```
illuhad: I think we could also allow `ranged_placeholder` as it should be just as general as `true_t` and `false_t` and not have any empty base classes.

fodinabor:
Have another look... just found https://devblogs.microsoft.com/cppblog/optimizing-the-layout-of-empty-base-classes-in-vs2015-update-2-3/ with `__declspec(empty_bases)`, which fixes our issue.

@hipsycl-ci-bot test the cleaner change please.

illuhad (approved):

PR #625, opened 2021-09-04
add information on cmake options

bkmgit:
Boost, C compiler and C++ compiler flags for cmake

### DIFF
--- a/doc/installing.md
+++ b/doc/installing.md
@@ -40,6 +40,7 @@ In order to successfully build and install hipSYCL, the following major requirem
 * python 3 (for the `syclcc` and `syclcc-clang` compiler wrappers)
 * `cmake`
 * the Boost C++ libraries (in particular `boost.fiber`, `boost.context` and for the unit tests `boost.test`)
+  * it may be helpful to set the `BOOST_ROOT` `cmake` variable to the path to the root directory of Boost you wish to use if `cmake` does not find it automatically
 
 If hipSYCL does not automatically configure the build for the desired clang/LLVM installation, the following cmake variables can be used to point hipSYCL to the right one:
 * `LLVM_DIR` must be pointed to your LLVM installation, specifically, the **subdirectory containing the LLVM cmake files**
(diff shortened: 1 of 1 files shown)


doc/installing.md
```diff
@@ -60,4 +60,10 @@ $ make install
 ```
 The default installation prefix is `/usr/local`. Change this to your liking.
 
+Other cmake variables that maybe useful to set are
+* `BOOST_ROOT` path to the root directory of Boost
```
fodinabor: Would make more sense around line 47, as it regards dependencies.

fodinabor (commented):
As to whether we should re-document `CMAKE_C{XX}_COMPILER` I don't have a too strong opinion, but feels somewhat unnecessary.. CMake takes the common `CC` and `CXX` env variables into account as well and a real quick search reveals these options as well..

doc/installing.md
```diff
@@ -60,4 +60,10 @@ $ make install
...
+* `CMAKE_C_COMPILER` C compiler to use
```
illuhad: I agree with @fodinabor that redocumenting `CMAKE_C_COMPILER` is definitely unnecessary since hipSYCL does not use `CMAKE_C_COMPILER` anywhere - adding it might maybe rather confuse people. Can you remove it?

doc/installing.md
```diff
@@ -60,4 +60,10 @@ $ make install
...
+* `CMAKE_CXX_COMPILER` C++ compiler to use
```
illuhad: Documenting `CMAKE_CXX_COMPILER` has a bit more value I think, because there is hipSYCL-specific additional meaning: It is also used to set the default C++ compiler for the CPU backend when using `syclcc` later on (which can later be modified using `HIPSYCL_CPU_CXX`). Can you maybe add this information?

doc/installing.md
bkmgit: Done.

doc/installing.md
bkmgit: done.

doc/installing.md
bkmgit: Moved it earlier.

illuhad (approved):
Thank you :)

PR #628, opened 2021-09-07
Implement `known_identity`

DieGoldeneEnte:
This implements `known_identity`, `has_known_identity` as well as corresponding tests.

Changes `buildins.hpp` consist of adding `static_cast<T>` to the scalar branch (and slight reformatting, since some lines are now longer). This doesn't change anything for scalar values, as T is returned anyways. For vectors the `static_cast` makes the branch valid, and doesn't change the return type, since vectors don't use the scalar branch.
In `HIPSYCL_BUILTIN_GENERATOR_UNARY_T_RET_INT` the return value needs to be casted to `int` first, to tell the compiler how the casting should work.

The vector class needed to be adjusted slightly to make it a literal type. For this the constructors are now labeled `constexpr` and `count_num_elements` was moved into the detail namespace (was a private vector-member before).
The `_storage` needs a default initialization, because writing to uninitialized elements is only allowed starting with C++20.

The actual identity implementation is done using a macro, which creates specializations for the default template for both `known_identity` and `has_known_identity`. If no identity is known, a compile time error is triggered.

This PR only includes the `known_identity` struct, but doesn't add code to use it for example when computing reductions.

### DIFF
--- a/include/hipSYCL/glue/generic/hiplike/hiplike_kernel_launcher.hpp
+++ b/include/hipSYCL/glue/generic/hiplike/hiplike_kernel_launcher.hpp
@@ -485,8 +485,8 @@ private:
         ceil_division(allocated_local_mem_size, alignment) *
         alignment;
 
-    // Dynamic local memory layout, reduction known:   [values...]
-    // Dynamic local memory layout, reduction unknown: [values...][initialized_flags...]
+    // Dynamic local memory layout, identity known:   [values...]
+    // Dynamic local memory layout, identity unknown: [values...][initialized_flags...]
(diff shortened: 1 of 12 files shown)


fknorr:
`known_identity` is already implemented according to spec in #578, but not for vector types. Maybe this PR should be rebased onto #578 as a hipSCYL specific extension for `sycl::vec`?

DieGoldeneEnte:
This applies the same/similar solution to the #578 (which should be merged first).

DieGoldeneEnte:
superseded by #778

PR #629, opened 2021-09-09
Fix multi_ptr(accessor) constructor

al42and:
It was using `detail::accessor::get_accessor_ptr` function, which was
removed in ee367866, but remained forward-declared.

We use `accessor::get_pointer()` directly now.


fodinabor:
LGTM, let's @hipsycl-ci-bot test the fix

illuhad (approved):
Thanks, LGTM!

PR #631, opened 2021-09-15
[Win32] `__device__ signbit(long double)` is available in clang >= 12.

fodinabor:
The workaround to add `__device__ signbit(long double)` ourselves is only necessary for Clang < 12 and breaks with Clang 12+ due to redeclaration. Thus extending the macro guard.


illuhad (approved):
LGTM

PR #633, opened 2021-09-28
Only vectorise innermost loop for Scoped Parallelism on the host

tomdeakin:
This change is important for good performance in both LLVM and GCC on CPUs. Without it, the performance is lower, particularly for GCC (and to a lesser degree for LLVM).

### DIFF
--- a/include/hipSYCL/glue/generic/host/iterate_range.hpp
+++ b/include/hipSYCL/glue/generic/host/iterate_range.hpp
@@ -100,20 +100,20 @@ void iterate_range_simd(const sycl::range<Dim> r, Function f) noexcept {
       f(sycl::id<Dim>{i});
     }
   } else if constexpr (Dim == 2) {
-    #ifdef _OPENMP
-#pragma omp simd collapse(2)
-#endif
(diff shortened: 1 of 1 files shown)


fodinabor:
I have a similar suspicion elsewhere, so LGTM.
Do you have any intuition as to why this makes a difference?
My initial guess would be, that the index might not be detectable as being in a unit stride anymore -> more masking / less efficient memory accesses or something in that direction?

Let's @hipsycl-ci-bot test this as well.

illuhad:
Hi Tom,

this is already contained in my local copy of the scoped parallelism branch (I just hadn't pushed it yet), along with some other minor improvements. Is it okay if we just use my version? Otherwise I'd need to split up the commit.

@fodinabor We found it out empirically. IMHO this code is as intended by the OpenMP spec. To me it looks like a limitation in LLVM.

tomdeakin:
@illuhad Your local copy is fine! Just need to not forget to push this 😄

illuhad:
@tomdeakin Done :)

PR #641, opened 2021-10-06
Add support for multiversioning scoped parallelism kernels

illuhad:
This adds support for multiversioned scoped parallelism kernels. This is necessary to handle cases where the group size is not divisible by the subgroup size.


illuhad:
@hipSYCL-ci-bot test

PR #642, opened 2021-10-06
Fix outlining of lambda kernels with auto arguments

### LINKED ISSUE #640: Clang plugin does not handle generic lambdas as kernel function
This example:
```c++
#include <sycl/sycl.hpp>

void myfn() {}

	sycl::queue q;
	q.submit([](sycl::handler& cgh) {
		cgh.parallel_for<class Foo>(sycl::range<1>{16},
		    // [=](sycl::item<1> itm) { myfn(); } // works
		    [=](auto itm) { myfn(); } // doesn't work
		);
	});
	return 0;
}
```
when compiled for CUDA, produces the error:
```
ptxas fatal   : Unresolved extern function '_Z4myfnv'
clang: error: ptxas command failed with exit code 255 (use -v to see invocation)
```
Using hipSYCL @ 6f4ccc76 and Clang 10.0.1. It does not appear to be a regression, I've reproduced it as far back as 2daf8407 (version 0.8.0).

illuhad:
Previously, lambda kernels with `auto` arguments were not outlined correctly. The reason is that during parsing, we first collect all user kernel invocations and outline them later when parsing is finished and the call tree can be expected to be complete. For this, we store the declarations of the `operator()` of the kernel functor arguments. When the kernel functor is a generic lambda however, this `operator()` may not exist yet since it hasn't yet been instantiated. (At this point, we have just started parsing the kernel invocation). If that happens, the kernel will not be registered for outlining later, and kernel code may not be emitted to device when it should be.

This PR resolves this by instead of storing the `operator()` of the user kernel, storing the invocation kernel of the kernel launcher (e.g. `hipsycl::glue::hiplike_dispatch::parallel_for`) that has been instantiated with the user lambda. This function already exists and can therefore be stored for outlining.

I was briefly concerned whether this would affect the implementation of hierarchical parallelism, since hierarchical tries to allocate declarations in the outer scope in local memory. Since the outer scope has now moved up in the call tree to the `parallel_for_work_group` kernel launcher, in principle it might happen that variables from there end up in local memory. In practice however this cannot happen, since as a simple heuristic, we only store variables of functions in local memory that have a `group` as argument.

Fixes #640

### DIFF
--- a/include/hipSYCL/compiler/Frontend.hpp
+++ b/include/hipSYCL/compiler/Frontend.hpp
@@ -118,7 +118,7 @@ class CompleteCallSet : public clang::RecursiveASTVisitor<CompleteCallSet> {
       // fixme: investigate where the invalid decls come from..
       if(!D)
         return true;
-
+      
       clang::Decl* DefinitionDecl = D;
       clang::FunctionDecl* FD = clang::dyn_cast<clang::FunctionDecl>(D);
 
@@ -246,20 +246,16 @@ public:
     if(!F) return true;
     if(!CustomAttributes::SyclKernel.isAttachedTo(F)) return true;
 
+    // Store user kernel for it to be marked as device code later on. We store
+    // the kernel invoke itself instead of user lambda since the user lambda()
+    // will not have an operator() that we can inspect for further outlining if
+    // it is generic (i.e., has auto parameter).
+    UserKernels.insert(F);
+
     auto KernelFunctorType = llvm::dyn_cast<clang::RecordType>(Call->getArg(0)
       ->getType()->getCanonicalTypeUnqualified());
-
-    // Store user kernel for it to be marked as device code later on
-    if(KernelFunctorType)
-    {
-      auto Methods = llvm::dyn_cast<clang::CXXRecordDecl>(
-        KernelFunctorType->getDecl())->methods();
-      for(auto&& M : Methods)
-      {
-        if(M->getNameAsString() == "operator()")
-          UserKernels.insert(M);
-      }
-    }
+    if(!KernelFunctorType)
+      return true;
 
     // Determine unique kernel name to be used for symbol name in device IR
     clang::FunctionTemplateSpecializationInfo* Info = F->getTemplateSpecializationInfo();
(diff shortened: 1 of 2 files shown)


illuhad:
@hipSYCL-ci-bot test

fknorr:
I realize that the group-parameter heuristic is already part of hipSYCL, but It got me thinking about this edge case:

```c++
#include <CL/sycl.hpp>

int main() {
    cl::sycl::buffer<int> b{{2}};

    cl::sycl::queue q;
    q.submit([&](cl::sycl::handler &cgh) {
        auto a = b.get_access<cl::sycl::access::mode::discard_write>(cgh);
        cgh.parallel_for_work_group<class k>(cl::sycl::range<1>{1}, cl::sycl::range<1>{2}, [=](cl::sycl::group<1> g) {
            [=] { // <- g is captured here and operator() does not take a group so ...
                int local_mem[2]; // <- ... this variable is not put into local memory
                g.parallel_for_work_item([&](cl::sycl::h_item<1> h) {
                    int i = h.get_global_id(0);
                    local_mem[i] = i + 1;
                });
                g.parallel_for_work_item([&](cl::sycl::h_item<1> h) {
                    int i = h.get_global_id(0);
                    a[i] = local_mem[1 - i];
                });
            }();
        });
    }).wait();

    auto a = b.get_access<cl::sycl::access::mode::read>();
    for (int i = 0; i < 2; ++i) {
        printf("[%d] = %d\n", i, a[i]);
    }
}
```

hipSYCL allocates `local_mem` in private memory and prints:
```
[0] = 0
[1] = 0
```

DPC++ places it in local memory and prints:
```
[0] = 2
[1] = 1
```

For hipSYCL, removing the `[=]{}()` wrapper produces the same result as DPC++.

The spec is not really clear on what should happen here, it talks about `parallel_for_work_group` and `parallel_for_work_item` _scopes_, so one could argue that this is not even properly defined.

Just something to keep in mind.

illuhad:
@fknorr You're right, this is an issue. My interpretation of hierarchical is that the variable in your example should be put in local memory - arguably, *the entire lambda* should be in local memory, since it is declared in the parallel for work group scope AND we would need to treat members of function objects the same as function arguments when applying the `sycl::group` heuristic.

However, the 1.2.1 hierarchical implementation in hipSYCL is incomplete anyway and has some even more severe issues. For example, if you initialize a variable that is going to be put in local memory, clang will just silently ignore the initialization, because local/`__shared__` memory cannot be initialized. This can silently lead to incorrect code and is very hard to figure out from a user's point of view.

Given the unclear future that hierarchical has, being discouraged in the 2020 spec, and because we have something far better with #619, I don't think this is something to worry about or put effort in fixing at the moment. But it's good to be aware of this, so thanks for the heads up :)

PR #643, opened 2021-10-08
Fix cmake config in case Boost is found using config mode

sbalint98:
If Boost is found with the Config mode of `find_package`, variables like `Boost_CONTEXT_LIBRARY_DEBUG` will not be set. The unset variable causes the `get_filename_component` to be called with the wrong number of arguments.

Similar to other portions of the cmake configuration, to other parts of the cmake configuration, this problem can be alleviated by an additional check that needs to be introduced, to check whether we are finding Boost via the config or the module mode.


CMakeLists.txt
```diff
@@ -202,12 +202,14 @@ else()
   set(DEFAULT_LINK_PREFIX "-l")
 endif()
 
-if(CMAKE_BUILD_TYPE MATCHES Debug)
-  get_filename_component(Boost_CONTEXT_LIBRARY_NAME ${Boost_CONTEXT_LIBRARY_DEBUG} NAME)
-  get_filename_component(Boost_FIBER_LIBRARY_NAME ${Boost_FIBER_LIBRARY_DEBUG} NAME)
-else()
-  get_filename_component(Boost_CONTEXT_LIBRARY_NAME ${Boost_CONTEXT_LIBRARY_RELEASE} NAME)
-  get_filename_component(Boost_FIBER_LIBRARY_NAME ${Boost_FIBER_LIBRARY_RELEASE} NAME)
+if(Boost_CONTEXT_LIBRARY_DEBUG)
```
fodinabor: If `Boost_CONTEXT_LIBRARY_RELEASE` isn't set coming from the CONFIG, we still have the problem, that this `DEBUG` variable will be unavailable for `CMAKE_BUILD_TYPE == Release` as well.. so probably `Boost_CONTEXT_LIBRARY_DEBUG OR Boost_CONTEXT_LIBRARY_RELEASE`?

fodinabor (commented):
And I hoped this would be always be set, if boost is found :/
Might it be the case that sometimes only `Boost_FIBER_LIBRARY_RELEASE` is found if installed from e.g. APT? Could we then just use whichever is found?

sbalint98:
How could the case happen where only one is found? I think you cannot mix config mode with module mode, so either both of these variables are set or none, since otherwise we would get an error by the `find_package`.

CMakeLists.txt
```diff
@@ -202,12 +202,14 @@ else()
...
+if(Boost_CONTEXT_LIBRARY_DEBUG OR Boost_CONTEXT_LIBRARY_RELEASE)
+  if(CMAKE_BUILD_TYPE MATCHES Debug)
+    get_filename_component(Boost_CONTEXT_LIBRARY_NAME ${Boost_CONTEXT_LIBRARY_DEBUG} NAME)
+    get_filename_component(Boost_FIBER_LIBRARY_NAME ${Boost_FIBER_LIBRARY_DEBUG} NAME)
+  else()
+    get_filename_component(Boost_CONTEXT_LIBRARY_NAME ${Boost_CONTEXT_LIBRARY_RELEASE} NAME)
+    get_filename_component(Boost_FIBER_LIBRARY_NAME ${Boost_FIBER_LIBRARY_RELEASE} NAME)
+  endif()
 endif()
 set(DEFAULT_WIN32_OMP_LINK_LINE "-L${Boost_LIBRARY_DIR} ${DEFAULT_LINK_PREFIX}${Boost_CONTEXT_LIBRARY_NAME} ${DEFAULT_LINK_PREFIX}${Boost_FIBER_LIBRARY_NAME} ${DEFAULT_OMP_FLAG}")
```
fodinabor, line 275: Okay, I guess that `Boost_CONTEXT_LIBRARY_*` both aren't set then.. we will have to get hold of `Boost_CONTEXT_LIBRARY_NAME` and `Boost_FIBER_LIBRARY_NAME` differently then, could we just integrate this in the if/elseif/else from line 168 on?
That would mean using ` get_filename_component(FIBER_LIBRARY_NAME} ${DEFAULT_OMP_FLAG}")
 "${FIBER_IMPORT_LIBRARY}" NAME)` and add the same for context?

CMakeLists.txt
```diff
@@ -167,17 +167,33 @@ endif()
...
+  get_filename_component(Boost_FIBER_LIBRARY_NAME "${FIBER_IMPORT_LIBRARY}" NAME)
   get_target_property(Boost_INCLUDE_DIR Boost::fiber INTERFACE_INCLUDE_DIRECTORIES)
 endif()
 
+if(Boost_CONTEXT_LIBRARY_DEBUG)
+  get_filename_component(Boost_CONTEXT_LIBRARY_NAME ${Boost_CONTEXT_LIBRARY_DEBUG} NAME)
+elseif(Boost_CONTEXT_LIBRARY_RELEASE)
+  get_filename_component(Boost_CONTEXT_LIBRARY_NAME ${Boost_CONTEXT_LIBRARY_RELEASE} NAME)
+else()
+  get_target_property(CONTEXT_IMPORT_LIBRARY Boost::fiber IMPORTED_LOCATION_DEBUG)
```
fodinabor: `Boost::context`

CMakeLists.txt
```diff
@@ -167,17 +167,33 @@ endif()
...
+  if(NOT CONTEXT_IMPORT_LIBRARY)
+    get_target_property(CONTEXT_IMPORT_LIBRARY Boost::fiber IMPORTED_LOCATION_RELEASE)
```
fodinabor: as above

CMakeLists.txt
```diff
@@ -167,17 +167,33 @@ endif()
...
+  endif()
+  get_filename_component(Boost_CONTEXT_LIBRARY_NAME "${FIBER_IMPORT_LIBRARY}" NAME)
```
fodinabor: `FIBER_IMPORT_LIBRARY`

CMakeLists.txt
```diff
@@ -167,17 +167,33 @@ endif()
 
 if(Boost_FIBER_LIBRARY_DEBUG)
   get_filename_component(Boost_LIBRARY_DIR ${Boost_FIBER_LIBRARY_DEBUG} DIRECTORY)
+  get_filename_component(Boost_FIBER_LIBRARY_NAME ${Boost_FIBER_LIBRARY_DEBUG} NAME)
 elseif(Boost_FIBER_LIBRARY_RELEASE)
   get_filename_component(Boost_LIBRARY_DIR ${Boost_FIBER_LIBRARY_RELEASE} DIRECTORY)
+  get_filename_component(Boost_FIBER_LIBRARY_NAME ${Boost_FIBER_LIBRARY_RELEASE} NAME)
 else()
   get_target_property(FIBER_IMPORT_LIBRARY Boost::fiber IMPORTED_LOCATION_DEBUG)
   if(NOT FIBER_IMPORT_LIBRARY)
```
fodinabor: hmm... as we now use this to get the actual library as well, we should consider the build type again... (this is important for windows at least, where you can't link a debug library against a release project and the other way around.. the CI doesn't catch this, as we use a self-built boost only including Release and only build hipSYCL for Release as well)

not sure what's a _nice_ way to do that correctly though... On Windows I'd probably just use
```CMake
if(CMAKE_BUILD_TYPE MATCHES Debug AND Boost_FIBER_LIBRARY_DEBUG)
  get_filename_component(Boost_LIBRARY_DIR ${Boost_FIBER_LIBRARY_DEBUG} DIRECTORY)
  get_filename_component(Boost_FIBER_LIBRARY_NAME ${Boost_FIBER_LIBRARY_DEBUG} NAME)
elseif(NOT CMAKE_BUILD_TYPE MATCHES Debug AND Boost_FIBER_LIBRARY_RELEASE)
  get_filename_component(Boost_LIBRARY_DIR ${Boost_FIBER_LIBRARY_RELEASE} DIRECTORY)
  get_filename_component(Boost_FIBER_LIBRARY_NAME ${Boost_FIBER_LIBRARY_RELEASE} NAME)
else()
  if(CMAKE_BUILD_TYPE MATCHES Debug)
    get_target_property(FIBER_IMPORT_LIBRARY Boost::fiber IMPORTED_LOCATION_DEBUG)
  else()
    get_target_property(FIBER_IMPORT_LIBRARY Boost::fiber IMPORTED_LOCATION_RELEASE)
  endif()
  ....
endif()
```
but that would unnecessarily break the build on most Unix distros where only the Release boost libraries are shipped and can be used with Debug builds as well...
And then it get's ugly fast.. either we branch on `if(WIN32)` to do the above or what we have right now, or we always add a fallback to the release library if not on Windows... Might be worth putting that into a CMake macro in an included file then, though... any better ideas?

fodinabor (commented):
I'm sorry... I fear that breaks Windows builds now.. see inline comments...

CMakeLists.txt
sbalint98: Mhmm I am not sure if this is what you have meant, but can you have a look at the current state of this PR? I used macros to simplify the code a bit, let me know what you think ;)

illuhad:
@sbalint98 Is this ready now?

sbalint98:
I am not sure if we want to merge this after all. The user can always define `CXX_FLAGS` and `LINK_LINE` variables, so it might not make too much sense to further complicate this CMake file...

PR #644, opened 2021-10-08
Make all test cases pass on HIP

illuhad:
Previously there were some minor issues that prevented clean runs of the test suite on HIP:
* Due to lack of proper shared allocations in ROCM, USM pointer query tests are expected to fail. This PR disables those tests when the device is a HIP device.
* There was a minor issue with instrumentation for prefetch operations in HIP, since those would not by default be enabled because earlier ROCm versions did not have the `hipMemPrefetchAsync` API. This PR enables utilizing this API by default and makes sure that instrumentation is possible even when prefetches are mapped to no-ops because the API is unavailable.

This PR is a prerequisite to integrating AMD GPUs in CI.

### DIFF
--- a/include/hipSYCL/sycl/device.hpp
+++ b/include/hipSYCL/sycl/device.hpp
@@ -36,7 +36,7 @@
 #include "types.hpp"
 #include "aspect.hpp"
 #include "info/info.hpp"
-#include "libkernel/backend.hpp"
+#include "backend.hpp"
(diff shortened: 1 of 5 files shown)


fodinabor:
LGTM .. hyped :)

illuhad:
Let's see if this works...

@hipSYCL-ci-bot test...

PR #645, opened 2021-10-11
Optimize data user tracking by replacing weaker dependencies

### LINKED ISSUE #630: Slow kernel submission in loop
**Bug summary**
A seemingly simple time series kernel gets considerably slow after couple hundred timesteps.

**To Reproduce**
Compile and execute the attached code. On my system, it usually starts to slow down after approximately 200 timesteps, in which some time steps taking up to 2 seconds.

A reset in the time taken to submit a new kernel can also be observed periodically.
```
Timestep 0 submission took 0 ms
...
Timestep 100 submission took 14 ms
...
Timestep 202 submission took 181 ms
...
Timestep 303 submission took 544 ms
...
Timestep 304 submission took 28 ms
...
Timestep 413 submission took 1658 ms
```

```cpp
#include <chrono>
#include <sycl/sycl.hpp>

int main() {
  float ha = 42;
  float hr = 1337;
  sycl::buffer<float, 1> ba(&ha, 1);
  sycl::buffer<float, 1> br(&hr, 1);

  sycl::queue queue;

  for (size_t t = 0; t < 1000; t++) {
    const auto before = std::chrono::steady_clock::now();
    queue.submit([&](sycl::handler &cgh) {
      auto a = ba.get_access<sycl::access::mode::read>(cgh);
      auto r = br.get_access<sycl::access::mode::read_write>(cgh);
      cgh.parallel_for(ba.get_range(), [=](sycl::id<1>) {
        (void)a;
        (void)r;
      });
    });
    std::cout << "Timestep " << t << " submission took "
              << std::chrono::duration_cast<std::chrono::milliseconds>(
                     std::chrono::steady_clock::now() - before)
                     .count()
              << " ms\n";
    //  queue.wait_and_throw(); // Makes things faster
  }
  return 0;
}
```

**Describe your setup**
hipSYCL @ 22e94f2
NVIDIA RTX 2070

On running with `perf`, the function `recursive_find` appears to be taking most of the time:
```
Children      Self        Shared Object             Symbol 
98.13%       66.33%    libhipSYCL-rt.so          [.] hipsycl::rt::(anonymous namespace)::recursive_find
```

However, the program seems to work as expected with manual queue synchronization.
Also, please note this behavior happens most of the time while sometimes the program works as expected even without manual synchronization.

illuhad:
The data user tracker is responsible for storing a list of all operations currently using a buffer. This list is used to calculate dependencies for new operations that come in. It is purged when operations finish. However, when many operations are submitted before they can finish, this list can grow considerably in size. This causes the `dag_builder` to potentially add a lot of dependencies to an operation during access conflict detection. This in turn considerably slows down the DAG optimization algorithms which try to find edges that can be elided, as the analysis problem becomes potentially very large.

This PR optimizes this scenario by removing nodes from the data user tracker when a new node is added that we know will already synchronize with the existing ones.

Fixes #630

### DIFF
--- a/include/hipSYCL/runtime/data.hpp
+++ b/include/hipSYCL/runtime/data.hpp
@@ -254,11 +254,23 @@ public:
   bool has_user(dag_node_ptr user) const;
 
   void release_dead_users();
+
+  template<class Predicate>
   void add_user(dag_node_ptr user, 
                 sycl::access::mode mode, 
                 sycl::access::target target, 
-                id<3> offset, 
-                range<3> range);
+                id<3> offset,
+                range<3> range,
+                Predicate replaces_user) {
+    std::lock_guard<std::mutex> lock{_lock};
+
+    _users.erase(std::remove_if(_users.begin(), _users.end(), replaces_user), 
+      _users.end());
+
+    _users.push_back(
+      data_user{std::weak_ptr<dag_node>(user), mode, target, offset, range});
+  }
+
 private:
   std::vector<data_user> _users;
   mutable std::mutex _lock;
--- a/src/runtime/dag_builder.cpp
+++ b/src/runtime/dag_builder.cpp
@@ -31,6 +31,7 @@
 #include "hipSYCL/runtime/operations.hpp"
 #include "hipSYCL/runtime/dag_builder.hpp"
 #include "hipSYCL/runtime/serialization/serialization.hpp"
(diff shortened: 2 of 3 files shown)


illuhad:
@hipSYCL-ci-bot test.

(Linux CI issues are due to issues with upstream LLVM package repositories and unrelated to this PR)

PR #648, opened 2021-10-13
Align vec store and load with spec

### LINKED ISSUE #647: vec::load and vec::store do not implement the SYCL 2020 spec
**Bug summary**

`sycl::vec<dataT,numElements>::load` is documented with `Loads the values at the address of ptr offset in elements of type dataT by numElements * offset, into the components of this SYCL vec.` at https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#_vec_interface.

The hipsycl implementation at https://github.com/illuhad/hipSYCL/blob/develop/include/hipSYCL/sycl/libkernel/vec.hpp#L528 does not scale `offset` with `numElements` leading to incorrect results.

The matching store operation has the same problem. Perhaps this is related to the change of meaning in sycl 2020 noted in the comment there?

**To Reproduce**
Use an `accessor<float> a` indexed by `i` and a `sycl::float2 f2` and attempt to `f2.load(i, a.get_pointer())`

**Expected behavior**

That it is straightforward to load(store) a `float2` from(to) an accessor of `float` without manually scaling `i` by `numElements`.

**Describe your setup**

hipSYCL develop branch commit 6f4ccc76e41aae running on ROCm 4.3 targeting an MI50

**Additional context**
GROMACS does a 3D-FFT that may or may not be in-place, so we have a 3D grid expressed as a 1-D buffer of `float` that logically contains `complex` values that are convenient to operate on as `sycl::float2`

illuhad:
Fixes #647

### DIFF
--- a/include/hipSYCL/sycl/libkernel/vec.hpp
+++ b/include/hipSYCL/sycl/libkernel/vec.hpp
@@ -531,14 +531,14 @@ public:
   HIPSYCL_UNIVERSAL_TARGET
   void load(size_t offset, multi_ptr<const T, AddressSpace> ptr) {
     for(int i = 0; i < N; ++i)
-      _data[i] = ptr.get()[offset + i];
+      _data[i] = ptr.get()[offset * N + i];
   }
 
   template <access::address_space AddressSpace>
   HIPSYCL_UNIVERSAL_TARGET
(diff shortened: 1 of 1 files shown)


illuhad:
@hipSYCL-ci-bot test

mabraham:
LGTM

PR #656, opened 2021-11-03
Wrap queue::_previous_submission in std::shared_ptr to retain reference semantics for queue

### LINKED ISSUE #655: No synchronization between tasks in copies of an in-order queue
**Bug summary**

Per [specs](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:reference-semantics), working with a copy of a SYCL runtime object should be the same as working with the original object itself:

> T must be copy constructible and copy assignable on the host application and within SYCL kernel functions in the case that T is a valid kernel argument. Any instance of T that is constructed as a copy of another instance, via either the copy constructor or copy assignment operator, must behave as-if it were the original instance and as-if any action performed on it were also performed on the original instance and must represent the same underlying native backend object as the original instance where applicable.

However, when one makes copies of an in-order `sycl::queue`, and submits tasks to those copies, the order of operations is not preserved. Apparently, the copy-constructor of `sycl::queue` object copies the `_previous_submission` field, and thus each copy of the queue object only ensures the order of operations submitted to that copy.

**To Reproduce**

- https://gist.github.com/al42and/b1a93f23922a1a5de563064451c85588

The code has a loop and an overcomplicated kernel to increase the chance of clearly demonstrating the problem. The essence is submitting tasks to two queue objects which are copies of the original in-order queue.

- Compile: `syclcc -O3 in_order_not_really.cpp -o in_order_not_really --hipsycl-targets=hip:gfx906`

- Get the runtime trace: `rocprof --stats --hip-trace --parallel-kernels ./in_order_not_really`

- Visualize it with https://ui.perfetto.dev/

image

**Expected behavior**

H2D copy, kernel, and D2H copy are executed in that particular order without overlapping,

**Describe your setup**
* hipSYCL develop (9ebaa0af), 
* ROCm 4.3, clang-13 from ROCm installation, and with MI50 GPU

illuhad:
Hopefully fixes #655

### DIFF
--- a/include/hipSYCL/sycl/queue.hpp
+++ b/include/hipSYCL/sycl/queue.hpp
@@ -411,7 +411,7 @@ public:
     if(is_in_order()) {
       std::lock_guard<std::mutex> lock{*_lock};
 
-      if(auto prev = this->_previous_submission.lock()){
+      if(auto prev = this->_previous_submission->lock()){
         if(!prev->is_complete()) {
           return std::vector<event>{event{prev, _handler}};
         }
@@ -864,7 +864,7 @@ private:
   template <class Cgf>
   rt::dag_node_ptr execute_submission(Cgf cgf, handler &cgh) {
     if (is_in_order()) {
-      auto previous = _previous_submission.lock();
+      auto previous = _previous_submission->lock();
       if(previous)
         cgh.depends_on(event{previous, _handler});
     }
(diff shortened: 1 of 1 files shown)


illuhad:
@hipsycl-ci-bot test.

illuhad:
CI issue seems unrelated to this PR.

al42and:
Can confirm that this PR resolved #655 for me. Thank you!

PR #657, opened 2021-11-12
Add __SYCL_DEVICE_ONLY__

sbalint98:
Previously only `SYCL_DEVICE_ONLY` was defined. However, the standard requires that `__SYCL_DEVICE_ONLY__`  is defined, which is added In this PR.

### DIFF
--- a/include/hipSYCL/sycl/libkernel/backend.hpp
+++ b/include/hipSYCL/sycl/libkernel/backend.hpp
@@ -62,6 +62,9 @@
 
 #ifdef HIPSYCL_LIBKERNEL_DEVICE_PASS
  #define SYCL_DEVICE_ONLY
+ #ifndef __SYCL_DEVICE_ONLY__
+  #define __SYCL_DEVICE_ONLY__
+ #endif
 #endif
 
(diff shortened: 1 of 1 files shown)


psalz:
I can't tell you how often I thought "I should create a PR for this"... Thank you :D

fodinabor (approved):
LGTM! :)

include/hipSYCL/sycl/libkernel/backend.hpp
```diff
@@ -62,6 +62,7 @@
 
 #ifdef HIPSYCL_LIBKERNEL_DEVICE_PASS
  #define SYCL_DEVICE_ONLY
+ #define __SYCL_DEVICE_ONLY__
```
illuhad: Only define `__SYCL_DEVICE_ONLY__` if it's not already defined. When compiling with DPC++ for SPIR-V, DPC++ will already handle it.

sbalint98: Ahh yeah you told me that, sorry

PR #660, opened 2021-11-15
Fix the packaging workflow and readme 

sbalint98:
This PR fixes the packaging workflow and updates the outdated documentation.

The primary problem for the failing packages was that the link and compilation lines were not updated, and therefore the workaround introduced in #559 were not applied to the packages.

 Furthermore, in the `CMakeLists.txt` quotation marks were added for the variables that might not always be set. In case the boost package has been found using the cmake configuration built with boost, the variables `${Boost_FIBER_LIBRARY_DEBUG}` won't be set; therefore, the `get_filename_component` command would receive an invalid number of arguments.


CMakeLists.txt
```diff
@@ -166,9 +166,9 @@ else()
 endif()
 
 if(Boost_FIBER_LIBRARY_DEBUG)
-  get_filename_component(Boost_LIBRARY_DIR ${Boost_FIBER_LIBRARY_DEBUG} DIRECTORY)
+  get_filename_component(Boost_LIBRARY_DIR "${Boost_FIBER_LIBRARY_DEBUG}" DIRECTORY)
 elseif(Boost_FIBER_LIBRARY_RELEASE)
-  get_filename_component(Boost_LIBRARY_DIR ${Boost_FIBER_LIBRARY_RELEASE} DIRECTORY)
+  get_filename_component(Boost_LIBRARY_DIR "${Boost_FIBER_LIBRARY_RELEASE}" DIRECTORY)
```
illuhad: I understand that the quotation marks prevent cmake from receiving an incorrect number of arguments in that call, but don't we just defer the problem to a later time then? i.e., wouldn't `${Boost_LIBRARY_DIR}` then end up empty, causing us to have an incomplete link line for `syclcc`? Shouldn't we just error in this case?
Also, how can this situation arise in the first place - shouldn't the `if(Boost_FIBER_LIBRARY_DEBUG)` etc prevent this already?

install/scripts/README.md
```diff
@@ -32,6 +32,14 @@ Two meta-packages in order to keep consistent with the previous packages:
 
 We require some additional software repos to be enabled (for example, `release-scl` and `epel` for centos 7 ). To make adding these easier, we provide scripts in the `install/scripts/add-hipsyc-repo` for all supported distributions that handles adding these repositories, as well as adding the hipSYCL repo.
 
+Since dpkg does not support upper case letters in the package names, the **ubuntu packages are spelled all lower case**. For example, to install the rocm base packages, one would need to install `hipsycl-base-rocm<-nightly>.
+
+### Legacy packages in the repository
+The hipSYCL package architecture has been updated to follow a plug-in backend architecture. This allows for only shipping the library files for each backend without the rest of the hipSYCL runtime. The packages for the old hipSYCL architecture are still available in the repository; Nonetheless, these are outdated and **not supported**. Please use the packages listed above. The unsupported packages are the following:
+* `hipSYCL-rocm`
+* `hipSYCL-rocm-nightly`
+* `hipSYCL-cuda-nightly`
```
illuhad, line 41: Would it maybe be better to remove them, and have the new packages "provide" the old ones, so that package manager can replace the old ones with the new ones?

install/scripts/README.md
```diff
@@ -19,7 +19,7 @@ Installing using the repositories is beneficial because the hipSYCL installation
 We provide the following packages in both versions:
 
 Base packages: 
-* `hipSYCL-base<-nightly>`
+* `hipSYCL-base<-nightly>` (contains: llvm 11.0.1, boost: 1.7x with cmake )
 * `hipSYCL-base-rocm<-nightly>`
 HipSYCL packages:
 * `hipSYCL-omp<-nightly>`
```
illuhad, line 25: Can you remind me why we haven have those combinations still? With the plugin architecture we should be able to install all backends independently.

install/scripts/spack-install/hipsycl.sh
```diff
@@ -59,8 +59,16 @@ cmake \
...
+-DCUDA_LINK_LINE="-Wl,-rpath=\$HIPSYCL_CUDA_LIB_PATH -L\$HIPSYCL_CUDA_LIB_PATH -lcudart" \
+-DCUDA_CXX_FLAGS="-U__FLOAT128__ -U__SIZEOF_FLOAT128__ -isystem \$HIPSYCL_PATH/include/hipSYCL/std/hiplike" \
+-DOMP_LINK_LINE="-L/opt/hipSYCL/boost/boost/lib -lboost_context -lboost_fiber -Wl,-rpath=/opt/hipSYCL/boost/boost/lib -fopenmp=libomp -L/opt/hipSYCL/llvm/llvm/lib  -Wl,-rpath=/opt/hipSYCL/llvm/llvm/lib " \
+-DOMP_CXX_FLAGS="-I/opt/hipSYCL/boost/boost/include  -fopenmp=libomp" \
+-DSEQUENTIAL_LINK_LINE="-L/opt/hipSYCL/boost/boost/lib -lboost_context -lboost_fiber -lomp -lpthread -Wl,-rpath=/opt/hipSYCL/boost/boost/lib -L/opt/hipSYCL/llvm/llvm/lib -Wl,-rpath=/opt/hipSYCL/llvm/llvm/lib " \
+-DSEQUENTIAL_CXX_FLAGS="-I/opt/hipSYCL/boost/boost/include" \
```
illuhad, line 70: We really should think about how this can be done in a better way in the long run... It's quite easy to break this if we change the default `syclcc` configuration and then forget about this.

install/scripts/README.md
sbalint98: That is true, these are for two reason: In order to keep consistent with the earlier packaging system, and to save the user typing hipsycl twice. Internaly `hipSYCL-omp-cuda` is just  a meta-package depending on `hipSYCL-omp` and `hipSYCL-cuda`

install/scripts/README.md
sbalint98: yes, we could do that

CMakeLists.txt
sbalint98: Sorry you are right the quotation marks are indeed not necessary here, and I agree that even if the code were accessible it wouldn't make a difference since we check in the if statement. I will remove these. Although on line 206-210 I think we need them, since `Boost_CONTEXT_LIBRARY_DEBUG` will only be set when cmake is found by the module mode

al42and: > I will remove these. 

I'd suggest keeping them. I doubt it is common for `Boost_FIBER_LIBRARY_RELEASE` to have spaces or other nasty symbols in it, but I see no harm in handling it properly (at least here).

PR #664, opened 2021-11-17
Fix async worker deadlocks by using notify_all()

### LINKED ISSUE #658: Deadlock when using hipSYCL in multithreaded applications
**Bug summary**

Trying to use hipSYCL from a multi-thread (POSIX threads) application occasionally leads to a never-ending `queue::wait()`. The more threads there are, the likelier it is.

The deadlock happens when calling `queue.wait()` from one of the threads. Looking at a stack trace, the thread is stuck on a `_condition_wait.wait` call inside `worker_thread::wait`, despite the `_enqueued_operations` being empty.

Apparently, this thread was not awoken by `_condition_wait.notify_one()` from the executor thread.

The issue is resolved by replacing all `notify_one()` with `notify_all()` in `async_worker.cpp`, but that may be a bit heavy-handed?

**To Reproduce**

Happens pretty regularly in GROMACS runs, with both CUDA and HIP backends. I can come up with a shorted reproducer if needed, but I hope the description would be sufficient to reason through.

**Expected behavior**

`worker_thread::wait()` should return as soon as there are no more tasks in the queue.

**Describe your setup**
* hipSYCL 46d6b3c (latest `develop`)
   * clang-11 from Ubuntu repos for CUDA build, clang-13 from ROCm for HIP build.
   * CUDA 11.4 (clang-11.0)
   * ROCm 4.3
* Tested with `cuda:sm_75` and `hip:gfx906` backends (separate builds)

illuhad:
Hopefully fixes #658 

However, it's still unclear if this is the best approach or if some of the `notify_all()` calls could stay `notify_one()` calls, especially when they are there to wake up the worker thread specifically.

### DIFF
--- a/src/runtime/generic/async_worker.cpp
+++ b/src/runtime/generic/async_worker.cpp
@@ -54,7 +54,7 @@ void worker_thread::wait()
   if(!_enqueued_operations.empty())
   {
     // Before going to sleep, wake up the other thread to avoid deadlocks
-    _condition_wait.notify_one();
+    _condition_wait.notify_all();
     // Wait until no operation is pending
     _condition_wait.wait(lock, [this]{return _enqueued_operations.empty();});
   }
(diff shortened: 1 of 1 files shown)


al42and:
Can confirm that it fixes my issue :+1:

PR #665, opened 2021-11-17
Add implicit conversion from multi_ptr<T> to multi_ptr<const T>

al42and:
Per SYCL 2020 (rev 4), the following conversion operator must be defined for `multi_ptr`:

```cpp
  // Implicit conversion to multi_ptr<const value_type, Space>.
  template<access::decorated DecorateAddress2>
  operator multi_ptr<const value_type, Space, DecorateAddress2>() const;
```

Without this change, the following code fails to compile:

```cpp
#include <SYCL/sycl.hpp>

int main()
{
    sycl::multi_ptr<float, sycl::access::address_space::local_space> ptr;
    sycl::multi_ptr<const float, sycl::access::address_space::local_space> const_ptr = ptr;
}
```


illuhad (approved):
Thanks!

PR #666, opened 2021-11-18
Add support for the CUDA backend to operate as a library for nvc++

illuhad:
This PR adds support for hipSYCL's CUDA backend to operate as a library for nvc++. This allows targeting NVIDIA GPUs with a vendor-supported compiler. In this mode, hipSYCL simply invokes nvc++ without any SYCL-specific compilation magic. It is then nothing more than any other library for nvc++. LLVM as a dependency is not required in this mode.

*As outlined further down, don't expect everything to work already. hipSYCL can currently trigger nvc++ ICEs in specific cases, which is why a couple of unit tests are currently disabled, and overall this PR needs more testing. See below for more details.*

## Potential use cases are:
* Day 1 support for new hardware
* Much simpler deployment on NVIDIA GPU HPC machines, where nvc++ might already be preinstalled.

**Note that this will not replace the existing clang-based backend, it will just add another option to target NVIDIA hardware.**

## About

NVIDIA's nvc++ compiler has support for automatic kernel outlining when compiling CUDA code, i.e. can automatically infer `__host__` and `__device__` attributes. This is the main component that hipSYCL's clang module implements; hence support for it in nvc++ itself means that we can support most of the existing functionality without having to add SYCL-specific compiler magic.
Consequently, SYCL can be implemented as a library for nvc++.

Note that we do not rely on C++ standard algorithms or OpenACC, but CUDA C++, which nvc++ also supports, even though it is not a well documented fact. We use the same CUDA backend as with the existing clang-based driver, with minute difference between the code paths. 

The most significant difference between the clang and nvc++ code paths is that nvc++ to my knowledge does not provide any means to modify data layout of classes depending on the compilation target. This means that some minor storage optimizations cannot be done in nvc++ mode and are disabled. I do not expect this to have a significant performance impact.

This vast majority of this PR is about refactoring how we specialize code paths for compilation targets. Since it is not possible to do this with `nvc++` using macros as in clang, we have introduced the `__hipsycl_if_target_<backend>(code)` macros that generalize both to clang as well as nvc++.

Also note that in nvc++ mode, the `HIPSYCL_LIBKERNEL_IS_DEVICE_PASS_HOST` and `HIPSYCL_LIBKERNEL_IS_DEVICE_PASS_CUDA` macros are not mutually exclusive, but active at the same time. In concordance with the SYCL specification, `SYCL_DEVICE_ONLY` is never defined in nvc++ mode. Use the `__hipsycl_if_target` macros for a cross platform approach to specialize code paths.

## How to use

When building hipSYCL, use `cmake -DNVCXX_COMPILER=/path/to/nvc++` to point it to an nvc++ installation. If clang-based CUDA support is not desired, use `-DWITH_CUDA_NVCXX_ONLY=ON` to only enable nvc++ support and remove the dependency on LLVM.

This new mode of operation can then be activated by passing `cuda-nvcxx` to `--hipsycl-targets`, e.g. `syclcc --hipsycl-targets="omp;cuda-nvcxx`. 

Specific GPUs can be targeted using `ccXY`, with `XY` being the respective compute capability. For example, `syclcc --hipsycl-targets="omp;cuda-nvcxx:cc80"`.

As mentioned in the *caveats* section below, currently an additional `-isystem` argument might be needed to get things to work properly.

### nvc++ compiler bugs
*During development I have encountered numerous nvc++ compiler bugs. hipSYCL can currently trigger nvc++ ICEs in certain cases, which is why a couple of unit tests are currently disabled. However the majority of the functionality should already work.*

These are these known issues:

* Certain nested lambdas patterns can make nvc++ EDG frontend crash. I have so far only observed this with scoped parallelism test cases. The workaround is to use capture-by-value instead of capture-by-reference. I have reported those parts for which I could create a reproducer: https://forums.developer.nvidia.com/t/nvc-21-9-ice-when-using-nested-lambdas/194724
* There is an LLVM backend error when using vector swizzles
* The `fmod` builtin and `fdim` builtins don't work correctly, causing `ptxas`/linker errors.
* For some reason, nvc++ behavior is different when adding `-isystem $HIPSYCL_INSTALL_PREFIX/include` to the compilation line. Without this argument, `nvc++` might issue a large number of - as it seems - incorrect warnings and might crash.

### Unsupported features
* Hierarchical parallelism is unsupported in this mode, as it requires dedicated compiler support. This is not a significant issue, as hierarchical parallelism is discouraged in SYCL 2020 anyway.


illuhad:
@hipSYCL-ci-bot test

illuhad:
This still has numerous unit tests failing, likely due to nvc++ bugs, but I think we should merge soon since this reorganizes quite a lot of code and I'd like to avoid future merge conflicts.

illuhad:
@hipSYCL-ci-bot test

PR #667, opened 2021-11-24
Fix linking errors with libstdc++ < 9

al42and:
This patch adds support for detecting if a special linking flag is needed when compiling on a system with old-ish libstdc++ (for example, CentOS 8, where libstdc++ 8 is used even with clang-12).

Without adding flag, hipSYCL itself compiles fine, but `syclcc` fails to link even a simple program (that does not use `std::filesystem` directly) with errors like:

```
/usr/local/bin/../lib//libhipSYCL-rt.so: undefined reference to `std::filesystem::__cxx11::path::parent_path() const'
/usr/local/bin/../lib//libhipSYCL-rt.so: undefined reference to `std::filesystem::__cxx11::directory_iterator::operator*() const'
/usr/local/bin/../lib//libhipSYCL-rt.so: undefined reference to `std::filesystem::__cxx11::path::_M_find_extension() const'
/usr/local/bin/../lib//libhipSYCL-rt.so: undefined reference to `std::filesystem::__cxx11::path::has_root_directory() const'
/usr/local/bin/../lib//libhipSYCL-rt.so: undefined reference to `std::filesystem::status(std::filesystem::__cxx11::path const&)'
/usr/local/bin/../lib//libhipSYCL-rt.so: undefined reference to `std::filesystem::__cxx11::path::_M_split_cmpts()'
/usr/local/bin/../lib//libhipSYCL-rt.so: undefined reference to `std::filesystem::equivalent(std::filesystem::__cxx11::path const&, std::filesystem::__cxx11::path const&)'
/usr/local/bin/../lib//libhipSYCL-rt.so: undefined reference to `std::filesystem::__cxx11::directory_iterator::operator++()'
/usr/local/bin/../lib//libhipSYCL-rt.so: undefined reference to `std::filesystem::__cxx11::directory_iterator::directory_iterator(std::filesystem::__cxx11::path const&, std::filesystem::directory_options, std::error_code*)'
/usr/local/bin/../lib//libhipSYCL-rt.so: undefined reference to `std::filesystem::__cxx11::path::has_filename() const'
```

It can be solved by adding `-lstdc++fs` flag to `syclcc` invocation, but it seems more logical to me to link it statically to libhipSYCL-rt.so.

Used FindFilesystem module from [CMake Community Modules](https://github.com/vector-of-bool/CMakeCM).


illuhad:
hmm... so technically one could argue that SYCL 2020 requires C++17, and if `std::filesystem` does not come with the supplied standard library, it's an unsupported setup.
But I do see the use case given the prevalence of old toolchains on HPC machines, and can see this improving convenience greatly.

I do suspect however that this patch is quiet specific to a certain libstdc++ version, and I wonder if we can make it more generic. For example:
* Very new `libstdc++` work out of the box
* Slightly older ones need `-lstdc++fs`
* Even older ones need `-lstdc++fs` as well as minor changes in the source code since they only have `std::tr1::filesystem`
* Even older ones don't have it at all.

I wonder if it might not be more generic to just add support to also use `boost::filesystem` (boost being a dependency anyway already) via a cmake switch? Then we might be able to cover even more versions of `libstdc++`.

al42and:
> hmm... so technically one could argue that SYCL 2020 requires C++17, and if std::filesystem does not come with the supplied standard library, it's an unsupported setup.

I don't like this rationale very much myself, but technically, the C++ standard says nothing about compiler switches. Just like requiring the `-std=c++17` switch for C++17 support, it is legal to require using `-lstc++fs`. So, libstdc++ 8 can claim to support std::filesystem: https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html#status.iso.2017, despite requiring extra command-line flags.

> I wonder if it might not be more generic to just add support to also use boost::filesystem (boost being a dependency anyway already) via a cmake switch? Then we might be able to cover even more versions of libstdc++.

I agree that it sounds like a good solution, and it will involve less CMake, which is always good :)

@pszi1ard is the one who initially encountered that issue, so he can probably evaluate which solution is preferable better than I.

illuhad:
> I don't like this rationale very much myself, but technically, the C++ standard says nothing about compiler switches. Just like requiring the -std=c++17 switch for C++17 support, it is legal to require using -lstc++fs. So, libstdc++ 8 can claim to support std::filesystem: https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html#status.iso.2017, despite requiring extra command-line flags.

Arrgh, this is terrible - but okay, I see your point :-)

> I agree that it sounds like a good solution, and it will involve less CMake, which is always good :)
> @pszi1ard is the one who initially encountered that issue, so he can probably evaluate which solution is preferable better than I.

Alright, then let's wait for his feedback.

pszi1ard:
Sorry for the delay @illuhad, as there are already boost dependencies, adding another is probably not a concern, but I would still recommend to use `libstdc++` whenever this is known to work (and plan to remove again the `boost::filesystem` when the requirements ensure that `std::filesystem` always works).

al42and:
@illuhad, FYI, I won't have time to work on rewriting it to use `boost::filesystem` in the near future. But I think this code does more good than harm even as it is now, so if you also consider it useful, I think it would be easier to merge it now, and perhaps get back to making a better solution later.

al42and:
@illuhad, FYI, I won't have time to work on rewriting it to use `boost::filesystem` in the near future. But I think this code does more good than harm even as it is now, so if you also consider it useful, I think it would be easier to merge it now, and perhaps get back to making a better solution later.

pszi1ard:
Current HPE Cray installations (including LUMI-G) which ship with default gcc 7.5, hence only have `experimental/filesystem`, so unfortunately even this fix wont' work OOB there. A few lines of code can fix the issue to enabling detection to work for experimental as well and to use it: https://termbin.com/wb0g

@illuhad can you comment whether you are OK with this PR the above fix?

illuhad:
@pszi1ard Please check out the branch https://github.com/illuhad/hipSYCL/tree/feature/flexible-filesystem-fallback ;)
We just need to put in the cmake side of things and we are good to go ;)

illuhad:
... if you want to work on this, feel free to merge my branch into your PR, or create a PR against my branch :)

al42and:
@pszi1ard, your fix can be simplified a bit, since `FindFilesystem.cmake` provides `CXX_FILESYSTEM_HEADER` (`filesystem` or `experimental/filesystem`) and `CXX_FILESYSTEM_NAMESPACE` (`std::filesystem` or `std::experimental::filesystem`). No need for `ifdef`'s. If we also want to support `boost::filesystem`, it won't be that easy, though.

illuhad:
@al42and I think I'd like to keep the door open for boost support, so I'd prefer if the cmake side were to just decide which filesystem provider to use (e.g. by defining some macro) and then linking against the appropriate libraries. See https://github.com/illuhad/hipSYCL/blob/9bfbccc7e3aae0aa2723ed0f7f62d31594455a48/src/runtime/backend_loader.cpp#L39

al42and:
> @al42and I think I'd like to keep the door open for boost support, so I'd prefer if the cmake side were to just decide which filesystem provider to use (e.g. by defining some macro) and then linking against the appropriate libraries. See

OK. I personally don't see much use for supporting *both* `boost::filesystem` and `std(::experimental)::filesystem`. Boost has the benefit of not having to bother with finding the right flags and includes, and since we already depend on Boost, there's little harm in requiring one more component, in addition to fibers and context. On the other hand, one of `std::filesystem` or `std::experimental::filesystem` is always provided by any C++17 compiler, so if we support both of them, there's unlikely any need for using boost::filesystem.

But anyway, I expanded the CMake module to look for Boost::filesystem if neither std::filesystem nor std::experimental::filesystem can be found.

In a few configurations I tested, it worked okay, but we might want to check that it helps @pszi1ard and probably do some more testing on different machines.

pszi1ard:
> In a few configurations I tested, it worked okay, but we might want to check that it helps @pszi1ard and probably do some more testing on different machines.

Tested on a couple of HPE Cray machines, works with `std::experimental::filesystem` detection.

src/runtime/FindFilesystem.cmake
```diff
@@ -0,0 +1,287 @@
+# Distributed under the OSI-approved BSD 3-Clause License.
+# See https://cmake.org/licensing for details.
+# This file is part of CMake Community Modules:
+# https://github.com/vector-of-bool/CMakeCM/blob/master/modules/FindFilesystem.cmake
+#[=======================================================================[.rst:
+
+FindFilesystem
```
illuhad: I wonder if this file might better be placed in our `cmake` directory?

src/runtime/CMakeLists.txt
```diff
@@ -5,6 +5,12 @@ set(HIPSYCL_RT_EXTRA_LINKER_FLAGS "")
 set(HIPSYCL_RT_SANITIZE "" CACHE STRING
     "Enable building hipSYCL runtime with specified sanitizers")
 
+list(PREPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}")
```
illuhad: I'm trying this on the Intel devcloud, because it also has an older toolsystem as a test case. There, cmake is also slightly older and does not yet have `PREPEND`. I guess it's plausible that when users have older toolchains so that they require this PR, they might also have an older cmake. Any reason why we cannot use `APPEND`?

src/runtime/CMakeLists.txt
```diff
@@ -5,6 +5,12 @@ set(HIPSYCL_RT_EXTRA_LINKER_FLAGS "")
...
+find_package(Filesystem REQUIRED Final Experimental Boost)
+set(CXX_FILESYSTEM_HEADER "${CXX_FILESYSTEM_HEADER}" PARENT_SCOPE)
+set(CXX_FILESYSTEM_NAMESPACE "${CXX_FILESYSTEM_NAMESPACE}" PARENT_SCOPE)
+list(POP_FRONT CMAKE_MODULE_PATH)
```
illuhad: Similarly, the pop commands are not available. Can we circumvent by storing and reassigning the state of `CMAKE_MODULE_PATH` before we modify it?

illuhad (commented):
This is awesome!

src/runtime/CMakeLists.txt
al42and: We can use `APPEND`, but since we have modified the `FindFilesystem.cmake`, it would be safer to have the path to our version first in the list. `INSERT` can do what we want and works with CMake 3.10 (the one I tested).

src/runtime/CMakeLists.txt
al42and: I'd rather use `REMOVE_AT`: that's one line instead of two for save/restore :)

src/runtime/FindFilesystem.cmake
al42and: Done :ok_hand:

illuhad (approved):
Thank you! You have just made life for a lot of SYCL cluster people a lot easier :-)

PR #674, opened 2021-12-08
Use quotation marks for argument of cmake get_filename_component()

illuhad:
Use quotation marks for argument of cmake `get_filename_component()`.

This can fix "incorrect number of arguments" cmake errors if the supplied paths are empty.


sbalint98 (approved):
LGTM, I think we can close #643 if this is merged.

PR #682, opened 2022-01-18
Compiler support to accelerate nd_range parallel_for on CPU

fodinabor:
This adds explicit compiler support for the nd_range parallel_for paradigm to speed it up by up to multiple orders of magnitude compared to the current Boost.Fiber-based implementation.

Exemplary performance comparison of https://github.com/UoB-HPC/sycl_dgemm
Shows fibers, the compiler supported CPU version `cbs_man`, the hierarchical implementation as well as an equivalent OpenCL implementation using differing POCL modes on 4 different machines (all from Isambard, big thanks to @simonmcs for letting us test there!).

sycl_dgemm csv

Same game for SYCL-Bench's reduction benchmark:
sycl-bench-red csv

Both examples show that there's a real benefit to using this instead of the fiber nd_range variant! Though, there's still some room for improvement to fully match the hierarchical performance.

Copying the high-level documentation:  
>While hipSYCL usually strives to reuse as much compiler infrastructure as possible, the nd_range parallel_for is not efficiently implementable without explicit compiler support.
>
>Therefore, the hipSYCL Clang plugin implements a set of passes to perform deep loop fission on nd_range parallel_for kernels that contain barriers. The continuation-based synchronization approach is employed to achieve good performance and  functional correctness (_Karrenberg, Ralf, and Sebastian Hack. "Improving performance of OpenCL on CPUs." International Conference on Compiler Construction. Springer, Berlin, Heidelberg, 2012. [https://link.springer.com/content/pdf/10.1007/978-3-642-28652-0_1.pdf](https://link.springer.com/content/pdf/10.1007/978-3-642-28652-0_1.pdf)_). A deep dive into how the implementation works and why this approach was chosen can be found in my [master thesis](https://joameyer.de/hipsycl/Thesis_JoachimMeyer.pdf).
>
>As Clang plugins are only officially supported on Linux, this functionality is currently supported by hipSYCL on Linux only as well.
By default, this compiler support is used, as long as Clang is used as host compiler as well. It can be explicitly disabled, by setting `-DWITH_ACCELERATED_CPU=OFF` at hipSYCL build time. If hipSYCL was built with the compiler support enabled, it can be explicitly en-/disabled by adding `--hipsycl-use-accelerated-cpu`/`--hipsycl-use-accelerated-cpu=false` to the syclcc command line. When using CMake, set `-DHIPSYCL_USE_ACCELERATED_CPU=` to override the default behavior.

This currently is tested with LLVM 11-13 in the CI but is also compatible with LLVM 14 at the time of writing.
I would claim this feature still is experimental, but encourage everyone to test it out and would love any feedback!
If someone uses non-XCode Clang on Mac, it should be possible to enable the Clang plugin there as well, but I can't test it, so I didn't do it...


illuhad:
Can you have a glance at the `syclcc` changes in #683 and check if it might make sense to utilize some of the infrastructure from there here as well, e.g. backend flavors?

fodinabor:
I guess one could add an `omp.accelerated` flavor of the OpenMP target that mimics/replaces the `--hipsycl-use-accelerated-cpu` flag. The omp invocation could then return a caveat warning if the host compiler is not the known clang as well.

The latter is a slight inconvenience I hoped might be easily solvable by your changes, but from looking at them, doesn't seem that way. This is as even though the `config.pure_cpu_compiler` is checked in `omp_invocation` whether it is clang or not. If it's not clang, the `nd_range` acceleration is not enabled, even if `--hipsycl-use-accelerated-cpu` is set. If another backend (e.g. CUDA or HIP) is active though, the host compiler is overridden by clang, thus the acceleration would work. Now the question is: if we have an `omp.accelerated` flavor or so (and also `--hipsycl-use-accelerated-cpu` if we keep that flag), should that override the host compiler to clang or not? If not, can we _nicely_ add another stage, where depending of the finally chosen host compiler can add more flags? (or is that even already covered by something like `_configure_multipass_extended_host_pass`?)

illuhad:
> I guess one could add an omp.accelerated flavor of the OpenMP target that mimics/replaces the --hipsycl-use-accelerated-cpu flag. The omp invocation could then return a caveat warning if the host compiler is not the known clang as well.

I think this would nicely mimic the explicit multipass logic: The `--hipsycl-use-accelerated-cpu` flag would enable CPU acceleration for all (we currently only really have one) host backends, while the `.accelerated` would allow for a more selective enabling if multiple host backends are available.

> Now the question is: if we have an omp.accelerated flavor or so (and also --hipsycl-use-accelerated-cpu if we keep that flag), should that override the host compiler to clang or not

I think so - this would mimic the behavior of the device backends. The user has explicitly requested the flavor, so we should make sure that things happen as requested, no?

> If not, can we nicely add another stage, where depending of the finally chosen host compiler can add more flags? (or is that even already covered by something like _configure_multipass_extended_host_pass?)

The explicit multipass PR adds a `set_host_compiler()` member function for the multipass backends which is invoked by the main driver before compilation so that multipass backends can set caveats accordingly. We could extend this mechanism to all backends, not only multipass backends. Then they could take decisions based on the host compiler.

CMakeLists.txt
```diff
@@ -160,6 +172,9 @@ endif()
 #Use add_definitions for now for older cmake versions
 cmake_policy(SET CMP0005 NEW)
 add_definitions(-DHIPSYCL_DEBUG_LEVEL=${HIPSYCL_DEBUG_LEVEL})
+if(WITH_ACCELERATED_CPU)
+  add_definitions(-DHIPSYCL_USE_ACCELERATED_CPU)
```
illuhad: Does this need to be set globally or just for the clang plugin target?

CMakeLists.txt
```diff
@@ -284,11 +299,11 @@ if(NOT CUDA_CXX_FLAGS)
 endif()
 
 if(NOT OMP_CXX_FLAGS) 
-  set(OMP_CXX_FLAGS "-I${Boost_INCLUDE_DIR} ${DEFAULT_OMP_FLAG}" CACHE STRING "Arguments passed to compiler to compile SYCL applications with OpenMP")
+  set(OMP_CXX_FLAGS "-I${Boost_INCLUDE_DIR} ${DEFAULT_OMP_FLAG} -D_ENABLE_EXTENDED_ALIGNED_STORAGE" CACHE STRING "Arguments passed to compiler to compile SYCL applications with OpenMP")
```
illuhad, line 304: Why is `_ENABLE_EXTENDED_ALIGNED_STORAGE` required?

bin/syclcc-clang
```diff
@@ -913,12 +930,15 @@ class omp_invocation:
...
+    self._use_plugin = config.use_accelerated_cpu and self._cxx == config.clang_path
 
   def get_compiler_preference(self):
     return (self._cxx, 1)
 
   def get_cxx_flags(self):
-    flags =["-D__HIPSYCL_ENABLE_OMPHOST_TARGET__"]
+    flags = ["-D__HIPSYCL_ENABLE_OMPHOST_TARGET__"]
+    if self._use_plugin and not sys.platform.startswith("win32"):
+      flags += ["-fplugin=" + config.hipsycl_plugin_path, "-fpass-plugin=" + config.hipsycl_plugin_path, "-DHIPSYCL_USE_ACCELERATED_CPU"]
```
illuhad: Maybe multiple lines? We also have the vague convention that internal macros set by `syclcc` to control compilation follow the form `__HIPSYCL_*__`. Would it make sense to apply this to `HIPSYCL_USE_ACCELERATED_CPU`?

doc/compilation.md
```diff
@@ -13,13 +13,32 @@ hipSYCL distinguishes two modes of compilation:
...
+| CUDA (nvc++) | Yes                   | No                  | No support for CPU nd-range acceleration |
+| HIP          | Yes                   | No                  |                                          |
+| SPIR-V       | No                    | Yes                 |                                          |
 
 **Note:** Explicit multipass requires building hipSYCL against a clang that supports `__builtin_unique_stable_name()` as described in the [installation documentation](installing.md).
 
+
+## Compiler support to accelerate nd_range parallel_for on CPUs
+While hipSYCL usually strives to reuse as much compiler infrastructure as possible,
+the nd_range parallel_for is not efficiently implementable without explicit compiler support.
```
illuhad: Maybe also add a sentence about scoped parallelism. I think the message should be:
1.) Use a model that is designed to be performance portable such as scoped,
2.) If this is not an option and you need nd_range (e.g. because of existing code base) the clang plugin is a viable option
3.) nd_range without clang plugin will be slow *as soon as barriers are present* due to issues with the `nd_range` model

include/hipSYCL/compiler/.clang-format
```diff
@@ -0,0 +1,2 @@
+BasedOnStyle: LLVM
+ColumnLimit: 120
```
illuhad: Can we reduce this a bit? I think for the most part our lines are kept way shorter.

include/hipSYCL/common/debug.hpp
```diff
@@ -113,5 +119,14 @@ class output_stream {
   HIPSYCL_DEBUG_STREAM(HIPSYCL_DEBUG_LEVEL_INFO, \
                       HIPSYCL_DEBUG_PREFIX_INFO)
 
+#define HIPSYCL_DEBUG_EXECUTE(level, content) \
+  if(level <= ::hipsycl::common::output_stream::get().get_debug_level()) \
+  {\
+    content;\
+  }
 
+#define HIPSYCL_DEBUG_EXECUTE_VERBOSE(content) HIPSYCL_DEBUG_EXECUTE(HIPSYCL_DEBUG_LEVEL_VERBOSE, content)
```
illuhad, line 128: Why do we need the `HIPSYCL_DEBUG_EXECUTE`? Isn't the same check effectively already don in `HIPSYCL_DEBUG_STREAM`?

include/hipSYCL/compiler/Frontend.hpp
```diff
@@ -438,6 +447,7 @@ class FrontendASTVisitor : public clang::RecursiveASTVisitor<FrontendASTVisitor>
   std::unordered_set<clang::FunctionDecl*> MarkedKernels;
   std::unordered_set<clang::FunctionDecl*> HierarchicalKernels;
   std::unordered_set<clang::FunctionDecl*> UserKernels;
+  std::unordered_set<clang::FunctionDecl*> NDKernels;
```
illuhad: I wonder if there is a way to separate device backend compilation logic and CPU compilation logic somewhat.. Maybe prefix NDKernels with Host, or LoopFission?

include/hipSYCL/compiler/cbs/IRUtils.hpp
```diff
@@ -0,0 +1,158 @@
...
+struct MDKind {
+  static constexpr const char Arrayified[] = "hipSYCL.arrayified";
+  static constexpr const char InnerLoop[] = "hipSYCL.loop.inner";
+  static constexpr const char WorkItemLoop[] = "hipSYCL.loop.workitem";
+};
+
+static constexpr const char BarrierIntrinsicName[] = "__hipsycl_barrier";
+static constexpr const char LocalIdGlobalNameX[] = "_ZN7hipsycl4glue12omp_dispatch20__hipsycl_local_id_xE";
+static constexpr const char LocalIdGlobalNameY[] = "_ZN7hipsycl4glue12omp_dispatch20__hipsycl_local_id_yE";
+static constexpr const char LocalIdGlobalNameZ[] = "_ZN7hipsycl4glue12omp_dispatch20__hipsycl_local_id_zE";
```
illuhad: Can we find a way to not hardcode things to OpenMP? It would be great if we could also just use this in say, a TBB backend

include/hipSYCL/compiler/cbs/SubCfgFormation.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+#ifndef HIPSYCL_SUBCFGFORMATION_HPP
+#define HIPSYCL_SUBCFGFORMATION_HPP
+
+#include "llvm/Analysis/LoopInfo.h"
+#include "llvm/IR/Function.h"
+#include "llvm/Pass.h"
+
+#if LLVM_VERSION_MAJOR <= 12
+#define HIPSYCL_NO_PHIS_IN_SPLIT
```
illuhad: Can you add a comment about the meaning of this? Is there something unsupported in LLVM <= 12?

include/hipSYCL/compiler/cbs/SyncDependenceAnalysis.hpp
```diff
@@ -0,0 +1,97 @@
...
+  unsigned size() const { return LoopPO.size(); }
+  const llvm::BasicBlock *getBlockAt(unsigned Idx) const { return LoopPO[Idx]; }
+};
+
+/// \brief Relates points of divergent control to join points in
+/// reducible CFGs.
+///
+/// This analysis relates points of divergent control to points of converging
+/// divergent control. The analysis requires all loops to be reducible.
+class SyncDependenceAnalysis {
```
illuhad, line 66: I understand that this is from LLVM >= 13 that we backport for compatibility reasons, right? Would it make sense to put such classes in a `compatibility` or `llvm_pre13_compatibility` namespace to not clutter the main `hipsycl::compiler` namespace with content that is only needed for old LLVM versions?

include/hipSYCL/compiler/cbs/UniformityAnalysis.hpp
```diff
@@ -0,0 +1,162 @@
+//===- rv/analysis/VectorizationAnalysis.h - divergence analysis --*- C++ -*-===//
+//
+// Adapted from the RV Project, under the Apache License v2.0 with LLVM Exceptions.
```
illuhad, line 3: This is from region vectorizer? Didn't you find that RV was still too unstable to rely on it?

include/hipSYCL/glue/kernel_names.hpp
```diff
@@ -92,17 +92,16 @@ struct kernel_name_traits<__hipsycl_unnamed_kernel, KernelBodyT> {
   using multiversioned_suggested_mangling_name =
       multiversioned_kernel_wrapper<KernelBodyT, MultiversionParameters...>;
 
-  template <typename... MultiversionParameters>
+template <typename... MultiversionParameters>
```
illuhad: This indentation looks a bit weird to me?

include/hipSYCL/glue/omp/omp_kernel_launcher.hpp
```diff
@@ -227,30 +259,58 @@ inline void parallel_for_kernel_offset(Function f,
...
+      for (size_t g_x = 0; g_x < num_groups[0]; ++g_x) {
+        for (size_t g_y = 0; g_y < num_groups[1]; ++g_y) {
+          for (size_t g_z = 0; g_z < num_groups[2]; ++g_z) {
+            const sycl::id<Dim> group_id{g_x, g_y, g_z};
+            iterate_nd_range_omp(f, std::move(group_id), num_groups, local_size, offset,
+              num_local_mem_bytes, &group_shared_memory_ptr, barrier_impl, reducers...);
+          }
+        }
+      }
+    }
```
illuhad: This whole pattern is effectively what is already implemented in `iterate_range_omp`, no? Can we reuse it here?

include/hipSYCL/runtime/omp/omp_hardware_manager.hpp
```diff
@@ -30,8 +30,10 @@
 
 #include "../hardware.hpp"
 
-namespace hipsycl {
-namespace rt {
+namespace hipsycl
+{
+namespace rt
+{
```
illuhad: We usually don't do line breaks before namespace opening brace

include/hipSYCL/sycl/libkernel/detail/local_memory_allocator.hpp
```diff
@@ -159,7 +159,7 @@ class host_local_memory
   static constexpr size_t _max_static_local_mem_size = 1024*32;
   inline static char* _local_mem;
   
-  alignas(alignof(std::max_align_t)) inline static char _static_local_mem
+  alignas(sizeof(double) * 16) inline static char _static_local_mem
```
illuhad, line 162: Why can we not have maximum alignment?

include/hipSYCL/sycl/libkernel/backend.hpp
```diff
@@ -89,8 +89,14 @@
 #ifdef __clang__
  #define HIPSYCL_FORCE_INLINE \
  __attribute__((always_inline)) __attribute__((flatten)) inline
+ #define HIPSYCL_ND_KERNEL [[clang::annotate("hipsycl_nd_kernel")]]
+ #define HIPSYCL_ND_KERNEL_LOCAL_SIZE_ARG [[clang::annotate("hipsycl_nd_kernel_local_size_arg")]]
+ #define HIPSYCL_BARRIER [[clang::annotate("hipsycl_barrier")]]
 #else
  #define HIPSYCL_FORCE_INLINE inline
+ #define HIPSYCL_ND_KERNEL
+ #define HIPSYCL_BARRIER
```
illuhad: Should we maybe clarify those macros to refer to loop splitting? Like `HIPSYCL_LOOP_SPLIT_TRIGGER` or something similar?  After all, these attributes trigger something that is fairly specific to CPU backends and don't affect barriers or nd kernels *in general*.

include/hipSYCL/sycl/libkernel/id.hpp
```diff
@@ -110,14 +110,16 @@ struct id {
 
   HIPSYCL_UNIVERSAL_TARGET
   id(const range<dimensions> &range) {
-    for(std::size_t i = 0; i < dimensions; ++i)
+    this->_data[0] = range[0];
+    for(std::size_t i = 1; i < dimensions; ++i)
```
illuhad, line 115: This is curious, can you add a comment why this is beneficial?

src/compiler/GlobalsPruningPass.cpp
```diff
@@ -29,7 +28,7 @@ void pruneUnusedGlobals(llvm::Module &M) {
     G->eraseFromParent();
   }
   HIPSYCL_DEBUG_INFO << "===> IR Processing: Pruning of globals complete, removed "
-                     << VariablesForPruning.size() << " global variable(s)." << std::endl;
+                     << VariablesForPruning.size() << " global variable(s).\n";
```
illuhad, line 31: Why the changes in this file?

src/compiler/HipsyclClangPlugin.cpp
```diff
@@ -40,38 +45,77 @@ namespace hipsycl {
...
 
 static llvm::RegisterStandardPasses
     RegisterGlobalsPruningPassOptimizerLast(llvm::PassManagerBuilder::EP_OptimizerLast,
-                                            registerGlobalsPruningPass);
+                                               registerGlobalsPruningPass);
+
+#if defined(HIPSYCL_USE_ACCELERATED_CPU) && !defined(_WIN32)
+static llvm::RegisterPass<SplitterAnnotationAnalysisLegacy>
+    splitterAnnotationReg("splitter-annot-ana", "hipSYCL splitter annotation analysis pass",
+                          true /* Only looks at CFG */, true /* Analysis Pass */);
```
illuhad, line 67: In PR #683, we relay information to the clang plugin using macros, i.e. `syclcc` will define a macro, and the clang plugin checks the existence of the macro as compiler option to adjust its behavior. Would it make sense to have a similar mechanism with the cbs plugin, so that we can disable support even though the clang plugin has been compiled with support for it?

src/runtime/omp/omp_hardware_manager.cpp
```diff
@@ -142,7 +142,11 @@ omp_hardware_context::get_property(device_uint_property prop) const {
     return std::numeric_limits<std::size_t>::max();
     break;
   case device_uint_property::max_group_size:
+#ifndef HIPSYCL_USE_ACCELERATED_CPU
```
illuhad: I think this is confusing. There are two different ways in which this macro is used: At compile time of hipSYCL, affecting the hipSYCL runtime and clang plugin, and secondly, set by syclcc if accelerated cpu is enabled. What happens if `syclcc` compiles without this macro because the user has disabled accelerated cpu, but the runtime and clang plugin have it baked in? Would it make sense to distinguish one macro the runtime and clang plugin whether there is *support*, and a second one to be used by `syclcc` whether it is *enabled*?

src/runtime/omp/omp_hardware_manager.cpp
```diff
@@ -266,22 +270,21 @@ omp_hardware_context::get_property(device_uint_property prop) const {
...
+std::vector<std::size_t> omp_hardware_context::get_property(device_uint_list_property prop) const
+{
+  switch(prop) {
+    case device_uint_list_property::sub_group_sizes:
+      return std::vector<std::size_t>{1};
+      break;
+  }
+  assert(false && "Invalid device property");
+  std::terminate();
+}
```
illuhad: This has just been moved down a couple of lines? :D

tests/sycl/group_functions/group_functions.hpp
```diff
@@ -209,7 +209,7 @@ elementType<T> local_value(size_t id, size_t offsetBase) {
 inline void create_bool_test_data(std::vector<char> &buffer, size_t local_size,
                                   size_t global_size) {
   BOOST_REQUIRE(global_size == 4 * local_size);
-  BOOST_REQUIRE(local_size + 10 < 2 * local_size);
+  BOOST_REQUIRE(local_size + 2 < 2 * local_size);
```
illuhad: Reason for change?

.github/workflows/linux-cbs.yml
```diff
@@ -0,0 +1,70 @@
...
+          echo ${CXXFLAGS}
+          
+          cmake -DCMAKE_CXX_COMPILER=/usr/bin/clang++-${{matrix.clang_version}} -DOMP_CXX_FLAGS="$OMP_CXX_FLAGS -fopenmp" -DCLANG_EXECUTABLE_PATH=/usr/bin/clang++-${{matrix.clang_version}} -DLLVM_DIR=/usr/lib/llvm-${{matrix.clang_version}}/cmake .. -DCMAKE_INSTALL_PREFIX=`pwd`/install
+          make -j2 install
+      - name: setup CPU tests with loop splitting
+        run: |
+          mkdir /home/runner/work/hipSYCL/hipSYCL/build/tests-cpu
+          cd /home/runner/work/hipSYCL/hipSYCL/build/tests-cpu
+          
+          cmake -DHIPSYCL_PLATFORM=cpu -DHIPSYCL_USE_ACCELERATED_CPU=true -DhipSYCL_DIR=/home/runner/work/hipSYCL/hipSYCL/build/install/lib/cmake/hipSYCL -DCMAKE_CXX_COMPILER=/usr/bin/clang++-${{matrix.clang_version}} -DCMAKE_CXX_FLAGS="$CXXFLAGS" /home/runner/work/hipSYCL/hipSYCL/tests
```
illuhad: Plz use `HIPSYCL_TARGETS` for new code ;)

.github/workflows/linux-cbs.yml
```diff
@@ -0,0 +1,70 @@
...
+          cd /home/runner/work/hipSYCL/hipSYCL/build/tests-cpu
+          make -j2
+      - name: run LIT tests on CPU
+        run: |
+          cd /home/runner/work/hipSYCL/hipSYCL/build/tests-cpu
+          LD_LIBRARY_PATH=/home/runner/work/hipSYCL/hipSYCL/build/install/lib make check
+      - name: run CPU tests
+        run: |
+          cd /home/runner/work/hipSYCL/hipSYCL/build/tests-cpu
+          LD_LIBRARY_PATH=/home/runner/work/hipSYCL/hipSYCL/build/install/lib ./sycl_tests
```
illuhad, line 70: This tests CPU exclusively, would there be value in also testing the use case when accelerated CPU and device backends are used simultanteously? (e.g. in hipSYCL-ci-bot?)

illuhad (commented):
I'm so excited! :D

CMakeLists.txt
fodinabor: It's actually Windows specific, but I figured it wouldn't hurt to keep it set generally while not cluttering the CMake file anymore..
It's a macro changing `std::aligned_storage`'s ABI to allow the necessary alignment (for `sycl::vec<double, 16>`)

include/hipSYCL/common/debug.hpp
fodinabor: yeah, it is.. the difference here is that we can guard the execution of anything.. which sometimes came in handy, e.g. `HIPSYCL_DEBUG_EXECUTE_VERBOSE(F.viewCFG());` is used several times, which writes the functions CFG to a file and opens a text editor to view it.

include/hipSYCL/compiler/cbs/IRUtils.hpp
fodinabor: Sure.. this is just a declaration-only global variable's mangled name... moving the variables to a different namespace (e.g. just `hipsycl::glue`) or marking them as `extern "C"` also removes the omp reference... alternatively we could again use `clang::annotate("...")` but I don't see a benefit there...

include/hipSYCL/compiler/.clang-format
fodinabor: Lowered to 100. Only touched new files (for fewer conflicts with #683 )

CMakeLists.txt
fodinabor: In this version globally, as the hw manager also needed to know this, but as I changed that, I also moved this.

include/hipSYCL/compiler/cbs/UniformityAnalysis.hpp
fodinabor: I pulled the uniformity analysis from RV into hipSYCL, yes.
The actual vectorization + RV's intrinsic were too unstable indeed, but with small adaptions from the original, the uniformity analysis has been pretty stable.

src/runtime/omp/omp_hardware_manager.cpp
fodinabor: yeah.. had it in the header for some period of time, which didn't change what I wanted to change though :)

include/hipSYCL/sycl/libkernel/detail/local_memory_allocator.hpp
fodinabor: `std::max_align_t` represents the alignment of `malloc` or similar, and is usually defined to `alignof(long double) == {8, 16}`. We need an alignment of `alignof(sycl::vec<double, 16>) == sizof(double) * 16`

include/hipSYCL/sycl/libkernel/id.hpp
fodinabor: hmm.. I loop rotated all id and range functions as it completely eliminates the loop already at the point in time, the CBS happens, and also to fix some values being misclassified as varying by the uniformity analysis while they could be marked uniform. I don't think I verified that this is necessary for all these functions and possibly got a bit over-ambitious, but for most it was actually beneficial.

src/compiler/GlobalsPruningPass.cpp
fodinabor: I changed the plugin's `HIPSYCL_DEBUG_INFO` and friends to use `llvm::outs()` instead of `std::cout` to get more deterministically interleaved output from LLVM and the hipSYCL plugin. `llvm::outs()` does not support `<< std::endl` thus I had to replace it with `\n`.

src/compiler/HipsyclClangPlugin.cpp
fodinabor: Sounds useful. It won't lead to functional changes, though, the plugin doesn't do anything if it doesn't find any nd_range annotated kernels, which only exist if `HIPSYCL_USE_ACCELERATED_CPU` (or its successor) is defined in the client code. Thus we could only reduce compiletime overhead slightly by checking the preprocessor definitions.
I'm not sure, if the right invocation context would be accessible from here either, so delaying this as possible optimization..

bin/syclcc-clang
fodinabor: Will do. It might make sense to just relay the `-DWITH_ACCELERATED_CPU` CMake Option to a Macro for the hipSYCL runtime and compiler, while `-D__HIPSYCL_USE_ACCELERATED_CPU__` might be a better fit in syclcc and thus the client code.

src/runtime/omp/omp_hardware_manager.cpp
fodinabor: Split those macros to `HIPSYCL_WITH_ACCELERATED_CPU` and `__HIPSYCL_USE_ACCELERATED_CPU__`
Additionally, as discussed offline, changed this instance to default to `1024` as it also is more factually correct for the group algos.

.github/workflows/linux-cbs.yml
fodinabor: I guess especially with the explicit multipass changes to syclcc it definitely would make sense to test at least the build...
But I'd delay that until explicit multipass and this are merged, so we can use the mix immediately.

fodinabor (commented):
I changed syclcc to override the default host compiler, fi the accelerated CPU flag is set.
I will look into merging the multipass and this PR at some other point in time (possibly after one is merged ;-))

CMakeLists.txt
illuhad: Okay, that's fine. Maybe add a comment though?

include/hipSYCL/compiler/cbs/UniformityAnalysis.hpp
illuhad: Okay. Are these small adaptions pointed out somewhere? It might be useful to know for maintenance in the future :)

include/hipSYCL/sycl/libkernel/detail/local_memory_allocator.hpp
illuhad: Ugh. That's good to know. @DieGoldeneEnte I believe this might have an impact on the local memory used for the group algorithms on CPU.

include/hipSYCL/sycl/libkernel/id.hpp
illuhad: Ok, please add a comment (at least where it is really important) so that we don't accidentally "simplify" this in the future.

src/compiler/GlobalsPruningPass.cpp
illuhad: Okay.

src/compiler/HipsyclClangPlugin.cpp
illuhad: Alright - I think it's not super urgent anyway because, as you point out, it shouldn't make a functional difference. It might be useful though if CBS triggers issues with future clang versions etc.

include/hipSYCL/compiler/cbs/SubCfgFormation.hpp
fodinabor: I kind of believe that was just for making my CI test both cases... removed it..

fodinabor:
@hipsycl-ci-bot test hopefully one last time....

bin/syclcc-clang
```diff
@@ -1193,7 +1236,16 @@ class compiler:
 
     for backend in self._targets:
       if backend == "omp":
-        self._backends.append(omp_invocation(config))
+        if config.use_accelerated_cpu or \
+                (not "cuda-nvcxx" in config.targets
+                 and not sys.platform.startswith("win32") and not sys.platform.startswith("darwin")):
```
illuhad: Should we not also check whether hipsycl-cpu-cxx is clang here?

fodinabor: fixed.

bin/syclcc-clang
```diff
@@ -1125,6 +1174,59 @@ class omp_invocation:
...
+    self._clang_path = config.clang_path
+
+  def get_compiler_preference(self):
+    return (self._clang_path, 10)
+
+  def get_host_pass_requirements(self):
+    return {
+      'requires-extended-host-pass' : False,
+      'extended-host-pass-providers' : [],
+      'conflicts' : ["cuda-nvcxx"],
```
illuhad:
```suggestion
      'conflicts' : ["cuda-nvcxx", "omp.library-only"],
```
Just noticed you can do `--hipsycl-targets="omp;omp.accelerated"` - this should probably be prevented, because we cannot actually have both simultaneously, right?

fodinabor: good catch :)

illuhad:
When compiling the unit tests (clang 13, `--hipsycl-targets="omp.accelerated;cuda:sm_61;hip:gfx906"`) I see warnings such as:
```
[hipSYCL Warning] Instruction not dominated   %61 = lshr i64 %63, 1 operand:   %63 = phi i64 [ %61, %60 ]
```

This seems to happen for multiple source files. Is this expected?

Unit tests seem to run correctly however.

fodinabor:
Yeah, that's some relict I got too used to seeing :)
That's handled, so I just made it emit that only with debug level info, as it still might be relevant for actual debugging purposes..

illuhad:
Alright, in it goes <3

illuhad (approved):

PR #683, opened 2022-01-21
Reimplement explicit multipass for clang 13+

### LINKED ISSUE #685: undeclared identifier '__hipsycl_warp_size' with --hipsycl-explicit-multipass
**Bug summary**
When a program is compiled using "--hipsycl-explicit-multipass" this error occurs. Probably introduced in 542a602fd93eac924168ea523641da3099feaec7.

```bash
HIPSYCL_TARGETS=cuda:sm_60 syclcc --hipsycl-explicit-multipass test.cpp
```
The output:
```bash
use of undeclared identifier '__hipsycl_warp_size'
      if constexpr(DivisorX % __hipsycl_warp_size == 0) {
                              ^
/root/build_hipsycl/bin/../include/hipSYCL/glue/cuda/../generic/hiplike/hiplike_kernel_launcher.hpp:295:52: error: use of undeclared identifier '__hipsycl_warp_size'
                         nested_range<static_range<__hipsycl_warp_size>>>;
                                                   ^
/root/build_hipsycl/bin/../include/hipSYCL/glue/cuda/../generic/hiplike/hiplike_kernel_launcher.hpp:297:16: error: use of undeclared identifier 'decomposition'
        return decomposition{};
               ^
/root/build_hipsycl/bin/../include/hipSYCL/glue/cuda/../generic/hiplike/hiplike_kernel_launcher.hpp:517:14: error: use of undeclared identifier 'hiplike'
  __device__ hiplike::local_reducer<ReductionDescriptor>
```

**To Reproduce**
Compile any program with "--hipsycl-explicit-multipass".  I tried on this one: https://gist.github.com/Luigi-Crisci/aa62113dc868f129cd2c1f09e1b72930

*My setup*
* HipSYCL at ee8cd86f95fe5604b6cf130d724828f7f5f0e121, compiled with Cuda backend
* Clang 11
* CXX compiler: G++ 9
* CUDA 10.1

illuhad:
This PR
* adds new infrastructure to `syclcc` to handle conflicting backends, and adds backend flavors that can be pass to `--hipsycl-targets` such as `cuda.explicit-multipass`
* adds infrastructure to support split compilation scenarios where the host compiler is not clang (in this case unnamed kernel lambdas are unsupported).
   - Clang plugin detects if `-D__HIPSYCL_SPLIT_COMPILER__` is defined and if so, mangles in the traditional way using kernel name tag/kernel body, and emits an error if unnamed kernel lambdas are used. This definition is automatically attached by `syclcc` if the host compiler does not match clang.
* Changes name mangling on clang 13+ (unless `__HIPSYCL_SPLIT_COMPILER__` is used) to rely on `__builtin_get_device_side_mangled_name()`. This builtin can only operate on CUDA/HIP `__global__` functions, so we use a dummy kernel that we use as template to generate kernel names to base our actual names on. In this mode, all kernel names are completely demangleable as `__hipsycl_kernel<KernelNameTag or KernelBodyT>`. Additionally, this builtin requires that we are in `-x cuda` or `-x hip` in the host pass. `syclcc` enforces that this is at least the case using `--cuda-host-only`.
* This builtin is used to support explicit multipass on clang 13+
* refactors the different mangling modes in the clang plugin

TODO: More testing required for different clang versions and modes of operations, since the implementation special cases different clang versions.


illuhad:
@hipSYCL-ci-bot test....

illuhad:
Seems GitHub has an internal issue with Windows CI.

illuhad:
PR now also adds support to compile the unit tests for all backends to one binary, and then select at runtime which backend to test using `HIPSYCL_VISIBILITY_MASK`. This in particular also allows to run the unit tests in an explicit multipass scenario, where the host pass might be mainly powered by a different backend compared to the device pass.

illuhad:
Ok, I believe this can go in now.

PR #687, opened 2022-01-27
Add buffer::size() and buffer::byte_size() functions

al42and:
They replace `buffer::get_count()` and `buffer::get_size()`, respectively, in SYCL2020: https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#_buffer_interface


illuhad (approved):
Thanks! :)

PR #691, opened 2022-01-29
Fix sycl::memory_environment() invalid memory access for private memory requests

illuhad:
Previously the `private_memory_access` class would always attempt to index its pointer based on the logical local id. While this is the correct approach on CPU where private memory allocations are widened to an array due to the CPU execution model in scoped parallelism, this causes an invalid memory access on device where private memory allocations are simple scalar variable declarations.
This PR fixes this.


illuhad:
@hipSYCL-ci-bot test.

PR #693, opened 2022-02-01
Apply small array optimizations to device_array

illuhad:
For small array instantiations of `device_array`, don't use a stack array member but instead regular member variables that are accessed using if-else if statements. This can help registerization on device if the indexed id cannot be inferred by the compiler at compile time.
`device_array` is used by core classes such as `id` and `range`, so this can have a massive impact.

A similar issue likely exists with our `sycl::vec` implementation, but that needs more investigation.


illuhad:
@hipSYCL-ci-bot test

PR #696, opened 2022-02-04
Don't submit work to async worker if DAG is empty

illuhad:
Previously, when flushing the DAG, it was only checked inside the async worker whether the DAG size is actually zero. This can lead to a substantial amount of no-op tasks being queued in the worker, which can slow things down.

This PR changes this so that work is only enqueued if we are sure that there is work.


illuhad:
@hipSYCL-ci-bot test

PR #697, opened 2022-02-05
[SYCL 2020][host] Missing negation

Michoumichmich:
Fixes a bug in sycl::none_of_group that happens only on host/omp and sub_groups.
Reproducer: 

```C++
    sycl::queue q{sycl::cpu_selector{}};
    q.submit([](sycl::handler& cgh) {
        sycl::stream os(256, 1024, cgh);
        cgh.parallel_for(sycl::nd_range<1>(32, 32), [=](sycl::nd_item<1> item) {
            if (sycl::none_of_group(item.get_sub_group(), true)) {
                os << "None of group" << sycl::endl;
            } else {
                os << "At least one of group" << sycl::endl; // Expected result.
            }
        });
    }).wait();
```

### DIFF
--- a/include/hipSYCL/sycl/libkernel/host/group_functions.hpp
+++ b/include/hipSYCL/sycl/libkernel/host/group_functions.hpp
@@ -238,7 +238,7 @@ inline bool __hipsycl_all_of_group(group<Dim> g, bool pred) {
 
 HIPSYCL_KERNEL_TARGET
 inline bool __hipsycl_all_of_group(sub_group g, bool pred) {
-  return pred;
+  return !pred;
 }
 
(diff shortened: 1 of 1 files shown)


DieGoldeneEnte (changes requested):
You are right none_of_group on cpu-subgroups should be negated. Your change adds the negation to all_of_group instead.
Could you please correct this?

Michoumichmich:
> You are right none_of_group on cpu-subgroups should be negated. Your change adds the negation to all_of_group instead. Could you please correct this?

Sorry about that, it's fixed!

illuhad (approved):

PR #701, opened 2022-02-09
Add isnan support for scalar float and double

sbalint98:
This PR adds support for sycl::isnan for scalar float and double

### DIFF
--- a/include/hipSYCL/sycl/libkernel/builtin_interface.hpp
+++ b/include/hipSYCL/sycl/libkernel/builtin_interface.hpp
@@ -54,6 +54,11 @@ HIPSYCL_BUILTIN T __hipsycl_acos(T x) noexcept {
 }
 
 template<class T>
+HIPSYCL_BUILTIN int __hipsycl_isnan(T x) noexcept {
+  HIPSYCL_RETURN_DISPATCH_BUILTIN(__hipsycl_isnan, x);
+}
+
+template<class T>
 HIPSYCL_BUILTIN T __hipsycl_acosh(T x) noexcept {
   HIPSYCL_RETURN_DISPATCH_BUILTIN(__hipsycl_acosh, x);
 }
--- a/include/hipSYCL/sycl/libkernel/builtins.hpp
+++ b/include/hipSYCL/sycl/libkernel/builtins.hpp
@@ -502,6 +502,24 @@ using ulonglong16 = vec<unsigned long long, 16>;
     }                                                                          \
   }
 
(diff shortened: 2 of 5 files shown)


include/hipSYCL/sycl/libkernel/builtin_interface.hpp
```diff
@@ -53,6 +53,11 @@ HIPSYCL_BUILTIN T __hipsycl_acos(T x) noexcept {
   HIPSYCL_RETURN_DISPATCH_BUILTIN(__hipsycl_acos, x);
 }
 
+template<class T>
+HIPSYCL_BUILTIN int __hipsycl_isnan(T x) noexcept {
```
illuhad, line 57: Per the spec, the return type would vary: 32-bit int for float and 64-bit int for double. The easiest way to accomplish this is probably to have auto return type here, or, we could also introduce the convention that our builtins are always 32 bit, and then handle the correct type in the main builtins.hpp header.

include/hipSYCL/sycl/libkernel/builtins.hpp
```diff
@@ -943,7 +943,8 @@ HIPSYCL_DEFINE_BUILTIN(fast_normalize, HIPSYCL_BUILTIN_OVERLOAD_SET_GENGEODOUBLE
 
 
 // ********************** relational functions *******************
-// TODO
+HIPSYCL_DEFINE_BUILTIN(isnan, HIPSYCL_BUILTIN_OVERLOAD_SET_SGENFLOAT,
+                       HIPSYCL_BUILTIN_GENERATOR_UNARY_T)
```
illuhad: Not sure this handles the return type correctly?

illuhad (commented):
Please also provide a SPIR-V implementation, as unmapped builtins *will* very likely break compilation otherwise (you have a call to a function that doesn't exist then in builtin_interface.hpp)

include/hipSYCL/sycl/libkernel/spirv/builtins.hpp
```diff
@@ -469,6 +469,9 @@ HIPSYCL_BUILTIN T __hipsycl_fast_normalize(T a) noexcept {
   return a / spirv_builtins::__hipsycl_fast_length(a);
 }
 
+// ****************** relational functions ******************
+HIPSYCL_DEFINE_SPIRV_BUILTIN(IsNan)
```
illuhad: Can you double check this? I think the builtin is called `__spirv_IsNan()` (at least that is used a couple of lines above), but `HIPSYCL_DEFINE_SPIRV_BUILTIN(name)` expands to `__spirv_ocl_name`.

illuhad (approved):
Thanks :)

PR #736, opened 2022-05-16
Add global kernel cache and HCF infrastructure

illuhad:
* Add common global kernel cache infrastructure
* Add unified HCF file format for kernels, tools, and documentation
* Reimplement explicit multipass for CUDA and SPIR-V to use new infrastructure
* Reimplement `syclcc` explicit multipass drivers to utilize unified HCF generation and integration header generation

TODO: Programs don't yet terminate cleanly on Level Zero due to a lifetime issue of the Level Zero backend shutting down before the cached SPIR-V kernels are unloaded.

SYCL-Bench task throughput benchmarks indicate that CUDA explicit multipass with the new kernel cache infrastructure outperforms integrated multipass by 30-40%.


include/hipSYCL/common/small_map.hpp
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  auto find(const Key& k) noexcept {
+    return std::find_if(
+        _v.begin(), _v.end(),
+        [&](const value_type &v) { return v.first == k; });
+  }
+
+  bool contains(const Key& k) const noexcept {
+    return find(k) == _v.cend();
```
al42and: Shouldn't it be

```suggestion
    return find(k) != _v.cend();
```

illuhad: Thanks!

illuhad:
It seems there is no easy fix for the Level Zero issue. This relates to the fundamental issue that we need to transition to a runtime that has its lifetime dependent on the existence of SYCL objects. This is a larger change, and should not block this PR.

illuhad:
In it goes.

PR #737, opened 2022-05-16
Update Windows version for CI

illuhad:
Let's see if this works...


illuhad:
It seems based on #742 that Windows CI just magically started to work again?

PR #739, opened 2022-05-20
Fix multi-threaded task processing

### LINKED ISSUE #728: dag_direct_scheduler: Direct scheduler does not support processing multiple unsubmitted nodes
**Bug summary**

When using hipSYCL runtime from multiple threads in the same process on a single CUDA GPU, occasionally the following error is thrown:

```
[hipSYCL Error] from /home/aland/hipSYCL/src/runtime/dag_direct_scheduler.cpp:313 @ submit(): dag_direct_scheduler: Direct scheduler does not support processing multiple unsubmitted nodes
============== hipSYCL error report ============== 
hipSYCL has caught the following undhandled asynchronous errors: 

   0. from /home/aland/hipSYCL/src/runtime/dag_direct_scheduler.cpp:313 @ submit(): dag_direct_scheduler: Direct scheduler does not support processing multiple unsubmitted nodes
The application will now be terminated.
```

The error is progressively more likely to occur with increasing number of threads (always happens with 9 threads, occasionally happens with four). Setting `HIPSYCL_RT_SCHEDULER=unbound` does **not** change the behavior. Setting `HIPSYCL_DEBUG_LEVEL=3` or 4 makes the error go away consistently. No warnings reported at level 2 besides the inability to load HIP plugin.

**To Reproduce**

I could not easily assemble a small reproducer, since it seems to depend on the multithreaded nature of the code and perhaps also on repeated re-initialization of the device backend with creation and destruction of SYCL queues for each test case.

Ubuntu 20.04, CUDA 11.6, hipSYCL develop (1046a78777a23ca75c6ea4e92291f1dbe36169ca):

```bash
wget https://ftp.gromacs.org/gromacs/gromacs-2022.tar.gz && tar zxf gromacs-2022.tar.gz && cd gromacs-2022/
mkdir build && cd build
cmake .. -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_C_COMPILER=clang-12 -DCMAKE_CXX_COMPILER=clang++-12 -DGMX_GPU=SYCL -DGMX_SYCL_HIPSYCL=ON -DHIPSYCL_TARGETS='cuda:sm_75'
nice cmake --build . --target tests -- -j$(nproc)
# A lot of "[hipSYCL Warning] CBS pipeline might not result in peak performance with old PM" warnings
./bin/mdrun-mpi-test -ntmpi 9
# Observe the failure. Happens in a random test case each time.
# Setting HIPSYCL_DEBUG_LEVEL=3 makes the problem go away
```

**Expected behavior**

hipSYCL runtime does not throw any errors, test passes.

**Describe your setup**
* hipSYCL develop (1046a78777a23ca75c6ea4e92291f1dbe36169ca)
   * Clang 12 from Ubuntu repos
   * Same Clang 12
   * CUDA 11.6
* How you have compiled your application and which arguments you have passed to `syclcc`. In particular, which backends and hardware you have compiled for.

**Additional context**

Unfortunately, I could not come up with a short reproducer, but perhaps the description would be enough to find the source of the bug? The issue is not critical for us, but random failures from the runtime are not nice.

illuhad:
Hopefully fixes #728 

The main change is that this introduces a mutex to lock `dag_manager::flush_async()`, thereby giving task submission atomic semantics.
Task submission involves two main steps:
1. Popping all DAG nodes that have been collected by the `dag_builder` from all threads
2. Enqueuing work to the worker thread to process and submit those nodes for execution.
When task submission from another thread jumps in between those steps, it can happen it grabs those DAG nodes for processing. In this case, since `flush_async()` did not have atomic semantics, an incorrect submission order can arise.

This can lead to nodes, that should have already been submitted, being submitted too late, and the `submit(); flush_sync()` pattern not actually guaranteeing that everything is submitted.

This PR fixes this. 

Also adds some minor improvements in this area, like improving diagnostic output, and additional debug assertions.

### DIFF
--- a/include/hipSYCL/runtime/dag_manager.hpp
+++ b/include/hipSYCL/runtime/dag_manager.hpp
@@ -28,6 +28,8 @@
 #ifndef HIPSYCL_DAG_MANAGER_HPP
 #define HIPSYCL_DAG_MANAGER_HPP
 
+#include <mutex>
+
 #include "dag.hpp"
 #include "dag_builder.hpp"
 #include "dag_direct_scheduler.hpp"
@@ -71,6 +73,9 @@ private:
   dag_direct_scheduler _direct_scheduler;
   dag_unbound_scheduler _unbound_scheduler;
   dag_submitted_ops _submitted_ops;
+
+  // Should only be used for flush_async()
+  std::mutex _flush_mutex;
 };
 
 class dag_build_guard
--- a/src/runtime/dag_builder.cpp
+++ b/src/runtime/dag_builder.cpp
@@ -269,14 +269,18 @@ dag dag_builder::finish_and_reset()
   dag final_dag = _current_dag;
   _current_dag = dag{};
 
-  final_dag.for_each_node([](dag_node_ptr node) {
-    HIPSYCL_DEBUG_INFO << "dag_builder: DAG contains operation: "
-                       << dump(node->get_operation()) << " @node " << node.get()
-                       << std::endl;
+  HIPSYCL_DEBUG_INFO << "dag_builder: DAG contains operations: " << std::endl;
+  int operation_index = 0;
+  final_dag.for_each_node([&](dag_node_ptr node) {
+    HIPSYCL_DEBUG_INFO << operation_index << ". " << dump(node->get_operation())
+                       << " @node " << node.get() << std::endl;
+
     for (dag_node_ptr req : node->get_requirements()) {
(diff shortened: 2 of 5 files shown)


al42and:
Can confirm, fixes #728 for me

fxzjshm:
Also confirm this fixes #728 for me, by testing multi-threading on a searching pipeline. (It finally runs faster, thank you.)

illuhad:
Thanks everyone for the confirmation and the patience!

Issue #726, opened 2022-04-06
RPATH with boost


pszi1ard:
**Describe the motivation for the feature request**
Currently setting LD_LIBRARY_PATH before launching an application that uses hipSYCL is requires.

**Describe the solution you'd like**
Not have to set LD_LIBRARY_PATH to boost.

sbalint98:
Can you share your syclcc.json? I believe boost should be put in the `-rpath` for the `omp_link_line` and the for the `sequential_link_line` as well. Can you also confirm that you don't set the `OMP_LINK_LINE` and `SEQUENTIAL_LINK_LINE` variables when configuring hipSYCL?

illuhad:
Yes, `-rpath` enters for `OMP_LINK_LINE`:
https://github.com/illuhad/hipSYCL/blob/2fe0d85d8cbf0ee7cd74be0a4b7f41b164849f94/CMakeLists.txt#L247

@pszi1ard Is it possible that this issue only triggers when *not* targeting `omp`, and can it be fixed for you using `--hipsycl-targets="omp;..."?

al42and:
I think I figured out what's going on. Not sure about the best way to fix it, though.

Turns out, Boost libraries are dynamically linked to other boost libraries without specifying RPath, and RPaths are not transitive: https://gitlab.kitware.com/cmake/cmake/-/issues/19626.

Example.

Here the `test` binary is dynamically linked to `libboost_fiber` and `libboost_context`. These libraries are not in `LD_LIBRARY_PATH`, but RPath is set correctly, so they are loaded just fine. However, libboost_fiber is linked to `libboost_filesystem`, and does not have RPath set, so when it gets loaded, it needs to load `libboost_filesystem` without RPath.

```console
$ syclcc -O1 --hipsycl-dryrun test.cpp -o test
[...]

PR #748, opened 2022-06-08
Fix fiinding boost library path for boost with cmake intgeration 

sbalint98:
For boost libraries that provide cmake integration, we extract the Library location from the exported target. Before this change, the entire path to the boost library was added to the -L flag. This change makes sure that only the directory is included. This is possibly a fix for #726


illuhad (approved):
Thanks!

PR #751, opened 2022-06-17
Fix signature of __hipsycl_atomic_store for double and float

al42and:
Not a major issue, just a matter of consistency and not having
misleading return types.


illuhad:
Whoopsie, this looks like a copy-paste error ;) Thanks for fixing!

illuhad (approved):

PR #763, opened 2022-07-08
[cbs] Fix compatibility issues with upstream Clang/LLVM

aaronmondal:
This commit adds two missing headers whose contents have
been moved in upstream Clang/LLVM. It also changes a call
from `PointerType::getElementType()` to
`Type::getPointerElementType()` as a temporary fix
until opaque pointers are introduced. The former was already
deprecated in Clang/LLVM 14 and has been removed in upstream.

### DIFF
--- a/src/compiler/cbs/LoopSimplify.cpp
+++ b/src/compiler/cbs/LoopSimplify.cpp
@@ -31,6 +31,7 @@
 #include "hipSYCL/compiler/cbs/SplitterAnnotationAnalysis.hpp"
 
 #include <llvm/Analysis/ScalarEvolution.h>
+#include <llvm/IR/Dominators.h>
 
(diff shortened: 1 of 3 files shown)


aaronmondal:
Fixed commit message. Is actually `Type::getPointerElementType`, not `PointerType::getPointerElementType`.

fodinabor (approved):
Hey, thank you!
Looks like a good first step in the right direction.
As far as I could tell, this should fix the build issues, but not the runtime issues.
I'll follow up with another PR that fixes those as well.

aaronmondal:
@fodinabor Yes. This only fixes compile-time issues. The clang plugin still fails (unregistered passes or something like that if i remember correctly). I'll create an issue with my findings so far 😊

PR #770, opened 2022-07-14
Add dedicated backend queues for inorder queues and priority queue support

illuhad:
* Introduces `rt::inorder_executor` for simple, straight-forward in-order execution
* Make `rt::multi_queue_executor` rely on multiple `inorder_executor`
* Add mechanism to runtime to create dedicated `inorder_executor` objects (which backends may or may not support)
* Change default behavior of in-order `queue` to request dedicated in-order executor, if supported by backend (otherwise, uses backend-provided executor). All fancy hipSYCL extensions like `hipSYCL_retarget` *should* still work, even if a dedicated inorder-executor has been requested, so no code should break. This allows expert users to have more control over scheduling decisions
* Add `hipSYCL_priority{int}` queue property. When used with dedicated in-order executor (i.e. in-order queue), passes this priority to HIP/CUDA stream creation methods. On HIP, uses `hipStreamCreateWithPriority`. I'm not sure if this actually does something or if we need to create the stream with CU mask to achieve the desired effect.

@al42and @pszi1ard this one is for you, apologies for the delay. Things were buggy for a long time, I hope it works now, although I would still consider it early access ;)

### DIFF
--- a/src/runtime/inorder_executor.cpp
+++ b/src/runtime/inorder_executor.cpp
@@ -92,6 +92,8 @@ get_maximum_execution_index_for_lane(const std::vector<dag_node_ptr> &nodes,
 inorder_executor::inorder_executor(std::unique_ptr<inorder_queue> q)
 : _q{std::move(q)} {}
 
+inorder_executor::~inorder_executor(){}
+
 bool inorder_executor::is_inorder_queue() const {
   return true;
 }
--- a/src/runtime/inorder_executor.cpp
+++ b/src/runtime/inorder_executor.cpp
@@ -1,0 +1,215 @@
+/*
+ * This file is part of hipSYCL, a SYCL implementation based on CUDA/HIP
+ *
+ * Copyright (c) 2022 Aksel Alpay
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <cassert>
+
+#include "hipSYCL/runtime/inorder_executor.hpp"
+#include "hipSYCL/runtime/inorder_queue.hpp"
+#include "hipSYCL/runtime/operations.hpp"
+#include "hipSYCL/runtime/serialization/serialization.hpp"
+
+namespace hipsycl {
+namespace rt {
+
+namespace {
+
+class queue_operation_dispatcher : public operation_dispatcher
+{
+public:
+  queue_operation_dispatcher(inorder_queue* q)
+  : _queue{q}
+  {}
+
+  virtual ~queue_operation_dispatcher(){}
+
+  virtual result dispatch_kernel(kernel_operation *op,
+                                 dag_node_ptr node) final override {
+
+    return _queue->submit_kernel(*op, node);
+  }
+
+  virtual result dispatch_memcpy(memcpy_operation *op,
+                                 dag_node_ptr node) final override {
+    return _queue->submit_memcpy(*op, node);
+  }
+
+  virtual result dispatch_prefetch(prefetch_operation *op,
+                                   dag_node_ptr node) final override {
+    return _queue->submit_prefetch(*op, node);
+  }
+
+  virtual result dispatch_memset(memset_operation *op,
+                                 dag_node_ptr node) final override {
+    return _queue->submit_memset(*op, node);
+  }
+
+private:
+  inorder_queue* _queue;
+};
+
+std::size_t
+get_maximum_execution_index_for_lane(const std::vector<dag_node_ptr> &nodes,
+                                     inorder_executor *executor,
+                                     inorder_queue* lane) {
+  std::size_t index = 0;
+  for (const auto &node : nodes) {
+    if (node->is_submitted() &&
+        node->get_assigned_device().get_backend() ==
+            lane->get_device().get_backend() &&
+        node->get_assigned_execution_lane().second == lane) {
+      if(node->get_assigned_execution_index() > index)
+        index = node->get_assigned_execution_index();
+    }
+  }
+  return index;
+}
+
+} // anonymous namespace
+
+inorder_executor::inorder_executor(std::unique_ptr<inorder_queue> q)
+: _q{std::move(q)} {}
+
+bool inorder_executor::is_inorder_queue() const {
+  return true;
+}
+
+bool inorder_executor::is_outoforder_queue() const {
+  return false;
+}
+
+bool inorder_executor::is_taskgraph() const {
+  return false;
+}
+
+void inorder_executor::submit_directly(dag_node_ptr node, operation *op,
+                                       const std::vector<dag_node_ptr> &reqs) {
+  
+  HIPSYCL_DEBUG_INFO << "inorder_executor: Processing node " << node.get()
+	  << " with " << reqs.size() << " non-virtual requirement(s) and "
+	  << node->get_requirements().size() << " direct requirement(s)." << std::endl;
+
+  assert(!op->is_requirement());
+
+  if (node->is_submitted())
+    return;
+
+  node->assign_to_execution_lane(
+      std::pair<std::size_t, void *>{0, _q.get()});
+  node->assign_execution_index(_num_submitted_operations);
+  ++_num_submitted_operations;
+
+  // Submit synchronization mechanisms
+  result res;
+  for (auto req : reqs) {
+    // The scheduler should not hand us virtual requirements
+    assert(!req->is_virtual());
+    assert(req->is_submitted());
+
+    // Nothing to do if we have to synchronize with
+    // an operation that is already known to have completed
+    if(!req->is_known_complete()) {
+      if (req->get_assigned_device().get_backend() !=
+          _q->get_device().get_backend()) {
+        HIPSYCL_DEBUG_INFO
+            << " --> Synchronizes with external node: " << req
+            << std::endl;
+        res = _q->submit_external_wait_for(req);
+      } else {
+        if (req->get_assigned_execution_lane().second == _q) {
+          HIPSYCL_DEBUG_INFO
+            << " --> (Skipping same-lane synchronization with node: " << req
+            << ")" << std::endl;
+          // Nothing to synchronize, the requirement was enqueued on the same
+          // inorder queue and will therefore be executed before
+          // the new node
+        } else {
+          assert(req->get_event());
+
+          HIPSYCL_DEBUG_INFO << " --> Synchronizes with other queue for node: "
+                            << req
+                            << std::endl;
+          // We only need to actually synchronize with the lane if this req
+          // is the operation that has been submitted *last* to the lane
+          // out of all requirements in reqs.
+          // (Follows from execution lanes being in-order queues)
+          //
+          // Find the maximum execution index out of all our requirements.
+          // Since the execution index is incremented after each submission,
+          // this allows us to identify the requirement that was submitted last.
+          inorder_queue *req_q = static_cast<inorder_queue *>(
+              req->get_assigned_execution_lane().second);
+          std::size_t maximum_execution_index =
+              get_maximum_execution_index_for_lane(reqs, this, req_q);
+          
+          if(req->get_assigned_execution_index() != maximum_execution_index) {
+            HIPSYCL_DEBUG_INFO
+                << "  --> (Skipping unnecessary synchronization; another "
+                   "requirement follows in the same inorder queue)"
+                << std::endl;
+          } else {
+            res = _q->submit_queue_wait_for(req->get_event());
+          }
+        }
+      }
+      if (!res.is_success()) {
+        register_error(res);
+        node->cancel();
+        return;
+      }
+    }
+  }
+
+  HIPSYCL_DEBUG_INFO
+      << "inorder_executor: Dispatching to lane " << _q << ": "
+      << dump(op) << std::endl;
+  
+  queue_operation_dispatcher dispatcher{_q.get()};
+  res = op->dispatch(&dispatcher, node);
+  if (!res.is_success()) {
+    register_error(res);
+    node->cancel();
+    return;
+  }
+
+  if (node->get_execution_hints()
+          .has_hint<hints::coarse_grained_synchronization>()) {
+    node->mark_submitted(_q->create_queue_completion_event());
+  } else {
+    node->mark_submitted(_q->insert_event());
+  }
+}
+
+inorder_queue* inorder_executor::get_queue() const {
+  return _q.get();
+}
+
+bool inorder_executor::can_execute_on_device(const device_id& dev) const {
+  return _q->get_device() == dev;
+}
+
+}
+}
--- a/src/runtime/inorder_executor.cpp
+++ b/src/runtime/inorder_executor.cpp
@@ -73,14 +73,13 @@ private:
 
 std::size_t
 get_maximum_execution_index_for_lane(const std::vector<dag_node_ptr> &nodes,
-                                     inorder_executor *executor,
                                      inorder_queue* lane) {
   std::size_t index = 0;
   for (const auto &node : nodes) {
     if (node->is_submitted() &&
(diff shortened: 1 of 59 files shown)


pszi1ard:
Thanks @illuhad I'll be testing this and previous changes in the coming days.

al42and:
I'm getting some crashes on at least two machines (ROCm 5.1.3 + MI100 and ROCm 4.5.2 + MI50) using the recent hwe-release-2022 branch of GROMACS.

Configure: `cmake ../.. -DCMAKE_C_COMPILER=$ROCM_HOME/llvm/bin/clang -DCMAKE_CXX_COMPILER=$ROCM_HOME/llvm/bin/clang++ -DGMX_GPU=SYCL -DGMX_SYCL_HIPSYCL=ON -DHIPSYCL_TARGETS='hip:gfx908'  -DGMX_BUILD_OWN_FFTW=ON`

Test: `HIPSYCL_RT_SCHEDULER=direct HIPSYCL_DEBUG_LEVEL=4 GMX_FORCE_UPDATE_DEFAULT_GPU=1 ./bin/mdrun-test -ntmpi 2 --gtest_filter=MultipleTimeSteppingIsNearSingleTimeSteppingPull/MtsComparisonTest.WithinTolerances/0`.

```
[hipSYCL Info] dag_builder: DAG contains operations: 
[hipSYCL Info] 0. Memset: @0x7f80d400e000 4 bytes of value  @node 0x429e0f0
[hipSYCL Info] dag_manager [async]: Flushing!
[hipSYCL Info] dag_manager [async]: Submitting node to scheduler!
[hipSYCL Info] inorder_executor: Processing node 0x429e0f0 with 0 non-virtual requirement(s) and 0 direct requirement(s).
[hipSYCL Info] inorder_executor: Dispatching to lane 0x424acb0: Memset: @0x7f80d400e000 4 bytes of value 
[hipSYCL Info] dag_manager [async]: DAG flush complete.
[hipSYCL Info] dag_manager: Checking DAG flush opportunity...
[hipSYCL Info] dag_manager: Submitting asynchronous flush...
[hipSYCL Info] dag_builder: DAG contains operations: 
[hipSYCL Info] 0. Memset: @0x7f80d400d000 4 bytes of value  @node 0x429e010
[hipSYCL Info]     --> requires node @0x429e0f0 Memset: @0x7f80d400e000 4 bytes of value 
[hipSYCL Info] queue: Constructed queue with node group id 21
[hipSYCL Info] dag_manager [async]: Flushing!
[hipSYCL Info] dag_manager [async]: Submitting node to scheduler!
[hipSYCL Info] inorder_executor: Processing node 0x429e010 with 0 non-virtual requirement(s) and 1 direct requirement(s).
[hipSYCL Info] inorder_executor: Dispatching to lane 0x424acb0: Memset: @0x7f80d400d000 4 bytes of value 
[hipSYCL Info] dag_manager [async]: DAG flush complete.
[hipSYCL Info] queue: Constructed queue with node group id 22
[hipSYCL Error] from /global/home/andreyaac/hipSYCL/src/runtime/hip/hip_queue.cpp:176 @ insert_event(): hip_queue: Couldn't record event (error code = HIP:400)
Segmentation fault
```

In the debug build, segfault is replaced by `/nethome/aland/hipSYCL/src/runtime/inorder_executor.cpp:151: virtual void hipsycl::rt::inorder_executor::submit_directly(hipsycl::rt::dag_node_ptr, hipsycl::rt::operation *, const std::vector<dag_node_ptr> &): Assertion `req->get_event()' failed.`

The trace for the failed API call goes to the first event recording after the creation of the new queue. Event pool returns a new event, not reuses an old one.

```
Thread 58 "mdrun-test" hit Breakpoint 1, hipsycl::rt::hip_queue::insert_event (this=0x7ffbfc037890) at /nethome/aland/hipSYCL/src/runtime/hip/hip_queue.cpp:185
185             __hipsycl_here(),
(gdb) bt
#0  hipsycl::rt::hip_queue::insert_event (this=0x7ffbfc037890) at /nethome/aland/hipSYCL/src/runtime/hip/hip_queue.cpp:185
#1  0x00007fff0279bbfc in hipsycl::rt::host_timestamped_event::host_timestamped_event (this=0x7ffc317e0d28, q=0x7ffbfc037890) at /nethome/aland/hipSYCL/include/hipSYCL/runtime/hip/../generic/host_timestamped_event.hpp:43
#2  0x00007fff02795128 in hipsycl::rt::hip_queue::hip_queue (this=0x7ffbfc037890, be=0x1e2ab70, dev=..., priority=0) at /nethome/aland/hipSYCL/src/runtime/hip/hip_queue.cpp:150
#3  0x00007fff027b3420 in std::make_unique<hipsycl::rt::hip_queue, hipsycl::rt::hip_backend*, hipsycl::rt::device_id&, int&> (__args=@0x7ffc317e0fcc: 0, __args=@0x7ffc317e0fcc: 0, __args=@0x7ffc317e0fcc: 0)
    at /usr/lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/unique_ptr.h:857
#4  0x00007fff027b2f7b in hipsycl::rt::hip_backend::create_inorder_executor (this=0x1e2ab70, dev=..., priority=0) at /nethome/aland/hipSYCL/src/runtime/hip/hip_backend.cpp:112
#5  0x00007ffff7a83ec7 in hipsycl::sycl::queue::init (this=0x7ffc317e15d0) at /nethome/aland/modules/hipSYCL/develop-20220801-69cbcf014-rocm4.5-pr770/bin/../include/SYCL/../hipSYCL/sycl/queue.hpp:995
#6  0x00007ffff7a83566 in hipsycl::sycl::queue::queue(hipsycl::sycl::context const&, std::vector<hipsycl::sycl::device, std::allocator<hipsycl::sycl::device> > const&, std::function<void (std::vector<std::__exception_ptr::exception_ptr, std::allocator<std::__exception_ptr::exception_ptr> >)> const&, hipsycl::sycl::property_list const&) (this=0x7ffc317e15d0, syclContext=..., devices=std::vector of length 1, capacity 1 = {...}, asyncHandler=..., propList=...)
    at /nethome/aland/modules/hipSYCL/develop-20220801-69cbcf014-rocm4.5-pr770/bin/../include/SYCL/../hipSYCL/sycl/queue.hpp:242
#7  0x00007ffff7a8aa32 in hipsycl::sycl::queue::queue (this=0x7ffc317e15d0, syclContext=..., devices=std::vector of length 1, capacity 1 = {...}, propList=...)
    at /nethome/aland/modules/hipSYCL/develop-20220801-69cbcf014-rocm4.5-pr770/bin/../include/SYCL/../hipSYCL/sycl/queue.hpp:209
#8  0x00007ffff7a7bf24 in hipsycl::sycl::queue::queue (this=0x7ffc317e15d0, syclContext=..., syclDevice=..., propList=...) at /nethome/aland/modules/hipSYCL/develop-20220801-69cbcf014-rocm4.5-pr770/bin/../include/SYCL/../hipSYCL/sycl/queue.hpp:189
#9  0x00007ffff7a7b8b8 in DeviceStream::DeviceStream (this=0x7ffbfc03ca80, deviceContext=..., useTiming=false) at /tmp/gromacs/src/gromacs/gpu_utils/device_stream_sycl.cpp:83
...
```

This happens when multiple GPUs are used simultaneously from either two threads or two processes (when compiled with real MPI). Works fine with the `develop` branch. I plan to keep debugging this problem, but perhaps there's some obvious issue you can see here?

illuhad:
Hm.. I vaguely remember having had some issues around that area that I had to debug, so maybe there's indeed something broken that was fixed by some of the other PRs that already went into develop.
I've looked at the code and at first glance, it looks correct. The event pool seems to be properly locked. Since this seems to be a multithreading issue, I'd suggest maybe trying thread sanitizer. With recent hipSYCL, the runtime can be conveniently instrumented with a sanitizer, by building e.g. with `-DHIPSYCL_RT_SANITIZE=thread` for thread sanitizer :-) You will still have to compile the client application using the appropriate `-fsanitize` flags though.

al42and:
> Since this seems to be a multithreading issue, I'd suggest maybe trying thread sanitizer.

Thanks for the idea, I'll try the sanitizer. I doubt it's the threading issue, though: (a) it's 100% repeatable, (b) is triggered at exactly the same place with real MPI, when we have two separate processes (one per GPU) instead of threads.

al42and:
Neither TSan nor ASan report any issues despite the problem persisting.

But I noticed that setting `HIP_VISIBLE_DEVICES=0` (making both threads use the same device) avoids the issue. Perhaps ROCm runtime is to blame?

Sidenote: curiously, a simple program launching a kernel complains about a data race in `async_worker` between at the program's end (#794), while no such thing is getting reported for GROMACS on the same machine.

illuhad:
Maybe it is related to synchronization mechanisms across devices. Does it also happen with CUDA on multiple devices?

al42and:
> Maybe it is related to synchronization mechanisms across devices. Does it also happen with CUDA on multiple devices?

Good point, I initially tested only a single device. Yes, it does, same "400 / invalid handle" error:

```
[hipSYCL Info] dag_manager [async]: DAG flush complete.
[hipSYCL Info] queue: Constructed queue with node group id 17
[hipSYCL Info] queue: Constructed queue with node group id 18
[hipSYCL Info] queue: Constructed queue with node group id 19
[hipSYCL Info] queue: Constructed queue with node group id 20
[hipSYCL Error] from /nethome/aland/hipSYCL/src/runtime/cuda/cuda_queue.cpp:185 @ insert_event(): cuda_queue: Couldn't record event (error code = CUDA:400)
```

illuhad:
Are you by chance using any backend interop in Gromacs that involves calls to `hipSetDevice()` or `cudaSetDevice()`?

al42and:
> Are you by chance using any backend interop in Gromacs that involves calls to hipSetDevice() or cudaSetDevice()?

No.

We do an intermediate runtime shutdown (after device detection, all SYCL objects leave scope; later the runtime is re-initialized), but that alone does not trigger the error.

EDIT: What we do, is that we create new in-order queues (on one device) and submit to other ones (on the other device) in parallel, and that's when the crash happens.

illuhad:
Yes, runtime restart should not be an issue. (You can set the environment variable `HIPSYCL_PERSISTENT_RUNTIME=1` to be sure)

My suspicion was because what *has* changed with queue construction in this PR, is that queue construction now takes place in application threads during the runtime of the program. And we have a thread-local cache of the index of the currently active CUDA/HIP device so that we don't have to constantly invoke `cuda/hipGet/SetDevice()`.
Now, when the user also does some backend interop which may change the active device, the value of the device index in the cache might no longer reflect the real active device, and then SYCL objects might be constructed with the wrong device active. This might explain your invalid resource handle errors and also why it only happens in multi-gpu scenarios.

This is usually not a big issue, because most of the SYCL runtime backend calls run in their own thread, and so have their own thread-local device index cache. The exception are USM memory management functions and now, in-order queue construction.

To be sure, you could try to update `_device`  with `hipGetDevice()` here: https://github.com/illuhad/hipSYCL/blob/8907351836d98098094df1d697c438db80dc3334/src/runtime/hip/hip_device_manager.cpp#L50

al42and:
> To be sure, you could try to update _device with hipGetDevice() here: 

Thanks, that resolves the problem.

illuhad:
Ok... This strongly indicates that there is interference with something else setting the CUDA or HIP device. Is it possible that maybe your rocFFT interop interferes?

illuhad:
Potentially related: https://github.com/illuhad/hipSYCL/issues/446, see in particular https://github.com/illuhad/hipSYCL/issues/446#issuecomment-767531620

al42and:
That happens before the rocFFT is even called. And also happens on CUDA, where we don't do any FFTs yet. Here are two logs for the successful run (when less queues are created) and a failed one (which crashes during the creation of those extra queues).

[log_fail.txt](https://github.com/illuhad/hipSYCL/files/9270147/log_fail.txt)
[log_pass.txt](https://github.com/illuhad/hipSYCL/files/9270148/log_pass.txt)
 
I have not come up with a repro yet, but I managed to cause a deadlock after runtime shutdown: https://gist.github.com/al42and/c04d87a0ceb7da8c1c461fdbc6555459. Behaves ok with `develop`. EDIT: happens even with the hipGetDevice workaround.

al42and:
> Potentially related: https://github.com/illuhad/hipSYCL/issues/446, see in particular https://github.com/illuhad/hipSYCL/issues/446#issuecomment-767531620

It indeed might be related, but the issue here manifests even when we use MPI, and each GPU is mostly used from its own process (although each process pokes all GPUs at the very start).

src/runtime/inorder_executor.cpp
```diff
@@ -0,0 +1,222 @@
...
+        index = node->get_assigned_execution_index();
+    }
+  }
+  return index;
+}
+
+} // anonymous namespace
+
+inorder_executor::inorder_executor(std::unique_ptr<inorder_queue> q)
+: _q{std::move(q)} {}
```
al42and: Don't we need

```suggestion
: _q{std::move(q)}, _num_submitted_operations{0} {}
```

illuhad: Absolutely, we have to initialize it to 0. But I don't think this can impact multi-gpu cases. Does it have an effect for your issue?

al42and:
Here are hipSYCL traces from two runs: with the original version of this PR and with the `hipGetDevice` fix. There are some changes in the handling of the devices long before the failure occurs:

[fail.rank_1.log](https://github.com/illuhad/hipSYCL/files/9270426/fail.rank_1.log)
[fixd.rank_1.log](https://github.com/illuhad/hipSYCL/files/9270427/fixd.rank_1.log)

src/runtime/inorder_executor.cpp
al42and: No impact, but I noticed it while looking at the code.

pszi1ard:
@illuhad I finally got things running LUMI-G and in multi-rank testing of GROMACS using of the current develop branch I ran into an issue which seems to have been introduced by this code.

Multi-rank runs with commit 14fe9a5 (so we have multiple queues per rank) throw this error:
```
:0:rocdevice.cpp            :2603: 288640691419 us: 14786: [tid:0x14a79c3b7700] Device::callbackQueue aborting with error : HSA_STATUS_ERROR_OUT_OF_RESOURCES: The runtime failed to allocate the necessary resources. This error may also occur when the core runtime library needs to spawn threads or create internal OS-specific events. code: 0x1008                 
:0:rocdevice.cpp            :2603: 288640695575 us: 14787: [tid:0x151c54fa7700] Device::callbackQueue aborting with error : HSA_STATUS_ERROR_OUT_OF_RESOURCES: The runtime failed to allocate the necessary resources. This error may also occur when the core runtime library needs to spawn threads or create internal OS-specific events. code: 0x1008
```

Here's the HIP debug log output:  [slurm-1423480.out.tar.gz](https://github.com/illuhad/hipSYCL/files/9386990/slurm-1423480.out.tar.gz) 

Neither the parent commit 598767ab nor v0.9.2 reproduces this.

illuhad:
I have seen this error only once before; namely when experimenting with running hipSYCL on the APU on my notebook (which is not really supported by ROCm, so probably something is going wrong inside ROCm there). As far as I am aware, current `develop` behaves correctly on supported ROCm devices. Do you have instructions how I could go about reproducing this?

Which ROCm version are you using? Is the `amd/5.0.2` module in the output referring to ROCm 5.0.2?
Can you confirm if you also have all of the related PRs merged in? I see that it seems to make a lot of calls to `hipEventCreate()` which might happen e.g. if #771 is missing.

pszi1ard:
> Do you have instructions how I could go about reproducing this?

For LUMI-G (or similarly set up machine) here are the instructions:
```
# $ module list 
#
# Currently Loaded Modules:
#   1) libfabric/1.15.0.0       4) xpmem/2.4.4-2.3_9.1__gff0e1d9.shasta       7) craype-x86-trento  10) cray-dsmml/0.2.2       13) PrgEnv-amd/8.3.3
#   2) craype-network-ofi       5) ModuleLabel/label                    (S)   8) amd/5.0.2          11) cray-mpich/8.1.18      14) cray-fftw/3.3.10.1
#   3) perftools-base/22.06.0   6) init-lumi/0.1                        (S)   9) craype/2.7.17      12) cray-libsci/22.08.1.1  15) craype-accel-amd-gfx90a

# build hipSYCL develop 03de8f8

 CMAKE_XOPTS="$CMAKE_XOPTS -DWITH_ACCELERATED_CPU=off -DHIPSYCL_NO_DEVICE_MANGLER=ON" 
build_type=Release

cmake ../ -DCMAKE_C_COMPILER=`which clang` -DCMAKE_CXX_COMPILER=`which clang++`  \
  -DCMAKE_INSTALL_PREFIX=${install_prefix} \
  -DBOOST_ROOT=${BOOST_HOME} -DWITH_CUDA_BACKEND=OFF -DWITH_ROCM_BACKEND=ON \
  -DCMAKE_BUILD_TYPE=${build_type} \
  ${CMAKE_XOPTS}

# build GROMACS

cmake ../ -DCMAKE_C_COMPILER=`which amdclang` -DCMAKE_CXX_COMPILER=`which amdclang++` -DCMAKE_PREFIX_PATH="/opt/cray/pe/fftw/3.3.8.12/x86_trento;/users/szpall/programs/hipsycl/dev-03de8f8_hip_rocm5.0.2" -DGMX_CYCLE_SUBCOUNTERS=ON -DBUILD_SHARED_LIBS=ON -DGMX_PREFER_STATIC_LIBS=OFF -DGMX_EXTERNAL_BLAS=OFF -DGMX_EXTERNAL_LAPACK=OFF -DGMX_INSTALL_NBLIB_API=OFF -DGMXAPI=OFF -DGMX_GPU=SYCL -DGMX_SYCL_HIPSYCL=ON -DHIPSYCL_TARGETS="hip:gfx90a" -DHIPSYCL_SYCLCC_EXTRA_ARGS="-ffast-math -munsafe-fp-atomics" -DGMX_MPI=ON  -DCMAKE_CXX_FLAGS="-I${MPICH_DIR}/include" -DMPI_CXX_COMPILER_INCLUDE_DIRS="${MPICH_DIR}/include" -DMPI_CXX_LINK_FLAGS="-L${MPICH_DIR}/lib -lmpi -L${CRAY_MPICH_ROOTDIR}/gtl/lib -lmpi_gtl_hsa"

# slurm script to run GROMACS:

#!/bin/bash                                                                                                                                                                                                                                                         
#SBATCH --time=00:10:00
#SBATCH --ntasks=2
#SBATCH --cpus-per-task=8                                                                                                                                                                                                        
#SBATCH --gpus-per-task=1
#SBATCH --exclude=


srun  --cpu-bind=none  --ntasks=2   $PATH_TO_GROMCAS/gmx_mpi mdrun  -nsteps 4000 -resetstep 2000 -pin on -quiet -v -noconfout   -nb gpu -update gpu -ntomp 8 -npme -1 -g test_1N_2x8_1TPC_hipSYCL-dev-03de8f8_nb-gpu_update-gpu_jID1438250
```

I have only reproduced this with MPI on LUMI and admittedly the only other machine where I have only MPI testing is out local 2xMi50 node where I used ROCm 4.5 and OpenMPI whereas this is 5.0.2 (drivers and runtime) with Cray-MPICH.

> Which ROCm version are you using? Is the amd/5.0.2 module in the output referring to ROCm 5.0.2?

Yes, this is 5.0.2 both driver and runtime. (current defaults on LUMI)

> Can you confirm if you also have all of the related PRs merged in? I see that it seems to make a lot of calls to hipEventCreate() which might happen e.g. if https://github.com/illuhad/hipSYCL/pull/771 is missing.

Yes, I'm using develop 03de8f8ec from Aug 9 which contains the PR.

PR #771, opened 2022-07-14
Use weak_ptr in node requirements list

illuhad:
Previously, `dag_node` has maintained `shared_ptr` objects to its requirements in the dependency graph. For deep dependency graphs, this can prevent nodes from being deleted even if the underlying operation has already completed since they might still be referenced.
So far, this has not been a huge issue. However, with the introduction of event pools in #757, this becomes important: If the DAG node cannot be deleted, its event stays alive and cannot be returned to the pool. In the worst case, this can prevent us from benefiting from the event pool at all, having to create individual events for each operation again.

To fix this, this PR changes the requirements list to `weak_ptr` so that DAG nodes can be deleted as soon as their operation has finished.

TODO: ~Seems to work on GPU, but still crashes on CPU. Presumably because CPU backend has its own worker thread to execute kernels, and this additional asynchronous behavior probably changes the lifetime guarantees of DAG nodes.~

### DIFF
--- a/include/hipSYCL/runtime/dag_node.hpp
+++ b/include/hipSYCL/runtime/dag_node.hpp
@@ -98,7 +98,7 @@ public:
   // Add requirement if not already present
   void add_requirement(dag_node_ptr requirement);
   operation* get_operation() const;
-  const std::vector<dag_node_ptr>& get_requirements() const;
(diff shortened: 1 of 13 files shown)


illuhad:
At last, this should be stable now :)

PR #773, opened 2022-07-15
[Plugin] Resolve version macros in HIPSYCL_STRINGIFY

aaronmondal:
The previous variant did not work with some preprocessors, falsely resolving macros like `#define HIPSYCL_VERSION_MINOR 9` to "HIPSYCL_VERSION_MINOR" instead of "9".


fodinabor (approved):
Good catch. Looks good to me.

illuhad (approved):

PR #775, opened 2022-07-16
Add missing sycl::nd_range::get_group_range function

al42and:
https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsubsec:nd-range-class


illuhad (approved):
Thanks!

PR #777, opened 2022-07-16
[Styling] Introduce pre-commit hooks

aaronmondal:
Currently it is not possible to "fix" PEP8 compliance without changing
large chunks of python code. This commit blindly runs yapf without
additional changes so that future commits can focus on improving the
actual content of the python files. Reproducible by running
```
pip install pre-commit
pre-commit install
pre-commit run -a
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,6 @@
+repos:
+- repo: https://github.com/google/yapf
+  rev: v0.32.0
+  hooks:
+    - id: yapf
+      name: "yapf"
--- a/bin/syclcc-clang
+++ b/bin/syclcc-clang
@@ -1,5 +1,4 @@
 #!/usr/bin/env python3
-
 '''
  *
  * This file is part of hipSYCL, a SYCL implementation based on CUDA/HIP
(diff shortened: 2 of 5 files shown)


aaronmondal:
@illuhad This commit may be too large to be commited by an outside contributor. It also introduces a level of indirection to git blame for the python files. Please feel free to close this if it is too much of a change.

Basic PEP8 compliance should be targeted nonetheless, but I am not sure how to tackle this without a commit as large as this one.

illuhad:
Most of the changes seem to be changing the indentation from 2->4. I cannot find a technical reason in the PEP for this. What is the rationale behind this? hipSYCL uses 2 space indentation everywhere, so it might be convenient for the project to also stick to this in the python files.

aaronmondal:
I think the [official guideline](https://peps.python.org/pep-0008/#indentation) (randomly?) decided on this, and the second most used guideline by [google](https://google.github.io/styleguide/pyguide.html#34-indentation) agreed. The subjectiveness of this decision is apparent in e.g. TensorFlow, which uses 2 space indentation in Python, effectively breaking the own guidelines :sweat_smile: 

I think the 4 spaces make sense since most (all?) formatters, linters and editors will default to 4 space indentation for Python. My initial plan was to use more expressive linters such as flake8 with wemake. It was not really possible to use these linters, since the actually relevant warnings got lost in all the indentation warnings.

The 4 space indentation is hardcoded into flake8, so a 2 space indentation would require disabling the related warnings `E111` with no way to explicitly enforce 2 space indentation.

A workaround would be to disable the flake8 indentation warnings and use a different linter to enforce a 2 space indentation. This additional complexity seems kind of unnecessary to me.

aaronmondal:
Closing this and will send a new PR to get pre-commit hooks running in a less invasive way.

illuhad:
One approach might perhaps be two split the PR in two: One for all the indentation changes, and one for the rest. This might make it easier to see what is actually changed.

aaronmondal:
@illuhad Yes it is probably better to do these things more incrementally. I sent another pr that fixes more trivial issues first. Then we can expand on that and introduce formatting etc more gentle/gradually.

PR #792, opened 2022-08-01
[infra] Introduce pre-commit hooks

aaronmondal:
This commit introduces basic style uniformity in the entire repository.
Most notably, it fixes all occurrances of trailing whitespaces and missing
newlines at the end of files.

I think it makes sense to defer the commit that documents the pre-commit hooks so that this commit only contains the changes made by the hooks, and the `.pre-commit-config.yaml` file.

I also deferred a commit unifying the C++ code via `clang-format`, since the formatting style should probably be discussed first.

While these hooks can (and should) be installed by contributors locally, a next step will be adding this to the CI pipelines so that they run before the actual test pipelines start. This way we can give contributors immediate feedback on style issues instead of them having to wait for 20-60 minutes for the full CI feedback.


aaronmondal:
Closing since too outdated. I'll recreate something like this at some point that starts off with just a single check. Then we can gradually introduce these changes over time.

PR #795, opened 2022-08-05
Avoid printing unprintable from memset_operation::dump

al42and:
Since `get_pattern` returns `unsigned char`, it was printed as such.
So, for example, when zeroing an array, a NULL byte was printed to
stderr in debug mode. Nothing serious, but not very helpful either.

Since using `memset` to fill the buffer with repetitions of a printable
character is not the most common usecase, now we print the pattern as an
integer.


illuhad (approved):
Thanks!

PR #796, opened 2022-08-05
Use device managers in allocators instead of setting device directly

illuhad:
It seems the CUDA and HIP allocators have invoked `cudaSetDevice()` and `hipSetDevice()` directly, which they should never do because it can invalidate the state of the `hip_device_manager` and `cuda_device_manager`, which are used to manage the active device by the rest of the runtime.

This is a terrible oversight. I don't know why these function calls are used here instead of the device managers; the files even include them but then don't use them. This seems to be a very old bug that may have been as old as when we introduced the current runtime.

There might be a case for invoking both the `device_manager::activate_device()` *as well as* `cuda/hipSetDevice()` directly, as this would make USM memory management functions more robust in backend interop scenarios where the user also sets device explicitly.

I believe this fixes the issues described in #770 

CC @al42and


al42and:
Can confirm, that fixes the issue I've been observing with #770 :+1:

PR #800, opened 2022-08-08
Manually check version of clang if ROCm is used.

### LINKED ISSUE #797: cmake options to select the ROCm LLVM
What is the cmake command options for selecting ROCm's LLVM (/opt/rocm/llvm/) instead of the LLVM installed by default ?  Thanks.

-- Building hipSYCL against LLVM configured from /usr/lib/llvm-10/cmake
-- Selecting clang: /usr/lib/llvm-10/bin/clang++

fodinabor:
This adds simplistic ROCm version discovery by relying on Clang's version output and adds an exemplary check for ROCm < 5 to enable `HAS_TYPED_PTR`. This should fix #797 and enable handling more issues more gracefully.

This assumes, that ROCm's Clang is built from a git branch called `roc-([0-9]+).([0-9]+).([0-9]+)`.
An alternative approach would be to first (somehow) detect whether we have a ROCm clang and then parsing the output `${CLANG_EXECUTABLE_PATH} "-v" "-x" "hip" "--rocm-path=${ROCM_PATH}" "/dev/null"` for the string `Found HIP installation: .+, version ([0-9]+).([0-9]+).([0-9]+)`.
Maybe we can detect whether we have AMD clang more easily from CMake? If not, we could rely on it reporting `AMD clang version ..`, which has to have been set using `-DCLANG_VENDOR="AMD clang"` during the LLVM build, though, which might not be the case if ROCm's llvm is built manually?

I was only able to verify any of those commands' stability with ROCm 5.2. Maybe someone can chip in with testing 4.5 and if it fails reporting the output of those two commands:
```bash
$ROCM_PATH/llvm/bin/clang --version
$ROCM_PATH/llvm/bin/clang -v -x hip < /dev/null |& grep HIP 
```

@al42and could you check with 4.5.2?


CMakeLists.txt
```diff
@@ -165,6 +165,18 @@ if(BUILD_CLANG_PLUGIN)
...
+      OUTPUT_VARIABLE CLANG_VERBOSE_OUTPUT)
+    string(REGEX MATCH "clang version [0-9]+.[0-9]+.[0-9]+ \\(.+ roc-([0-9]+).([0-9]+).([0-9]+)" AMD_CLANG_VERSION ${CLANG_VERBOSE_OUTPUT})
+
+    if(AMD_CLANG_VERSION)
+      set(ROCM_VERSION_MAJOR ${CMAKE_MATCH_1})
+      set(ROCM_VERSION_MINOR ${CMAKE_MATCH_2})
+      set(ROCM_VERSION_PATCH ${CMAKE_MATCH_3})
+      message(STATUS "AMD clang version: ${ROCM_VERSION_MAJOR}.${ROCM_VERSION_MINOR}.${ROCM_VERSION_PATCH}")
+    endif()
+  endif()
```
al42and, line 185: Would also be nice to automatically enable the device mangler fix:

```suggestion
  endif()
  if(${LLVM_VERSION_MAJOR} EQUAL 13 AND AMD_CLANG_VERSION AND NOT DEFINED ${HIPSYCL_NO_DEVICE_MANGLER})
      message(STATUS "Enabling workarounds for AMD clang 13.x.x: HIPSYCL_NO_DEVICE_MANGLER")
      set(HIPSYCL_NO_DEVICE_MANGLER TRUE)
  endif()
```

al42and:
Version detection works fine with 4.2.0, 4.3.0, and 4.5.2.

Does **not** work with 3.9.0 and 4.1.0 (clang does not print anything to differentiate it from upstream), but those versions are ancient, so whatever.

Would've been nice if hipSYCL were to auto-apply the `HIPSYCL_NO_DEVICE_MANGLER` fix in this case (see my comment), but not a dealbreaker.

CMakeLists.txt
illuhad: It's not really a fix though, just a workaround at the expense of other functionality... I don't think this should be applied automatically because it *will* break any explicit multipass compilation flow, leading to silent kernel launch failures due to mismatch of the mangled kernel name between the host binary and the compiled device binary.

So, this should probably be opt-in, with clear warning signs...

al42and: If we have Clang 13 from ROCm 4.5.2 (and likely earlier versions as well), hipSYCL would not compile unless the device mangler functionality is disabled. As long as we trust the version detection, I don't see how **not** setting `HIPSYCL_NO_DEVICE_MANGLER` is better for functionality.

But if you think emitting a warning is better, that's fine too. Certainly better than the compile error :)

Alternatively, the most robust approach, IMHO, would be to use `check_cxx_source_compiles`, but it requires carefully setting the flags to use the correct LLVM. I tried a bit and have not managed to make it work.

fodinabor: I just added an error in CMake (using your condition) if `HIPSYCL_NO_DEVICE_MANGLER` is not set, as it'd error in the compilation anyways (IIRC?).

illuhad: @al42and The issue is that it will break half of hipSYCL's compilation flows in a very subtle and hard-to-debug way. And cmake warnings are often not something that sticks out very much.

We've decided that the best course of action is to inform the user that  they need to set `HIPSYCL_NO_DEVICE_MANGLER` in case we detect the problem. In that way, the user is forced to explicitly acknowledge that they do not want explicit multipass support, including all of the consequences.

CMakeLists.txt
```diff
@@ -175,6 +175,11 @@ if(BUILD_CLANG_PLUGIN)
       set(ROCM_VERSION_MINOR ${CMAKE_MATCH_2})
       set(ROCM_VERSION_PATCH ${CMAKE_MATCH_3})
       message(STATUS "AMD clang version: ${ROCM_VERSION_MAJOR}.${ROCM_VERSION_MINOR}.${ROCM_VERSION_PATCH}")
+
+      if(${LLVM_VERSION_MAJOR} EQUAL 13 AND NOT DEFINED ${HIPSYCL_NO_DEVICE_MANGLER})
+        message(SEND_ERROR "AMD clang version not compatible with hipSYCL explicit multipass. "
+                           "To disable explicit multipass set -DHIPSYCL_NO_DEVICE_MANGLER=ON.")
```
illuhad: Can we also mention that this breaks simultaneous compilation for both CUDA and HIP? This is probably when most people are going to come into contact with explicit multipass.

CMakeLists.txt
al42and: Ok, makes sense.

illuhad (approved):
Awesome!

illuhad:
FWIW, this does not seem to work on Arch Linux with the rocm-arch AUR packages (https://github.com/rocm-arch/rocm-arch/blob/master/rocm-llvm/PKGBUILD):
```
$ /opt/rocm/llvm/bin/clang++ --version
clang version 14.0.0 (https://aur.archlinux.org/rocm-llvm.git bc7001d688249409ad3ba05561595e7bec515bdf)
```
Not a priority since Arch Linux is not an official supported ROCm target, but worth pointing out.

fodinabor:
Ah, I've been using https://aur.archlinux.org/packages/rocm-llvm on Manjaro, which apparently poses as the "official" AMD clang.
Might be worth considering one of the alternative approaches: 
- Use a two-step process: figuring out that we have a ROCm clang (path starts with either `/opt/rocm/` or `${ROCM_PATH}`, or check whether `clang++ --version` contains either `rocm`, `roc-(version)` or `AMD clang`) and then use `$ROCM_PATH/llvm/bin/clang -v -x hip < /dev/null |& grep HIP` to get the version.
- Rely on the HIP CMake integration

I don't know much about when the CMake integration was added / how reliably it can be used.
The first approach I'd expect to also theoretically support some older ROCm versions.
Is that something worth pursuing?

illuhad:
hu, I think this is the package from the ROCm-Arch project and the same one that I am using. Do you get a different output there?

I think getting ROCm version from `clang -v -x hip` is probably a good idea since it offloads version detection to clang. However in practice we might have to give it additional arguments such as `--rocm-path` if ROCm lives in a non-standard location.

Another idea to check for ROCm-clang might be to check whether clang lives in a subdirectory of `$ROCM_PATH`. It's not bulletproof, but since the other ideas seem to be heuristics as well, this is maybe one that might work well in practice?

fodinabor:
I do get
```
$ /opt/rocm/llvm/bin/clang++ --version
AMD clang version 14.0.0 (https://github.com/RadeonOpenCompute/llvm-project roc-5.2.0 22204 50d6d5d5b608d2abd6af44314abc6ad20036af3b)
```
For a refined version using the 2-stage approach see #807

PR #804, opened 2022-08-09
Add HIPSYCL_EXT_QUEUE_PRIORITY flag

al42and:
Follow-up to #770, to show that we have a custom queue property available.


illuhad (approved):
Thanks!

PR #806, opened 2022-08-09
Fix CMake error with ROCm 4.5 Clang

al42and:
It should be `DEFINED var`, not `DEFINED ${var}`.

Follow-up to #800.


fodinabor (approved):
Copy & pasta :)
Thanks!

PR #819, opened 2022-08-31
Add minimal install script

illuhad:
This can be useful for training courses etc. when people just need to get a SYCL implementation asap with minimal dependencies.

### DIFF
--- a/install/scripts/hipsycl-minimal-install.sh
+++ b/install/scripts/hipsycl-minimal-install.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+set -e
+
+echo "This will install hipSYCL into the current directory in a VERY minimal configuration:"
+echo "The installation will only support CPU and no LLVM compiler acceleration of SYCL kernels."
+echo "For production use and performance, this may not be ideal, but if you just quickly want to have a SYCL implementation, it might be perfect :-)"
+echo ""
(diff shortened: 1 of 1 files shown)


Momellouky:
Hello, 
I have tried to run this script on ubuntu 22.04 LTS. I am facing a problem with the following include (#include "clang/AST/ASTContext.h"). It says that No such file or directory. 
The complete error message: 
fatal error: clang/AST/ASTContext.h: No such file or directory
   38 | #include "clang/AST/ASTContext.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.

illuhad:
It's possible that, if cmake finds an LLVM installation, the new SSCP compiler is enabled which then might fail if you are lacking a sufficiently new LLVM installation (in your case, probably the llvm `-dev` packages are missing).

Can you try:
1. Open the minimal install script
2. Edit the line where `cmake` is invoked and add `-DWITH_SSCP_COMPILER=OFF`
3. Run the script again

Momellouky:
Hi.
Yes, you are right. In fact now it works perefectly. It compiles and everything is fine. 
Thank you.

illuhad:
Thanks for the feedback, I have opened PR #1009 to fix this issue in the script.

PR #824, opened 2022-09-06
Add HIPSYCL_SYCLCC_EXTRA_COMPILE_OPTIONS

al42and:
Similar to `HIPSYCL_SYCLCC_EXTRA_ARGS`, but will only be applied to files explicitly listed in `SOURCES` argument of `add_sycl_to_target`.

Examples of flags where it can be useful are various `-Wno-...` and `-ffast-math` (which can break some code on host, but can be desired for device kernels).

I'm open to changing the name of the variable.

Testing: https://gist.github.com/al42and/2c21e842702f45b9be44d793a77a7d69, there `mkdir build && cmake -DHIPSYCL_TARGETS=omp -Bbuild/ && cmake --build build/`


illuhad:
This is good in principle, but we need to come up with another name. It's not really device-only, as it also affects host compilation. Maybe a name to focus on the fact that it affects the SYCL *compiler*, not linker?

al42and:
It's setting `COMPILE_OPTIONS`, so how about `HIPSYCL_SYCLCC_EXTRA_COMPILE_OPTIONS`?

illuhad:
I like it!

Issue #774, opened 2022-07-16
Information descriptors are enum values, must be structs


al42and:
**Bug summary**

Per [2020r5 spec](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:information-descriptors), the info descriptors (like `sycl::info::device::device_type`) must be `struct`s:

```cpp
namespace sycl {
namespace info {
namespace platform {
struct profile;
}}}
```

In hipSYCL, they are [enum values](https://github.com/illuhad/hipSYCL/blob/2d937998228826f183a670d927749a4e6cb93a11/include/hipSYCL/sycl/info/platform.hpp#L41):

```cpp
namespace sycl {
namespace info {
enum class platform : unsigned int
{
  profile,
}}}
```

**To Reproduce**

```cpp
// Compile with syclcc test.cpp --hipsycl-targets=omp
#include <sycl/sycl.hpp>
int main() {
  using dt = sycl::info::device::device_type;
  return 0;
}
```

**Expected behavior**

The code above should compile fine. In practice, it produces an error:

`test.cpp:4:14: error: no type named 'device_type' in 'hipsycl::sycl::info::device'; did you mean 'hipsycl::sycl::info::device_type'?`

**Describe your setup**
* Latest `develop` version of hipSYCL, 2d937998228826f183a670d927749a4e6cb93a11
* Describe the dependencies that hipSYCL sits on top of in your setup: 
   * clang-14

**Additional context**

Pretty minor thing and easy to work around. Just wanted to record it.

aaronmondal:
Potentially related is that
> Information query descriptors have been changed to structures under namespaces named accordingly. param_traits has been removed and the return type of an information query is now contained in the descriptor.

at the bottom of [Appendix D](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#cha:what-changed-from).

In hipSYCL:
https://github.com/illuhad/hipSYCL/blob/f31fd92f9cb90852e75025c7b7d9ca748f15c31a/include/hipSYCL/sycl/info/param_traits.hpp#L38-L44

illuhad:
Yes, this is all due to changes from SYCL 1.2.1 to 2020. For most applications, the practical effect of this is fairly limited, so it wasn't a priority to do all the refactoring work to change this to SYCL 2020 until now.

PR #832, opened 2022-09-22
Fix device::max_work_item_sizes

nmnobre:
Hi @illuhad,

I've noticed that `sycl::device::get_info<sycl::info::device::max_work_item_sizes>()` returns the allowed per-dimension number of work items, i.e. threads, not on a block, i.e. a work group, but on a grid of blocks. It seems the standard is asking for the former? I don't know if you can test it easily, but our friends at [intel](https://github.com/intel/llvm/blob/0cd04144d9ca83371c212e8e4709a59c968291b9/sycl/plugins/cuda/pi_cuda.cpp#L878-L879) seem to have the same opinion.

Currently, I've only changed the behaviour for the cuda backend to illustrate the idea, but we should change the others too before eventually merging.

Cheers,
-Nuno


illuhad:
Thanks for bringing this up. I've had a look at the spec, and I think you are right. 

However, I think the `max_global_size`  property of the runtime should remain the same, since it correctly returns the max global size as the name suggests. We should probably introduce new properties for the max group size in each dimension, `max_group_size0, max_group_size1` etc.

nmnobre:
> However, I think the `max_global_size` property of the runtime should remain the same, since it correctly returns the max global size as the name suggests. We should probably introduce new properties for the max group size in each dimension, `max_group_size0, max_group_size1` etc.

Sounds good, I think it looks better now.

src/runtime/omp/omp_hardware_manager.cpp
```diff
@@ -141,6 +141,15 @@ omp_hardware_context::get_property(device_uint_property prop) const {
   case device_uint_property::max_global_size2:
     return std::numeric_limits<std::size_t>::max();
     break;
+  case device_uint_property::max_group_size0:
+    return std::numeric_limits<std::size_t>::max();
```
fodinabor: I think we want to honor the 1024 elements limit as with `max_group_size` - this is due to the (admittedly arbitrary) limit present in the accelerated CPU backend.

nmnobre: You are quite right, that was an oversight on my part.

src/runtime/cuda/cuda_hardware_manager.cpp
```diff
@@ -219,6 +219,15 @@ cuda_hardware_context::get_property(device_uint_property prop) const {
   case device_uint_property::max_global_size2:
     return _properties->maxThreadsPerBlock * _properties->maxGridSize[2];
```
al42and: It seems that `device_uint_property::max_global_sizeX` are not correct (should be `_properties->maxThreadsDim[X] * _properties->maxGridSize[X]`). However, as far as I can tell, they are not used now, so we can safely remove them.

include/hipSYCL/sycl/device.hpp
```diff
@@ -313,11 +313,11 @@ HIPSYCL_SPECIALIZE_GET_INFO(device, max_work_item_dimensions)
...
   std::size_t size0 = static_cast<std::size_t>(get_rt_device()->get_property(
-      rt::device_uint_property::max_global_size0));
+      rt::device_uint_property::max_group_size0));
   std::size_t size1 = static_cast<std::size_t>(get_rt_device()->get_property(
-      rt::device_uint_property::max_global_size1));
+      rt::device_uint_property::max_group_size1));
   std::size_t size2 = static_cast<std::size_t>(get_rt_device()->get_property(
-      rt::device_uint_property::max_global_size2));
+      rt::device_uint_property::max_group_size2));
   return id<3>{size0, size1, size2};
```
al42and, line 321: Pre-existing problem: given how we "flip" the launch dimensions for CUDA, HIP (https://github.com/illuhad/hipSYCL/blob/f3177b86c7731c27805dfff51c1efa2e7b96cb1f/include/hipSYCL/glue/generic/hiplike/hiplike_kernel_launcher.hpp#L366), and ZE (https://github.com/illuhad/hipSYCL/blob/f3177b86c7731c27805dfff51c1efa2e7b96cb1f/include/hipSYCL/glue/ze/ze_kernel_launcher.hpp#L381), should we flip the order here as well?

src/runtime/cuda/cuda_hardware_manager.cpp
illuhad: Good catch, we should fix them.

include/hipSYCL/sycl/device.hpp
illuhad: I'd say, no. Index flipping is a problem of the kernel headers and the glue code, not of the runtime library which is currently completely unaware that something like this happens. Also, even if we wanted to, we couldn't expose the right flipping here because the correct flip depends on the dimensionality of the query. I.e. `max_size[0]` would depend on whether we are asking for a 1D, 2D, or 3D kernel.

The reason why SYCL 2020 changed the device descriptors to structs instead of enums was precisely to carry the information about the dimensionality. Because I don't want a templated query interface for the runtime, this is another reason why it probably should be handled by the headers. We will have to expose though whether a backend has vectorization across first index as a runtime property so that the device queries in the headers can do the right thing.

src/runtime/cuda/cuda_hardware_manager.cpp
nmnobre: Done.

include/hipSYCL/sycl/device.hpp
nmnobre: I didn't change anything as per @illuhad's request but I should add that the intel runtime does also flip the dimensions on a request to `sycl::device::get_info<sycl::info::device::max_work_item_sizes>()`.

illuhad: @nmnobre To clarify: Eventually those queries should also include the flip in hipSYCL. I was just talking about *where* it would eventually need to go. This requires distinguishing between the SYCL headers, where `get_info<>()` lives and the actual hipSYCL runtime library `libhipSYCL-rt` and its backend plugins. I was just saying that in the hipSYCL architecture, that flip should probably be implemented inside `get_info<>`, not the runtime library itself.

For this we probably need to move to the SYCL 2020 notion where structs are used instead of enums for those queries (#774). In SYCL 2020, `max_work_item_sizes<Dim>` is templated with the dimensionality so that the flip can be included correctly.

If DPC++ also flips without knowing the dimensionality, I would consider it confusing and not part of SYCL 2020 since this query does not exist anymore without template parameter for the dimension.
https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_device_information_descriptors

illuhad (approved):

include/hipSYCL/sycl/device.hpp
al42and: @illuhad, yes, you're right; no way to fix it currently.

nmnobre: > If DPC++ also flips without knowing the dimensionality, I would consider it confusing and not part of SYCL 2020 since this query does not exist anymore without template parameter for the dimension. https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_device_information_descriptors

I've checked and, my bad, they have in fact transitioned to use structs for the device descriptors a couple months back and they do require the template parameter as per SYCL 2020. I was using an old version which still used the enums and no dimension specification. For clarity, the flip happened then and it still does now. :-)

PR #834, opened 2022-09-26
Add __HIPSYCL_FORCE_INLINE_ALL__ macro to force inlining

al42and:
If `__HIPSYCL_FORCE_INLINE_ALL__` is defined, the hipSYCL plugin will attach always_inline attribute to all functions called from kernels.

For some reason, adding the `flatten` attribute to kernels does not help.

For GROMACS on NVIDIA devices, this speeds up some kernels up to 9x:

Results on RTX3060, with and without setting this flag (`-DSYCL_CXX_FLAGS_EXTRA=-D__HIPSYCL_FORCE_INLINE_ALL__`). Native CUDA shown for reference.

  | With inlining | Default | Native CUDA
-- | -- | -- | --
NbnxmKernelPruneOnly<bool=0> | 42.143us | 97.431us | 40.149us
BondedKernel<bool=0, bool=0> | 58.143us | 285.26us | 57.636us
PmeSplineAndSpreadKernel<int=4, bool=1, bool=1, bool=1, bool=1, int=1, bool=0, ThreadsPerAtom=0, int=32> | 219.91us | 759.48us | 221.41us
PmeGatherKernel<int=4, bool=1, bool=1, int=1, bool=0, ThreadsPerAtom=0, int=32> | 88.017us | 793.58us | 88.826us

</body>

</html>

### DIFF
--- a/include/hipSYCL/compiler/Frontend.hpp
+++ b/include/hipSYCL/compiler/Frontend.hpp
@@ -325,6 +325,11 @@ public:
 
   void applyAttributes()
   {
+    const auto& PPMacros = Instance.getPreprocessor().getPreprocessorOpts().Macros;
+    const bool ForceInlineAll = std::any_of(PPMacros.begin(), PPMacros.end(),
+            [](const auto& v){ return v.first == "__HIPSYCL_FORCE_INLINE_ALL__";});
+    HIPSYCL_DEBUG_INFO << "AST processing: Adding always_inline attribute to all device functions\n";
+
     for(auto F : MarkedHostDeviceFunctions)
     {
       // Strictly speaking, setting these attributes is not even necessary!
@@ -361,6 +366,9 @@ public:
         {
           RD->addAttr(clang::CUDAHostAttr::CreateImplicit(Instance.getASTContext()));
           RD->addAttr(clang::CUDADeviceAttr::CreateImplicit(Instance.getASTContext()));
+        }
+        if (ForceInlineAll && !RD->hasAttr<clang::AlwaysInlineAttr>()) {
+          RD->addAttr(clang::AlwaysInlineAttr::CreateImplicit(Instance.getASTContext()));
         }
       }
 


aaronmondal:
@al42and Interesting! What would be potential drawbacks to always enabling this for GPUs? Are there any known kinds of kernels where forceinlining would a bad idea? If the potential codesize increase is only an issue in edgecases, would it be a good idea to set this behavior as default and conditionally disable it via a `__HIPSYCL_NO_FORCE_INLINE_ALL__` macro or similar.

I find it very strange that `flatten` does not work though. According to the clang docs:
> The flatten attribute causes calls within the attributed function to be inlined unless it is impossible to do so, for example if the body of the callee is unavailable or if the callee has the noinline attribute.

If `flatten` doesn't work, could this be a bug in Clang?

al42and:
> What would be potential drawbacks to always enabling this for GPUs? Are there any known kinds of kernels where forceinlining would a bad idea?

I'm wary of messing with the default compiler behavior (including inlining heuristics), so I'm making this feature opt-in. I believe it would be beneficial to have it enabled in most cases, but I'd rather not fix things that are not broken. If the hipSYCL team thinks that it's good to have enabled by default, sure.

We can also, like in CUDA, add the `always_inline` attribute to all SYCL library functions, and let the users annotate their kernels and functions as they wish.

> If flatten doesn't work, could this be a bug in Clang?

Perhaps, I am not that familiar with LLVM internals to quickly debug the issue. For the record, I was using something like the following in the `for(auto F : MarkedKernels)` loop:

```cpp
      if (!F->hasAttr<clang::FlattenAttr>()) {
          F->addAttr(clang::FlattenAttr::CreateImplicit(Instance.getASTContext()));
      }
```

illuhad:
> @al42and Interesting! What would be potential drawbacks to always enabling this for GPUs? Are there any known kinds of kernels where forceinlining would a bad idea? If the potential codesize increase is only an issue in edgecases, would it be a good idea to set this behavior as default and conditionally disable it via a __HIPSYCL_NO_FORCE_INLINE_ALL__ macro or similar.

First of all, this would most likely not have any effect on AMD GPUs because we compile (hipcc does the same) with `-mllvm -amdgpu-early-inline-all=true -mllvm -amdgpu-function-calls=false`. So on AMD, the amdgpu LLVM backend should already inline everything. This however does not have performance reasons, but is due to the fact that some AMD hardware/ROCm versions don't support the notion of functions, and so the only way to express them is to inline all function calls. I don't know to what extent this still is the case, but last I checked hipcc still did it by default.

I remember seeing at least one case in one of the ROCm github repositories where the AMD behavior of inlining everything in HIP code caused performance issues. Couldn't find the issue with a quick search just now unfortunately. So, inlining everything is not necessarily always beneficial.

Code size on GPUs can be a real problem for very complex kernels because there is typically a hard limit to the size of kernels.

I would also expect that the nvptx backend already inlines whenever it thinks that it makes sense. The fact that there seem to be cases where it fails to recognize such opportunities hint at some of the heuristics being incomplete, or maybe there's a bug. So, the root problem should probably be addressed at the LLVM level, not by forcing a workaround for a particular issue unconditionally on all kernels and all backends.

Lastly, GPU backends are not the only citizens in town. Applying clang attributes to the AST will also potentially affect e.g. CPU backends, where the optimal inlining behavior might be very different, and inlining everything might just thrash the cache.

So, long story short, I'm also leaning towards not enabling this by default. There are plenty of examples where hipSYCL CUDA delivers similar performance as native CUDA without inlining everything, and we don't know what the impact would be there. I don't know what exactly is going on in LLVM with the Gromacs kernels that causes such performance deltas, but this is definitely the exception, not the rule.

I'm also surprised that flatten does not seem to do anything.

@al42and Which LLVM version(s) did you test and observe this performance behavior with?

al42and:
> I would also expect that the nvptx backend already inlines whenever it thinks that it makes sense. The fact that there seem to be cases where it fails to recognize such opportunities hint at some of the heuristics being incomplete, or maybe there's a bug. So, the root problem should probably be addressed at the LLVM level, not by forcing a workaround for a particular issue unconditionally on all kernels and all backends.

Agree, there is some inefficiency in the LLVM heuristics. When compiling native CUDA with Clang (no SYCL involvement), we got ~4x speed-up by forcibly inlining some functions. For comparison, NVCC was inlining those functions automatically (and got good performance). See https://gitlab.com/gromacs/gromacs/-/merge_requests/3113. So, LLVM can certainly see some improvements.

Recent versions of Clang have [`-fgpu-inline-threshold`](https://reviews.llvm.org/D99233) flag, which might be helpful, although I have not evaluated it yet.

> I'm also surprised that flatten does not seem to do anything.

Perhaps I was doing something wrong with the attribute. It needs to be applied to the `__global__` function, which is generated by the hipSYCL plugin, so I could not easily test it by just manually adding the flag :(

> @al42and Which LLVM version(s) did you test and observe this performance behavior with?

Clang 15 from LLVM's apt repos.

illuhad:
> Recent versions of Clang have [-fgpu-inline-threshold](https://reviews.llvm.org/D99233) flag, which might be helpful, although I have not evaluated it yet.

Ah, that sounds interesting.

> Perhaps I was doing something wrong with the attribute. It needs to be applied to the __global__ function, which is generated by the hipSYCL plugin, so I could not easily test it by just manually adding the flag :(

It does not generate the function, it just changes a preexisting function into a global function by attaching the attribute :-)
These functions are located here (all that marked as `__sycl_kernel` - the plugin just replaces `__sycl_kernel` with `__global__`):
https://github.com/illuhad/hipSYCL/blob/c3156fa5f76a08ab05b2436f9c45a4715489f09d/include/hipSYCL/glue/generic/hiplike/hiplike_kernel_launcher.hpp#L205
You could try to add flatten there.

> Clang 15 from LLVM's apt repos.

Okay, thanks for the info.

al42and:
Based on our testing, `-fgpu-inline-threshold` works just fine, so I'm closing it.

PR #838, opened 2022-09-30
replace activemask with ballot function

### LINKED ISSUE #823: Incorrect use of `__activemask`
**Bug summary**

In some group functions for CUDA backend, hipSYCL uses `__activemask` intrinsic to detect currently active threads, e.g. here: https://github.com/illuhad/hipSYCL/blob/develop/include/hipSYCL/sycl/libkernel/cuda/group_functions.hpp#L99

However, per the section "Active Mask Query" in https://developer.nvidia.com/blog/using-cuda-warp-level-primitives/, that is not correct. Listing 5 there is pretty much how hipSYCL does things now. Even if all the threads in a sub-group call the group primitive, it is not guaranteed that `__activemask()` will return all-1s. 

**To Reproduce**

Relies on the de-synchronization within the warp, so NVIDIA Volta or newer is needed.

**Expected behavior**

SYCL specs 2020v5 does not explicitly say what happens if not all threads call `reduce_over_group`, but the wording seems to suggest that it must be called by all threads in a (sub)group:

> Returns: The result of combining all the values of x specified by each work-item in the group using the operator binary_op, where the values are combined according to the generalized sum defined in standard C++.

So, for various warp-level primitives it should be okay to use `const unsigned int activemask = 0xffffffffff;`

- EDIT: not really, we are allowed to launch < 32 work-items overall, so our sub-group will not be a full warp. Pain.

Note: in HIP some inline assembly is used which I'm not willing to dig into, but it seems that something similar is going on there.

**Describe your setup**

* hipSYCL develop, CUDA 11.6, Clang 12

DieGoldeneEnte:
This is done to prevent retrieving the wrong thread mask. This is also implemented in hip, but most likely not necessary.

I tested this on an AMD GPU, but didn't do any performance measurements.

fixes #823

### DIFF
--- a/include/hipSYCL/sycl/libkernel/cuda/group_functions.hpp
+++ b/include/hipSYCL/sycl/libkernel/cuda/group_functions.hpp
@@ -96,7 +96,7 @@ __device__ T __hipsycl_reduce_over_group(sub_group g, T x,
                                          BinaryOperation binary_op) {
   const size_t       lid        = g.get_local_linear_id();
   const size_t       lrange     = g.get_local_linear_range();
-  const unsigned int activemask = __activemask();
+  const unsigned int activemask = __ballot_sync(0xffffffff, 1);
 
   auto local_x = x;
 
@@ -114,7 +114,7 @@ __device__ T __hipsycl_exclusive_scan_over_group(sub_group g, V x, T init,
                                                  BinaryOperation binary_op) {
   const size_t       lid        = g.get_local_linear_id();
   const size_t       lrange     = g.get_local_linear_range();
-  const unsigned int activemask = __activemask();
+  const unsigned int activemask = __ballot_sync(0xffffffff, 1);
 
   auto local_x = x;
 
@@ -146,7 +146,7 @@ __device__ T __hipsycl_inclusive_scan_over_group(sub_group g, T x,
                                                  BinaryOperation binary_op) {
   const size_t       lid        = g.get_local_linear_id();
   const size_t       lrange     = g.get_local_linear_range();
-  const unsigned int activemask = __activemask();
+  const unsigned int activemask = __ballot_sync(0xffffffff, 1);
 
   auto local_x = x;
 
--- a/include/hipSYCL/sycl/libkernel/hip/group_functions.hpp
+++ b/include/hipSYCL/sycl/libkernel/hip/group_functions.hpp
@@ -93,8 +93,7 @@ template <typename T, typename BinaryOperation>
 __device__ T __hipsycl_reduce_over_group(sub_group g, T x,
                                          BinaryOperation binary_op) {
   auto     local_x = x;
-  uint64_t activemask;
-  asm("s_mov_b64 %0, exec" : "=r"(activemask));
+  uint64_t activemask = __ballot(1);
 
   auto lid = g.get_local_linear_id();
 
@@ -119,8 +118,7 @@ __device__ T __hipsycl_inclusive_scan_over_group(sub_group g, T x,
   auto         local_x = x;
   const size_t lid     = g.get_local_linear_id();
 
-  uint64_t activemask;
-  asm("s_mov_b64 %0, exec" : "=r"(activemask));
+  uint64_t activemask = __ballot(1);
 
   size_t lrange = g.get_local_linear_range();
 


DieGoldeneEnte:
@hipSYCL-ci-bot test please :)

fodinabor:
Tested on a mixed GPU setup, seems good.
Both HIP and CUDA group_reduce and group_inclusive_scan seem to be in the same performance ballpark as before (according to your sycl-bench fork :))

include/hipSYCL/sycl/libkernel/cuda/group_functions.hpp
```diff
@@ -96,7 +96,7 @@ __device__ T __hipsycl_reduce_over_group(sub_group g, T x,
                                          BinaryOperation binary_op) {
   const size_t       lid        = g.get_local_linear_id();
   const size_t       lrange     = g.get_local_linear_range();
-  const unsigned int activemask = __activemask();
+  const unsigned int activemask = __ballot_sync(0xffffffff, 1);
```
al42and, line 99: I think `const unsigned int activemask = 0xffffffff;` would have the same effect.

- If all threads in the warp are entering this function, the result is the same. `__ballot_sync` will synchronize them, but I don't think it's necessary since `__hipsycl_shuffle_impl` uses `__shfl_sync` with the full mask, so all the cross-lane operations are properly synchronized.
- If only some threads in the warp are entering this function, then calling `__ballot_sync(0xffffffff, 1)` is, to the best of my knowledge, undefined behavior (UB). It should be called before the divergence. But it is not a big deal, since we're using `__shfl_sync` with the full mask in `__hipsycl_shuffle_impl`, so it was UB already.

DieGoldeneEnte: I'm not sure if
```
__ballot_sync(__activemask(), 1):
```
would be valid, they use something similar in their [blog](https://developer.nvidia.com/blog/using-cuda-warp-level-primitives/) at "Opportunistic Warp-level Programming".

If that isn't defined I personally am more comfortable with the original logic, maybe add a sync in front (which also doesn't guarantee the mask to be correct).
Just using `0xffffffff` for the active mask is in my opinion worse, since we would use the result of the shuffle which is also undefined, but the old version never failed the tests for me at least.

Since the spec expects us to also allow incomplete groups/sub_groups the only real fix (if __activemask() inside a *_sync function doesn't work) is to rewrite the functions to work without using the active mask. I'll check that option as well.

al42and: > would be valid, they use something similar in their [blog](https://developer.nvidia.com/blog/using-cuda-warp-level-primitives/) at "Opportunistic Warp-level Programming".

I think they are using `__ballot_sync` **before** any divergence occurs, to get a mask of said divergence.

`__ballot_sync(mask, 1)` will either produce the mask itself (if the mask is correct) or, if the mask is too broad, will likely hang (although I think it's not officially defined). So, per my understanding `__ballot_sync(__activemask(), 1)` is the same as `__activemask()`.

> Since the spec expects us to also allow incomplete groups/sub_groups the only real fix (if __activemask() inside a *_sync function doesn't work) is to rewrite the functions to work without using the active mask. I'll check that option as well.

Perhaps [`coalesced_threads()`](https://developer.nvidia.com/blog/cooperative-groups/) can be the answer. But I have not read deeply about cooperative groups, so don't know if there are any pitfalls.

We can also check all the conditional happening before the current sub-group collective operation, and insert `__ballot_sync` before them to keep the mask updated. But that looks complicated and wasteful.

Long story short, having partial warps is very complicated. For starters, I suggest getting the full warps right.

>  Just using 0xffffffff for the active mask is in my opinion worse, since we would use the result of the shuffle which is also undefined, but the old version never failed the tests for me at least.

On CUDA, we're using the full mask for shuffle intrinsic:

https://github.com/illuhad/hipSYCL/blob/9e78270f5c5e85d3f030e565628edc33eee61f94/include/hipSYCL/sycl/libkernel/generic/hiplike/warp_shuffle.hpp#L90

https://github.com/illuhad/hipSYCL/blob/9e78270f5c5e85d3f030e565628edc33eee61f94/include/hipSYCL/sycl/libkernel/generic/hiplike/warp_shuffle.hpp#L146

- In the case of having only part of the warp, the old code will not work anyway (hanging at `__shfl_sync`). But it also potentially has a data race with a full warp.
- Using the full mask here will at least guarantee that the full warp case works fine, since we will be correctly applying the `binary_op`.
- Using `__ballot_sync` will also work, but it's more instructions with no benefit, as far as I can tell.

DieGoldeneEnte: I spend some more time reading up on how this could be solved.

I couldn't find any mention in the [programming guide](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#warp-shuffle-functions) or the [PTX ISA](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#data-movement-and-conversion-instructions-shfl-sync) that `0xFFFFFFFF` as mask leads to undefined behavior, as long as only data from active threads is used.

While __ballot_sync using `0xFFFFFFFF` as mask is undefined behavior if not all threads are active, `__match_any_sync` does not have this limitation (at least it isn't mentioned in the [programming guide](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#warp-match-functions) or [PTX ISA](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#parallel-synchronization-and-communication-instructions-match-sync)).

So replacing `__ballot_sync` with `__match_any_sync` should solve our issue.

I need to test my code, if everything works I'll update the PR.

al42and: > I couldn't find any mention in the [programming guide](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#warp-shuffle-functions) or the [PTX ISA](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#data-movement-and-conversion-instructions-shfl-sync) that 0xFFFFFFFF as mask leads to undefined behavior, as long as only data from active threads is used.

From https://developer.nvidia.com/blog/using-cuda-warp-level-primitives/:

> Don’t just use `FULL_MASK` (i.e. `0xffffffff` for 32 threads) as the mask value. If not all threads in the warp can reach the primitive according to the program logic, then using `FULL_MASK` may cause the program to hang.

Also:

> Don’t just use `__activemask()` as the mask value. `__activemask()` tells you what threads happen to be convergent when the function is called, which can be different from what you want to be in the collective operation.

> Do analyze the program logic and understand the membership requirements. Compute the mask ahead based on your program logic.

DieGoldeneEnte: Using FULL_MASK shouldn't be a problem for us, since the SYCL standard requires the group_functions to be run by all threads

> These group functions act as synchronization points and must be encountered in converged control flow by all work-items in the group. If one work-item in a group calls a group function, then all work-items in that group must call exactly the same function under the same set of conditions --- calling the same function under different conditions (e.g. in different iterations of a loop, or different branches of a conditional statement) results in undefined behavior. 


Regarding the `__activemask()`, I should have specified, I don't plan on using `__activemask()`, instead I'll use FULL_MASK.

al42and: > Using FULL_MASK shouldn't be a problem for us, since the SYCL standard requires the group_functions to be run by all threads

Are we guaranteed to have a full warp? E.g., what if we have only 16 threads total? 

From the description of `sycl::sub_group::get_local_range()` it seems that such things are allowed:

> Return a range representing the size of the sub-group. This size may be less than the value returned by `get_max_local_range()`, depending on the position of the sub-group within its parent work-group and the manner in which sub-groups are constructed by the implemen­tation.

DieGoldeneEnte: We are not guaranteed to have a full warp, that's why we need to use the mask. According to the documentation shuffle and match instructions only synchronize non-exited threads.

> shfl.sync will cause executing thread to wait until all non-exited threads corresponding to membermask have executed shfl.sync with the same qualifiers and same membermask value before resuming execution.

The same is true for the match instruction. When looking at the PTX ISA even `__ballot_sync` might work

> vote.sync will cause executing thread to wait until all non-exited threads corresponding to membermask have executed vote.sync with the same qualifiers and same membermask value before resuming execution.

I interpreted the description in the programming guide differently before, but it only seems to be a problem if different masks or instructions are used.

al42and: >  shfl.sync will cause executing thread to wait until all non-exited threads corresponding to membermask have executed shfl.sync with the same qualifiers and same membermask value before resuming execution.

Thanks, I missed the "non-exited" part. Then that should be fine; sorry for doubting.

> The same is true for the match instruction. When looking at the PTX ISA even __ballot_sync might work

Agree, `__ballot_sync` should be ok.

DieGoldeneEnte: It's so easy to miss or misread some condition for these low-level functions.

Happy you also had an eye on it.

illuhad (approved):
Thanks!

PR #845, opened 2022-10-02
Remove extraneous `kernel_configuration` include

philipturner:
There is no such file located at this header path. Removing these lines allowed the generic LLVM branch to compile.


illuhad:
Sorry, I forgot to add that file when I committed those changes :-) I have just pushed it, so this PR should no longer  be necessary.

Issue #767, opened 2022-07-11
Alternatives for non-standard sub_group::shuffle_up/down/xor and yet-to-be-implemented atomic_fence


nmnobre:
Hi @illuhad,

I've been trying to compile a project written in SYCL which also uses a few DPC++ extensions.

1) I'm aware [`sub_group::shuffle_up/down/xor`](https://intel.github.io/llvm-docs/doxygen/structcl_1_1sycl_1_1ext_1_1oneapi_1_1sub__group.html) are not part of the standard, but I was wondering if I could emulate their behaviour with standard features currently implemented in hipSYCL. Do you have any ideas you'd be happy to share?

2) Since `atomic_fence` hasn't yet been implemented, I've been thinking of biting the bullet and switching my code to use `mem_fence` instead. In your experience, is this the best course of action or are there better alternatives?

Thank you very much for your time,
-Nuno

illuhad:
1.) Yes, e.g. `shift_group_left/right`: https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#_shift_left_and_shift_right

2.) We could just implement it/you could submit a PR, it shouldn't be too difficult :-)  We currently only support relaxed memory order on GPUs anyway, so there would not be a difference to a `mem_fence` for the same memory_scope.

nmnobre:
1. Worked like a charm, thanks! 🙂
2. How do we expect the current `mem_fence` implementation to work when targeting nvc++?

illuhad:
It invokes one of the `__threadfence()` CUDA functions (depending on the fence scope argument to mem_fence). Here's the implementation detail: https://github.com/illuhad/hipSYCL/blob/c0808ade082d0e08ddfa603fd1cffb3db57910c7/include/hipSYCL/sycl/libkernel/detail/mem_fence.hpp

There's no difference between the nvc++ path and the clang CUDA path.

nmnobre:
I'd have thought so, but it's actually invoking `__hipsycl_if_target_host()` which would do nothing, if not for the weird link time undefined reference to `__builtin_is_device_code`, presumably from `if target (nv::target::is_host)`.

illuhad:
It should only execute `__hipsycl_if_target_host()` for the host code - the device binary should be compiled using the `__hipsycl_if_target_hiplike()` path, which is defined here:
[...]

PR #849, opened 2022-10-12
Update comment on nvc++ workaround in mem_fence()

nmnobre:
Hi @illuhad,

It seems Nvidia has indeed fixed the compiler errors with empty if target statements.
See #767 (happy to close if you are) and #798.

Cheers :)

### DIFF
--- a/include/hipSYCL/sycl/libkernel/detail/mem_fence.hpp
+++ b/include/hipSYCL/sycl/libkernel/detail/mem_fence.hpp
@@ -52,8 +52,8 @@ struct mem_fence_impl
                           __spv::MemorySemanticsMask::WorkgroupMemory);
     );
     // TODO What about CPU?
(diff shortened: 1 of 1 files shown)


nmnobre:
On second thought, I think this is overcomplicating things and adds nothing. Besides, it's the wrong place for a fix like this...
I would suggest either:

1) Adding a note to the comment saying it only applies to versions <= 22.7;
2) Editing `cuda_backend.hpp` to add `printf("");` or `int dummy; (void)dummy;` and moving the compiler bug comment there;
3) Amending the documentation to say we require version 22.9 or newer.

illuhad:
I think it would suffice to just add to the comment that this only affects these older nvc++ versions. We don't need to put in a lot of effort just to be able to express a no-op in an `if target` path :-)

illuhad (approved):
Thanks!

PR #859, opened 2022-11-04
Use -isystem instead of -I for hipSYCL headers to avoid warnings with high warning levels

illuhad:
@krasznaa Please check if this fixes the warning issue for you.

### DIFF
--- a/bin/syclcc-clang
+++ b/bin/syclcc-clang
@@ -1665,7 +1665,7 @@ class compiler:
   @property
   def common_cxx_flags(self):
     args = [
-      "-isystem" + self._hipsycl_include_path,
+      "-isystem", self._hipsycl_include_path,
       "-D__HIPSYCL__"
     ] + self._common_compiler_args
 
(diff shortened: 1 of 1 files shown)


bin/syclcc-clang
```diff
@@ -1665,7 +1665,7 @@ class compiler:
   @property
   def common_cxx_flags(self):
     args = [
-      "-I" + self._hipsycl_include_path,
+      "-isystem" + self._hipsycl_include_path,
```
krasznaa: I thought that one would have to leave a space between `-isystem` and the path. :confused: But this does seem to work as-is. :+1: At least in my own test the warnings are gone with this included.

krasznaa (approved):
Ahh, didn't realise that I could even approve this PR. :stuck_out_tongue: (And it's good that I can't... I thought I may have found a mistake in the repository's setup.)

bin/syclcc-clang
illuhad: hm.. good catch, I thought so too :P

krasznaa: Just to confirm that the updated/fixed formalism also works as desired in my test.

illuhad: Thanks! :)

PR #860, opened 2022-11-07
Add rt::settings::set method

al42and:
Using environment variables is not always convenient.

It might lead to strange things when an option is changed in the middle of the execution, but that's not a widely advertised user-facing feature, so I think it's not much more dangerous than allowing changing internal parameters using environment variables in the first place.

### DIFF
--- a/include/hipSYCL/runtime/settings.hpp
+++ b/include/hipSYCL/runtime/settings.hpp
@@ -197,6 +197,42 @@ public:
     return typename setting_trait<S>::type{};
   }
 
+  template <setting S> void set(typename setting_trait<S>::type value) {
+    if constexpr(S == setting::debug_level){
+      _debug_level = value;
+    } else if constexpr (S == setting::scheduler_type) {
+      _scheduler_type = value;
+    } else if constexpr (S == setting::visibility_mask) {
+      _visibility_mask = value;
+    } else if constexpr (S == setting::dag_req_optimization_depth) {
+      _dag_requirement_optimization_depth = value;
+    } else if constexpr (S == setting::mqe_lane_statistics_max_size) {
+      _mqe_lane_statistics_max_size = value;
+    } else if constexpr (S == setting::mqe_lane_statistics_decay_time_sec) {
+      _mqe_lane_statistics_decay_time_sec = value;
+    } else if constexpr (S == setting::default_selector_behavior) {
+      _default_selector_behavior = value;
+    } else if constexpr (S == setting::hcf_dump_directory) {
+      _hcf_dump_directory = value;
+    } else if constexpr (S == setting::persistent_runtime) {
+      _persistent_runtime = value;
+    } else if constexpr (S == setting::max_cached_nodes) {
+      _max_cached_nodes = value;
+    } else if constexpr(S == setting::sscp_failed_ir_dump_directory) {
+      _sscp_failed_ir_dump_directory = value;
+    } else if constexpr(S == setting::gc_trigger_batch_size) {
+      _gc_trigger_batch_size = value;
+    } else if constexpr(S == setting::ocl_no_shared_context) {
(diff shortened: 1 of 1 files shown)


illuhad:
Thanks, the main issue I see here is that some of these settings might be totally unaffected by changes during runtime, because they are only queried once during startup. I also don't think we can actually guarantee long term that things won't crash for some of these values when they are suddenly changed.
It might be very confusing when changing some settings has an effect, changing others does not, and changing others may even lead to crashes.
Even if changing one does work, it might be considered an implementation detail, and maybe change in other versions.

I wonder if it might not be better to add the possibility to change the default values by somehow passing them through `syclcc` flags.. or by making it read a config file in the application directory that the hipSYCL runtime library would read when starting up.

illuhad:
... or we could just say that changing these settings is only possible in conjunction with a reset of the runtime, i.e. that it will never affect the current instance of the runtime, but only the next launch of the runtime.

This could probably be easily implemented by just changing the default values, not the actual values with `set()`.

The hipSYCL runtime will start as soon as SYCL objects are constructed, and shutdown when there are no more SYCL objects around. If the application invokes `set_default() before any SYCL functionality is used, it should be farily well defined.

al42and:
>  Even if changing one does work, it might be considered an implementation detail, and maybe change in other versions.

Just like the environment variables, no?

> I wonder if it might not be better to add the possibility to change the default values by somehow passing them through syclcc flags.. or by making it read a config file in the application directory that the hipSYCL runtime library would read when starting up.

I think that's less than ideal. While we only plan to use static value at first, being able to choose in the runtime what to do can be nice. E.g., the flush lag is less of an issue in some cases, so we might want to set different `max_cached_nodes` values depending on the GROMACS's offloading scheme.

> ... or we could just say that changing these settings is only possible in conjunction with a reset of the runtime, i.e. that it will never affect the current instance of the runtime, but only the next launch of the runtime.

Yes, I think that should be fine. We anyway have a runtime shutdown after the device detection (not intentionally, but all SYCL objects go out of scope), so we can even make the choice of the values based on which device we're running on. 

I'll update the PR tomorrow.

Slightly offtopic regarding the initialization: is it possible to disable the creation of the extra device streams at the runtime initialization? They are supposed to be used for the DAG-based scheduler, but four of them are created on each device even if no out-of-order queues are created. Not a big deal, but it pollutes the profiler timeline (e.g., 4 streams, 4 devices, two runtime initialization, making 32 extra streams that are never used). Does not seem to have any negative effect on the performance, though, so did not want to open an issue because of it.

al42and:
> Thanks, the main issue I see here is that some of these settings might be totally unaffected by changes during runtime, because they are only queried once during startup. I also don't think we can actually guarantee long term that things won't crash for some of these values when they are suddenly changed.

It seems things are a bit more complicated. The main settings holder is a static variable in `hipsycl::rt::application::get_settings()` function. So, it's getting initialized not when the runtime is initialized but only at the program start (loading of `libhipSYCL-rt.so`). 

~~We can try re-initializing the `settings` when the `application` is being constructed, do you think that would be ok?~~ We can't, it's static object. So, when `runtime` is created?

illuhad:
Ah, damn. Right. I guess we could reinitialize it when the runtime starts, but we might have to think about thread safety then. The runtime is not the only user of `settings`, the SYCL headers might access it as well.

al42and:
Hi! 

@illuhad, do you have an idea of how best to proceed with this? Shall the creation of the `settings` object be moved to the runtime initialization? Or can we just allow changing it, with all the possible footguns that it entails?

For us, even with #1054, the performance with `HIPSYCL_RT_MAX_CACHED_NODES=100` is ~15% slower for small systems than when a lower value is used (`0` is not always the best choice). So, being able to change setting from the code would be really good.

PR #868, opened 2022-11-16
Remove comparing `my_id` since it does not exist in class item

nilsfriess:
The comparison operators of `item` compares the member variable `my_id` but this member does not exist in the `item` class. This change removes this comparison (and keeps the ones that are correct).


illuhad (approved):
Good catch! :-) Is there a context here, how did you find it?

nilsfriess:
The SYCL CTS tests the comparison operators, that's how I found it (it's actually not the only thing that makes the compilation of the `item` tests fail but the other errors don't seem to have such an easy fix)

PR #875, opened 2022-11-23
Rename `global_mem_cache_type::write_only` to `read_write`

### LINKED ISSUE #432: info :: global_mem_cache_type differs from sycl/1.2.1 spec.
**Bug summary**
The definition of the `enum class cl::sycl::info::global_mem_cache_type` has `write_only`, which is not in the sycl/1.2.1 spec, and is missing `read_write`, which is in the spec.

**To Reproduce**
Look at the source code and the spec? :-)

```c++
#include <CL/sycl.hpp>
int main()
{
  cl::sycl::device device;
  auto foo = device.get_info<cl::sycl::info::device::global_mem_cache_type>() == cl::sycl::info::global_mem_cache_type::read_write ;
  return 0;
}
```

**Expected behavior**
hipSYCL and sycl/1.2.1 match.
No compile time warning error: 'read_write' is not a member of 'hipsycl::sycl::info::global_mem_cache_type'

I find write_only in 3 header files, all related to global_mem_cache
```
./include/hipSYCL/runtime/hardware.hpp:45:  global_mem_cache_write_only,
./include/hipSYCL/sycl/device.hpp:456:          rt::device_support_aspect::global_mem_cache_write_only))
./include/hipSYCL/sycl/device.hpp:457:    return info::global_mem_cache_type::write_only;
./include/hipSYCL/sycl/info/device.hpp:165:enum class global_mem_cache_type : int { none, read_only, write_only };
```

**Additional context**
Looking at SYCL specification:  Version 1.2.1  Document Revision: 7  Revision Date: April 27, 2020
Still in my first week of learning / playing with SYCL, so please be kind if I'm clueless. :-)

nilsfriess:
According to Table 25 in section A.3 of the [SYCL 2020 Specification](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#appendix.device.descriptors), the enum `global_mem_cache_type` should contain the entries `none`, `read_only`, and `read_write`. The last one was called `write_only` in hipSYCL, this PR renames it (and also other "linked"  names to be consistent with naming).

Should fix #432

### DIFF
--- a/src/runtime/cuda/cuda_hardware_manager.cpp
+++ b/src/runtime/cuda/cuda_hardware_manager.cpp
@@ -165,7 +165,8 @@ bool cuda_hardware_context::has(device_support_aspect aspect) const {
     return false;
     break;
   case device_support_aspect::global_mem_cache_read_write:
-    return false;
+    // NVIDIA GPUs have read/write cache at least since Fermi architecture
+    return true;
     break;
   case device_support_aspect::images:
     return false;
--- a/src/runtime/cuda/cuda_hardware_manager.cpp
+++ b/src/runtime/cuda/cuda_hardware_manager.cpp
@@ -164,7 +164,7 @@ bool cuda_hardware_context::has(device_support_aspect aspect) const {
   case device_support_aspect::global_mem_cache_read_only:
     return false;
     break;
-  case device_support_aspect::global_mem_cache_write_only:
+  case device_support_aspect::global_mem_cache_read_write:
     return false;
     break;
   case device_support_aspect::images:
--- a/src/runtime/hip/hip_hardware_manager.cpp
+++ b/src/runtime/hip/hip_hardware_manager.cpp
@@ -172,7 +172,8 @@ bool hip_hardware_context::has(device_support_aspect aspect) const {
     return false;
     break;
   case device_support_aspect::global_mem_cache_read_write:
-    return false;
+    // AMD GPUs have read/write cache at least since GCN1 architecture
+    return true;
     break;
   case device_support_aspect::images:
     return false;
--- a/src/runtime/hip/hip_hardware_manager.cpp
+++ b/src/runtime/hip/hip_hardware_manager.cpp
@@ -171,7 +171,7 @@ bool hip_hardware_context::has(device_support_aspect aspect) const {
   case device_support_aspect::global_mem_cache_read_only:
     return false;
     break;
-  case device_support_aspect::global_mem_cache_write_only:
+  case device_support_aspect::global_mem_cache_read_write:
     return false;
     break;
   case device_support_aspect::images:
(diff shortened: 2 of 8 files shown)


src/runtime/cuda/cuda_hardware_manager.cpp
```diff
@@ -164,7 +164,7 @@ bool cuda_hardware_context::has(device_support_aspect aspect) const {
   case device_support_aspect::global_mem_cache_read_only:
     return false;
     break;
-  case device_support_aspect::global_mem_cache_write_only:
+  case device_support_aspect::global_mem_cache_read_write:
     return false;
```
illuhad: Now that the semantics has changed from write-only to read-write - I suspect we will have to return true for CUDA, since I think CUDA hardware has read-write cache. But better check to be sure ;)

src/runtime/hip/hip_hardware_manager.cpp
```diff
@@ -171,7 +171,7 @@ bool hip_hardware_context::has(device_support_aspect aspect) const {
   case device_support_aspect::global_mem_cache_read_only:
     return false;
     break;
-  case device_support_aspect::global_mem_cache_write_only:
+  case device_support_aspect::global_mem_cache_read_write:
```
illuhad, line 174: Same question here ;)

src/runtime/omp/omp_hardware_manager.cpp
```diff
@@ -83,7 +83,7 @@ bool omp_hardware_context::has(device_support_aspect aspect) const {
   case device_support_aspect::global_mem_cache_read_only:
     return false;
     break;
-  case device_support_aspect::global_mem_cache_write_only:
+  case device_support_aspect::global_mem_cache_read_write:
```
illuhad, line 86: CPUs definitely have read-write cache

src/runtime/ze/ze_hardware_manager.cpp
```diff
@@ -263,7 +263,7 @@ bool ze_hardware_context::has(device_support_aspect aspect) const {
   case device_support_aspect::global_mem_cache_read_only:
     return false;
     break;
-  case device_support_aspect::global_mem_cache_write_only:
+  case device_support_aspect::global_mem_cache_read_write:
```
illuhad, line 266: Same question here

src/runtime/ze/ze_hardware_manager.cpp
```diff
@@ -263,8 +263,8 @@ bool ze_hardware_context::has(device_support_aspect aspect) const {
   case device_support_aspect::global_mem_cache_read_only:
     return false;
     break;
-  case device_support_aspect::global_mem_cache_write_only:
-    return false;
+  case device_support_aspect::global_mem_cache_read_write:
+    return true;
```
nilsfriess, line 267: I couldn't actually find a reference stating that all devices that can be used with Level Zero have read/write cache but dpc++ returns true here, and I think we can trust that this is true then.

illuhad: Agreed :)

illuhad (approved):
From my side this looks good. Is this ready to go @nilsfriess ?

nilsfriess:
Yes it is :)

PR #876, opened 2022-11-23
Remove unnecessary/wrong consts

nilsfriess:
This PR removes a few `const`s, namely:
- in `include/hipSYCL/sycl/libkernel/item.hpp`, the `const` for the member `offset` is removed to make objects of type item copyable;
- in `include/hipSYCL/sycl/libkernel/nd_range.hpp`, the `const` for the members is removed for the same reason as above;
- in `include/hipSYCL/sycl/queue.hpp`, the `const` in the second argument `secondaryQueue` of the `submit` method is removed, since inside the method, a non-const function of `secondaryQueue` is called, therefore causing a compile error.


illuhad (approved):
Thanks!

PR #877, opened 2022-11-24
Set `local_size` in reduction to default value

### LINKED ISSUE #857: Hang on device reduction kernel with 1 work item per group.
**Bug summary**

A generic reduction kernel hangs on the device when I supply an nd_range with only 1 work item per group.
I only observe the hang when running on the device, and only if using a reduction.

Thread backtrace shows it's stuck in the buffer destructor:
```
#0  0x000015554ebb370c in pthread_cond_wait@@GLIBC_2.3.2 ()
   from /lib64/libpthread.so.0
#1  0x000015554fef9420 in __gthread_cond_wait (__mutex=<optimized out>, 
    __cond=<optimized out>)
    at /ptmp/peint/build-cray-gcc-20210813.230140.804215673/cray-gcc/BUILD/snos_objdir/x86_64-suse-linux/libstdc++-v3/include/x86_64-suse-linux/bits/gthr-default.h:865
#2  std::__condvar::wait (__m=..., this=<optimized out>)
    at /ptmp/peint/build-cray-gcc-20210813.230140.804215673/cray-gcc/BUILD/snos_objdir/x86_64-suse-linux/libstdc++-v3/include/bits/std_mutex.h:155
#3  std::condition_variable::wait (this=<optimized out>, __lock=...)
    at ../../../../../cpe-gcc-11.2.0-202108140355.9bf1fd589a5c1/libstdc++-v3/src/c++11/condition_variable.cc:41
#4  0x0000155555542b3b in hipsycl::rt::worker_thread::wait() ()
   from /ccs/proj/stf006/rogersdd/crusher/lib/libhipSYCL-rt.so
#5  0x000000000028113b in hipsycl::sycl::detail::buffer_impl::~buffer_impl() ()
#6  0x00000000002808e9 in void __gnu_cxx::new_allocator<hipsycl::sycl::detail::buffer_impl>::destroy<hipsycl::sycl::detail::buffer_impl>(hipsycl::sycl::detail::buffer_impl*) ()
#7  0x000000000028088d in void std::allocator_traits<std::allocator<hipsycl::sycl::detail::buffer_impl> >::destroy<hipsycl::sycl::detail::buffer_impl>(std::allocator<hipsycl::sycl::detail::buffer_impl>&, hipsycl::sycl::detail::buffer_impl*) ()
#8  0x00000000002805c6 in std::_Sp_counted_ptr_inplace<hipsycl::sycl::detail::buffer_impl, std::allocator<hipsycl::sycl::detail::buffer_impl>, (__gnu_cxx::_Lock_policy)2>::_M_dispose() ()
#9  0x00000000002669f3 in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release() ()
#10 0x00000000002669aa in std::__shared_count<(__gnu_cxx::_Lock_policy)2>::~__shared_count() ()
#11 0x0000000000287bb9 in std::__shared_ptr<hipsycl::sycl::detail::buffer_impl, 
...
```

**To Reproduce**
Compile and run the following code:
```
#include <stdio.h>
#include <stdlib.h>
#include <CL/sycl.hpp>

namespace sycl = cl::sycl;

int main(int argc, char *argv[]) {
    int N = 0;
    int groups = 4;
    int width = 1;
    if(argc > 1) {
        groups = atoi(argv[1]);
    }
    if(argc > 2) {
        width = atoi(argv[2]);
    }

    sycl::queue q;

    if(1) {
      sycl::buffer<int,1> ret(&N, 1);
        sycl::nd_range<1> rng(groups*width, width);
        sycl::accessor<int, 1> d_ret(ret, cgh);//, sycl::read_write);
        const size_t count = 1 << 6;

        cgh.parallel_for(rng,
                sycl::reduction(d_ret, sycl::plus<int>()),
                [count=count](sycl::nd_item<1> it, auto &ans) {
            sycl::group<1> g = it.get_group();
            printf("count %d id %d range %d\n",
                    count, g.get_group_linear_id(),
                    g.get_group_linear_range());
            ans += 1;
        });
      });
    }

    printf("Launched %d workers.\n", N);
    return 0;
}
```

Passing a value >1 for the width works, but with `width=1`, it deadlocks.

**Expected behavior**
The code should run successfully for all device-supported values of groups and width.

**Describe your setup**
* I installed commit 2d937998228826f183a670d927749a4e6cb93a11 from July 13 and am running on AMD gfx90a.

   * host compiler cce/14.0.0
   * boost 1.79.0
   * ROCm version 5.1.0
   * llvm includes /opt/rocm-5.1.0/llvm/lib/clang/14.0.0/include

* How you have compiled your application: `syclcc --hipsycl-gpu-arch=gfx90a -o err err.cpp`

**Additional context**

Error discovered while running device-accelerated tests on a hash table implementation <https://github.com/ORNL/SYCLhash>.

nilsfriess:
Passing a `local_size` of 1 in a `parallel_for` reduction previously caused an infinite loop. This PR sets the `local_size` to a default value of 128 if the user passed 1.

Fixes #857

### DIFF
--- a/include/hipSYCL/glue/generic/hiplike/hiplike_kernel_launcher.hpp
+++ b/include/hipSYCL/glue/generic/hiplike/hiplike_kernel_launcher.hpp
@@ -420,6 +420,11 @@ determine_reduction_stages(sycl::range<Dimensions> global_size,
   sycl::range<Dimensions> current_num_groups = num_groups;
   sycl::range<Dimensions> current_num_work_items = global_size;
 
+  for (int i=0; i < Dimensions; ++i)
+    // Check if local_size == 1, this would cause an infinite loop
+    if (local_size[i] == 1)
+      local_size[i] = 128;
+  
   while(current_num_groups.size() > 1) {
 
     current_num_work_items = current_num_groups;
--- a/include/hipSYCL/glue/generic/hiplike/hiplike_kernel_launcher.hpp
+++ b/include/hipSYCL/glue/generic/hiplike/hiplike_kernel_launcher.hpp
@@ -426,8 +426,8 @@ determine_reduction_stages(sycl::range<Dimensions> global_size,
   for (int i=1; i<Dimensions; ++i)
     local_size[i] = 1;
   // Check if local_size == 1, this would cause an infinite loop
-  if (local_size[0] == 1)
-    local_size[0] = 128;
+  if (local_size[Dimensions - 1] == 1)
+    local_size[Dimensions - 1] = 128;
 
   num_groups[Dimensions - 1] = num_groups.size();
   for (int i=1; i<Dimensions; ++i)
--- a/include/hipSYCL/glue/generic/hiplike/hiplike_kernel_launcher.hpp
+++ b/include/hipSYCL/glue/generic/hiplike/hiplike_kernel_launcher.hpp
@@ -417,15 +417,22 @@ determine_reduction_stages(sycl::range<Dimensions> global_size,
   stages.push_back(reduction_stage<Dimensions>{
       local_size, num_groups, global_size});
 
+  // Reduce ranges of the form (N,M,L) to (N*M*L,1,1)
+  global_size[0] = global_size.size();
+  for (int i=1; i<Dimensions; ++i)
+    global_size[i] = 1;
+
+  local_size[0] = local_size.size();
+  for (int i=1; i<Dimensions; ++i)
+    local_size[i] = 1;
+  // Check if local_size == 1, this would cause an infinite loop
+  if (local_size[0] == 1)
+    local_size[0] = 128;
+
   sycl::range<Dimensions> current_num_groups = num_groups;
   sycl::range<Dimensions> current_num_work_items = global_size;
-
-  for (int i=0; i < Dimensions; ++i)
-    // Check if local_size == 1, this would cause an infinite loop
-    if (local_size[i] == 1)
-      local_size[i] = 128;
   
-  while(current_num_groups.size() > 1) {
+  while(current_num_groups[0] > 1) {
 
     current_num_work_items = current_num_groups;
     current_num_groups =
@@ -693,7 +700,6 @@ public:
         k(handle);
 
       } else {
-
         sycl::range<Dim> grid_range =
             hiplike_dispatch::determine_grid_configuration(
                 global_range, effective_local_range);
--- a/include/hipSYCL/glue/generic/hiplike/hiplike_kernel_launcher.hpp
+++ b/include/hipSYCL/glue/generic/hiplike/hiplike_kernel_launcher.hpp
@@ -419,18 +419,18 @@ determine_reduction_stages(sycl::range<Dimensions> global_size,
 
   // Reduce ranges of the form (N,M,L) to (1,1,N*M*L)
   global_size[Dimensions - 1] = global_size.size();
-  for (int i=1; i<Dimensions; ++i)
+  for (int i=0; i<Dimensions-1; ++i)
     global_size[i] = 1;
 
   local_size[Dimensions - 1] = local_size.size();
-  for (int i=1; i<Dimensions; ++i)
+  for (int i=0; i<Dimensions-1; ++i)
     local_size[i] = 1;
   // Check if local_size == 1, this would cause an infinite loop
   if (local_size[Dimensions - 1] == 1)
     local_size[Dimensions - 1] = 128;
 
   num_groups[Dimensions - 1] = num_groups.size();
-  for (int i=1; i<Dimensions; ++i)
+  for (int i=0; i<Dimensions-1; ++i)
     num_groups[i] = 1;
 
   sycl::range<Dimensions> current_num_groups = num_groups;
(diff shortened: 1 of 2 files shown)


include/hipSYCL/glue/generic/hiplike/hiplike_kernel_launcher.hpp
```diff
@@ -420,6 +420,11 @@ determine_reduction_stages(sycl::range<Dimensions> global_size,
   sycl::range<Dimensions> current_num_groups = num_groups;
   sycl::range<Dimensions> current_num_work_items = global_size;
 
+  for (int i=0; i < Dimensions; ++i)
+    // Check if local_size == 1, this would cause an infinite loop
+    if (local_size[i] == 1)
+      local_size[i] = 128;
```
illuhad: So, if Dimensions is greater than 1, we end up with (128, 128, ...) as work group size. This probably will break on CUDA and HIP, since those devices typically support at most 1024 elements in a work group.

Wouldn't it be sufficient to execute the reduction kernel always as a one-dimensional thing?

illuhad:
Thinking about it some more, would it maybe make sense to add the reduction with group size of 1 to our test cases?

include/hipSYCL/glue/generic/hiplike/hiplike_kernel_launcher.hpp
nilsfriess: Yes that makes sense, I didn't really think about what happens if `Dimensions > 1`. But I don't quite understand what you mean by executing the kernel as a 1D thing

tests/sycl/reduction.cpp
```diff
@@ -327,6 +328,12 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(two_reductions, T, large_test_types) {
   test_two_reductions<T>(128*128, 128);
 }
 
+#ifdef BOOST_SIGACTION_BASED_SIGNAL_HANDLING
+BOOST_AUTO_TEST_CASE(local_size_one_reduction, * boost::unit_test::timeout(2)) {
```
nilsfriess: I have added a test case that has a timeout of 2 seconds to detect if the reduction is stuck in an infinite loop.

The `#ifdef` checks if boost is actually able to force stop the test if it times out, otherwise the test would be reported as failed but the test suite would still continue execution (in the infinite loop). I don't know however when `BOOST_SIGACTION_BASED_SIGNAL_HANDLING` is actually defined, on my setup on jarvis, for instance, it's not. So I'm not sure if this test makes sense this way.

include/hipSYCL/glue/generic/hiplike/hiplike_kernel_launcher.hpp
illuhad: I meant just executing the kernel with a 1 dimensional parallel iteration space, i.e. using as a group size something like `(size, 1, 1)` if `Dimensions == 3`.
(Please check whether it needs to be `(size, 1, 1)` or `(1, 1, size)` - SYCL works in terms of `(1, 1, size)`, but CUDA/HIP in terms of (`size, 1, 1)`, so we flip the indices somewhere in the kernel launcher).

nilsfriess: I'm not sure if I understand. 

Say we have a reduction over a `nd_range<2>({N, M}, {1, 16})`.  Then in the first dimension we still have the problem that computing the reduction stage sizes ends in an infinite loop (because the number of groups in the first step will be computed as `num_groups[0] = N / local_size[0]`, then this will again be divided by `local_size[0]` etc. until `num_groups[0] == 1` but `local_size[0] == 1` so always `num_groups[0] == N`).

illuhad: That's true, but I think this problem only affects when we launch dedicated reduction kernels after the user-provided kernels here:
https://github.com/illuhad/hipSYCL/blob/56df01807eeb81cde76c4e0d6266d78e365ccf24/include/hipSYCL/glue/generic/hiplike/hiplike_kernel_launcher.hpp#L852

I think these could be run as 1D, since all they have to do is collect values from previous kernels and reduce them further.
In fact, looking at the code, they seem to be invoked as 1D already - so can we maybe just somehow remove the `Dimensions > 1` code paths in `determine_reduction_stages()`?.
I think all that code actually needs to care about is the total group size (`x * y * z`) so that it knows how many items will be reduced to one.

nilsfriess: Okay, I see, makes sense :)

I've changed it so that ranges of the form `(N,M,L)` are transformed to `(N*M*L,1,1)` (or `(N,M)` to `(N*M,1)` in case of 2D) before computing the reduction stages.

nilsfriess:
I've now also added tests for two- and three-dimensional reductions (both "general" reductions as well as reductions with `local_size = 1`. I also removed the the timeout test described above again, I guess it doesn't really make sense to have such a test if it isn't even supported on the system we're mostly using to test things.

PR #882, opened 2022-11-29
WIP: Fix MacOS build CI

normallytangent:


.github/workflows/macos.yml
```diff
@@ -16,17 +16,32 @@ jobs:
...
         brew install libomp
+        brew list libomp
         brew install boost
         set -e
     - name: build
       run: |
         mkdir build && cd build
-        cmake .. -DCMAKE_INSTALL_PREFIX=/Users/runner/work/hipSYCL/hipSYCL/build/install
+        cmake .. \
+        -DOpenMP_libomp_LIBRARY=/usr/local/Cellar/libomp/15.0.5/lib/libomp.a \
```
fodinabor: Obviously, just try to get it going as a first step, but in the longer run, you probably will want to detect this path automatically, as it will change with every LLVM update.
One example that might work: 
```
OMP_ROOT=`brew list libomp | grep libomp.a | sed -E "s/^(.*)/lib/libomp.a$/\1/"`
```
and then use `$OMP_ROOT/lib/libomp.a` for the CMake invocation.

illuhad:
Now that even the compiler-accelerated CPU backend has been reported working in https://github.com/illuhad/hipSYCL/pull/883, maybe we could test that in CI on Mac too?

.github/workflows/macos.yml
normallytangent: You're right, thanks, I'll implement this. I actually kept adding the hardcoded cmake paths for the specific errors I saw in each new iteration and it snowballed into this long list

.github/workflows/macos.yml
```diff
@@ -16,17 +16,33 @@ jobs:
...
+        -DOpenMP_C_LIB_NAMES=libomp \
+        -DOpenMP_C_FLAGS="-Xclang -fopenmp" \
+        -DOpenMP_C_INCLUDE_DIR=/usr/local/Cellar/libomp/15.0.5/include \
+        -DOpenMP_CXX_LIB_NAMES=libomp \
+        -DOpenMP_CXX_FLAGS="-Xclang -fopenmp" \
+        -DOpenMP_CXX_INCLUDE_DIR=/usr/local/Cellar/libomp/15.0.5/include \
+        -DWITH_ACCELERATED_CPU=ON \
```
illuhad: If we enable accelerated CPU, wouldn't we also need to install LLVM first? Or does it also work with Apple clang?

illuhad: (And are there even Mac packages for LLVM? Compiling it in CI could  take forever..)

fodinabor: Guess it *should* work, but will be a similar situation as with ROCm... versioning is quite different for Apple clang...

illuhad: Meaning it probably won't work then :P I guess then the question is whether we want to support Apple clang with accelerated cpu to the extent that we test it in CI..

fodinabor: yeah, that's your decision, I guess.. 😇 
at least brew should cover your upstream LLVM needs :) `brew install llvm`

normallytangent: It doesn't work, with Apple clang at least. I had already tested this yesterday, but again now after your edit, with accelerated cpu 'on' I get the [same error](https://github.com/illuhad/hipSYCL/actions/runs/3594049471/jobs/6051809253). Should I  then install llvm additionally? Or do I remove accelerated cpu?

illuhad: Ok - I think then it's probably best to focus on LLVM for accelerated CPU for now, and Apple clang only for the library-only compilation flow.

illuhad: @normallytangent Let's focus first on getting the Apple clang with library-only flow to work. Once this is done we can consider adding another CI flow for accelerated CPU using LLVM :)

normallytangent: @illuhad I tried to build the tests with omp.library-only version in my [second last commit](https://github.com/illuhad/hipSYCL/pull/882/commits/2699bf3244db805207924bfcae1bb47c9b37632b), it did not work, I keep getting the error that the file omp.h is not found. I'm not sure how to debug this, I use the same flags for building hipSYCL. If I understood it correctly, OpenMP is a hard requirement. Perhaps so is LLVM?

illuhad: No, LLVM is *not* a requirement.
The cmake infrastructure for the tests does not use the cmake  `FindOpenMP` module. So the flags you have added to the cmake invocation for the tests will not be taken into account. You will have to work `-I` and `-L` flags directly (try `CMAKE_CXX_FLAGS`).
The tests assume that hipSYCL has been built and configured fully, so that compiling with `syclcc` "just works". Because of this, they don't pull in any additional cmake infrastructure to check for hipSYCL dependencies.

So, a better solution than adding flags to the tests cmake  would be to adapt `OMP_CXX_FLAGS` when building hipSYCL to take into account OpenMP paths. This would then go into the `syclcc` configuration, and be applied automatically. But let's first try to get the basics working :)

fodinabor: @normallytangent note the difference: `OMP_CXX_FLAGS` vs `OpenMP_CXX_FLAGS` where the first one is consumed when configuring the hipSYCL build (not the tests), setting the default for `syclcc` which is then automatically used for building the tests.
`OpenMP_CXX_FLAGS` are unused by hipSYCL.

normallytangent: Thank you so much! I was so confused why the flags used during building hipsycl were not being used by when building the tests. If I understand correctly, I need to give both the OMP_CXX_FLAGS and OpenMP_CXX_FLAGS. One for cmake to find the libomp package, and the latter for configuring hipSYCL

fodinabor: Looking at https://github.com/illuhad/hipSYCL/blob/develop/src/runtime/CMakeLists.txt#L181, I think hipSYCL indeed uses the `FindOpenMP` module? So yeah, those `OpenMP_` variables will likely have an effect on whether or not that `find_package` is happy. Might be that setting `OpenMP_CXX_INCLUDE_DIR` would be what you'd want to do then? (cf. https://cmake.org/cmake/help/latest/module/FindOpenMP.html, https://github.com/Kitware/CMake/blob/master/Modules/FindOpenMP.cmake)

On a slightly related note (seeing your last commit): I would guess it might make sense to add `OMP_CXX_FLAGS_EXTRA` (and similarly for `*_LINK_*` and the other backends)  options to hipSYCL's main CMakeLists.txt that are just appended to their respective `OMP_CXX_FLAGS`? (so that one doesn't have to override the full commandline, e.g. messing up the boost include paths)

normallytangent: @fodinabor I had originally added `OpenMP_CXX_INCLUDE_DIR`. Later I removed it, when I started passing the path via `-I` flag to `OpenMP_CXX_FLAGS`. This [commit](https://github.com/illuhad/hipSYCL/actions/runs/3637769969/workflow) had built successfully. I've since been unable to reproduce it when I reduce the number of CMAKE options. Another solution probably would be to export `LDFLAGS` and `CPPFLAGS` for libomp but I didn't really try it on the CI. 

Thanks for the idea about how to append those flags, I was actually finding a way to do just this, but from the user's point of view.

fodinabor: Mhh.. how about providing `OMP_ROOT` as such `-DOpenMP_ROOT=$OMP_ROOT` to the hipSYCL build? (alternatively `-DCMAKE_PREFIX_PATH=$OMP_ROOT` might also work...?)
Maybe that still needs `-DCMAKE_CXX_FLAGS=-I$OMP_ROOT/include` for the tests though..?
Is that maybe sufficient?

Setting `OMP_CXX_FLAGS` as currently is, seems brittle (with the boost issue..), but with the `_EXTRA` thing you could do `-DOMP_CXX_FLAGS_EXTRA=-I$OMP_ROOT/include`.

.github/workflows/macos.yml
```diff
@@ -20,13 +20,26 @@ jobs:
...
+        OMP_ROOT=`brew list libomp | grep libomp.a | sed -E "s/\/lib\/.*//"`
         mkdir build && cd build
-        cmake .. -DCMAKE_INSTALL_PREFIX=/Users/runner/work/hipSYCL/hipSYCL/build/install
+        cmake .. \
+        -DOpenMP_libomp_LIBRARY=$OMP_ROOT/lib/libomp.a \
+        -DOpenMP_C_LIB_NAMES=libomp \
+        -DOpenMP_CXX_LIB_NAMES=libomp \
+        -DOpenMP_C_FLAGS="-Xclang -fopenmp -I$OMP_ROOT/include" \
+        -DOpenMP_CXX_FLAGS="-Xclang -fopenmp -I$OMP_ROOT/include" \
+        -DCMAKE_INSTALL_PREFIX=/Users/runner/work/hipSYCL/hipSYCL/build/install
```
fodinabor:
```suggestion
        cmake .. \
        -DOpenMP_ROOT=$OMP_ROOT \
        -DCMAKE_INSTALL_PREFIX=/Users/runner/work/hipSYCL/hipSYCL/build/install
```

.github/workflows/macos.yml
```diff
@@ -20,13 +20,26 @@ jobs:
...
       run: |
+        OMP_ROOT=`brew list libomp | grep libomp.a | sed -E "s/\/lib\/.*//"`
         mkdir build/tests-cpu && cd build/tests-cpu
-        cmake -DHIPSYCL_TARGETS=omp -DhipSYCL_DIR=/Users/runner/work/hipSYCL/hipSYCL/build/install/lib/cmake/hipSYCL /Users/runner/work/hipSYCL/hipSYCL/tests
+        cmake \
+        -DHIPSYCL_TARGETS=omp.library-only \
+        -DCMAKE_C_FLAGS="-Xclang -fopenmp -I$OMP_ROOT/include -L$OMP_ROOT/lib" \
+        -DCMAKE_CXX_FLAGS="-Xclang -fopenmp -I$OMP_ROOT/include -L$OMP_ROOT/lib" \
+        -DhipSYCL_DIR=/Users/runner/work/hipSYCL/hipSYCL/build/install/lib/cmake/hipSYCL \
+        /Users/runner/work/hipSYCL/hipSYCL/tests
```
fodinabor:
```suggestion
        cmake \
        -DHIPSYCL_TARGETS=omp.library-only \
        -DCMAKE_CXX_FLAGS=-I$OMP_ROOT/include \
        -DhipSYCL_DIR=/Users/runner/work/hipSYCL/hipSYCL/build/install/lib/cmake/hipSYCL \
        /Users/runner/work/hipSYCL/hipSYCL/tests
```

fodinabor (commented):
I believe this can be even shorter?

fodinabor (approved):
LGTM :)

illuhad (approved):
@normallytangent Is this good to go from your side?

normallytangent:
> @normallytangent Is this good to go from your side?

Yes, it is ready to merge. Should I merge it? If yes, then should I 'create a merge commit'?

illuhad:
I merge :)

PR #883, opened 2022-11-30
Fix hipSYCL clang plugin path on MacOS

illuhad:
@normallytangent Does this help with accelerated CPU on Apple?

### DIFF
--- a/bin/syclcc-clang
+++ b/bin/syclcc-clang
@@ -696,7 +696,7 @@ class syclcc_config:
     if sys.platform.startswith('win32'):
       return os.path.join(self.hipsycl_installation_path, "bin", "hipSYCL_clang.dll")
     elif sys.platform == "darwin":
-      return os.path.join(self.hipsycl_installation_path, "lib", "hipSYCL_clang.dylib")
+      return os.path.join(self.hipsycl_installation_path, "lib", "libhipSYCL_clang.dylib")
     else:
       return os.path.join(self.hipsycl_installation_path, "lib", "libhipSYCL_clang.so")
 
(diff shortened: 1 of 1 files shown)


normallytangent:
Confusingly, it does not. I tested it with accelerated cpu set to On. I get the same error I showed earlier but now if complains it can't find hipSYCL_clang.dylib file even though it is  already there at the same path the error message is about.

illuhad:
That's weird... Can you please send me the error, or post it somewhere?

bin/syclcc-clang
```diff
@@ -695,6 +695,8 @@ class syclcc_config:
   def hipsycl_plugin_path(self):
     if sys.platform.startswith('win32'):
       return os.path.join(self.hipsycl_installation_path, "bin", "hipSYCL_clang.dll")
+    elif sys.platform == "darwin":
+      return os.path.join(self.hipsycl_installation_path, "lib", "hipSYCL_clang.dylib")
```
normallytangent: The file name is missing the string "lib". Adding this in line 699 fixes the issue on mac, i.e. "**lib**hipSYCL_clang.dylib".

illuhad: Ugh... I feel stupid :( Thanks!

Does that mean that accelerated CPU on mac is really working now?

fodinabor:
Uhh, nice! :)

One thing that's slightly related, which I learned in the meantime, and I just wanted to note: the plugin-style dynamic libraries like `libhipSYCL_clang.so` or `libhipSYCL_omp.so` usually are (should be) declared with `add_library(<name> MODULE <src>)`.
See https://cmake.org/cmake/help/latest/command/add_library.html and maybe https://stackoverflow.com/questions/4845984/difference-between-modules-and-shared-libraries
Doesn't make much of a functional difference. On MacOS `MODULE` libraries will be `.so` files instead of `.dylib`, though.

bin/syclcc-clang
normallytangent: Yes, it's working now!

normallytangent (approved):
I tested it once more, all the tests pass with accelerated cpu.

PR #884, opened 2022-12-01
[SYCL 2020] Initial marray implementation

illuhad:
Totally untested, may or may not work, still needs test cases, and no support for builtins yet.

@nilsfriess Maybe this can help you?


illuhad:
I think this can go in now. marray support for math builtins will have to be a subsequent PR.

PR #888, opened 2022-12-07
Ensure that `device{} == device{default_selector{}}`

nilsfriess:
The SYCL 2020 specification requires that a default constructed device is the same as a device constructed with the default selector. This PR implements this by forward declaring the `sycl::device` default constructor and defining it in `device_selector.hpp` where `default_selector_v` is available. The default constructor then just delegates to the already defined constructor that takes any device selector as its argument.


illuhad (approved):
LGTM!

PR #901, opened 2023-01-07
[WIP] Add runtime components for Metal backend

philipturner:
This is a work-in-progress draft of the Metal backend. It currently adds the `metal_hardware_context` class and half-implements the Metal allocator. Many components, such as the allocator and blit encoder, have working prototypes written in Swift. They're located at [metal-usm](https://github.com/philipturner/metal-usm). I just need to translate them to C++.


CMakeLists.txt
```diff
@@ -115,7 +115,7 @@ else()
   set(WITH_ACCELERATED_CPU false)
 endif()
 
-if(WITH_SSCP_COMPILER) # sanitize input
+if(WITH_SSCP_COMPILER OR WITH_METAL_BACKEND) # sanitize input
```
illuhad, line 118: Do we really need to special case metal here? `WITH_SSCP_COMPILER` is automatically enabled if LLVM is sufficiently new and it's not explicitly disabled. `WITH_XXX_BACKEND` refers to runtime backends, not necessarily compiler infrastructure.

The other backends work as follows:
* `WITH_SSCP_COMPILER` is activated by default for sufficiently new LLVM, or disabled explicitly by user.
* `WITH_XXX_BACKEND` affects whether the runtime backend is built
* `llvm-to-backend`-backend-specific infrastructure for the SSCP compiler is built if `WITH_XXX_BACKEND` is enabled.

bin/syclcc-clang
```diff
@@ -697,6 +697,8 @@ class syclcc_config:
   def hipsycl_plugin_path(self):
     if sys.platform.startswith('win32'):
       return os.path.join(self.hipsycl_installation_path, "bin", "hipSYCL_clang.dll")
+    elif sys.platform == "darwin":
+      return os.path.join(self.hipsycl_installation_path, "lib", "libhipSYCL_clang.dylib")
```
illuhad, line 701: We already have this fix in current develop. I can merge that into `generic-llvm-sscp` branch if you like.

src/runtime/CMakeLists.txt
```diff
@@ -185,6 +185,40 @@ if(WITH_LEVEL_ZERO_BACKEND)
...
+  
+  target_include_directories(rt-backend-metal PRIVATE ${HIPSYCL_SOURCE_DIR}/include)
+  target_link_libraries(rt-backend-metal PRIVATE hipSYCL-rt)
+  
+  target_compile_options(rt-backend-metal PRIVATE ${HIPSYCL_RT_EXTRA_CXX_FLAGS})
+  target_link_libraries(rt-backend-metal PRIVATE ${HIPSYCL_RT_EXTRA_LINKER_FLAGS})
+  
+  if(WITH_SSCP_COMPILER)
+    target_compile_definitions(rt-backend-metal PRIVATE -DHIPSYCL_WITH_SSCP_COMPILER)
+    target_link_libraries(rt-backend-metal PRIVATE llvm-to-spirv)
```
illuhad, line 214: Wouldn't you want to link against a future `llvm-to-msl` here (which might internally reuse some SPIR-V infrastructure?)

src/runtime/cuda/cuda_hardware_manager.cpp
```diff
@@ -325,7 +325,7 @@ cuda_hardware_context::get_property(device_uint_property prop) const {
     return 8; // TODO
     break;
   case device_uint_property::global_mem_cache_line_size:
-    return 128; //TODO
+    return 128; // TODO
```
illuhad, line 328: Maybe remove this change from this PR, or open a separate PR for it?

src/runtime/hip/hip_hardware_manager.cpp
```diff
@@ -333,7 +333,7 @@ hip_hardware_context::get_property(device_uint_property prop) const {
     return 8; // TODO
     break;
   case device_uint_property::global_mem_cache_line_size:
-    return 128; //TODO
+    return 128; // TODO
```
illuhad, line 336: Same here?

src/runtime/metal/metal_allocator.cpp
```diff
@@ -0,0 +1,243 @@
...
+  uint64_t to_count = (0 > (input - 1)) ? 0 : (input - 1);
+  return 1 << (64 - std::__countl_zero(to_count));
+}
+
+int64_t round_down_to_power_of_2(int64_t input) {
+  uint64_t to_count = input;
+  return 1 << (64 - 1 - std::__countl_zero(to_count));
+}
+
+}
```
illuhad, line 242: I would like to see some more explanations in this file eventually, but I understand that this is work-in-progress :)

src/runtime/omp/omp_hardware_manager.cpp
```diff
@@ -247,7 +247,7 @@ omp_hardware_context::get_property(device_uint_property prop) const {
     return 8; // TODO
     break;
   case device_uint_property::global_mem_cache_line_size:
-    return 64; //TODO
+    return 64; // TODO
```
illuhad, line 250: Other PR?

src/runtime/omp/omp_queue.cpp
```diff
@@ -218,9 +218,9 @@ result omp_queue::submit_memcpy(memcpy_operation &op, dag_node_ptr node) {
     std::size_t total_num_bytes = op.get_num_transferred_bytes();
 
     bool is_src_contiguous =
-        is_contigous(src_offset, transferred_range, src_allocation_shape);
+        is_contiguous(src_offset, transferred_range, src_allocation_shape);
     bool is_dest_contiguous =
-        is_contigous(dest_offset, transferred_range, dest_allocation_shape);
+        is_contiguous(dest_offset, transferred_range, dest_allocation_shape);
```
illuhad, line 223: Can we have a separate PR for this typo?

illuhad (commented):
That looks like good progress! I cannot really comment too much on the logic for the Apple stuff, since I know way too little about it :(

CMakeLists.txt
philipturner: I want to signify that the Metal backend doesn't work without SSCP. For example, CMake should handle the following combos as:
- `WITH_SSCP_COMPILER` and `WITH_METAL_BACKEND` - builds GPU backend on ARM, error on x86
- `WITH_SSCP_COMPILER` - builds CPU-only everywhere
- `WITH_METAL_BACKEND` - error because Metal requires SSCP

For the first, I'd like to set `WITH_METAL_BACKEND` to true behind-the-scenes. It would only be true on ARM, but it would force Metal to be accessible on those platforms. I fear that developers will explicitly enable ze/cuda/rocm in their developer workflows, but forget Metal because macOS isn't their priority (they don't even know Metal exists). Therefore Metal will have 2nd-class support. Also, the goal of SSCP is reducing build complexity and the need to think about separate backends.

bin/syclcc-clang
philipturner: Yes, please do.

src/runtime/CMakeLists.txt
philipturner: I'm not yet sure how `llvm-to-msl` would work. Hopefully we can share code with `llvm-to-spirv`, then add additional code that structurizes it before `spirv-to-msl`. 

Perhaps the `llvm-to-spirv` code can accept check macros for Level Zero or Metal backends. We might want to enable Apple-specific intrinsics like `simdgroup_matrix_multiply_accumulate` which aren't part of `spirv_ocl`. Conversely, some Level Zero intrinsics might be inappropriate for Apple. We would disable those with Metal but not for ze.

src/runtime/cuda/cuda_hardware_manager.cpp
philipturner: All of the minor nits could go into a PR. Also, could you make a GitHub action that ensures each source file ends with a newline? Xcode automatically adds one to any file you view.

src/runtime/metal/metal_allocator.cpp
philipturner: I was in the middle of translating my [prototype](https://github.com/philipturner/metal-usm/blob/2f98211939ec7110bf50a755764a5f33f9aa637f/MetalAllocator/MainFile.swift#L302-L979) to C++, and lost interest halfway. This is the most iffy part of the entire backend. @woachk might be better able to review this.

I was having trouble with `std::__countl_zero` though. There's no easy way like `clz` in Metal or `.leadingZeroBitCount` in Swift. What form do you prefer for hipSYCL?

philipturner:
Another concern was how to configure metal-cpp. I'd like CMake to download it from the internet, maybe run `singleheader.py`, and include the result in CMake include paths. It's not good to hard-code the header into the hipSYCL source tree.

Also, metal-cpp requires exactly one file to include `{MTL,NS,CA}_PRIVATE_IMPLEMENTATION` before importing headers. No less, no more. Any way to auto-generate a special file for this in CMakeLists, or a good place in the source-tree to place it? `glue`?

CMakeLists.txt
illuhad: So, metal is not unique with regards to only being able to run kernels from the SSCP compiler. The same is kind of true for Intel GPUs and the Level Zero backend (technically it can also still be targeted using the clang SYCL toolchain, but that functionality is very limited, is already behind the SSCP compiler in terms of support and will be replaced eventually).

I think there are use cases where runtime backend capabilities can be needed without the ability to compile user kernels. For example, a user might want to run SYCL kernels on the CPU exclusively, but might want to interop with some other Metal code, and would like to synchronize those operations with SYCL operations. This can be achieved using backend interoperability in SYCL, if the runtime has a Metal backend.

I think your use cases can be addressed using sensible defaults and auto-detection of system capabilities, can't they?

I'm not sure I understand your concerns correctly, but if you are concerned that many people might not be using/enabling the SSCP compiler, I can assure you that it is a crucial component and focus for the future of the project. Once it is mature enough, the idea is to position it as the default compilation flow and let it take center stage. Because of its unified code representation, it wouldn't matter if people ultimately run on metal or some other backends as long as they use the SSCP compiler.

bin/syclcc-clang
illuhad: Done :)

src/runtime/CMakeLists.txt
illuhad: `llvm-to-spirv` generates generic OpenCL SPIR-V, so there's nothing Level Zero-specific to it. You could either use `llvm-to-spirv` internally for `llvm-to-msl`, or just share some code. The builtins are implemented in the libkernel bitcode libraries. Those could have an implementation for Apple (or new ones added).

illuhad:
> Another concern was how to configure metal-cpp. I'd like CMake to download it from the internet, maybe run singleheader.py, and include the result in CMake include paths. It's not good to hard-code the header into the hipSYCL source tree.

Sure, we also do something similar for the llvm-spirv translator, which gets automatically downloaded and built by cmake

> Also, metal-cpp requires exactly one file to include {MTL,NS,CA}_PRIVATE_IMPLEMENTATION before importing headers. No less, no more. Any way to auto-generate a special file for this in CMakeLists, or a good place in the source-tree to place it? glue?

glue is for code that connects two or more of the hipSYCL components (runtime, compiler, headers). Where would metal-cpp be needed? If it's just needed for the runtime, you could also put it inside the metal directory of the runtime.

CMakeLists.txt
philipturner: > but if you are concerned that many people might not be using/enabling the SSCP compiler

My concern is that many people use the SSCP compiler (no one's going to use the old workflow), but few people use the SSCP compiler + the Metal backend. Macs would be stuck in CPU-only because Metal is disabled by default. That's already the case with CUDA/ROCM/ZE workflows - you must explicitly *enable* each backend. I don't want Metal to be like that.

For example, some enters the hipSYCL directory and types `cmake .. -DWITH_SSCP_COMPILER -DWITH_HIP_BACKEND`. Could you provide a breakdown of what happens with each OS and with each hardware vendor?

src/runtime/metal/metal_hardware_manager.cpp
```diff
@@ -0,0 +1,641 @@
...
+// TODO: Make a special file to declare these implementations.
+// You must define these *before* including Metal.hpp.
+#define NS_PRIVATE_IMPLEMENTATION
+#define MTL_PRIVATE_IMPLEMENTATION
+#define CA_PRIVATE_IMPLEMENTATION
+
+#include <Metal/Metal.hpp>
+#include <IOKit/IOKitLib.h>
+#include <CoreFoundation/CoreFoundation.h>
+#include <sys/sysctl.h>
```
philipturner, line 42: > glue is for code that connects two or more of the hipSYCL components (runtime, compiler, headers). Where would metal-cpp be needed? If it's just needed for the runtime, you could also put it inside the metal directory of the runtime.

metal-cpp is just needed for compile-time. It's a wrapper around the Objective-C Metal binaries, which ship with macOS.

Declaring these `#defines` in some random source file doesn't make much sense. We could make `metal_cpp_implementation.cpp` in this `runtime/metal` directory. But how is this part of the runtime?

src/runtime/CMakeLists.txt
philipturner: Is it appropriate to place the `find_library` invocations in the CMakeLists of `src/runtime`, or do we need to extract it somewhere else? How do/should we mention them in the compile flags listed by `syclcc --hipsycl-version`?

philipturner: Is it appropriate to place the `find_library` invocations in the CMakeLists of `src/runtime`, or do we need to extract it somewhere else? How do/should we mention them in the compile flags listed by `syclcc --hipsycl-version`?

CMakeLists.txt
illuhad: > My concern is that many people use the SSCP compiler (no one's going to use the old workflow), but few people use the SSCP compiler + the Metal backend. Macs would be stuck in CPU-only because Metal is disabled by default. That's already the case with CUDA/ROCM/ZE workflows - you must explicitly enable each backend. I don't want Metal to be like that.

This is absolutely not true. Those backends will automatically be enabled whenever CUDA or ROCm installations are found by cmake (this is not yet the case for Level Zero, but that is only because it was super experimental with the old compilation flow anyway).
This is pretty much the best we can do because we *have* to rely on preinstalled CUDA or ROCm anyway because those dependencies are too complex to ship with hipSYCL. If cmake cannot find it, we can only assume it's not installed or installed in some non-standard path, in which case the user must provide explicit cmake options anyway.

If shipping Metal dependencies with hipSYCL is simpler, or we can just assume it to be available on Mac, I am not opposed to just enabling it unconditionally for those platforms.

> For example, some enters the hipSYCL directory and types cmake .. -DWITH_SSCP_COMPILER -DWITH_HIP_BACKEND. Could you provide a breakdown of what happens with each OS and with each hardware vendor?

* `-DWITH_SSCP_COMPILER` (enabled by default whenever cmake finds an LLVM >= 14 on the system): 
  * If no sufficiently new LLVM is found it errors, otherwise it builds the core SSCP compiler infrastructure
* `-DWITH_ROCM_BACKEND` (enabled by default whenever cmake finds ROCm):
  * If no ROCm is found it errors, otherwise builds `librt-backend-hip` and `llvm-to-amdgpu` since `WITH_SSCP_COMPILER` is set.
 
I don't think there are any particular OS-specific branches regarding the HIP backend or SSCP compiler, but of course AMD supports ROCm only on Linux currently, so it will not be found or be installable on other OSes. But that is an AMD problem, not a hipSYCL problem ;)

src/runtime/metal/metal_allocator.cpp
illuhad: You could use `__builtin_clz()` which should be widely available on many compilers (since Metal backend is only going to be built on Apple platforms, it would be sufficient if that builtin is supported by the compilers there. clang has it). C++ 20 introduces `std::countl_zero`.

src/runtime/metal/metal_hardware_manager.cpp
illuhad: I had assumed that this metal-cpp provides runtime functionality, such as hardware management, submitting tasks etc. In that case the only user would be the hipSYCL metal runtime backend, so it would make sense to put it there.

Or is that not the case? Is it something that is needed by kernels?

philipturner: > Or is that not the case? Is it something that is needed by kernels?

It might also export C++ symbols. For example, we might permit Metal interop with SYCL code. That requires exposing the `MTL::CommandQueue`, `MTL::Buffer` etc. objects through a C++ API. Some of our shipped headers must import metal-cpp.

illuhad: Do we need metal-cpp for that, or would it be sufficient to just forward declare the exported types, and require that the user include metal-cpp by themselves if they want to use such interop? This is what the other backends currently do for interop.

philipturner: That might be the best idea.

illuhad:
Can you reopen against `develop`, as the SSCP compiler is now merged?

philipturner:
I want to reopen once I've made more progress, which may be a while (winter break is ending). Meanwhile, would you mind correcting the spelling/formatting nits? 

Also, could you add something automated that ensures each file ends with a newline? That will minimize unrelated diffs when people edit in Xcode.

PR #905, opened 2023-01-12
[Doc] Update install-llvm.md

nmnobre:
Since e092e65, the build process requires `libLLVM.so` (`-lLLVM`) due to a cmake call to `llvm_config` in the new llvm-to-backend module:

https://github.com/illuhad/hipSYCL/blob/21d99ef0434d81cb071cc38e3999e8dda1cf21a9/src/compiler/llvm-to-backend/CMakeLists.txt#L30

This just asks the user to make sure to generate said library when compiling LLVM.


illuhad (approved):
Thanks! I was thinking of whether we should clarify that this is only needed when the SSCP compiler is enabled, but since it's enabled by default with LLVM >= 14, this might be more confusing than helpful...

nmnobre:
It's unlikely anyone building from source today or in the future is using anything below version 15, but we can add a note, let me have a go.

illuhad:
That's true (it's even supported starting LLVM 14), but there might be users who explicitly disable it because they either don't need it, or they build against an LLVM in-between stable versions (e.g. ROCm clang) where LLVM APIs might be in a transitional state that breaks our code, or because they use tip-of-the-tree LLVM which I think currently might not work with the SSCP compiler, at least for the SPIR-V backend.

illuhad:
Thanks, LGTM!

PR #907, opened 2023-01-13
Update `sycl::vec` class to reflect SYCL 2020 requirements 

nilsfriess:
This PR introduces the following small changes in the `sycl::vec` class to reflect the SYCL 2020 spec:
- A `size` method is added which replaces the `get_count` method (which is now marked deprecated)
- A `byte_size` method is added which replaces the `get_size` (which is now marked deprecated)
- The signature of the `as` method is changed: according to Table 144 of the SYCL 2020 spec, the template parameter `asT` should refer to the whole vector type that the vec is reinterpreted as. Previously, `asT` referred to the type of the new vec's elements.

These all came up when trying to compile the CTS `vec_api` test; it is still not compiling, but the remaining errors seem to be be CTS-related, not hipSYCL-related.

### DIFF
--- a/include/hipSYCL/sycl/libkernel/generic/hiplike/builtins.hpp
+++ b/include/hipSYCL/sycl/libkernel/generic/hiplike/builtins.hpp
@@ -473,7 +473,7 @@ HIPSYCL_HIPLIKE_BUILTIN T __hipsycl_dot(T a, T b) noexcept {
 template <class T, std::enable_if_t<!std::is_arithmetic_v<T>, int> = 0>
 HIPSYCL_HIPLIKE_BUILTIN typename T::element_type __hipsycl_dot(T a, T b) noexcept {
   typename T::element_type result = 0;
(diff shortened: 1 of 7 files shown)


illuhad:
Thanks! Have you made a survey whether we use those deprecated functions somewhere internally in hipSYCL? E.g. I strongly suspect that `get_count()` is used by the builtin implementations to iterate over `vec` elements. It would be unfortunate if users get a deprecated-warning triggered by our own headers :-)

nilsfriess:
No, I didn't really think about that... 
I've now replaced `get_count` by `size` in the builtins and the tests that use it.

Actually, there is [one function in the group function tests](https://github.com/illuhad/hipSYCL/blob/develop/tests/sycl/group_functions/group_functions.hpp#L212) that still uses it but this function isn't actually used anywhere so maybe this function can just be deleted anyway?

illuhad:
@DieGoldeneEnte can you comment on the function in the group function tests?

illuhad (approved):

PR #916, opened 2023-01-18
Fix zero dimensional accessors

nilsfriess:
Zero dimensional accessors can be used to access the first element of a 1D buffer. However, the methods in the accessor class expected the `dimensions` of the accessor and the buffer to match, causing a compilation error when trying to define a zero dimensional accessor. 
To fix this, a compile time constant `adjDimensions` is introduced that is set to 1 if `dimensions == 0` and set to the value of `dimensions` if `dimensions > 0`. This is then used instead of `dimensions` where necessary.


nilsfriess:
Superseded by #1072

PR #919, opened 2023-01-20
Apply LLVM EH/RTTI settings to LLVM dependent libs/tools.

fodinabor:
Also document that SSCP requires LLVM to have been built with RTTI atm.

This is just to document the current state and enforce it before having more obscure issues later.
One might want to consider whether the `typeid` call in `LLVMToBackendTranslator` can somehow be replaced to remove that requirement.

I don't know what distros typically build their LLVM with, with or without RTTI. Manjaro e.g. does with..


illuhad:
Hang on, I think I have an idea on how to fix this in the first place.

fodinabor:
Closing as superseeded by #920

PR #920, opened 2023-01-20
[SSCP] Avoid using typeid in LLVMToBackend to allow RTTI-less LLVM

illuhad:
@fodinabor Does this do the trick?


fodinabor (approved):
Sneaky ;-)
Works for me :+1:

PR #926, opened 2023-01-22
Explicitly convert paths to `std::string` 

nilsfriess:
I am currently trying to fix the Windows Github CI and clang complained that it can't convert `fs::path` types to `std::string` (no idea why that's not a problem on Linux). This PR just calls the `.string()` method on the `fs::path` objects to explicitly convert them to `std::string`s.

### DIFF
--- a/src/common/filesystem.cpp
+++ b/src/common/filesystem.cpp
@@ -68,8 +68,8 @@ std::string get_install_directory() {
   
 #endif
   if(paths.empty() || !fs::is_directory(paths.back()))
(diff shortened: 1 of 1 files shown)


al42and:
> no idea why that's not a problem on Linux

`fs::path` can only be implicitly converted to a "native" string, which is `std::string` on Linux, and `std::wstring` on Windows. Quite annoying. Ref: https://en.cppreference.com/w/cpp/filesystem/path/native

Just FYI.

illuhad (approved):
Thanks!

PR #927, opened 2023-01-22
install-rocm.md: Problem with ROCm 5.0 and SSCP

al42and:
Clang bundled with ROCm 5.0.2 is incompatible with SSCP, causing a compilation error.

### DIFF
--- a/doc/install-rocm.md
+++ b/doc/install-rocm.md
@@ -8,7 +8,7 @@ Please install ROCm 4.0 or later as described in the ROCm readme. Make sure to a
 * **Such configurations typically work, but are generally less tested.**
 * Also note that the LLVM distributions shipping with ROCm are not official LLVM releases, and depending on when the upstream development was last merged, may have slightly diverging functionality. There are multiple known cases where this causes problems: 
   * The clang 13 from ROCm 4.5 lacks functionality that is present in official clang 13 releases and that hipSYCL's clang 13 code paths need. In that case you will need to set `-DHIPSYCL_NO_DEVICE_MANGLER=ON` when compiling hipSYCL. This will however break [explicit multipass](compilation.md) support.
-  * Similarly, the clang 14 from ROCm 5.0 lacks functionality that is present in official clang 14 releases that are required by hipSYCL's compiler acceleration for CPU targets. You can work around those issues by setting `-DWITH_ACCELERATED_CPU=OFF` at the expense of reduced kernel performance on CPUs.
+  * Similarly, the clang 14 from ROCm 5.0 lacks functionality that is present in official clang 14 releases. You can work around those issues by setting `-DWITH_ACCELERATED_CPU=OFF -DWITH_SSCP_COMPILER=OFF` at the expense of reduced kernel performance on CPUs and lack of [SSCP](compilation.md) support.
 
 *Note: hipSYCL is by default configured to utilize the ROCm compilation flags that apply for recent clang and ROCm versions. If you are using an older clang (<= 10) or ROCm < 4, you might have to adjust `-DROCM_CXX_FLAGS` (not recommended!).*
 


illuhad (approved):
Ugh, thanks. Not surprising that this doesn't work, but still sad. Given the role of SSCP as the upcoming new "default" compilation  flow once it is more stable, we may have to rephrase this eventually to discourage ROCm clang more strongly.
By the way, SSCP should be more robust with respect to ROCm compatibility because it just feeds hipRTC with LLVM IR, which then generates amdgcn. So, hipSYCL does not even need to be built against ROCm clang to take advantage of the ROCm LLVM backend :-)

al42and:
> Ugh, thanks. Not surprising that this doesn't work, but still sad. Given the role of SSCP as the upcoming new "default" compilation flow once it is more stable, we may have to rephrase this eventually to discourage ROCm clang more strongly.

It works fine with newer ROCm's, at least 5.4 that I have. It's that 14-but-not-quite version that's shipped with ROCm 5.0 specifically that is the problem. And unfortunately, some HPC installations of interest still use ROCm 5.0.2.

> So, hipSYCL does not even need to be built against ROCm clang to take advantage of the ROCm LLVM backend :-)

But we lose things like `#ifdef __gfx90a__` and other architecture-specific optimizations, don't we?

illuhad:
> It works fine with newer ROCm's, at least 5.4 that I have. It's that 14-but-not-quite version that's shipped with ROCm 5.0 specifically that is the problem. And unfortunately, some HPC installations of interest still use ROCm 5.0.2.

Yeah, so the usual problem with in-between clang versions :(

> But we lose things like #ifdef __gfx90a__ and other architecture-specific optimizations, don't we?

Not necessarily. While macros will be unsupported for this purpose, the SSCP flow is designed to provide similar mechanism as e.g. nvc++ `if target()` to support target specialization. There is a lot of infrastructure already in place for this, and it is supported already for `__hipsycl_if_target_host()/__hipsycl_if_target_device`. APIs for device-specific specialization are not yet publicly available but will come too :)

al42and:
> There is a lot of infrastructure already in place for this, and it is supported already for __hipsycl_if_target_host()/__hipsycl_if_target_device

It is still somewhat limited. E.g., constructs like `__hipsycl_if_target_device(atomicAdd(ptr, 1.0F););` don't work with `generic`, while working fine in explicit- and implicit multipass modes. And I don't think just declaring (but not defining) these functions for host pass will work, will it?

illuhad:
> It is still somewhat limited. E.g., constructs like __hipsycl_if_target_device(atomicAdd(ptr, 1.0F);); don't work with generic, while working fine in explicit- and implicit multipass modes. And I don't think just declaring (but not defining) these functions for host pass will work, will it?

Well, you cannot use *HIP* source code in kernels. But you will likely be able to use inline assembly, and calling builtins from AMD device bitcode libraries like OCML or OCKL (https://github.com/RadeonOpenCompute/ROCm-Device-Libs) already works today by just declaring them (as long as they are only used on the AMD path).
It might also be possible to link against bitcode libraries generated the clang HIP toolchain but that is still more unclear.

In other words, in your example you could not call `atomicAdd` because that is defined only inside HIP header files, and we cannot recreate HIP headers on the fly. You should however be able to just invoke the builtins that `atomicAdd` invokes under the hood. (assuming of course that you don't just want to call the SYCL atomics in the first place :) )

illuhad:
> And I don't think just declaring (but not defining) these functions for host pass will work, will it?

There is no dedicated host pass, it's a unified host-device pass :) But I get what you are asking: What will happen when linking the host binary and there are undefined function declarations, which only get resolved on the device side.

The short answer is: This is supported, because device-only code is removed from the host binary by the SSCP compiler. So it will detect that these declarations are only used inside kernels, are therefore device code, and then remove them from the host code.

Because of this, we can actually even support `SYCL_EXTERNAL` device functions that are resolved dynamically in device code from a library that is opened via `dlopen()`. So you could even have a plugin architecture for kernels :D

al42and:
> Well, you cannot use HIP source code in kernels. But you will likely be able to use inline assembly, and calling builtins from AMD device bitcode libraries like OCML or OCKL (https://github.com/RadeonOpenCompute/ROCm-Device-Libs) already works today by just declaring them (as long as they are only used on the AMD path).

Yeah, unless those are arch-specific. Or, well, if one wants to target both AMD and NVIDIA devices. But that, as I understood, will also be doable soon :+1: 

> assuming of course that you don't just want to call the SYCL atomics in the first place :) 

The last time I checked (late fall 2022), clang/llvm could not use "unsafe" AMD atomics when they were hidden under all the layers of SYCL, while plain `atomicAdd` were optimizable, with the right compiler flags.

> The short answer is: This is supported, because device-only code is removed from the host binary by the SSCP compiler. So it will detect that these declarations are only used inside kernels, are therefore device code, and then remove them from the host code.

Ok. I thought the code for the host was always generated, my bad.

Re-declaring all the functions is an annoyance, but yeah, that *is* pretty cool, as long as JIT can optimize the code well enough. Unfortunately, GROMACS refuses to build for `generic` target as-is, and I don't have time to fix all the problems now and check out its performance :(

illuhad:
> Yeah, unless those are arch-specific. Or, well, if one wants to target both AMD and NVIDIA devices. But that, as I understood, will also be doable soon

Not sure what you are referring to exactly. AMD OCKL/OCML builtins work across all AMD GPUs, but need to be configured for the appropriate ISA by linking with the right ROCm bitcode libraries. JIT takes care of that automatically.
Of course they won't work on NVIDIA, but that's expected if you start to write backend-specific code paths :)

> The last time I checked (late fall 2022), clang/llvm could not use "unsafe" AMD atomics when they were hidden under all the layers of SYCL, while plain atomicAdd were optimizable, with the right compiler flags.

Well in that case it might be a better idea to first try to invoke our `__hipsycl_sscp_*` builtins for atomics directly once they are there before trying to reimplement things ;)

> Re-declaring all the functions is an annoyance, but yeah, that is pretty cool, as long as JIT can optimize the code well enough. Unfortunately, GROMACS refuses to build for generic target as-is, and I don't have time to fix all the problems now and check out its performance :(

We already have headers for AMD ocmkl/ockl and NVIDIA libdevice - so you could just include those, I think, and not redeclare yourself :)

I don't expect it to work with Gromacs just yet. Atomics and group algorithms are not yet implemented which I think you need.
As for performance - I'm working on an IWOCL paper on SSCP at the moment, and so far it looks promising. Especially on AMD I am seeing cases where it outperforms the old compilation flow by 20% without substantial regressions so far. And I have not yet even started to look at optimizing SSCP perf in particular :)
JIT goes through LLVM -O3 pipeline after linking (so you always effectively get lto), then it goes to hipRTC which also optimizes.

al42and:
> Not sure what you are referring to exactly. AMD OCKL/OCML builtins work across all AMD GPUs, but need to be configured for the appropriate ISA by linking with the right ROCm bitcode libraries. JIT takes care of that automatically.

I was referring to things like `__ockl_atomic_add_noret_f32`, which are valid on all architectures, but only make sense on certain ones. 

> Of course they won't work on NVIDIA, but that's expected if you start to write backend-specific code paths :)

Well, in multipass mode, one can already both have the cake (portability) and eat it too (backend-specific code paths) :)

> We already have headers for AMD ocmkl/ockl and NVIDIA libdevice - so you could just include those, I think, and not redeclare yourself :)

Currently, not so much. We cannot include NVIDIA headers if there's no CUDA installed on the machine, so we cannot compile on HPC systems that only have ROCm stack available. More PP macros can solve the problem, of course, but that is ugly.

> Atomics and group algorithms are not yet implemented which I think you need.

That can be danced around with intrinsics/builtins. A bigger problem would be that we rely on knowing the sub-group size at compile time. E.g., we do reduction manually, and we need to know both the number of steps to do and the size of the local memory buffer (and whether the buffer is needed at all). That is what can be hard to optimize, I think.

>  As for performance - I'm working on an IWOCL paper on SSCP at the moment, and so far it looks promising. Especially on AMD I am seeing cases where it outperforms the old compilation flow by 20% without substantial regressions so far. And I have not yet even started to look at optimizing SSCP perf in particular :)

:+1:. Do you have any idea why JIT works better than AOT compilation?

illuhad:
> I was referring to things like __ockl_atomic_add_noret_f32, which are valid on all architectures, but only make sense on certain ones.

Ok. Sure, but we are talking about creating optimized code paths for particular HW already, so having to specialize to some extent already is expected, no?

>  Well, in multipass mode, one can already both have the cake (portability) and eat it too (backend-specific code paths) :)

Only if you ignore Intel hardware ;) With SSCP Intel is just as well supported as AMD and NV. In multipass.. it's complicated ;)

> Currently, not so much. We cannot include NVIDIA headers if there's no CUDA installed on the machine, so we cannot compile on HPC systems that only have ROCm stack available. More PP macros can solve the problem, of course, but that is ugly.

Yes, that's true for multipass, but I was talking about SSCP where we e.g. already ship ockl headers as part of hipSYCL:
https://github.com/illuhad/hipSYCL/blob/develop/include/hipSYCL/sycl/libkernel/sscp/builtins/amdgpu/ockl.hpp
So I think you could just include that file and invoke those builtins directly without having to declare them yourself.

> That can be danced around with intrinsics/builtins. A bigger problem would be that we rely on knowing the sub-group size at compile time. E.g., we do reduction manually, and we need to know both the number of steps to do and the size of the local memory buffer (and whether the buffer is needed at all). That is what can be hard to optimize, I think.

You would have to do something like:
```
if(__hipsycl_sscp_get_subgroup_max_size() == 32){
  run<32>();
else if(__hipsycl_sscp_get_subgroup_max_size() == 64) {
  run<64>();
}
```
Alternatively, you could query the backend and specialize based on that. But I don't think it would make a difference, because those queries return constants which should allow SSCP runtime kernel outlining to throw away to unneeded paths.

> +1. Do you have any idea why JIT works better than AOT compilation?

I don't know what exactly is the cause in this particular case, but there are some differences that could play a role here:

* SSCP goes through LLVM -O3 and then hipRTC, which does some optimizations as well - there might be differences here compared to just going LLVM -O3 AOT.
* SSCP has its own mapping of builtins, i.e. it directly calls ockl/ocml functions. There might be differences compared to the HIP C++ functions that the multipass model invokes
* SSCP generates somewhat different LLVM IR. For example, it passes pointers as individual arguments into the kernel with addrspace(1) annotation. It then runs some address space inference passes. Multipass just passes in the kernel lambda as one blob. The SSCP setup might allow for more efficient address space inference.

al42and:
> Sure, but we are talking about creating optimized code paths for particular HW already, so having to specialize to some extent already is expected, no?

Yes, fair enough. Does not make sense to complain about a known limitation which is soon to be fixed anyway :)

> Yes, that's true for multipass, but I was talking about SSCP where we e.g. already ship ockl headers as part of hipSYCL:
> https://github.com/illuhad/hipSYCL/blob/develop/include/hipSYCL/sycl/libkernel/sscp/builtins/amdgpu/ockl.hpp

Missed them. That's convenient.

> But I don't think it would make a difference, because those queries return constants which should allow SSCP runtime kernel outlining to throw away to unneeded paths.

We'll see. I'm a bit skeptical given how often we have to fight with the compiler over similarly "minor" things, but maybe these conditionals will be optimized just fine.

illuhad:
> Missed them. That's convenient.

Our `__hipsycl_sscp_*` builtin implementations need to have declarations too so they can call backend builtins, that's why we have those headers :-)

> We'll see. I'm a bit skeptical given how often we have to fight with the compiler over similarly "minor" things, but maybe these conditionals will be optimized just fine.

Well, the host-device code separation relies on a similar mechanism using `if(__hipsycl_sscp_is_device){}`, so that better be reliable ;)
The only thing that could be an issue in this case is that in the AMD path, there's an additional if-statement based on `__oclc_wavefrontsize64` to support both wave32 and wave64 modes on AMD.  `__oclc_wavefrontsize64` only gets resolved by hipRTC, after our kernel reoutlining steps already have already run. However, the AMD bitcode libraries like ockl themselves rely heavily on being able to optimize precisely these if-statements, so hipRTC will have to ensure this too.

If hipRTC is not thorough enough here, we could however always insert the correct value ourselves during JIT. Bottom line is that it should be possible to support this either way.

PR #931, opened 2023-01-24
Resolve -Wpessimizing-move warning

al42and:
std::move is not necessary when returning things or dealing with temporaries.


illuhad (approved):
Thanks!

PR #936, opened 2023-01-26
Fix parsing backend string

al42and:
Nothing major, just any invalid value was parsed as OpenMP instead of triggering an error.

Also removed an unused variable.


illuhad (approved):
Thanks!

PR #937, opened 2023-01-26
[Doc] Tweak instructions to build w/ LLVM

nmnobre:
Hi @illuhad,

Minor change again.

It turns out when installing LLVM from source you don't get the symlink in `/path/to/llvm/cmake` to `/path/to/llvm/lib/cmake/llvm` and specifying `-DLLVM_DIR=/path/to/llvm/lib/cmake` does not seem to work.

Curiously, `-DLLVM_ROOT=/path/to/llvm/cmake` (which, if it exists, will just point to `/path/to/llvm/lib/cmake/llvm`) and `-DLLVM_ROOT=/path/to/llvm/lib/cmake` (notice I didn't need to specify the llvm subdirectory as above for `LLVM_DIR`) both work...
I think using `LLVM_ROOT`, if the user specifies whichever path ending in `cmake` within the LLVM installation directory, it should (hopefully) always work, increasing the chances of success at first try.

I think it's at least the second time I've spent more than a couple of minutes fighting this out, so I thought I'd share it, even if it's just for my future self.

Cheers,
-Nuno

### DIFF
--- a/doc/install-llvm.md
+++ b/doc/install-llvm.md
@@ -36,7 +36,7 @@ It is generally not necessary to compile LLVM by yourself. However, if you wish 
 When invoking cmake, the Open SYCL build infrastructure will attempt to find LLVM automatically (see below for how to invoke cmake).
 
 If Open SYCL does not automatically configure the build for the desired clang/LLVM installation, the following cmake variables can be used to point Open SYCL to the right one:
-* `-DLLVM_DIR=/path/to/llvm/cmake` must be pointed to your LLVM installation, specifically, the **subdirectory containing the LLVM cmake files**. 
+* `-DLLVM_DIR=/path/to/llvm/cmake` must be pointed to your LLVM installation, specifically, the **subdirectory containing the LLVM cmake files**. Note that different LLVM installations may have the LLVM cmake files in different subdirectories that don't necessarily end with `cmake` (e.g. it might also be `/path/to/llvm/lib/cmake/llvm`). Alternatively, you can try `-DLLVM_ROOT` which might be more forgiving.
 
 Verify from the cmake that the selected `clang++` and include headers match the LLVM that you have requested. Example output:
 ```
--- a/doc/install-rocm.md
+++ b/doc/install-rocm.md
@@ -2,7 +2,7 @@
 
 Please install ROCm 4.0 or later as described in the ROCm readme. Make sure to also install HIP (runtime libraries and headers).
 
-*Note: Newer ROCm versions may require building Open SYCL against newer clang versios as well. For example, ROCm 4.5 requires clang 13+.*
+*Note: Newer ROCm versions may require building Open SYCL against newer clang versions as well. For example, ROCm 4.5 requires clang 13+.*
 
 *Note: Instead of building Open SYCL against a regular clang/LLVM, it is also possible to build Open SYCL against the clang/LLVM that ships with ROCm. This can be interesting if other available clang/LLVM installations are not new enough to work with the ROCm installation.* 
 * **Such configurations typically work, but are generally less tested.**


doc/install-llvm.md
```diff
@@ -36,7 +36,7 @@ It is generally not necessary to compile LLVM by yourself. However, if you wish
 When invoking cmake, the Open SYCL build infrastructure will attempt to find LLVM automatically (see below for how to invoke cmake).
 
 If Open SYCL does not automatically configure the build for the desired clang/LLVM installation, the following cmake variables can be used to point Open SYCL to the right one:
-* `-DLLVM_DIR=/path/to/llvm/cmake` must be pointed to your LLVM installation, specifically, the **subdirectory containing the LLVM cmake files**. 
+* `-DLLVM_ROOT=/path/to/llvm/cmake` must be pointed to your LLVM installation, specifically, the **subdirectory containing the LLVM cmake files**.
```
illuhad:
```suggestion
* `-DLLVM_DIR=/path/to/llvm/cmake` must be pointed to your LLVM installation, specifically, the **subdirectory containing the LLVM cmake files**. Note that different LLVM installations may have the LLVM cmake files in different subdirectories that don't necessarily end with `cmake` (e.g. it might also be `/path/to/llvm/lib/cmake/llvm`). Alternatively, you can try `-DLLVM_ROOT` which might be more forgiving.
```

Maybe something like this? `*_DIR` seems to be the standard way in cmake to specify the path to dependencies, so assuming that the path is correct, this might be more reliable or natural than alternatives.

nmnobre: Sure, I like it. :-)

illuhad (approved):

Momellouky:
Hi. 

I am trying to run this script in my local machine (Ubuntu 22.04 LTS). To be honest, I am not experienced with cmake. So I am facing some issues that I couldn't solve. 

The issue is the following: 
It seems that cmake doesn't reconize the path to clang++. ASM. In fact, I am pretty confused and I couldn't reach a conclusion of what causes the issue, so I couldn't solve it. 
The trace of the execution is the following: 

```
Cloning LLVM llvmorg-11.0.1
Cloning into '/home/wannatry/git/llvm-vanilla'...
remote: Enumerating objects: 5357981, done.
remote: Counting objects: 100% (3194/3194), done.
remote: Compressing objects: 100% (571/571), done.
remote: Total 5357981 (delta 2799), reused 2764 (delta 2622), pack-reused 5354787
Receiving objects: 100% (5357981/5357981), 2.41 GiB | 1.43 MiB/s, done.
Resolving deltas: 100% (4392593/4392593), done.
Note: switching to '43ff75f2c3feef64f9d73328230d34dac8832a91'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

Updating files: 100% (94299/94299), done.
-- The C compiler identification is unknown
-- The CXX compiler identification is unknown
-- The ASM compiler identification is unknown
-- Didn't find assembler
CMake Error at CMakeLists.txt:57 (project):
  The CMAKE_C_COMPILER:

    clang

  is not a full path and was not found in the PATH.

  Tell CMake where to find the compiler by setting either the environment
  variable "CC" or the CMake cache entry CMAKE_C_COMPILER to the full path to
  the compiler, or to the compiler name if it is in the PATH.


CMake Error at CMakeLists.txt:57 (project):
  The CMAKE_CXX_COMPILER:

    clang++

  is not a full path and was not found in the PATH.

  Tell CMake where to find the compiler by setting either the environment
  variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
  to the compiler, or to the compiler name if it is in the PATH.


CMake Error at CMakeLists.txt:57 (project):
  No CMAKE_ASM_COMPILER could be found.

  Tell CMake where to find the compiler by setting either the environment
  variable "ASM" or the CMake cache entry CMAKE_ASM_COMPILER to the full path
  to the compiler, or to the compiler name if it is in the PATH.


-- Warning: Did not find file Compiler/-ASM
-- Configuring incomplete, errors occurred!
See also "/home/wannatry/git/llvm-vanilla/build/CMakeFiles/CMakeOutput.log".
See also "/home/wannatry/git/llvm-vanilla/build/CMakeFiles/CMakeError.log".
```

Thank you

nmnobre:
Hi, thanks for reaching out.

> I am trying to run this script in my local machine (Ubuntu 22.04 LTS). To be honest, I am not experienced with cmake. So I am facing some issues that I couldn't solve.

I'm a bit lost, what script are you referring to?

> The issue is the following: It seems that cmake doesn't reconize the path to clang++.

Right, so the recommended (and probably easiest) way of doing things is to use `apt` as per the instructions [here](https://github.com/OpenSYCL/OpenSYCL/blob/develop/doc/install-llvm.md).
However, it appears you were trying to build (quite an old version of) LLVM from source.
If you really want to do that, if you first need a set of compilers and development tools, which people usually get by installing `build-essential` via `apt`.

Cheers,
-Nuno

Momellouky:
Hi @nmnobre 

I apologize, it was my bad I forgot to mention which script I am referring to. I am trying to run the `install-llvm.sh `script. 
However, the instructions from this [link](https://github.com/OpenSYCL/OpenSYCL/blob/develop/doc/install-llvm.md), works fine. 
I am trying to understand what I was missing to report it (it may be useful for someone in the future). I guess the issue was the fact that I didn't assign the cmake the arguments it needs. 

Thank you.

illuhad:
Hi, the `install-llvm.sh` script is part of our old packaging infrastructure that is planned to be redesigned from scratch. It is quite old and outdated, and not intended for users. We should perhaps remove it.

Momellouky:
Hi, 
Thank you for your feedback. I was trying to install opensycl in my local machine using scripts. 
However, the manual install works fine. 
Thank you.

nmnobre:
>  I guess the issue was the fact that I didn't assign the cmake the arguments it needs.

Yes, looking at the script, you'd need to set `HIPSYCL_BASE_CC` and `HIPSYCL_BASE_CXX` to some compiler you actually have installed. But, as Aksel (@illuhad) said, it's probably not worth the trouble given it's such an old of version of things.

PR #941, opened 2023-02-08
[SSCP] Add atomic support

illuhad:
WIP. Adds support for atomics to the SSCP compiler.

- [x] Builtin infrastructure support
- [x] Implementation for llvm-to-amdgpu backend
- [x] Implementation for llvm-to-spirv backend
- [x] Implementation for llvm-to-ptx backend


src/libkernel/sscp/amdgpu/atomic.cpp
```diff
@@ -0,0 +1,638 @@
...
+}
+
+HIPSYCL_SSCP_BUILTIN void __hipsycl_sscp_atomic_store_i64(
+    __hipsycl_sscp_address_space as, __hipsycl_sscp_memory_order order,
+    __hipsycl_sscp_memory_scope scope, __hipsycl_int64 *ptr, __hipsycl_int64 x) {
+  return __atomic_store_n(ptr, x, builtin_memory_order(order));
+}
+
+
+// ********************** atomic store ***************************
```
sbalint98:
```suggestion
// ********************** atomic load ***************************
```

src/libkernel/sscp/amdgpu/atomic.cpp
```diff
@@ -0,0 +1,638 @@
...
+    __hipsycl_sscp_address_space as, __hipsycl_sscp_memory_order order,
+    __hipsycl_sscp_memory_scope scope, __hipsycl_int64 *ptr, __hipsycl_int64 x) {
+  return __hip_atomic_fetch_xor(ptr, x, builtin_memory_order(order),
+                                builtin_memory_scope(scope));
+}
+
+
+
+// TODO: It seems that at least on gfx90a, there are additional unsafe atomic
+// operations that can yield a performance improvement. Should we expose those?
```
sbalint98, line 351: For reference I think this GROAMCS issue is related. I think we should definitely expose them, however would it be possible to use the `__builtin_amdgcn_ds_atomic_fadd_f32` https://gitlab.com/gromacs/gromacs/-/merge_requests/3493

illuhad: Usually the set of enabled clang builtins depends on the frontend. It's likely that amdgcn builtins won't work by default with SSCP because the clang frontend does not know that it will run on anything else but the host CPU.

We could just create a simple `__hip/opensycl_sscp_amdgcn_ds_atomic_fadd_f32()` declaration and implement this inside `amdgpu/atomic.cpp` with a simple call to the amdgcn builtin. As long as this function is only invoked when JITing for amdgpu (and a GPU that supports it), it should be fine with SSCP.

illuhad:
This PR now provides complete implementations of atomic operations for SPIR-V, PTX, and amdgcn.

illuhad:
Passes all current atomic tests on all backends. CC @RaulPPelaez

PR #946, opened 2023-02-09
Logo: Add text-to-path version of logo, and double check image size

### LINKED ISSUE #945: SVG version of the logo is not rendered consistently
**Bug summary**

Firefox 109 on Ubuntu 22.04 renders the SVG logo differently from the PNG:

- Cropped on the right
- Different font

Above: PNG from README; below: SVG opened separately in the same browser.

image

On Chrome Mobile, the fonts are correct, but the "L" is still cropped.

**To Reproduce**

Open SVG and PNG side-by-side on different systems.

**Expected behavior**

No perceptible differences between SVG and PNG versions.

**Additional context**

I suspect converting the text to paths or embedding the font would help.

illuhad:
Hopefully fixes #945 ?


al42and:
- `doc/img/logo/logo-color-no-text.svg` looks fine.
- `doc/img/logo/logo-color.svg` is still cropped, perhaps because the font is wider.

illuhad:
Ok, this is possible. It uses the Trade Gothic font, if you don't have it, maybe it doesn't fit. I would like to keep the version with the text in case some (hopefully minor) adjustments to the text have to made in the future.

al42and:
There is a way to embed fonts in the SVG, to make them both editable and consistently renderable: https://dzx.fr/blog/svg-font-embedding/

But since Trade Gothic requires licensing, this might be not possible from a legal standpoint.

illuhad:
Hm, okay. I'm not opposed to eventually changing the font, but for the sake of continuitiy in the logo appearance it might be a good idea to stick with it for a while. If it's ok, I'd merge this is a workaround for the time being.

PR #948, opened 2023-02-10
[Renaming] Remove internal syclcc references to hipSYCL; accept Open SYCL arguments

illuhad:
* Remove all internal usage of hipSYCL in syclcc that have no side effects;
* Migrate to `---opensycl-*` command line arguments and `OPENSYCL_*` environment variables
* Add mechanism to also accept the old `--hipsycl-*` flags and `HIPSYCL_*` environment variables


bin/syclcc-clang
```diff
@@ -394,13 +403,28 @@ class syclcc_config:
...
+
+  def _upgrade_legacy_arg(self, arg):
+    if(arg.startswith("--hipsycl")):
+      return arg.replace("--hipsycl", "--opensycl")
+    return arg
+
+  def _upgrade_legacy_environ_var(self, arg):
+    if(arg.startswith("HIPSYCL_")):
+      return arg.replace("HIPSYCL_", "OPENSYCL_")
+    return arg
```
sbalint98, line 415: Maybe we should warn about these being used, and that they might get removed in a future release?

illuhad: Not before we have not updated the entire infrastructure. E.g. cmake still uses `--hipsycl` and we wouldn't want all users to get bombarded with warnings for no reason ;)

PR #954, opened 2023-02-22
Check number of `Args...` of `vec` constructor in template parameter to allow SFINAE

nilsfriess:
Previously, the `vec(const Args &...)` constructor was sometimes hiding the `vec(const T& value)` constructor even in cases where it is not applicable since `count_num_elements != N` was only checked inside the constructor (i.e., after the compiler had already chosen this constructor) . Now `count_num_elements == N` is checked in the template arguments, allowing to choose the correct constructor using SFINAE.

### DIFF
--- a/include/hipSYCL/sycl/libkernel/builtins.hpp
+++ b/include/hipSYCL/sycl/libkernel/builtins.hpp
@@ -388,8 +388,9 @@ using ulonglong16 = vec<unsigned long long, 16>;
 #define HIPSYCL_BUILTIN_GENERATOR_TRINARY_T_T_T(T, name, impl_name)            \
   HIPSYCL_BUILTIN T name(T a, T b, T c) noexcept {                             \
     if constexpr (std::is_arithmetic_v<T>) {                                   \
-      return impl_name(detail::data_element(a, 0), detail::data_element(b, 0), \
-                       detail::data_element(c, 0));                            \
+      return static_cast<T>(impl_name(detail::data_element(a, 0),              \
+                                      detail::data_element(b, 0),              \
+                                      detail::data_element(c, 0)));            \
     } else {                                                                   \
       T result;                                                                \
       for (int i = 0; i < detail::builtin_type_traits<T>::num_elements; ++i) { \
@@ -405,8 +406,8 @@ using ulonglong16 = vec<unsigned long long, 16>;
 #define HIPSYCL_BUILTIN_GENERATOR_BINARY_T_T(T, name, impl_name)               \
   HIPSYCL_BUILTIN T name(T a, T b) noexcept {                                  \
     if constexpr (std::is_arithmetic_v<T>) {                                   \
-      return impl_name(detail::data_element(a, 0),                             \
-                       detail::data_element(b, 0));                            \
+      return static_cast<T>(impl_name(detail::data_element(a, 0),              \
+                                      detail::data_element(b, 0)));            \
     } else {                                                                   \
       T result;                                                                \
       for (int i = 0; i < detail::builtin_type_traits<T>::num_elements; ++i) { \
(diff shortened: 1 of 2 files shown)


include/hipSYCL/sycl/libkernel/vec.hpp
```diff
@@ -182,6 +182,21 @@ template <class Vector_type, class Function>
...
+
+template <typename Arg, typename T>
+struct get_size<Arg, T,
+                std::void_t<decltype(Arg::size())>,
+                std::enable_if_t<std::is_same_v<typename Arg::element_type, T>>> {
+  static constexpr int size = Arg::size();
+};
+
+template <typename Arg, typename T>
+constexpr int count_num_elements = get_size<Arg, T>::size;
```
illuhad, line 205: Is a `static constexpr` member function not sufficient here?

include/hipSYCL/sycl/libkernel/builtins.hpp
```diff
@@ -507,7 +511,10 @@ using ulonglong16 = vec<unsigned long long, 16>;
   typename detail::builtin_type_traits<T>::alternative_data_type<int64_t> name(\
       T a) noexcept {                                                          \
     if constexpr (std::is_arithmetic_v<T>) {                                   \
-      return impl_name(detail::data_element(a, 0));                            \
+      return static_cast<                                                      \
+          typename detail::builtin_type_traits<T>                              \
+                         ::alternative_data_type<int64_t>>                     \
+                             (impl_name(detail::data_element(a, 0)));          \
```
illuhad, line 517: Are the changes in this file related?

illuhad (commented):
Good catch! I wonder if the recently `marray` class has the same problem, since it has a similar constructor?

include/hipSYCL/sycl/libkernel/vec.hpp
nilsfriess: For some reason, I couldn't make it work with a `static constexpr` member function (and I don't really understand why, that's one reason why I opened it as a draft PR for now) but I'll check that again.

include/hipSYCL/sycl/libkernel/builtins.hpp
nilsfriess: Yes, and I only half understand why. Without the explicit casts, the compiler complains when `T` is a `sycl::vec`: in this case, it does not know how to convert from the return type of `impl_name` (for example `double`) to `T` (for example `sycl::vec<double, 2>`). The `vec(const Args &...args)` constructor is now not applicable anymore because one `double` is not enough for a `sycl::vec` with two elements and the `vec(const T &value)` constructor is not applicable because it is marked explicit. 

The funny thing is that this happens in the `if constexpr` branch that will never be executed for `T == sycl::vec` because `sycl::vec` is not an arithmetic type. However, the C++ standard apparently requires both branches in an `if constexpr` to be compilable if the `if constexpr` is not inside a template (not sure what the reason for this is). 

And by explicitly casting here, the `explicit` constructor of the `vec` class is now a possible option and the compiler is happy (even though it will never call this constructor).

include/hipSYCL/sycl/libkernel/vec.hpp
nilsfriess: A `static constexpr` function would work but not with the `if constexpr` as before, for a similar reason as in my other comment: when both branches of a `if constexpr` inside a function that is used in a template parameter fail, then this is not just a substitution failure but will cause a compilation error, so you can't really use `if constexpr` for SFINAE (at least that's how I understand it, I'm not really 100% sure).

nilsfriess:
Solving the zero-dim accessor ambiguity issues discussed earlier are now solved as follows: If the `vec(Args...)` constructor is passed a type that is neither a scalar type nor a class that has both `element_type` and a `size()` method (for example, a zero-dimensional accessor), this does not cause a compile error but these types are just given a size -1. This way, the `count_num_elements` variable will never equal `N` in this case, so SFINAE can do its job.
So we don't really need to check if the `Args` are convertible to `T` explicitly, giving them a size that doesn't make sense is enough.

illuhad (approved):

PR #957, opened 2023-02-28
Add half int constructor

### LINKED ISSUE #956: Ambiguous type conversion in sycl::half
**Bug summary**
When running the tests in SYCL-CTS, as well as [while building OneMKL](https://github.com/normallytangent/oneMKL/actions/runs/4282543086/jobs/7456952634) with Open SYCL with CPU backend, I get an error of the following type.

```
Users/vsanc/opensycl/build/install/include/sycl/../hipSYCL/sycl/libkernel/half.hpp:50:11: error: call to 'create' is ambiguous
  : _data{fp16::create(f)} {}
```
If I understand correctly, at least inside SYCL-CTS, it breaks because of the [implicit cast of `int`](https://github.com/KhronosGroup/SYCL-CTS/blob/be5b4853fee384555979ffb179654bb34390ea5f/tests/common/common.h#L249), also I do not see a matching constructor in `half.hpp` as of #885.

normallytangent:
Fixes #956


include/hipSYCL/sycl/libkernel/detail/half_representation.hpp
```diff
@@ -125,6 +127,10 @@ inline half_storage create(double f) noexcept {
   return truncate_from(f);
 }
 
+inline half_storage create(int i) noexcept {
+  return i;
+}
```
illuhad: Do we really need changes in `half_storage`? Would it not be sufficient to have a `half(int x)` constructor that just does `fp16::create(static_cast<float>(x))`?

include/hipSYCL/sycl/libkernel/half.hpp
```diff
@@ -65,6 +68,10 @@ class half {
     return fp16::promote_to_double(_data);
   }
 
+  operator int() const {
+    return fp16::promote_to_int(_data);
```
illuhad: `return static_cast<int>(fp16::promote_to_float(_data))`?

I think turning a half into an int is not really a promotion, it's a conversion.

include/hipSYCL/sycl/libkernel/detail/half_representation.hpp
normallytangent: True, we don't need it. I was making changes by blindly following the errors.

include/hipSYCL/sycl/libkernel/detail/half_representation.hpp
```diff
@@ -116,6 +115,9 @@ inline double promote_to_double(half_storage h) noexcept {
   return static_cast<double>(promote_to_float(h));
 }
 
+inline int promote_to_int(half_storage h) noexcept {
```
illuhad: This function is no longer used, and can be removed, no?

illuhad (approved):
Thanks, looks good! For the issue around the explicit constructors, you can either address this in this PR or in a subsequent PR - whatever you prefer :)

PR #965, opened 2023-03-07
add clz builtin

tdavidcl:
Hi,

I've tried to implement the `clz` builtin in OpenSYCL by mimicking the way it was done for `mul24`. 

The part where I'm unsure whether the definition is correct are : 

```c++
HIPSYCL_DEFINE_BUILTIN(clz, HIPSYCL_BUILTIN_OVERLOAD_SET_GENINTEGER,
                       HIPSYCL_BUILTIN_GENERATOR_UNARY_T)
```
and 
```c++
template<class T>
HIPSYCL_BUILTIN T __hipsycl_clz(T x) noexcept {
  HIPSYCL_RETURN_DISPATCH_BUILTIN(__hipsycl_clz, x);
}
```

Also do you have a mechanism in place to check the correctness of builtin functions ?


illuhad:
Nice, more awesome PRs! :D

> HIPSYCL_DEFINE_BUILTIN(clz, HIPSYCL_BUILTIN_OVERLOAD_SET_GENINTEGER,
>                       HIPSYCL_BUILTIN_GENERATOR_UNARY_T)

This looks plausible to me.

> template<class T>
> HIPSYCL_BUILTIN T __hipsycl_clz(T x) noexcept {
>  HIPSYCL_RETURN_DISPATCH_BUILTIN(__hipsycl_clz, x);
> }

That too.

> Also do you have a mechanism in place to check the correctness of builtin functions ?

You can add a test case. We do test builtin correctness for math functions here: https://github.com/OpenSYCL/OpenSYCL/blob/develop/tests/sycl/math.cpp
But we don't yet have a similar file for integer builtins. You can add it :)

There seems to be an issue with the current `enable_if` logic in the host path. Maybe it is easier just using overloads?

src/libkernel/sscp/ptx/integer.cpp
```diff
@@ -38,3 +38,19 @@ HIPSYCL_SSCP_BUILTIN __hipsycl_int32 __hipsycl_sscp_mul24_s32(__hipsycl_int32 a,
 HIPSYCL_SSCP_BUILTIN __hipsycl_uint32 __hipsycl_sscp_mul24_u32(__hipsycl_uint32 a, __hipsycl_uint32 b) {
   return __nv_umul24(a, b);
 }
+
+extern "C" __hipsycl_int32 __nv_clz(__hipsycl_int32);
+extern "C" __hipsycl_int64 __nv_clzll(__hipsycl_int64);
```
illuhad: Do we actually need that declaration here? I think you could just include https://github.com/OpenSYCL/OpenSYCL/blob/develop/include/hipSYCL/sycl/libkernel/sscp/builtins/ptx/libdevice.hpp
which contains declarations for the full functionality of NVIDIA libdevice, including clz.

include/hipSYCL/sycl/libkernel/sscp/builtins.hpp
```diff
@@ -405,6 +405,28 @@ HIPSYCL_BUILTIN T __hipsycl_clamp(T x, T minval, T maxval) noexcept {
     sscp_builtins::__hipsycl_max(x, minval), maxval);
 }
 
+
+
+template<class T,std::enable_if_t<(std::is_same_v<T, __hipsycl_uint8> || std::is_same_v<T, __hipsycl_int8>), int> = 0>
+HIPSYCL_BUILTIN T __hipsycl_clz(T x) noexcept {
+  return __hipsycl_sscp_clz_u8(x);
```
illuhad: Your sscp builtin implementations operate on and return unsigned types in all cases. So I don't think you can treat them in the same code path, as you would need `static_cast` to convert between them.

Also, `__hipsycl_*` types are mainly intended for low-level stuff that enters our bitcode libraries. This stuff is still regular C++, so you can use normal C++ types here :)

tdavidcl: I've change it to :

```c++
template <class T,
          std::enable_if_t<
              (std::is_same_v<T, unsigned int> || std::is_same_v<T, int> ||
               std::is_same_v<T, unsigned short> || std::is_same_v<T, short> ||
               std::is_same_v<T, unsigned char> ||
               std::is_same_v<T, signed char> || std::is_same_v<T, char>),
              int> = 0>
HIPSYCL_BUILTIN T __hipsycl_clz(T x) noexcept {
  return __hipsycl_sscp_clz_u32(static_cast<__hipsycl_uint32>(x));
}

template <class T, std::enable_if_t<(std::is_same_v<T, unsigned long long> ||
                                     std::is_same_v<T, long long>),
                                    int> = 0>
HIPSYCL_BUILTIN T __hipsycl_clz(T x) noexcept {
  return __hipsycl_sscp_clz_u64(static_cast<__hipsycl_uint64>(x));
}
```

tdavidcl:
> There seems to be an issue with the current `enable_if` logic in the host path. Maybe it is easier just using overloads?

I've changed the declaration now the checks passes on my computer at least

Currently on the host side I use a declaration like this

```c++

template <class T,
          std::enable_if_t<
              (std::is_same_v<T, unsigned int> || std::is_same_v<T, int> ||
               std::is_same_v<T, unsigned short> || std::is_same_v<T, short> ||
               std::is_same_v<T, unsigned char> ||
               std::is_same_v<T, signed char> || std::is_same_v<T, char>),
              int> = 0>
HIPSYCL_BUILTIN T __hipsycl_clz(T x) noexcept {
  return __builtin_clz(x);
}

template <class T, std::enable_if_t<(std::is_same_v<T, unsigned long> ||
                                     std::is_same_v<T, long>),
                                    int> = 0>
HIPSYCL_BUILTIN T __hipsycl_clz(T x) noexcept {
  return __builtin_clzl(x);
}

template <class T, std::enable_if_t<(std::is_same_v<T, unsigned long long> ||
                                     std::is_same_v<T, long long>),
                                    int> = 0>
HIPSYCL_BUILTIN T __hipsycl_clz(T x) noexcept {
  return __builtin_clzll(x);
}
```

which does work but is rather long. Would this be ok ?

illuhad:
I feel like this could be simplified. Would maybe something like this work?
```c++
template<class T>
T __hipsycl_clz(T x) {
  if constexpr(std::is_same_v<T, long long> || std::is_same_v<T, unsigned long long>)
    return __builtin_clzll(x);
  else if constexpr(std::is_same_v<T, long> || std::is_same_v<T unsigned long>)
    return __builtin_clzl(x);
  else
   return __builtin_clz(x); // do we need static_cast<T>(...) for the return statements?
}
```

illuhad:
CI says nvc++ is not happy:

```c++

"/home/runner/work/OpenSYCL/OpenSYCL/build/install/bin/../include/CL/../hipSYCL/sycl/libkernel/builtin_interface.hpp", line 534: error: no instance of overloaded function "hipsycl::sycl::detail::hiplike_builtins::__hipsycl_clz" matches the argument list
            argument types are: (char)
    HIPSYCL_RETURN_DISPATCH_BUILTIN(__hipsycl_clz, x);
    ^
          detected during instantiation of "T hipsycl::sycl::detail::__hipsycl_clz(T) noexcept [with T=char]" at line 788 of "/home/runner/work/OpenSYCL/OpenSYCL/build/install/bin/../include/CL/../hipSYCL/sycl/libkernel/builtins.hpp"

"/home/runner/work/OpenSYCL/OpenSYCL/build/install/bin/../include/CL/../hipSYCL/sycl/libkernel/builtin_interface.hpp", line 534: error: no instance of overloaded function "hipsycl::sycl::detail::hiplike_builtins::__hipsycl_clz" matches the argument list
            argument types are: (signed char)
    HIPSYCL_RETURN_DISPATCH_BUILTIN(__hipsycl_clz, x);
    ^
          detected during instantiation of "T hipsycl::sycl::detail::__hipsycl_clz(T) noexcept [with T=signed char]" at line 788 of "/home/runner/work/OpenSYCL/OpenSYCL/build/install/bin/../include/CL/../hipSYCL/sycl/libkernel/builtins.hpp"

"/home/runner/work/OpenSYCL/OpenSYCL/build/install/bin/../include/CL/../hipSYCL/sycl/libkernel/builtin_interface.hpp", line 534: error: no instance of overloaded function "hipsycl::sycl::detail::hiplike_builtins::__hipsycl_clz" matches the argument list
            argument types are: (unsigned char)
    HIPSYCL_RETURN_DISPATCH_BUILTIN(__hipsycl_clz, x);
    ^
          detected during instantiation of "T hipsycl::sycl::detail::__hipsycl_clz(T) noexcept [with T=unsigned char]" at line 788 of "/home/runner/work/OpenSYCL/OpenSYCL/build/install/bin/../include/CL/../hipSYCL/sycl/libkernel/builtins.hpp"

"/home/runner/work/OpenSYCL/OpenSYCL/build/install/bin/../include/CL/../hipSYCL/sycl/libkernel/builtin_interface.hpp", line 534: error: no instance of overloaded function "hipsycl::sycl::detail::hiplike_builtins::__hipsycl_clz" matches the argument list
            argument types are: (short)
    HIPSYCL_RETURN_DISPATCH_BUILTIN(__hipsycl_clz, x);
    ^
          detected during instantiation of "T hipsycl::sycl::detail::__hipsycl_clz(T) noexcept [with T=short]" at line 788 of "/home/runner/work/OpenSYCL/OpenSYCL/build/install/bin/../include/CL/../hipSYCL/sycl/libkernel/builtins.hpp"

"/home/runner/work/OpenSYCL/OpenSYCL/build/install/bin/../include/CL/../hipSYCL/sycl/libkernel/builtin_interface.hpp", line 534: error: no instance of overloaded function "hipsycl::sycl::detail::hiplike_builtins::__hipsycl_clz" matches the argument list
            argument types are: (unsigned short)
    HIPSYCL_RETURN_DISPATCH_BUILTIN(__hipsycl_clz, x);
    ^
          detected during instantiation of "T hipsycl::sycl::detail::__hipsycl_clz(T) noexcept [with T=unsigned short]" at line 788 of "/home/runner/work/OpenSYCL/OpenSYCL/build/install/bin/../include/CL/../hipSYCL/sycl/libkernel/builtins.hpp"

"/home/runner/work/OpenSYCL/OpenSYCL/build/install/bin/../include/CL/../hipSYCL/sycl/libkernel/builtin_interface.hpp", line 534: error: no instance of overloaded function "hipsycl::sycl::detail::hiplike_builtins::__hipsycl_clz" matches the argument list
            argument types are: (long)
    HIPSYCL_RETURN_DISPATCH_BUILTIN(__hipsycl_clz, x);
    ^
          detected during instantiation of "T hipsycl::sycl::detail::__hipsycl_clz(T) noexcept [with T=long]" at line 788 of "/home/runner/work/OpenSYCL/OpenSYCL/build/install/bin/../include/CL/../hipSYCL/sycl/libkernel/builtins.hpp"

"/home/runner/work/OpenSYCL/OpenSYCL/build/install/bin/../include/CL/../hipSYCL/sycl/libkernel/builtin_interface.hpp", line 534: error: no instance of overloaded function "hipsycl::sycl::detail::hiplike_builtins::__hipsycl_clz" matches the argument list
            argument types are: (unsigned long)
    HIPSYCL_RETURN_DISPATCH_BUILTIN(__hipsycl_clz, x);
    ^
          detected during instantiation of "T hipsycl::sycl::detail::__hipsycl_clz(T) noexcept [with T=unsigned long]" at line 788 of "/home/runner/work/OpenSYCL/OpenSYCL/build/install/bin/../include/CL/../hipSYCL/sycl/libkernel/builtins.hpp"

```

include/hipSYCL/sycl/libkernel/host/builtins.hpp
```diff
@@ -528,6 +528,31 @@ HIPSYCL_BUILTIN T __hipsycl_clamp(T x, T minval, T maxval) noexcept {
...
 
+template <class T,
+          std::enable_if_t<
+              (std::is_same_v<T, unsigned int> || std::is_same_v<T, int> ||
+               std::is_same_v<T, unsigned short> || std::is_same_v<T, short> ||
+               std::is_same_v<T, unsigned char> ||
+               std::is_same_v<T, signed char> || std::is_same_v<T, char>),
+              int> = 0>
+HIPSYCL_BUILTIN T __hipsycl_clz(T x) noexcept {
+  return __builtin_clz(x);
```
illuhad: It might also be a good idea to use:
```
#if __has_builtin(__builtin_clz)
  (use builtin)
#else
 (some slow fallback)
#endif
```
I don't know how widespread those builtins are outside of Linux.

tdavidcl: I've changed it to 

```c++
  #if __has_builtin(__builtin_clz)
    return __builtin_clz(x);
  #else
    // mbit = 1000 0000 ...
    constexpr T mBit = 1 << ((CHAR_BIT*sizeof(T)) - 1);
    T count = 0;
    while (!(x & mBit))
    {
      x = (x << 1);
      count ++;
    }
    return count;
  #endif
```

tdavidcl:
> CI says nvc++ is not happy:

actually after checking the hip/cuda doc specify `clz` for 32 bit ints and `clzll`  for 64 bit ints.

So I've changed it to :

```c++
template<class T>
HIPSYCL_HIPLIKE_BUILTIN T __hipsycl_clz(T x) noexcept {

  // use __clzll or __clz by checking the bit lenght because
  // the nvidia/hip documentation mention clz as 32 bits and clzll as 64
  
  if constexpr (sizeof(T)*CHAR_BIT == 64){
    return __clzll(static_cast<__hipsycl_int64>(x));
  }

  return __clz(static_cast<__hipsycl_int32>(x));
}
```

It should also solve the instantiation issues.

include/hipSYCL/sycl/libkernel/generic/hiplike/builtins.hpp
```diff
@@ -396,6 +396,19 @@ HIPSYCL_HIPLIKE_BUILTIN T __hipsycl_clamp(T x, T minval, T maxval) noexcept {
     hiplike_builtins::__hipsycl_max(x, minval), maxval);
 }
 
+template<class T>
+HIPSYCL_HIPLIKE_BUILTIN T __hipsycl_clz(T x) noexcept {
+
+  // use __clzll or __clz by checking the bit lenght because
+  // the nvidia/hip documentation mention clz as 32 bits and clzll as 64
+  
+  if constexpr (sizeof(T)*CHAR_BIT == 64){
```
illuhad: `sizeof(T) == 8`?

include/hipSYCL/sycl/libkernel/generic/hiplike/builtins.hpp
```diff
@@ -396,6 +396,19 @@ HIPSYCL_HIPLIKE_BUILTIN T __hipsycl_clamp(T x, T minval, T maxval) noexcept {
...
+    return __clzll(static_cast<__hipsycl_int64>(x));
+  }
+
+  return __clz(static_cast<__hipsycl_int32>(x));
```
illuhad, line 430: Better put this in the else branch, so that it actually gets removed when the constexpr condition is true.

include/hipSYCL/sycl/libkernel/host/builtins.hpp
```diff
@@ -528,6 +528,67 @@ HIPSYCL_BUILTIN T __hipsycl_clamp(T x, T minval, T maxval) noexcept {
...
+  #else
+    // mbit = 1000 0000 ...
+    constexpr T mBit = 1 << ((CHAR_BIT*sizeof(T)) - 1);
+    T count = 0;
+    while (!(x & mBit))
+    {
+      x = (x << 1);
+      count ++;
+    }
+    return count;
```
illuhad: Can we have this in some helper function e.g. in `detail` subnamespace that can just be called here, instead of repeating this logic in every definition of the function?

include/hipSYCL/sycl/libkernel/host/builtins.hpp
```diff
@@ -528,6 +528,67 @@ HIPSYCL_BUILTIN T __hipsycl_clamp(T x, T minval, T maxval) noexcept {
...
+               std::is_same_v<T, unsigned short> || std::is_same_v<T, short> ||
+               std::is_same_v<T, unsigned char> ||
+               std::is_same_v<T, signed char> || std::is_same_v<T, char>),
+              int> = 0>
+HIPSYCL_BUILTIN T __hipsycl_clz(T x) noexcept {
+  #if __has_builtin(__builtin_clz)
+    return __builtin_clz(x);
+  #else
+    // mbit = 1000 0000 ...
+    constexpr T mBit = 1 << ((CHAR_BIT*sizeof(T)) - 1);
```
illuhad: `m_bit` for consistency with the naming style otherwise used in this file?

include/hipSYCL/sycl/libkernel/generic/hiplike/builtins.hpp
tdavidcl: Technically the standard doesn't enforce a char to be 8 bits but this one is guaranteed to be correct `sizeof(T)*CHAR_BIT`. I don't know if that's the case for any architectures that may be of use tho.

include/hipSYCL/sycl/libkernel/host/builtins.hpp
tdavidcl: For now I've put it in `hipsycl::sycl::detail::host_builtins` next to the others clz, is that what you meant ?

```c++
template<class T, std::enable_if_t<std::is_integral_v<T>,int> = 0>
inline T __safe_clz(T x) noexcept {
  // m_bit = 1000 0000 ...
  constexpr T m_bit = 1 << ((CHAR_BIT*sizeof(T)) - 1);
  T count = 0;
  while (!(x & m_bit))
  {
    x = (x << 1);
    count ++;
  }
  return count;
}
```

include/hipSYCL/sycl/libkernel/host/builtins.hpp
```diff
@@ -528,6 +529,56 @@ HIPSYCL_BUILTIN T __hipsycl_clamp(T x, T minval, T maxval) noexcept {
   return std::min(std::max(x, minval), maxval);
 }
 
+template<class T, std::enable_if_t<std::is_integral_v<T>,int> = 0>
+inline T __fallback_clz(T x) noexcept {
```
illuhad: If starting with `__`, we should probably prefix with `__hipsycl`. Underscore names are reserved in C++ for the compiler implementation, and we'd like to avoid trouble with the compiler :-) But we could also just not have `__` here, as it's just a host helper function, not an actual builtin.

include/hipSYCL/sycl/libkernel/generic/hiplike/builtins.hpp
illuhad: That's technically true, but I don't think any architecture that we target has non-8 bit char width. I think we also assume `CHAR_BIT==8` at other places. And likely heterogeneous compute will fall completely apart if the device does not have `CHAR_BIT` of 8 :-)
It's okay for me if you prefer it this way, just want to let you know that assuming `CHAR_BIT==8` is a legitimate option if it can simplify code :-)

include/hipSYCL/sycl/libkernel/generic/hiplike/builtins.hpp
```diff
@@ -396,6 +396,20 @@ HIPSYCL_HIPLIKE_BUILTIN T __hipsycl_clamp(T x, T minval, T maxval) noexcept {
     hiplike_builtins::__hipsycl_max(x, minval), maxval);
 }
 
+template<class T>
+HIPSYCL_HIPLIKE_BUILTIN T __hipsycl_clz(T x) noexcept {
+
+  // use __clzll or __clz by checking the bit lenght because
```
keryell:
```suggestion
  // use __clzll or __clz by checking the bit length because
```

include/hipSYCL/sycl/libkernel/generic/hiplike/builtins.hpp
keryell: At some point, if you want to run for example on a Zuse 3 or its emulator, you might have 22-bit characters for example. ;-)
```suggestion
  if constexpr (sizeof(T)*CHAR_BIT > 32) {
```
to handle an Honeywell GCOS 6 with 36 bit integers. ;-)

illuhad:
> Also do you have a mechanism in place to check the correctness of builtin functions ?

Can you confirm whether you intend to also add a test case, or if this PR should be considered for final review/testing as is?

tdavidcl:
I think I won't be able to work on this one for a few weeks, so I won't be able to add testing to this as of now... 

> Can you confirm whether you intend to also add a test case, or if this PR should be considered for final review/testing as is?

So do you prefer to wait until i add some test case or do you want to review it now ?

include/hipSYCL/sycl/libkernel/sscp/builtins.hpp
```diff
@@ -405,6 +405,26 @@ HIPSYCL_BUILTIN T __hipsycl_clamp(T x, T minval, T maxval) noexcept {
...
 }
 
+
+template <class T,
+          std::enable_if_t<
+              (std::is_same_v<T, unsigned int> || std::is_same_v<T, int> ||
+               std::is_same_v<T, unsigned short> || std::is_same_v<T, short> ||
+               std::is_same_v<T, unsigned char> ||
+               std::is_same_v<T, signed char> || std::is_same_v<T, char>),
+              int> = 0>
```
illuhad, line 427: `std::is_integral_v<T> && !std::is_signed_v<T> && sizeof(T) <= 4` ?

src/libkernel/sscp/amdgpu/integer.cpp
```diff
@@ -37,3 +37,16 @@ HIPSYCL_SSCP_BUILTIN __hipsycl_int32 __hipsycl_sscp_mul24_s32(__hipsycl_int32 a,
 HIPSYCL_SSCP_BUILTIN __hipsycl_uint32 __hipsycl_sscp_mul24_u32(__hipsycl_uint32 a, __hipsycl_uint32 b) {
   return __ockl_mul24_u32(a, b);
 }
+
+HIPSYCL_SSCP_BUILTIN __hipsycl_uint8 __hipsycl_sscp_clz_u8(__hipsycl_uint8 a){
+  return __ockl_clz_u8(a);
+}
+HIPSYCL_SSCP_BUILTIN __hipsycl_uint16 __hipsycl_sscp_clz_u16(__hipsycl_uint16 a){
+  return __ockl_clz_u16(a);
+}
```
illuhad, line 46: Why do we have u8 and u16 implmentations here? The header does not define these builtins, and the SYCL interface does not use them. If there is some benefit to being able to use the OCKL u8 and u16 functions, we should expose 8 and 16-bit versions and use them in the SYCL headers.

src/libkernel/sscp/spirv/integer.cpp
```diff
@@ -37,4 +37,23 @@ HIPSYCL_SSCP_BUILTIN __hipsycl_int32 __hipsycl_sscp_mul24_s32(__hipsycl_int32 a,
...
+__hipsycl_int32 __spirv_ocl_clz(__hipsycl_int32 a);
+__hipsycl_int64 __spirv_ocl_clz(__hipsycl_int64 a);
+__hipsycl_uint32 __spirv_ocl_clz(__hipsycl_uint32 a);
+__hipsycl_uint64 __spirv_ocl_clz(__hipsycl_uint64 a);
+
+HIPSYCL_SSCP_BUILTIN __hipsycl_uint8 __hipsycl_sscp_clz_u8(__hipsycl_uint8 a){
+  return __spirv_ocl_clz(a);
+}
+HIPSYCL_SSCP_BUILTIN __hipsycl_uint16 __hipsycl_sscp_clz_u16(__hipsycl_uint16 a){
+  return __spirv_ocl_clz(a);
```
illuhad: Same here?

illuhad:
> So do you prefer to wait until i add some test case or do you want to review it now ?

I'll start reviewing now and merge once tests are there too :-)

tdavidcl:
I've added the test, and cleaned the clz implementation. 
The tests pass on omp mode, but i don't have hardware/config to test on amd & nvidia right now.

illuhad:
Thank you, I have just merged our new self-hosted CI with NVIDIA and AMD GPUs - so if you rebase on current `develop`, the tests will be run on GPUs :)

tdavidcl:
I fixed the typo for nvcc, but i'm suprised that nvcc doesn't have `__builtin_clz`, would `__clz`,  `__clzll` work since it is compiling some CUDA ?

tdavidcl:
I'm pushing a test to try using the using the __clz builtin in nvcc in host mode, if it does not work i will stick with the previous version now that all tests are passing

illuhad:
> I fixed the typo for nvcc, but i'm suprised that nvcc doesn't have __builtin_clz, would __clz, __clzll work since it is compiling some CUDA ?

We don't support nvcc, only nvc++ :-) (those are two very different compilers).
I'm not sure if the CUDA intrinsic would work in the host path - most likely not, so we'd have to use the fallback clz.

tdavidcl:
> I fixed the typo for nvcc, but i'm suprised that nvcc doesn't have __builtin_clz, would __clz, __clzll work since it is compiling some CUDA ?
> 
> We don't support nvcc, only nvc++ :-) (those are two very different compilers). I'm not sure if the CUDA intrinsic would work in the host path - most likely not, so we'd have to use the fallback clz.

yeah my bad i meant nvc++. The test in math.cpp with nvc++ passes, with the fix, tho i'm not sure if the __clz or fallback_clz version was used

edit : after checking in the doc it says `__device__​ int __clz ( int  x )`, so yes not on the host

tdavidcl:
So i reverted to the previous commit. With the test passing on all config (+ self hosted), is everything ok for final review ?

include/hipSYCL/sycl/libkernel/host/builtins.hpp
```diff
@@ -533,30 +533,11 @@ HIPSYCL_BUILTIN T __hipsycl_clamp(T x, T minval, T maxval) noexcept {
 template<class T, std::enable_if_t<std::is_integral_v<T>,int> = 0>
 inline T fallback_clz(T x) noexcept {
 
-    #if __has_builtin(__clz) && __has_builtin(__clzll)
+  std::bitset<sizeof(T)*CHAR_BIT> bset(x);
+  int idx = 0;
+  while(!bset[sizeof(T)*CHAR_BIT - idx -1]){idx++;}
```
illuhad, line 539: I just noticed, what happens if x is all bits set to 0? Would this then turn into an infinite loop, potentially accessing invalid memory?

tdavidcl: yes actually ... 
i've changed the loop to
```c++
  while(!bset[sizeof(T)*CHAR_BIT - idx -1] && idx<sizeof(T)*CHAR_BIT){idx++;}
```
https://godbolt.org/z/K61x6PvhG

illuhad (approved):
Thanks!

illuhad:
Hm.. it seems that there might still be an issue on CPU: https://github.com/OpenSYCL/OpenSYCL/actions/runs/4883593430/jobs/8716435004?pr=965

tdavidcl:
Yes, but i can not reproduce the issue ... 
The previous version with the issue in the null case was working, so i've rolled back to it with an additional check on `x==0`

https://godbolt.org/z/ToKsdoTKM

illuhad:
The only potential issue I see is that it would probably evalute `!bset[sizeof(T)*CHAR_BIT - idx -1]` before it would evaluate the `idx<sizeof(T)*CHAR_BIT`. So in the step when the second condition evaluates to false, it could happen that it accessed the invalid memory one time (thus triggering UB) for the first condition before finding out that it should not have done that.

Let's see if it works now :)

tdavidcl:
it's odd that the macos test doesn't work, it was ok here : https://github.com/OpenSYCL/OpenSYCL/actions/runs/4883180762/jobs/8714601525

and the only difference is this line : 
```c++
if(x==0){return sizeof(T)*CHAR_BIT;}
```

and the test pass on my MAC (M1) with appleClang :'(

illuhad:
Mac also failed previously, before the fix. So I wonder why changing to `if(x==0)` still has not resolved the issue entirely... For reference, this is how Mac CI works: https://github.com/OpenSYCL/OpenSYCL/blob/develop/.github/workflows/macos.yml
If you have a Mac, can you try following these commands to see if you can reproduce? Or maybe try address sanitizer to see if it's still doing some invalid access.

tdavidcl:
it is still correct on my Mac, I've added bunch of print statements, can you try running it only the Mac OS workflow?

illuhad:
I've invited you into the OpenSYCL organization; the workflows will then run automatically for you.

tdavidcl:
> I've invited you into the OpenSYCL organization; the workflows will then run automatically for you.

Oh thanks a lot, it will speed up the process for sure ^^

tdavidcl:
ok I found the issue : in the failing tests you have this ...

```
castvar = 0
__builtin_clz(castvar) = 73911 (unsigned int)
__builtin_clz(0) = 73896 (unsigned int)
```

the clz builtin looks broken in the Mac OS workflow, also it does not look consistent and hard to check ...

also I found this : https://stackoverflow.com/questions/19527897/how-undefined-are-builtin-ctz0-or-builtin-clz0
love it ...

illuhad:
Good find!

So basically we should always
```
if(x == 0)
  return sizeof(T)*CHAR_BIT
else
  return __builtin_clz(x)
```

whenever using the builtin?

I do wonder though when this case is so undefined, whether the SYCL specification even mandates a specific result here? I can kind of see that the concept of leading or trailing zeros falls apart somewhat in the case of a value of 0.

Good thing we noticed this edge case :)

tdavidcl:
The SYCL standard actually says : 

```
geninteger clz(geninteger x)

Returns the number of leading 0-bits in x, starting at the most significant bit position. 
If x is 0, returns the size in bits of the type of x or component type of x, if x is a vector type.
```
So I've added the check for __builtin_clz

In cuda : 

```
__device__ ​ int __clz ( int  x )
Return the number of consecutive high-order zero bits in a 32-bit integer.
```
so this one is fine 

I'm not 100% sure for the others

illuhad:
My guess is that SPIR-V will probably be aligned with OpenCL, which is probably aligned with SYCL. No idea for AMD. So maybe also add the check for those two backends to be sure?

illuhad:
... Or maybe just add the check to the high-level builtin interface, so that the backends don't need to be concerned about this case anymore?

tdavidcl:
> My guess is that SPIR-V will probably be aligned with OpenCL, which is probably aligned with SYCL. No idea for AMD. So maybe also add the check for those two backends to be sure?

As you guessed 
From https://registry.khronos.org/SPIR-V/specs/unified1/OpenCL.ExtendedInstructionSet.100.html
```
clz

Returns the number of leading 0 bits in x, starting at the most significant bit position. 
If x is 0, returns the size in bits of the type of x or component type of x, if x is a vector. 
```
So spir-v should be fine 

> ... Or maybe just add the check to the high-level builtin interface, so that the backends don't need to be concerned about this case anymore?

Actually we can just add the check in tests, also I don't know also why in the first place the value was 0 in the inputs, it shouldn't when i'm looking at the test.

tdavidcl:
From RadeonCompute gits:

https://github.com/RadeonOpenCompute/clang/blob/a09d37e345861d68f9768939e485d265f4fcb0ce/include/clang/Basic/TargetInfo.h#L776

Looks like they are aware and check for it :), so looks like we only have to be careful with the host.

PR #966, opened 2023-03-08
rewrite of group algorithms

DieGoldeneEnte:
The group algorithms were restructured and partially rewritten to make use of the known_identity.
Where possible functions are implemented using other group function variants to reduce code duplication.
This PR also implements functions for scoped parallelism V2 implemented using the ND-range variants.

The tests are also rewritten and only check functions that aren't implemented using others.


include/hipSYCL/sycl/libkernel/group_functions.hpp
```diff
@@ -63,272 +62,377 @@
...
-  HIPSYCL_RETURN_DISPATCH_GROUP_ALGORITHM(__hipsycl_exclusive_scan_over_group,
-                                          g, x, init, binary_op);
-}
-
-template<typename Group, typename T, typename BinaryOperation,
-          std::enable_if_t<is_group_v<std::decay_t<Group>>, bool> = true>
-HIPSYCL_BUILTIN
-T exclusive_scan_over_group(Group g, T x, BinaryOperation binary_op) {
-  HIPSYCL_RETURN_DISPATCH_GROUP_ALGORITHM(__hipsycl_exclusive_scan_over_group,
-                                          g, x, binary_op);
```
illuhad, line 221: Where did `exclusive_scan_over_group` without `init` go? The spec says we need to have this overload. I don't see it anymore, am I missing something?

DieGoldeneEnte: I put all backend independent implementations in line 253+. Those are functions for example without init, which can be implemented using the identity and version with init or where an index just needs to be linearized. The exclusive scan without init for example is in line 378.

PR #967, opened 2023-03-08
Change sycl::noinit to sycl::no_init

RaulPPelaez:
I got a debug warning about using `sycl::noinit`, after a while I realized the standard changed the name to sycl::no_init.
I changed the message printed by the warning and a reference to sycl::noinit in the docs.


illuhad (approved):
Thanks for noticing!

PR #969, opened 2023-03-10
Add operators for +,-,*,/ for `half` with other scalar types

nilsfriess:
This also required removing the `explicit` keyword for the constructors and adding three new constructors for `unsigned int`, `unsigned long` and `long`.

### DIFF
--- a/include/hipSYCL/sycl/libkernel/half.hpp
+++ b/include/hipSYCL/sycl/libkernel/half.hpp
@@ -162,18 +162,24 @@ public:
     // operator +,-,*,/ for combinations of half and other types
 #define OP_FOR_TYPE(op, type)                                         \
   friend half operator op(const half lhs, const type rhs) {           \
-    return lhs op half(rhs);                                          \
+    half res = rhs;                                                   \
+    res = lhs op res;                                                 \
+    return res;                                                       \
   }                                                                   \
                                                                       \
-  friend half operator op(const type lhs, const half rhs) {           \
-    return half(lhs) op rhs;                                          \
+  friend type operator op(const type lhs, const half rhs) {           \
+    type res = rhs;                                                   \
+    res = lhs op res;                                                 \
+    return res;                                                       \
   }
 
 #define OP(op)                                                        \
   OP_FOR_TYPE(op, int)                                                \
   OP_FOR_TYPE(op, unsigned int)                                       \
   OP_FOR_TYPE(op, long)                                               \
+  OP_FOR_TYPE(op, long long)                                          \
   OP_FOR_TYPE(op, unsigned long)                                      \
+  OP_FOR_TYPE(op, unsigned long long)                                 \
   OP_FOR_TYPE(op, float)                                              \
   OP_FOR_TYPE(op, double)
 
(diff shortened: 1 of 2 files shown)


illuhad:
Can you comment more on the use case? What is the code you attempt to make work?

nilsfriess:
Things like
```
sycl::half a(1);
float b = 1;
auto res = a * b;
```
currently do not compile because the multiplication is ambiguous (the compiler could first create a `half` from `b` and then use `operator*(const half& a, const half& b)` or convert `a` to `float` and use the built-in `operator*(float, float)`).
There are a few places in the CTS where scalars are multiplied with/divided by `half` and to compile the CTS it would probably suffice to just implement the operators for `float`. I just added the other ones for completeness.

illuhad:
The SYCL WG agreed that `half` should indeed only have the `float` conversion operator: https://github.com/KhronosGroup/SYCL-CTS/issues/559#issuecomment-1481490154

include/hipSYCL/sycl/libkernel/half.hpp
```diff
@@ -64,56 +68,58 @@ class half {
...
-      return __hipsycl_sscp_half_add(a._data, b._data),
-      return fp16::create(__hadd(fp16::as_cuda_half(a._data), fp16::as_cuda_half(b._data))),
+      data = fp16::builtin_add(a._data, b._data),
+      data = __hipsycl_sscp_half_add(a._data, b._data),
+      data = fp16::create(__hadd(fp16::as_cuda_half(a._data), fp16::as_cuda_half(b._data))),
       // HIP uses compiler builtin addition for native _Float16 type
-      return fp16::builtin_add(a._data, b._data),
-      return fp16::builtin_add(a._data, b._data))
+      data = fp16::builtin_add(a._data, b._data),
+      data = fp16::builtin_add(a._data, b._data));
```
illuhad, line 82: Couldn't you simplify this along the lines of:
```c++
return detail::create_half(__hipsycl_backend_switch(
  fp16::builtin_add(a._data, b._data),
  ....
));
```

nilsfriess: Unfortunately not, because `__hipsycl_backend_switch` adds semicolons to its arguments, so this becomes something like 
```
return detail::create_half(fp16::builtin_add(a._data, b._data););
```
which does not compile (this could of course be fixed by removing the semicolons in the definition of `__hipsycl_backend_switch` and adding them in the places where the macro is used). However, for the sscp backend, `__hipsycl_backend_switch` becomes an if statement, which again doesn't compile, and we can't just change that to a ternary operator because sometimes `__hipsycl_backend_switch` is used with return statements as its arguments and these are not allowed as expressions in the ternary operator.

illuhad: Ugh... even if SSCP might in principle work with ternary operator, it probably won't work with nvc++, where we need `if target()` branching.. So you are totally right :(

illuhad:
Can you resolve merge conflicts?

illuhad (approved):

PR #960, opened 2023-03-05
Tweaks and fixes for math built-ins

nmnobre:
Hi,

Simple and self-explanatory :)

Cheers,
-Nuno

PR #975, opened 2023-03-14
[SSCP] Fix sinpi/cospi builtins for SPIR-V targets

al42and:
For PTX and AMDGPU they were fixed in b6171ba0acc3 (#960).


illuhad (approved):
Thanks!

PR #977, opened 2023-03-15
Fix Windows GitHub CI

nilsfriess:
This PR fixes the Windows CI.
There are still a few TODOs (but at least the CPU tests compile):

- [ ] I've commented out [this](https://github.com/nilsfriess/OpenSYCL/blob/85a65f9bc12909cd6af8146bfe2210038534957a/include/hipSYCL/compiler/Frontend.hpp#L259) because the function `setDeviceMangleContext` does not exist in LLVM 15 (I think it was removed in LLVM13). Maybe @fodinabor knows if it is fine to just remove this?
- [ ] ~Compilation of tests with CUDA~ 
- [ ] ~Add more CUDA versions~
- [x] CPU tests are compiled but not executed
- [ ] Upload prebuilt LLVM on Uni servers, not Dropbox

Edit: Just compiling and running the tests for CPU for now


fodinabor:
Re the `setDeviceMangleContext`, I am unsure, if needed. Will try to test with a GPU soon-ish..
But I assume, it will only be necessary for LLVM < 13, since the other macro paths do create the mangle context in place anyways?

illuhad:
Maybe @fodinabor can have a look at this?

illuhad:
@fodinabor are you still planning to look at this? If not it's fine, I'd just like to merge this soonish as it should be a massive quality of life improvement once contributors are no longer spammed with emails about Windows CI failures :-)

fodinabor:
Sorry for the delay...
So it seems that with an LLVM 15 version that does have the NVPTX backend included (not the case for the binary version used in the CI), we can build the tests successfully, and some tests even pass (`device_compilation_tests` e.g.) but running `sycl_tests` still has some issues... so far most issues I found seem to be related to type ABI mismatches on CPU and GPU: `long` on MS ABI is 4 byte, but 8 byte on GPU.
But not sure if that's the only issue.
That being said, I haven't seen any indication so far that there might be mangling issues, so I guess go ahead with merging this?

illuhad:
Thanks for the feedback!

>  so far most issues I found seem to be related to type ABI mismatches on CPU and GPU: long on MS ABI is 4 byte, but 8 byte on GPU.

yikes... That means that basic code has the chance of blowing up, not even talking about data structures.. even `buffer<long>` is not going to work. We should really consider focusing on MinGW..
(SSCP also solves this issue :- ) )

illuhad (approved):

PR #980, opened 2023-03-20
[CI] Add LLVM 16.

fodinabor:
Since LLVM 16 is released now, add it to the CI.

Annoyingly, the packages seem to have broken symlinks, so there are a few extra lines to create working ones. Upstream bug: https://github.com/llvm/llvm-project/issues/61550


illuhad (approved):
Thanks!

PR #982, opened 2023-03-22
Append environment variable for flags in syclcc

### LINKED ISSUE #978: syclcc not taking into account environment variables for flags
**Bug summary**

`syclcc` does not take into account environment variables for flags anymore. e.g. `OPENSYCL_DRYRUN=1 syclcc` does not work. For non-flags it works, e.g. `OPENSYCL_TARGETS=generic syclcc`.

This is probably a regression from #948.

**To Reproduce**

Run  `OPENSYCL_DRYRUN=1 syclcc --help` and observe that `is-dryrun` is not enabled according to the output.

normallytangent:
Fixes #978


illuhad (approved):
Thanks!

PR #989, opened 2023-03-28
Addendum to #832 for CUDA/HIP devices

nmnobre:
Hi Aksel,

While checking what we did in #832, I realised we overlooked a fix for one of the device properties for the HIP backend.
`maxThreadsDim[x]` seems to be the same for all `x`, at least on the MI200s, so the cast is actually the most important bit here... since the multiplication is otherwise performed in 32 bits 🥶.

Cheers,
-N

### DIFF
--- a/src/runtime/cuda/cuda_hardware_manager.cpp
+++ b/src/runtime/cuda/cuda_hardware_manager.cpp
@@ -219,14 +219,16 @@ cuda_hardware_context::get_property(device_uint_property prop) const {
     return _properties->multiProcessorCount;
     break;
   case device_uint_property::max_global_size0:
-    return _properties->maxThreadsDim[0] * _properties->maxGridSize[0];
+    return static_cast<std::size_t>(_properties->maxThreadsDim[0]) *
+                                    _properties->maxGridSize[0];
     break;
   case device_uint_property::max_global_size1:
-    return _properties->maxThreadsDim[1] * _properties->maxGridSize[1];
+    return static_cast<std::size_t>(_properties->maxThreadsDim[1]) *
+                                    _properties->maxGridSize[1];
     break;
   case device_uint_property::max_global_size2:
-    return _properties->maxThreadsDim[2] * _properties->maxGridSize[2];
-    break;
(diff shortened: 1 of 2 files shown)


illuhad:
Thanks! Would we not also need to do the same change for CUDA? I think it has the same code:
https://github.com/OpenSYCL/OpenSYCL/blob/442311b26fcc3669c7acc346361c6bb273ea2b0d/src/runtime/cuda/cuda_hardware_manager.cpp#L222

nmnobre:
> Thanks! Would we not also need to do the same change for CUDA? I think it has the same code:
> 
> https://github.com/OpenSYCL/OpenSYCL/blob/442311b26fcc3669c7acc346361c6bb273ea2b0d/src/runtime/cuda/cuda_hardware_manager.cpp#L222

~~Unnecessary, the multiplication doesn't overflow there... unless you want the code to be literally the same...
Oh, wow, but according to the [docs](https://docs.nvidia.com/cuda/cuda-runtime-api/structcudaDeviceProp.html#structcudaDeviceProp) both operands are actually of `int` type though... How is it working? Can you give it a try? Wrong docs, of course, [this](https://github.com/llvm/llvm-project/blob/faafd26c4d58e1292eeb01179d82a375ed2a47d3/llvm/lib/Target/NVPTX/NVVMIntrRange.cpp#L77-L85) should be more relevant, but still... `unsigned`? That shouldn't be enough, can you give it a try?~~

I've compiled Open SYCL from scratch and it was overflowing on the CUDA path, so I fixed that as well... Thanks.
So weird, I have an older installation where this isn't a problem?... I can't seem to reproduce it though, compiling that same version again gets me an overflow... I'm pretty sure I tested it when we were working on #832 though... Sorry, please check the cast is needed when you have some time... For me, it now seems to be a must, it won't work without it for either CUDA or HIP devices.

illuhad:
Thanks! It might depend on what these queries actually return, i.e. the target device, and potentially driver settings or similar.

I think the CUDA docs were indeed the right ones, not the LLVM docs, since this is a query for the CUDA runtime library. The LLVM nvptx backend should not be involved here :-)

PR #990, opened 2023-03-28
Add `buffer(Container)` constructor

nilsfriess:
This PR adds a missing constructor to the buffer class (see the description in the spec [here](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_buffer_interface)). This constructor allows to construct buffers from containers for which `std::data(container)` and `std::size(container)` are valid, e.g., `std::vector` or `std::array`.

As described by the spec, `write_back` is disabled if `std::data(container)` does return a const pointer, and enabled otherwise. Because of this extra handling, we cannot just delegate to another constructor.

### DIFF
--- a/include/hipSYCL/sycl/buffer.hpp
+++ b/include/hipSYCL/sycl/buffer.hpp
@@ -307,23 +307,23 @@ private:
 };
 
 template <typename, typename = void>
-struct HasData : std::false_type {};
+struct has_data : std::false_type {};
 
 template <typename Container>
-struct HasData<Container, std::void_t<decltype(std::data(std::declval<Container>()))>>
+struct has_data<Container, std::void_t<decltype(std::data(std::declval<Container>()))>>
   : std::true_type {};
 
 template <typename, typename = void>
-struct HasSize : std::false_type {};
+struct has_size : std::false_type {};
 
 template <typename Container>
-struct HasSize<Container, std::void_t<decltype(std::size(Container{}))>>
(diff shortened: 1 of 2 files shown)


include/hipSYCL/sycl/buffer.hpp
```diff
@@ -600,6 +621,49 @@ class buffer : public detail::property_carrying_object
...
+    default_policies dpol;
+
+    // If std::data returns non-const pointer, enable write_back
+    if constexpr (isConstContainer)
+      dpol.writes_back = false;
+    else
+      dpol.writes_back = true;
+    
+    dpol.destructor_waits = true;
+    dpol.use_external_storage = true;
```
illuhad: Are we sure we don't want external storage in any case (i.e. never copy input data to internal storage)? For the pointer-based constructors, we:
* Use external storage and write-back for non-const pointers
* Copy to internal storage and disable write-back for const pointers

nilsfriess: I think I just misunderstood what `use_external_storage` is supposed to be, I hope I got it right now

illuhad (commented):
Thanks, I think it's good now :)

I have only one minor nitpick: In the SYCL headers, we use `CamelCase` only for template arguments, and `snake_case` everywhere else, including type traits or `constexpr` variables - similarly to the C++ standard library.

illuhad (approved):

PR #993, opened 2023-03-30
Add unary +/- operators for `half`

normallytangent:


include/hipSYCL/sycl/libkernel/half.hpp
```diff
@@ -158,6 +158,14 @@ class half {
     return a._data != b._data;
   }
 
+  friend half& operator+(half& a) noexcept {
+    return a;
+  }
+
+  friend half& operator-(half& a) noexcept {
+    return -a;
```
illuhad: This probably casts `a` to float, applies the sign, and converts back to `half`. I wonder if something like this might be more efficient:
```
constexpr half negative = -1.0; // half constructor might have to become `constexpr` for this
return negative * a;
```
Maybe @nilsfriess has an opinion too?

illuhad: Also, is this not just an infinite loop, since you call `half::operator-` inside `half::operator-`?

nilsfriess: Maybe we just directly flip the sign bit in the 16 bit representation? Something like
```
half operator-() {
    constexpr __hipsycl_uint16 sign_mask = 0x8000;
    _data ^= sign_mask;
    return *this;
  }
```

illuhad: Good idea. That would be even better if we are certain that this is the right bitmask :-)

nilsfriess: At least that's what is used in the [fp16 implementation we use](https://github.com/OpenSYCL/OpenSYCL/blob/develop/include/hipSYCL/sycl/libkernel/detail/fp16/fp16.h#L51) (there the 16bit representation is already shifted into the upper part of a 32 bit number but the sign bit that is extracted is the leftmost bit of the number). Just to be completely sure I checked the IEEE754 standard document and that confirms that the sign bit is always the leftmost bit :)

illuhad: Yeah, that adds up :) @normallytangent can you make that change? And we might still want to add a test case :)

normallytangent: Thank you for the inputs, it's taken me some time to understand it but I think I can make the change.  I incorporated @nilsfriess's suggestion. Now I'm working on the corresponding tests

PR #994, opened 2023-03-31
[CBS] Dynamically sized stack arrays.

fodinabor:
Backporting fixes done for upstreaming CBS to POCL.
This is mostly using an llvm::Value as alloca size for the local stack arrays -> in theory we don't have to limit the CPU work-group size to 1024 anymore.
This also allows for larger work-item private arrays, if the actual work-group size is < 1024.

Additionally, some small fixes like the insertion point into the function entry block is now always the terminator, not the first non-phi-or-debug.

Also upstream the nasty uninitialized memory fix for `sycl_dgemm` 😇


illuhad (approved):
Thanks!

PR #995, opened 2023-03-31
[CI] Fix CI for forks still using hipSYCL as repo name 😇 

fodinabor:
Use the `GITHUB_WORKSPACE` variable instead of hardcoding that path.


illuhad (approved):
Thanks!

PR #998, opened 2023-04-04
[CBS] Fix creating wi-loop for barrier-free kernel, if the kernel has…

fodinabor:
… more than one exit.

This e.g. is triggered if there are `assert`s inside the kernel. The barrier case can handle this already.


illuhad (approved):
Thanks!

Issue #1000, opened 2023-04-05
Failed to compile OpenSYCL with LLVM 17


AlbertXingZhang:
**Bug summary**
The following error prevents compiling OpenSYCL with LLVM which is built from its latest git main branch.
```
[1/57] Building CXX object src/compiler/CMakeFiles/opensycl-clang.dir/GlobalsPruningPass.cpp.obj
FAILED: src/compiler/CMakeFiles/opensycl-clang.dir/GlobalsPruningPass.cpp.obj 
[...]

PR #1003, opened 2023-04-05
[LLVM] Fix compat with upstream LLVM

fodinabor:
This fixes a few API / header incompatibilitíes introduced in upstream LLVM.

However, building the tests targeting SSCP is currently still crashing the compiler:
https://github.com/fodinabor/hipSYCL-LLVM-upstream-ci/actions/runs/4618580379/jobs/8166214068#step:11:46
Didn't investigate that too much.
Maybe @illuhad has an intuition about what might go wrong there? 

#1000

### DIFF
--- a/include/hipSYCL/sycl/libkernel/detail/fp16/bitcasts.h
+++ b/include/hipSYCL/sycl/libkernel/detail/fp16/bitcasts.h
@@ -12,7 +12,7 @@ HIPSYCL_UNIVERSAL_TARGET // So that CUDA calls are possible
 static inline float fp32_from_bits(__hipsycl_uint32 w) {
 #if defined(__OPENCL_VERSION__)
 	return as_float(w);
-#elif HIPSYCL_LIBKERNEL_IS_DEVICE_PASS_CUDA == 1
+#elif HIPSYCL_LIBKERNEL_IS_DEVICE_PASS_CUDA == 1 && !defined(HIPSYCL_LIBKERNEL_CUDA_NVCXX)
 	return __uint_as_float((unsigned int) w);
 #elif defined(__INTEL_COMPILER)
 	return _castu32_f32(w);
@@ -31,7 +31,7 @@ HIPSYCL_UNIVERSAL_TARGET // So that CUDA calls are possible
 static inline __hipsycl_uint32 fp32_to_bits(float f) {
 #if defined(__OPENCL_VERSION__)
 	return as_uint(f);
-#elif HIPSYCL_LIBKERNEL_IS_DEVICE_PASS_CUDA == 1
+#elif HIPSYCL_LIBKERNEL_IS_DEVICE_PASS_CUDA == 1 && !defined(HIPSYCL_LIBKERNEL_CUDA_NVCXX)
 	return (__hipsycl_uint32) __float_as_uint(f);
 #elif defined(__INTEL_COMPILER)
 	return _castf32_u32(f);
@@ -50,7 +50,7 @@ HIPSYCL_UNIVERSAL_TARGET // So that CUDA calls are possible
 static inline double fp64_from_bits(__hipsycl_uint64 w) {
 #if defined(__OPENCL_VERSION__)
 	return as_double(w);
-#elif HIPSYCL_LIBKERNEL_IS_DEVICE_PASS_CUDA == 1
+#elif HIPSYCL_LIBKERNEL_IS_DEVICE_PASS_CUDA == 1 && !defined(HIPSYCL_LIBKERNEL_CUDA_NVCXX)
 	return __longlong_as_double((long long) w);
 #elif defined(__INTEL_COMPILER)
 	return _castu64_f64(w);
@@ -69,7 +69,7 @@ HIPSYCL_UNIVERSAL_TARGET // So that CUDA calls are possible
 static inline __hipsycl_uint64 fp64_to_bits(double f) {
 #if defined(__OPENCL_VERSION__)
 	return as_ulong(f);
-#elif HIPSYCL_LIBKERNEL_IS_DEVICE_PASS_CUDA == 1
+#elif HIPSYCL_LIBKERNEL_IS_DEVICE_PASS_CUDA == 1 && !defined(HIPSYCL_LIBKERNEL_CUDA_NVCXX)
 	return (__hipsycl_uint64) __double_as_longlong(f);
 #elif defined(__INTEL_COMPILER)
 	return _castf64_u64(f);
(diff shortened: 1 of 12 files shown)


illuhad:
hm.. not really. We'd need to look at the IR it's actually generating. Maybe it's related to opaque pointers or something like that.

illuhad:
nvc++ is also failing.

include/hipSYCL/sycl/libkernel/detail/fp16/bitcasts.h
```diff
@@ -1,7 +1,7 @@
 #pragma once
-#include "hipSYCL/sycl/libkernel/host/host_backend.hpp"
 #ifndef FP16_BITCASTS_H
 #define FP16_BITCASTS_H
+#include "hipSYCL/sycl/libkernel/host/host_backend.hpp"
```
illuhad: Looking at this again, I don't think this should be included here - in fact, I wonder if it might be responsible for messing up compilation with nvc++ (`libkernel/backend.hpp` should already include host backend). Could this be a preexisting bug in this header?

See the comment here:
https://github.com/OpenSYCL/OpenSYCL/blob/7cff8d4b8ca5d818d67b6fc6a23b9f14b1302e01/include/hipSYCL/sycl/libkernel/backend.hpp#L35

fodinabor: Hmm.. I think the root cause for that ending up in there is the fact that we have a circular dependency here, where sscp_backend.hpp includes the (fp16) builtins and the builtins reference the backend again, but never reaching the definitions of the `HIPSYCL_UNIVERSAL_TARGET` macro.
```
.. include/hipSYCL/sycl/libkernel/backend.hpp
... include/hipSYCL/sycl/libkernel/cuda/cuda_backend.hpp
... include/hipSYCL/sycl/libkernel/hip/hip_backend.hpp
... include/hipSYCL/sycl/libkernel/spirv/spirv_backend.hpp
... include/hipSYCL/sycl/libkernel/sscp/sscp_backend.hpp
.... include/hipSYCL/glue/llvm-sscp/ir_constants.hpp
..... include/hipSYCL/glue/llvm-sscp/s1_ir_constants.hpp
..... include/hipSYCL/glue/llvm-sscp/s2_ir_constants.hpp
.... include/hipSYCL/sycl/libkernel/sscp/builtins/core.hpp
..... include/hipSYCL/sycl/libkernel/sscp/builtins/builtin_config.hpp
...... include/hipSYCL/sycl/libkernel/sscp/builtins/../../memory.hpp
...... include/hipSYCL/sycl/libkernel/detail/half_representation.hpp
....... include/hipSYCL/sycl/libkernel/backend.hpp
....... include/hipSYCL/sycl/libkernel/detail/bit_cast.hpp
....... include/hipSYCL/sycl/libkernel/detail/int_types.hpp
....... include/hipSYCL/sycl/libkernel/detail/fp16/fp16.h
........ include/hipSYCL/sycl/libkernel/detail/fp16/bitcasts.h
......... include/hipSYCL/sycl/libkernel/backend.hpp
```

That being said, I "fixed this" by just moving the includes after the macros in `sscp_backend.hpp`. I think the other changes are still applicable.

fodinabor:
Re SSCP failing with LLVM 17: one assert is triggered upon setting the dummy int value as content: https://github.com/OpenSYCL/OpenSYCL/blob/7cff8d4b8ca5d818d67b6fc6a23b9f14b1302e01/src/compiler/sscp/TargetSeparationPass.cpp#L260
Upon removing that, debug-built LLVM is able to compile `device_compilation_tests` but then the S2 fails..
Also, LLVM in Release mode still crashes in S1.. :/

```diff
--- a/src/compiler/sscp/TargetSeparationPass.cpp
+++ b/src/compiler/sscp/TargetSeparationPass.cpp
@@ -257,7 +257,7 @@ std::unique_ptr<llvm::Module> generateDeviceIR(llvm::Module &M,
   // SscpIsHostIdentifier can also be used in device code.
   S1IRConstantReplacer DeviceSideReplacer{
       {{SscpIsHostIdentifier, 0}, {SscpIsDeviceIdentifier, 1}},
-      {{SscpHcfContentIdentifier, 0 /* Dummy value */}, {SscpHcfObjectSizeIdentifier, 0}},
+      {{SscpHcfObjectSizeIdentifier, 0}},
       {{SscpHcfContentIdentifier, std::string{}}}};
   DeviceSideReplacer.run(*DeviceModule, DeviceMAM);
```

S2 failure, this branch: An assertion is triggered, since the GV `_ZTVN10__cxxabiv120__si_class_type_infoE.original` has no initializer.

S2 failure after applying #1020 to this branch:
```
[hipSYCL Error] from /home/joachimm/Projekte/opensycl/include/hipSYCL/glue/llvm-sscp/jit.hpp:251 @ compile(): jit::compile: Encountered errors:
0: LLVMToAmdgpu: hiprtcLinkComplete() failed
```
Thx HIP ^^

illuhad:
Thanks for digging into this... We should probably explicitly cast the `0` values to the actual type that it uses. I think it's `size_t`. Currently installing LLVM 17, maybe I can fix the issues :)

fodinabor:
well, `SscpHcfContentIdentifier` really should never get an int as value but an empty string only, right?

illuhad:
Yes, the initialization with 0 is clearly wrong, and probably an oversight since it's followed by initalization with empty string. ~I'm just concerned whether the other variables might be problematic too since the code might try to initialize them as `int` when actually they (or some of them) are `size_t` due to the type of the `0` literal.~

EDIT: The different arguments of the replacer are differently typed, so the arguments should already correctly take into account the difference between int and uint64 arguments.

src/compiler/llvm-to-backend/CMakeLists.txt
```diff
@@ -108,7 +108,9 @@ if(WITH_SSCP_COMPILER)
       TOOL spirv/LLVMToSpirvTool.cpp)
 
     # Install LLVM-SPIRV translator for llvm-to-spirv
-    set(LLVMSPIRV_BRANCH llvm_release_${LLVM_VERSION_MAJOR}0)
+    if(NOT LLVMSPIRV_BRANCH)
+      set(LLVMSPIRV_BRANCH llvm_release_${LLVM_VERSION_MAJOR}0)
```
illuhad, line 112: For me this fails when level zero backend is enabled: Upstream llvm-spirv translator does not have a release branch for the current LLVM development version, so that branch does not exist yet for 17. In that case, we should probably fall back to the `main` branch, although I'm unsure how to best do this in cmake here.

fodinabor: Jup, I added that so that you can manually override it with `main`:  https://github.com/fodinabor/hipSYCL-LLVM-upstream-ci/blob/main/.github/workflows/weekly.yml#L75

illuhad:
This commit:
https://github.com/llvm/llvm-project/commit/ae77aceba5ad6ee575d3d79eb0259624322b19f4
has removed `SyncDependenceAnalysis.h`, so for really current LLVM 17 distributions, another workaround is required :(

illuhad (approved):

PR #1004, opened 2023-04-05
[CI] Tidy up and test with ubuntu 22.04

nmnobre:
Hi @illuhad,

This was a bit harder than I imagined but I think it's looking good.
The last commit is there for you to see why specific configurations are being excluded from the job matrix.
Ideally I'd keep to a single ROCm version, but testing two different versions has its merits too I suppose.

Cheers,
-Nuno

### DIFF
--- a/.github/workflows/linux-cbs.yml
+++ b/.github/workflows/linux-cbs.yml
@@ -4,11 +4,11 @@ on: [push, pull_request]
 
 jobs:
   test:
-    name: clang ${{ matrix.clang_version }}, ${{ matrix.os }}
+    name: clang ${{ matrix.clang }}, ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        clang_version: [11, 12, 13, 14, 15, 16]
+        clang: [11, 12, 13, 14, 15, 16]
         os: [ubuntu-20.04]
     steps:
       - uses: actions/checkout@v2
@@ -18,11 +18,11 @@ jobs:
         run: |
           wget https://apt.llvm.org/llvm.sh
           chmod +x llvm.sh
-          sudo ./llvm.sh ${{matrix.clang_version}}
-          sudo apt install libclang-${{matrix.clang_version}}-dev clang-tools-${{matrix.clang_version}} libomp-${{matrix.clang_version}}-dev llvm-${{matrix.clang_version}}-dev
+          sudo ./llvm.sh ${{matrix.clang}}
+          sudo apt install libclang-${{matrix.clang}}-dev clang-tools-${{matrix.clang}} libomp-${{matrix.clang}}-dev llvm-${{matrix.clang}}-dev
           sudo python -m pip install lit
-          sudo ln -s /usr/bin/FileCheck-${{matrix.clang_version}} /usr/bin/FileCheck
-          if [[ "${{matrix.clang_version}}" == "16" ]]; then
+          sudo ln -s /usr/bin/FileCheck-${{matrix.clang}} /usr/bin/FileCheck
+          if [[ "${{matrix.clang}}" == "16" ]]; then
             sudo rm -r /usr/lib/clang/16*
             sudo ln -s /usr/lib/llvm-16/lib/clang/16 /usr/lib/clang/16
           fi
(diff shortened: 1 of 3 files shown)


illuhad (commented):
Thanks Nuno! It seems that it fails on more configurations than what you are excluding, is that correct? The Rocm `__oclc_abi_version` linker failure is a known issue that happens due to a bug either in ROCm or clang with certain ROCm and clang versions. One known fix is to compile using `-Xclang -mlink-bitcode-file -Xclang /path/to/rocm/amdgcn/bitcode/oclc_abi_version_400.bc`, i.e. manually the link the required bitcode file.

nmnobre:
> Thanks Nuno! It seems that it fails on more configurations than what you are excluding, is that correct?

Nope, the exclusions do cover all failing configurations. :-)

> The Rocm `__oclc_abi_version` linker failure is a known issue that happens due to a bug either in ROCm or clang with certain ROCm and clang versions. One known fix is to compile using `-Xclang -mlink-bitcode-file -Xclang /path/to/rocm/amdgcn/bitcode/oclc_abi_version_400.bc`, i.e. manually the link the required bitcode file.

I see, I'll have a look, that's happening with clang 13 and 14, and ROCm 5.3.

nmnobre:
> One known fix is to compile using `-Xclang -mlink-bitcode-file -Xclang /path/to/rocm/amdgcn/bitcode/oclc_abi_version_400.bc`, i.e. manually the link the required bitcode file.

I was able to confirm that adding that line to the compiler flags does fix it.
I set `OPENSYCL_ROCM_CXX_FLAGS` by copying the default flags and appending said line... but the result is both long and ugly.
I was hoping I'd be able to append some extra flags to the default set of flags, I thought maybe `OPENSYCL_SYCLCC_EXTRA_COMPILE_OPTIONS` would be the way to go, but it doesn't seem to work. How do you suggest I add those flags?

illuhad:
> I was able to confirm that adding that line to the compiler flags does fix it.
I set OPENSYCL_ROCM_CXX_FLAGS by copying the default flags and appending said line... but the result is both long and ugly.
I was hoping I'd be able to append some extra flags to the default set of flags, I thought maybe OPENSYCL_SYCLCC_EXTRA_COMPILE_OPTIONS would be the way to go, but it doesn't seem to work. How do you suggest I add those flags?

My guess is that `OPENSYCL_SYCLCC_EXTRA_COMPILE_OPTIONS` only works for integrated multipass, but not explicit multipass.
[I think that option just appends additional flags to the `syclcc` invocation - but for explicit multipass, `syclcc` does an extra device compiler invocation with its own set of flags]

I'm not sure if there's a convenient way to circumvent the issue in CI. Would it make sense to revert to the latest ROCm version where this is not an issue?

nmnobre:
> I'm not sure if there's a convenient way to circumvent the issue in CI. Would it make sense to revert to the latest ROCm version where this is not an issue?

I've tried ROCm 5.2.5 and unfortunately apt [reports unsatisfiable dependencies](https://github.com/nmnobre/OpenSYCL/actions/runs/4810707222/jobs/8563702745) on 22.04.
Additionally, ROCm 5.4.3 still [has the same linking problem](https://github.com/nmnobre/OpenSYCL/actions/runs/4810617667/jobs/8563481672) with clang 13 and 14.

I understand ideally we'd drop 20.04 altogether, but seeing we can't easily do that, would you be happy with:
- 🚂 20.04 w/ clang 11 to 14 and w/ ROCm 4.0.1 (this is a subset of the config we're already using)
- 🚄 22.04 w/ clang 15 and 16 and w/ ROCm 5.3.3 (as a replacement for 20.04 w/ clang 15 and 16 and w/ ROCm 4.0.1)

This would keep the same number of jobs we're already using at the expense of also dropping the following, even though they are working:
- 20.04 w/ clang 15 and 16 and w/ ROCm 4.0.1 and 5.3.3

illuhad:
Yeah we should probably keep 20.04 for some more time. 

I suppose there is not much that we can do otherwise apart from what you suggest :(

nmnobre:
So, to recap and for the avoidance of doubt, we are excluding, since they are not working, the following 14 configs:

- clang 11 to 14 w/ ROCm 5.4.3 since there are bitcode, header or linking incompatibilities (there are also no available packages for clang 11 and 12 on ubuntu 22.04, thus the exclusion for the CBS tests);
- ubuntu 22.04 w/ ROCm 4.0.1 since when installing ROCm there are unsatisfiable dependencies.

In addition, I've added a new commit which excludes the following four, working configs:

- clang 15 to 16 w/ ubuntu 20.04 since mostly redundant. 

We are thus left with six, one for each clang version, of the total 24 possible job configs.

Let me know if you'd like to avoid the same kind of redundancy in the CBS tests as well, or if you'd rather keep all the working tests, including the working ones I excluded above, just so we exhaust the space as best as we can.

illuhad:
That looks okay to me.

> Let me know if you'd like to avoid the same kind of redundancy in the CBS tests as well, or if you'd rather keep all the working tests, including the working ones I excluded above, just so we exhaust the space as best as we can.

Maybe @fodinabor would like to comment on this?

fodinabor:
IMO, a single Ubuntu version with the CBS tests should be fine since they are mostly concerned with testing the different `LLVM_VERSION_MAJOR` code paths... Could also do the 11, 12 on 20.04 and everything else on 22.04 to test both? 🤷🏼

nmnobre:
> IMO, a single Ubuntu version with the CBS tests should be fine since they are mostly concerned with testing the different `LLVM_VERSION_MAJOR` code paths... Could also do the 11, 12 on 20.04 and everything else on 22.04 to test both? 🤷🏼

I agree. Done.

illuhad (approved):
LGTM, thanks!

PR #1007, opened 2023-04-09
Implement math builtin frexp, modf, sincos

fxzjshm:
Implemented frexp, modf and sincos using existing math builtins for hiplike and ze backend.

Re-implementation of sincos is intended for better performance, although haven't benchmarked.

Tested ROCm backend on gfx906, CUDA backend on sm_86; currently I have no access to Intel GPUs, so cannot test ze backend...

### DIFF
--- a/include/hipSYCL/sycl/libkernel/spirv/builtins.hpp
+++ b/include/hipSYCL/sycl/libkernel/spirv/builtins.hpp
@@ -122,9 +122,10 @@ HIPSYCL_DEFINE_SPIRV_BUILTIN2(fmod)
 template<class T>
 T __hipsycl_fract(T x, T* ptr) noexcept;
 
-// Unsupported
 template<class T, class IntPtr>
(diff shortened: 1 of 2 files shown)


include/hipSYCL/sycl/libkernel/spirv/builtins.hpp
```diff
@@ -165,7 +166,9 @@ HIPSYCL_BUILTIN T __hipsycl_minmag(T x, T y) noexcept {
 
 // Not yet supported
```
illuhad: I think we can remove this comment now :-)

illuhad (commented):
Thanks! I think the SSCP compiler already supports these builtins. So, for Intel, you could just see what the SPIR-V SSCP builtins do - if they do the same thing as you are doing here, it's probably fine :-)

fxzjshm:
Oh accidentally forgot that... Thanks for reviewing!

illuhad (approved):
Thanks!

PR #1008, opened 2023-04-12
Implement iterators for the accessor class

nilsfriess:
This PR implements an iterator class for accessors as required in Section 4.7.6.7. of the SYCL 2020 spec. 

The implementation of the iterator is relatively simple, but arguably not optimal w.r.t. performance: the iterator stores a linear id that is relative to the accessor's offset. When the iterator is dereferenced, a `sycl::id` is computed that is again relative to the offset and this id is then transformed back to a linear id taking the offset into account using the accessor's `operator[](sycl::id)` (with the changes made in #992).

### DIFF
--- a/include/hipSYCL/sycl/libkernel/accessor.hpp
+++ b/include/hipSYCL/sycl/libkernel/accessor.hpp
@@ -1223,17 +1223,29 @@ public:
     return iterator::make_end(this);
   }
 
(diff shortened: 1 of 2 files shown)


illuhad:
Can you resolve conflicts? (and maybe merge in develop to get GPU CI :) )

illuhad (approved):
Thanks!

PR #1014, opened 2023-04-19
Enhance build system

bioinfornatics:
Dear team,

Few codes for a quick review.

This works aims to enhance the build system (without break the actual mechanism) in order to
1. be compatible with linux distribution that use lib64 as directory 
2. allow to not write a rpath into library (this is not necessary, read more [here](https://docs.fedoraproject.org/en-US/packaging-guidelines/#_beware_of_rpath))

So with this update user can provide those `cmake` options:
```
 -DLIB_SUFFIX=64
 -D CMAKE_INSTALL_RPATH_USE_LINK_PATH:BOOL=FALSE
```


src/CMakeLists.txt
```diff
@@ -1,10 +1,13 @@
...
 
 add_subdirectory(common)
 
-list( APPEND CMAKE_INSTALL_RPATH
-  ${CMAKE_INSTALL_PREFIX}/lib
-  ${CMAKE_INSTALL_PREFIX}/lib/hipSYCL/llvm-to-backend)
+if(CMAKE_INSTALL_RPATH_USE_LINK_PATH)
+  list( APPEND CMAKE_INSTALL_RPATH
+    ${CMAKE_INSTALL_PREFIX}/lib${LIB_SUFFIX}
+    ${CMAKE_INSTALL_PREFIX}/lib${LIB_SUFFIX}/hipSYCL/llvm-to-backend)
```
illuhad, line 9: Based on the documentation you have linked, I think most (all?) of our libraries fall under "internal libraries" where rpath is allowed. E.g. I'm fairly certain that disabling rpath unconditionally would probably break binaries using the generic SSCP compiler, as the `llvm-to-backend` libraries would no longer be found.

CMakeLists.txt
```diff
@@ -416,7 +416,7 @@ include(GNUInstallDirs)
 
 
 set(HIPSYCL_INSTALL_CMAKE_DIR
-  "lib/cmake/hipSYCL" CACHE PATH "Install path for legacy CMake config files")
+  "lib${LIB_SUFFIX}/cmake/hipSYCL" CACHE PATH "Install path for legacy CMake config files")
```
illuhad, line 419: I wonder, if we generalize this, do you think there might be use cases for having a lib-directory that starts without `lib`?

bioinfornatics: as fedora packager I see often this pattern into cmake file `lib${LIB_SUFFIX}` 
on mac os it is `/usr/local/lib` where `/usr/local` is the prefix
on linux  redhat  based it is `lib${LIB_SUFFIX}` depnding if it is a 32 library or a 64 
on linux debian based it is `lib`

src/CMakeLists.txt
bioinfornatics: so can we list which are internal libraries ?
we should put them into a specific dir `%{_libdir}/adaptivecpp/` and update rpath accordingly

illuhad: We do not have a single library that is meant to be directly linked against by users.

The only library we have that outside programs are linked against is our runtime library, `libacpp-rt`. Even in that case the linkage step and rpath-setup is handled internally by our compiler, `acpp`. It's not related at all to the cmake rpath settings.
Because our compiler and runtime library need to be matched exactly (the runtime library does *not* have a stable ABI), I don't think there's a very compelling use case at the moment for not rpath-ing it.

bioinfornatics:
if someone is enough skilled to patch and follow best practices, feel free to do it.
Thanks for your insight @illuhad .

PR #1016, opened 2023-04-21
Expose `Dimensions` template paramter for `{nd_}range`, `{nd_,h_}item` and `id`

nilsfriess:
In the newest revision of the SYCL 2020 spec, a `static constexpr int dimensions` member was added to a few classes to expose their `Dimensions` template parameter. In our case, the template parameter was typically lowercase, so I had to change it to `Dimensions` (with an uppercase D). I also changed it to uppercase in place where it wasn't strictly necessary (for instance in forward declarations) just to make the names of the template parameters consistent.

### DIFF
--- a/include/hipSYCL/sycl/libkernel/h_item.hpp
+++ b/include/hipSYCL/sycl/libkernel/h_item.hpp
@@ -37,55 +37,56 @@
 namespace hipsycl {
 namespace sycl {
 
-template<int dimensions>
+template<int Dimensions>
 struct group;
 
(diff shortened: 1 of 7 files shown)


illuhad (commented):
Good catch!
What is the state of the scoped parallelism classes (e.g. `sp_group`), should we add it there as well?

nilsfriess:
I added it to `sp_item`, `sp_group` already has it :)

illuhad (approved):
Thanks!

PR #1018, opened 2023-04-23
Add a few missing operators to `id` class

nilsfriess:


include/hipSYCL/sycl/libkernel/id.hpp
```diff
@@ -283,6 +283,45 @@ struct id {
...
+  HIPSYCL_ID_BINARY_OP_SIZE_T(<=)
+  HIPSYCL_ID_BINARY_OP_SIZE_T(>=)
+
+  // Unary operators +,-
+  friend id operator+(const id& rhs) {
+    return rhs;
+  }
+
+  friend id operator-(const id& rhs) {
+    return -1*rhs;
```
illuhad, line 299: Does that make sense? `id` stores `size_t`, so an unsigned type.

include/hipSYCL/sycl/libkernel/id.hpp
```diff
@@ -283,6 +283,45 @@ struct id {
   HIPSYCL_ID_BINARY_OP_SIZE_T(&)
   HIPSYCL_ID_BINARY_OP_SIZE_T(|)
   HIPSYCL_ID_BINARY_OP_SIZE_T(^)
+  HIPSYCL_ID_BINARY_OP_SIZE_T(&&)
+  HIPSYCL_ID_BINARY_OP_SIZE_T(||)
+  HIPSYCL_ID_BINARY_OP_SIZE_T(<)
+  HIPSYCL_ID_BINARY_OP_SIZE_T(>)
+  HIPSYCL_ID_BINARY_OP_SIZE_T(<=)
+  HIPSYCL_ID_BINARY_OP_SIZE_T(>=)
```
illuhad, line 291: These operators return `bool`. Do we need a `static_cast` to get it to return a `size_t` without warnings?

include/hipSYCL/sycl/libkernel/id.hpp
nilsfriess: I just added what's in the spec, but you're totally right, this makes no sense...
I will open an issue in the SYCL-Docs repo and ask about this

include/hipSYCL/sycl/libkernel/id.hpp
nilsfriess: The C++ standard defines what should happen when converting a bool to an integer type (`false` -> `0`, `true` -> `1`), so I think there will be no warnings here. And at least clang was happy

include/hipSYCL/sycl/libkernel/id.hpp
nilsfriess: Hmm, reading up on this a bit, I think this might make sense or is at least well-defined. The C++ standard says
> The negative of an unsigned quantity is computed by subtracting its value from $2^n$, where n is the number of bits in the promoted operand.

So I guess we can just keep it to be conformant

illuhad (approved):

PR #1028, opened 2023-04-26
[SSCP] Fix SSCP issues for LLVM 17

illuhad:
* Fix initialization of IR constants as discussed in #1003 
* Fix potential UB in HostKernelNameExtractionPass, although that was probably not responsible for the S1 compiler crash in #1003
* Invalidate LLVM analysis cache in S1 compilation before invoking LLVM default pipelines. This is a bit hacky. Is there a better way of making things more reliable short of running everything through LLVM pass builder?

With this PR, I can compile the unit tests and run SSCP applications with LLVM 17. So far I've only tried S2 compilation on CUDA, so not sure if the AMD S2 crash reported in #1003 is already handled.

Depends on #1003.

### DIFF
--- a/include/hipSYCL/compiler/Frontend.hpp
+++ b/include/hipSYCL/compiler/Frontend.hpp
@@ -484,12 +484,14 @@ private:
 
     // Need to iterate over all attributes to support the case
     // where multiple annotate attributes are present.
-    for(auto* Attr : f->getAttrs()) {
-      if(auto* AAttr = clang::dyn_cast<clang::AnnotateAttr>(Attr)) {
-        if (AAttr->getAnnotation() == "hipsycl_nd_kernel") {
-          markAsNDKernel(f);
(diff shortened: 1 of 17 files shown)


fodinabor:
Thx, indeed fixes the S1 compilation.
AMD S2 still fails.
(rocm 5.4.3 for my APU: gfx90c)

illuhad:
Which LLVM version is bundled with your ROCm version? Since LLVM bitcode is forwards compatible, but not backwards compatible, there can be issues when trying to inject LLVM IR from a newer LLVM into a ROCm that was built against an older LLVM. Since we are talking about LLVM 17, it's quite likely that this happens here...

Can you try setting the environment variable `HIPSYCL_SSCP_FAILED_IR_DUMP_DIRECTORY=/some/directory` to obtain the transformed IR that failed the hipRTC JIT, and then try to disassemble it using ROCm's `llvm-dis` to see if it can read it? (and to be sure that it is valid, also try to disassemble with LLVM 17)

PR #1029, opened 2023-04-29
[CI] Add GPU-based workflows & testing to CI

illuhad:
This PR adds (or rather, resurrects) GPU-based testing in CI using our own self-host infrastructure at Uni Heidelberg. We had some infrastructure for this in the past, but it turned out to be unstable -- hopefully this new approach remedies the issues.

GPU testing in CI has now become even more important than earlier, because our SSCP flow relies heavily on runtime code compilation - which means we don't even stress the full compiler stack unless we actually run SSCP binaries on a GPU.
While the unit tests are not yet in a shape to allow for SSCP testing, this PR is the first step towards this goal.

## Backend software infrastructure

The backend infrastructure to deploy and manage the github runners in a containerized manner is contained in the `OpenSYCL/ci` repository, which is currently only accessible for members of the OpenSYCL organization.

## Container for test environment

The PR here contains the workflow as well as the definition of the singularity container image that is used on our infrastructure.

Currently, when the container definition changes, the container needs to be rebuilt and deployed manually. 

This is a bit cumbersome, so ideally we could have some additional mechanism that automatically rebuilds the image when the file is change and maybe stores it on dockerhub or similar... The runners could then just pull that image (although they'd still need to be restarted somehow..).
In the best case, we could then even share the images between the self-hosted flows and the github flows to remove redundancy. If somebody has experience with these kinds of things and wants to look into this, please let me know :-)

## Hardware

Currently testing is enabled on a dedicated node with 32 cores and 4 NVIDIA 1080 Ti `sm_61` GPUs with the `omp` and SMCP `cuda` compilation flow.  Each GPU is used exclusively by one GitHub runner.
I also have AMD and Intel hardware available which I plan to hook up into CI as part of this PR soon.

## Security

**Important:** GitHub very prominently warns that adding self-hosted CI runners to public repositories like ours can pose a security risk, because if someone with malicious intent opens a PR, arbitrary code from that PR will be executed on our own machines.
Because of this, and even though I have taken some safeguards on our machines, I have changed the repository setting such that **all PRs from non-members of the organization will require approval from a member before workflows will be run**.

I'd appreciate if @OpenSYCL members could help me approve workflows where necessary :-) But please have a quick glance at the code before approving just to check if there is no bogus stuff in there that could be a security hazard. Whoever approves, vouches for the code ;)

Membership in the OpenSYCL organization is not intended to be exclusive; if you want in please let me know. I'm open to adding people as long as we have some idea who that person is, in order to have accountability regarding the security aspect. And bonus points for people who have contributed/we have collaborated with in the past :-)

## Workflow todos

The current workflow is pretty bare bones. There are more things we should test:

- [x] Explicit multipass testing
- [x] AMD hardware testing
- [ ] nvc++ testing?

Probably in a separate PR:
- [ ] SSCP testing
- [ ] Intel testing (depends on SSCP testing)

### DIFF
--- a/.github/workflows/linux-self-hosted.yml
+++ b/.github/workflows/linux-self-hosted.yml
@@ -9,6 +9,7 @@ jobs:
       matrix:
         clang_version: ['15']
         cuda: ['11.0']
+        nvhpc_version: ['22.11']
     steps:
     - uses: actions/checkout@v2
     - name: build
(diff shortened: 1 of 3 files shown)


illuhad:
We might also want to look into whether we can integrate automated performance regression testing, such as https://github.com/benchmark-action/github-action-benchmark

illuhad:
Alright, I think this is good enough for a first stab. I had to remove nvc++ testing because compiling is super slow with it (45min for the tests!) and it also miscompiles some of our tests which I had forgotten about. So we'd first have to address this before we can start testing with nvc++.

fodinabor:
Might be interesting to check out testing with arm as well? 🤔
https://resources.github.com/arm-gihub-actions-beta/

illuhad:
Definitely. But it seems we currently cannot do this, as the "larger-hosted runners" feature, which is a prerequisite, seems to be only available for enterprise accounts. In any case I cannot select the OpenSYCL org :(

PR #1031, opened 2023-05-01
[SYCL2020] Implement sycl::atomic_fence

### LINKED ISSUE #767: Alternatives for non-standard sub_group::shuffle_up/down/xor and yet-to-be-implemented atomic_fence

nmnobre:
Hi Aksel,

Less than a year later, I'm trying to bring it full circle here. 🥳
This (properly) closes #767.

Let me know what you think so far.
I haven't implemented the SSCP path because I could only find existing implementations for barriers but not memory fences, and so I think I should wait for you to tell me where and how you want this. 🙂

Cheers,
-N

### DIFF
--- a/include/hipSYCL/sycl/libkernel/atomic_fence.hpp
+++ b/include/hipSYCL/sycl/libkernel/atomic_fence.hpp
@@ -1,0 +1,111 @@
+/*
+ * This file is part of hipSYCL, a SYCL implementation based on CUDA/HIP
+ *
+ * Copyright (c) 2023 Aksel Alpay and contributors
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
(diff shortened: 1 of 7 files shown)


include/hipSYCL/sycl/libkernel/atomic_fence.hpp
```diff
@@ -0,0 +1,111 @@
...
+                          __spv::MemorySemanticsMask::SequentiallyConsistent |
+                          __spv::MemorySemanticsMask::CrossWorkgroupMemory |
+                          __spv::MemorySemanticsMask::WorkgroupMemory);
+    );
+    // TODO What about CPU?
+    // Empty __hipsycl_if_target_* breaks at compile time w/ nvc++ 22.7 or
+    // older, so comment out that statement for now.
+    //__hipsycl_if_target_host(/* todo */);
+  }
+};
```
illuhad: What is the benefit of these structs, when we do runtime if-else in `atomic_fence()` anyway?

illuhad (commented):
Good idea to work on this! There's already https://github.com/OpenSYCL/OpenSYCL/blob/develop/include/hipSYCL/sycl/libkernel/detail/mem_fence.hpp
which does basically the same thing, except that it comes from the SYCL 1.2.1 days. Can we implement the old `mem_fence` on top of the newer one?

For SSCP I would suggest looking at barrier - every barrier is effectively a combination of a memfence and a blocking synchronization primitive. So probably we can factor the `mem_fence` out from there :)

include/hipSYCL/sycl/libkernel/atomic_fence.hpp
nmnobre: Little perhaps.
It was mostly to keep it similar to `mem_fence.hpp` and perhaps bring more clarity to which of the order and scope combinations we actually support.

nmnobre:
> Can we implement the old `mem_fence` on top of the newer one?

Done.

> For SSCP I would suggest looking at barrier - every barrier is effectively a combination of a memfence and a blocking synchronization primitive. So probably we can factor the `mem_fence` out from there :)

Done.

~Can I get the declaration of `__spirv_MemoryBarrier` from `spirv_ops.hpp` in `fence.cpp` or better not?~ I did not, just in case we want a clear separation.

~As written, the semantics for `seq_cst` + `device` for  explicit multipass SPIR-V and SSCP SPIR-V is different as the memory semantics flags differ. I suspect the former includes `MemorySemanticsMask::WorkgroupMemory` because `mem_fence` could (and still can) use `fence_space::global_and_local`, but we were using it even with `fence_space::global_space`. In the barrier implementation we do not include `MemorySemanticsMask::WorkgroupMemory`. What should we do here?~

I've completely rewritten my implementation so it reflects my understanding of the intended semantics.
There are now no differences for SPIR-V.
For HIP-like devices, the SSCP path might be more precise, in the sense that the IR fence primitives we end up calling expose more choices than just `__threadfence_block()` and `__threadfence()`.
It may well be those IR primitives only implement sequential consistency at the workgroup and device scopes as well - I dunno - but we don't care about that... Indeed, if that were to change, SSCP compilations would automatically be taking advantage.

include/hipSYCL/sycl/libkernel/atomic_fence.hpp
nmnobre: Completely rewritten, structs gone.

PR #1032, opened 2023-05-02
[CI] Fix Windows CI (again)

nilsfriess:
Not really sure what exactly caused the CI to fail again, because this file `mt.exe` that it complained about is part of the Windows SDK and that should definitely be installed on the runners... Anyway, passing `embed-manifest-via=linker` when configuring boost seems to fix it (not 100% sure what it does, I think it just tells the boost installer not to use `mt.exe`).

I also removed everything CUDA-related (since we're not testing with the CUDA backend currently) and changed `clang_version` to `clang` where it was still missing from #1004.


illuhad (approved):
Thanks!

PR #1033, opened 2023-05-02
Allow non-default-construbtible types for `buffer(Container)` and `buffer(Iterator, Iterator)` constructors 

nilsfriess:
The previous implementations required the `Container` or `T` to be default-constructible. Now it also works for non-default-constructible.

### DIFF
--- a/include/hipSYCL/sycl/buffer.hpp
+++ b/include/hipSYCL/sycl/buffer.hpp
@@ -608,13 +608,8 @@ public:
       copy_host_content(reinterpret_cast<T*>(contiguous_buffer.data()));
     } else {
       std::vector<T> contiguous_buffer;
-      if constexpr (std::is_default_constructible_v<T>) {
-        contiguous_buffer.reserve(num_elements);
-        std::copy(first, last, contiguous_buffer.begin());
-      } else {
-        for(auto it = first; it != last; ++it)
-          contiguous_buffer.emplace_back(*it);
-      }
+      contiguous_buffer.reserve(num_elements);
+      std::copy(first, last, contiguous_buffer.begin());
       copy_host_content(contiguous_buffer.data());
     }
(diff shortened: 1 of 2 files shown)


include/hipSYCL/sycl/buffer.hpp
```diff
@@ -607,8 +607,14 @@ class buffer : public detail::property_carrying_object
       std::copy(first, last, reinterpret_cast<T*>(&(contiguous_buffer[0])));
       copy_host_content(reinterpret_cast<T*>(contiguous_buffer.data()));
     } else {
-      std::vector<T> contiguous_buffer(num_elements);
-      std::copy(first, last, contiguous_buffer.begin());
+      std::vector<T> contiguous_buffer;
+      if constexpr (std::is_default_constructible_v<T>) {
+        contiguous_buffer.reserve(num_elements);
+        std::copy(first, last, contiguous_buffer.begin());
+      } else {
```
nilsfriess: Here I wasn't really sure if we need the if/else because the way it is done in the else branch works of course also for types that are default-constructible although it is a bit less performant, I guess.

illuhad: Hang on, so the issue is that if it's not default constructible, we cannot reserve the memory in the vector? (does reserve also construct empty objects?) I wonder if there's a way to have a single branch using `std::copy`. Just manually allocating so that no object is initially created and then using placement new as needed?

nilsfriess: Oh I think I just confused `std::vector::resize` (which resizes and fills the new space with default-constructed elements) and `std::vector::reserve` (which only allocates new space). So this definitely does not need two branches, sorry for the confusion

illuhad (approved):
Thanks!

Issue #769, opened 2022-07-13
Make OpenSYCL live at HEAD


aaronmondal:
Inspired by recent events (#765), (#766), (#709), I thought it would be a good idea to kick off a discussion related to upstream compatibilities for hipSYCL.

# Main Point
HipSYCL is currently built and tested against various versions of Clang/LLVM (11 - 14 at the time of writing). I feel like it would make sense to add something like a "weekly build" (because nightly is overkill for the amount of commits in this repo).

I also think that it would make sense to provide a clear "current main compatibility target", and potentially set that target to "upstream".

# Suggestions
* A weekly build could use the latest version of Clang/LLVM, ROCm, CUDA. This would help detect outdated/changed APIs early so that stable hipSYCL releases are viable with newer toolchains for longer.
* If I am not mistaken, the current linux CI pipelines all test against Ubuntu. This would probably not be a good choice for a build that tests against the latest versions of, well, anything :sweat_smile: A rolling release distro is probably wanted here, so that things like recent glibc, recent python etc are implicitly used.
* If upstream toolchains are used for the development branch of hipSYCL, backwards compatibility may become simpler to maintain if it just means disabling unsupported features for older toolchains.
* Instead of treating all toolchains equally, setting the "default" compatibility target for hipSYCL to upstream Clang/LLVM may provide a clearer way of handling backwards compatibility in general. Something like having the code feel like "this is my default code path, but if you use an older toolchain I have a bunch of macros ready to fall back to your older toolchain". Currently it feels more like "If you have this toolchain I'll do this. If you have that toolchain I'll do that". This makes it harder to decide whether to *ENABLE* new features or *DISABLE* old features.

# Potential Issues
While I *think* the changes required to keep things always upstream-compatible are manageable, but I may be completely wrong here :sweat_smile: 
* CUDA seems to never change anything anyways, so probably not a lot of issues here.
* For the CPU backends, breaking changes in Clang/LLVM can be fixed in hipSYCL.
* The ROCm experience is horrible. Last time I checked, they used ancient LLVM versions and a codebase split across multiple codependent repos. Trivial fixes can take *forever* to be merged, since AMD is using different internal development branches than the ones on GitHub. Considering how many issues in hipSYCL complain about some kind of issue with ROCm, this may require an issue on its own.

# Related
[...]

PR #1040, opened 2023-05-05
Fix build with ROCm Clang 5.5.0

al42and:
It identifies as Clang 16, but lacks one new enum.

### DIFF
--- a/src/compiler/cbs/PipelineBuilder.cpp
+++ b/src/compiler/cbs/PipelineBuilder.cpp
@@ -70,7 +70,7 @@ void registerCBSPipelineLegacy(llvm::legacy::PassManagerBase &PM) {
 }
 #endif // LLVM_VERSION_MAJOR < 16
 
-#if defined(ROCM_CLANG_VERSION_MAJOR) && ROCM_CLANG_VERSION_MAJOR == 5 && ROCM_CLANG_VERSION_MINOR == 5 && ROCM_CLANG_VERSION_PATCH == 0
+#if defined(ROCM_CLANG_VERSION_MAJOR) && ROCM_CLANG_VERSION_MAJOR == 5 && ROCM_CLANG_VERSION_MINOR == 5
 #define IS_ROCM_CLANG_VERSION_5_5_0
 #endif
 
--- a/src/compiler/cbs/PipelineBuilder.cpp
+++ b/src/compiler/cbs/PipelineBuilder.cpp
@@ -70,6 +70,10 @@ void registerCBSPipelineLegacy(llvm::legacy::PassManagerBase &PM) {
 }
 #endif // LLVM_VERSION_MAJOR < 16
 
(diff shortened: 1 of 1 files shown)


illuhad:
@fodinabor I think you had some ideas about the handling of such cases, do you want to comment?

al42and:
The other alternative is to check the presence of the feature using CMake's [`try_compile`](https://cmake.org/cmake/help/latest/command/try_compile.html) and set the PP macros from there rather than relying on checking the LLVM version. This might be more robust when dealing with non-mainline clang releases and #769. That also means more CMake :scream:

fodinabor:
Mhh... unsure if I had any new insights there... do you remember in what direction they were going? ^^

In theory, I like the `try_compile` "feature test" approach, since it'd allow us to support Intel LLVM versions and others that are at least as arbitrary as ROCm versions... otoh it'd likely kill CMake configure times...

src/compiler/cbs/PipelineBuilder.cpp
```diff
@@ -70,6 +70,10 @@ void registerCBSPipelineLegacy(llvm::legacy::PassManagerBase &PM) {
 }
 #endif // LLVM_VERSION_MAJOR < 16
 
+#if defined(ROCM_CLANG_VERSION_MAJOR) && ROCM_CLANG_VERSION_MAJOR == 5 && ROCM_CLANG_VERSION_MINOR == 5 && ROCM_CLANG_VERSION_PATCH == 0
```
fodinabor: I think this is too restrictive. I don't remember ROCm ever pulling in upstream updates for their compiler version in a patch version... so I'd rather just check major and minor. here..

al42and:
```suggestion
#if defined(ROCM_CLANG_VERSION_MAJOR) && ROCM_CLANG_VERSION_MAJOR == 5 && ROCM_CLANG_VERSION_MINOR == 5
```

al42and:
> In theory, I like the `try_compile` "feature test" approach, since it'd allow us to support Intel LLVM versions and others that are at least as arbitrary as ROCm versions... otoh it'd likely kill CMake configure times...

CMake is pretty fast. GROMACS does a lot of `try_compile` checks to detect supported compiler flags, available I/O and signal handling functions, Python and MPI stuff, etc, and it all takes ~10 seconds on a local SSD and around ~20 seconds in our CI. Not instantaneous, but not too bad either, given how much is checked. It is also possible to skip tests when knowing the reported version number is enough.

src/compiler/cbs/PipelineBuilder.cpp
al42and: > I don't remember ROCm ever pulling in upstream updates for their compiler version in a patch version...

True. 5.5.1 is out, and this patch is still relevant.

fodinabor (approved):
For now I'm happy with it... adding `try_compile`s for some of these might be a good long-term fix, as it also allows to handle other "in-between" versions like Intel's DPC++ ..

illuhad (approved):

PR #1054, opened 2023-05-31
Optimize submission process for eager submission case

illuhad:
This PR optimizes a couple of current pain points when `HIPSYCL_RT_MAX_CACHED_NODES=0` (eager submission). With this PR, for a loop that submits empty kernels with hipSYCL coarse grained events on an inorder queue, I now achieve roughly 80% of the task throughput compared to `HIPSYCL_RT_MAX_CACHED_NODES=100`.

This corresponds to a 3x to 4x increase of task throughput in this scenario.

In more detail, this PR
* Removes a backend event state query from the submission path that was overlooked. The query is now replaced with a lookup to the event state cache, which should be sufficient.
* Removes the usage of `signal_channel` to notify waiting user threads that instrumentations are ready, and uses a spin lock instead. This is not ideal, but solves a current problem: Even if a task does not use instrumentation, the signal_channel had to be set up so that we can inform waiting user threads that no instrumentations are there. As it turns out, the `std::promise`/`std::future` mechanism that `signal_channel` uses is costly to set up. So we pay for something in the submission of every task even if most tasks are not being instrumented. The spin lock wastes CPU cycles when instrumentations are actually needed, but is almost free to setup, and therefore performs better in the majority of cases when we don't need instrumentations.
* Lastly, it completely changes the way garbage collection works. GC is needed for two reasons: Firstly, SYCL builds a global task graph behind the scenes, and old completed nodes must be removed from time to time so that it does not grow to a size where it starts to slow down performance. Secondly, we need to ensure that buffers stay alive as long as we have kernels operating on them. So we need to remove nodes once they have completed such that associated buffers can be freed. Previously, a GC task was spawned for each submission batch that would take care of GC for these particular nodes. This becomes a problem when we are in eager submission scenario. Then, one GC job is spawned per task. All these small GC jobs cause a lot of overhead and lock contention in our data structures. With this PR, a GC job is only spawned
   * If there is not a GC job running already and
   * If we are in a `flush_sync()` (this will happen when the user waits on a queue or event, and so we know that submission bursts are likely over)  or if after a submission the number of in-flight nodes has exceeded `HIPSYCL_RT_GC_TRIGGER_BATCH_SIZE`.

@al42and @pszi1ard I could not yet try this with Gromacs to see to what extent the benefits translate to real world apps, but I imagine it could be interesting for you :)

### DIFF
--- a/include/hipSYCL/runtime/dag_submitted_ops.hpp
+++ b/include/hipSYCL/runtime/dag_submitted_ops.hpp
@@ -34,6 +34,7 @@
 
 #include "dag_node.hpp"
 #include "generic/async_worker.hpp"
+#include "hints.hpp"
 
 namespace hipsycl {
 namespace rt {
@@ -44,12 +45,7 @@ class dag_submitted_ops
 public:
   // Asynchronously waits on the nodes to complete, and, once complete,
   // removes them (and other completed) nodes from the submitted list.
-  //
-  // All nodes in the provided argument vector must have been registered
-  // previously with update_with_submission()
-  //
-  // For best performance, the provided nodes should be in submission order.
-  void async_wait_and_unregister(const std::vector<dag_node_ptr>& nodes);
+  void async_wait_and_unregister();
   void update_with_submission(dag_node_ptr single_node);
   
   void wait_for_all();
@@ -58,9 +54,12 @@ public:
 
   bool contains_node(dag_node_ptr node) const;
 
+  std::size_t get_num_nodes() const;
+
   ~dag_submitted_ops();
 private:
   void purge_known_completed();
+  void copy_node_list(std::vector<dag_node_ptr>& out) const;
 
   std::vector<dag_node_ptr> _ops;
   mutable std::mutex _lock;
--- a/include/hipSYCL/runtime/dag_submitted_ops.hpp
+++ b/include/hipSYCL/runtime/dag_submitted_ops.hpp
@@ -57,6 +57,8 @@ public:
   std::vector<dag_node_ptr> get_group(std::size_t node_group);
 
   bool contains_node(dag_node_ptr node) const;
+
+  ~dag_submitted_ops();
 private:
   void purge_known_completed();
 
--- a/doc/env_variables.md
+++ b/doc/env_variables.md
@@ -16,3 +16,4 @@
 * `HIPSYCL_PERSISTENT_RUNTIME`: If set to 1, hipSYCL will use a persistent runtime that will continue to live even if no SYCL objects are currently in use in the application. This can be helpful if the application consists of multiple distinct phases in which SYCL is used, and multiple launches of the runtime occur.
 * `HIPSYCL_RT_MAX_CACHED_NODES`: Maximum number of nodes that the runtime buffers before flushing work.
 * `HIPSYCL_SSCP_FAILED_IR_DUMP_DIRECTORY`: If non-empty, hipSYCL will dump the IR of code that fails SSCP JIT into this directory.
+* `HIPSYCL_RT_GC_TRIGGER_BATCH_SIZE`: Number of nodes in flight that trigger a garbage collection job to be spawned
--- a/include/hipSYCL/runtime/generic/async_worker.hpp
+++ b/include/hipSYCL/runtime/generic/async_worker.hpp
@@ -33,7 +33,7 @@
 #include <condition_variable>
 #include <mutex>
 #include <functional>
-#include <queue>
+#include <deque>
 
 
 namespace hipsycl {
@@ -60,6 +60,9 @@ public:
   /// execution in the worker thread.
   /// \param f The function to enqueue for execution
   void operator()(async_function f);
+  // Enqueues a user-specified function at the *front*
+  // of the queue.
+  void push_front(async_function f);
 
   /// \return The number of enqueued operations
   std::size_t queue_size() const;
@@ -80,7 +83,7 @@ private:
   std::condition_variable _condition_wait;
   mutable std::mutex _mutex;
 
-  std::queue<async_function> _enqueued_operations;
+  std::deque<async_function> _enqueued_operations;
 };
 
 }
--- a/include/hipSYCL/runtime/generic/async_worker.hpp
+++ b/include/hipSYCL/runtime/generic/async_worker.hpp
@@ -33,7 +33,7 @@
 #include <condition_variable>
 #include <mutex>
 #include <functional>
-#include <deque>
+#include <queue>
 
 
 namespace hipsycl {
@@ -60,9 +60,6 @@ public:
   /// execution in the worker thread.
   /// \param f The function to enqueue for execution
   void operator()(async_function f);
-  // Enqueues a user-specified function at the *front*
-  // of the queue.
-  void push_front(async_function f);
 
   /// \return The number of enqueued operations
   std::size_t queue_size() const;
@@ -83,7 +80,7 @@ private:
   std::condition_variable _condition_wait;
   mutable std::mutex _mutex;
 
-  std::deque<async_function> _enqueued_operations;
+  std::queue<async_function> _enqueued_operations;
 };
 
 }
--- a/include/hipSYCL/runtime/instrumentation.hpp
+++ b/include/hipSYCL/runtime/instrumentation.hpp
@@ -82,13 +82,16 @@ inline constexpr bool is_instrumentation_v
 
 class instrumentation_set {
 public:
+  instrumentation_set()
+  : _registration_complete{false} {}
+
   /// Wait for the given instrumentation to make results available. 
   /// Returns nullptr if the given instrumentation was not set up.
   template<typename Instr> const std::shared_ptr<Instr> get() const {
     // First wait until all instrumentations have been set up
-    if(!_registration_complete_signal.has_signalled()){
-      _registration_complete_signal.wait();
-    }
+    
+    while(!_registration_complete)
+      ;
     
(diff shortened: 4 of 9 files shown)


pszi1ard:
@illuhad Thanks for the effort, this looks very interesting, I'll let you know when I had a chance to give it a try!

include/hipSYCL/runtime/dag_submitted_ops.hpp
```diff
@@ -44,21 +45,21 @@ class dag_submitted_ops
...
   std::vector<dag_node_ptr> get_group(std::size_t node_group);
 
   bool contains_node(dag_node_ptr node) const;
+
+  std::size_t get_num_nodes() const;
+
+  ~dag_submitted_ops();
 private:
   void purge_known_completed();
+  void copy_node_list(std::vector<dag_node_ptr>& out) const;
```
al42and, line 62: Stylistic question:

We have `wait_for_all()` and `wait_for_group(std::size_t node_group)`.

Since we have `std::vector<dag_node_ptr> get_group(std::size_t node_group);`, it would, IMO, be more consistent to rename this function to `std::vector<dat_node_ptr> get_all()`. Both functions can be `const`.

al42and:
Hi!

We did some testing on MI250. This patch clearly improves things for small systems, up to 6% increase of the overall performance (here we're comparing the best results over a few tested configurations, including different values of `HIPSYCL_RT_MAX_CACHED_NODES`):

Chart showing the speedup of this patch as a function of system size; performance improvements range are up to 6% for small systems of less than 100000 atoms

Thank you! :+1:

<hr/>

The CPU usage also clearly goes down. When GROMACS uses hipSYCL+ROCm, there are 5 extra threads spawned: one HSA worker thread, and four hipSYCL threads. The first two hipSYCL threads do nothing in our case, the third thread is a worker doing **submission** to the underlying runtime, while the fourth thread does **GC** (and perhaps other things). Sorry for the messy nomenclature.

Even for the lazy submission case (`HIPSYCL_RT_MAX_CACHED_NODES=100`), the CPU usage by the GC thread is noticeably lowered by this patch:

image

For the eager submission (`HIPSYCL_RT_MAX_CACHED_NODES=0`), the difference is much more noticeable:

image

Note: the CPU utilization measurements are very approximate (looking at htop and choosing average-ish value), and the error-bars are my visual estimate of how bad at visual estimation I am :)

Note 2: The CPU utilization by the HSA thread is not changed significantly; it's around 95% for small system and gradually drops to 60% for larger systems.

This brings a follow-up question: can we assume that both hipSYCL threads would be happy to share the same core or PU? Can this CPU usage be further reduced? Even with this patch and for large system sizes, the CPU usage of hipSYCL runtime is not-insignificant. E.g., on LUMI, we have 8 cores per GPU, one is already reserved for HSA, and deciding how much we must further reserve for hipSYCL is kind-of a big deal.

P.S.: Looking forward to your DevSummit talk!

illuhad:
Thanks for the feedback! That sounds great :)


> The CPU usage also clearly goes down. When GROMACS uses hipSYCL+ROCm, there are 5 extra threads spawned: one HSA worker thread, and four hipSYCL threads. The first two hipSYCL threads do nothing in our case, the third thread is a worker doing submission to the underlying runtime, while the fourth thread does GC (and perhaps other things). Sorry for the messy nomenclature.

That's quite right, there's one submission thread and one GC thread. The CPU backend creates one thread as its OpenMP "execution queue". Not sure where the last thread is coming from from the top of my head. In any case I'd indeed expect two active threads (submission and GC) in your case.

> Even for the lazy submission case (HIPSYCL_RT_MAX_CACHED_NODES=100), the CPU usage by the GC thread is noticeably lowered by this patch:

That's good to know - I hadn't looked at this case and did not expect a change here :-)

> Note 2: The CPU utilization by the HSA thread is not changed significantly; it's around 95% for small system and gradually drops to 60% for larger systems.

Do you have an idea what the HSA thread does when it spikes, or if there are some particular patterns that trigger its high CPU utilization that we should avoid?

> This brings a follow-up question: can we assume that both hipSYCL threads would be happy to share the same core or PU? 

Try it :) In theory the GC thread should not have high demands, and might be fine running on the same core. In practice there is also some locking going on between submission and GC thread (submission thread produces new nodes, and GC thread removes them again), so especially in the eager case it is likely that there is little concurrency between them anyway.

> Can this CPU usage be further reduced? Even with this patch and for large system sizes, the CPU usage of hipSYCL runtime is not-insignificant. E.g., on LUMI, we have 8 cores per GPU, one is already reserved for HSA, and deciding how much we must further reserve for hipSYCL is kind-of a big deal.

Maybe. This PR contains the straight-forward optimizations that I was able to find in a couple of days of profiling and experimentation, but there are some tuning knobs. The first thing I'd recommend is trying to play with the environment variable `HIPSYCL_RT_GC_TRIGGER_BATCH_SIZE` which controls how many nodes we can have before GC kicks in. The default is 128. A larger value means less frequent but larger GC jobs. Note that currently, when the application waits on something like a `queue::wait()` or `event::wait()` , GC is spawned in any case. The assumption is that when the code is waiting, a submission burst is over and that might be a good time to clean up. I think we might be able to loosen this if it causes too many GC jobs.

Independently I think in the long term we need to look into the memory allocation behavior and reducing the required mallocs during a submission. That requires larger changes and migration to an object pool architecture, so is not a quick change.

EDIT: I wonder what would happen if we dispatched submission and GC to the same thread. Semantically, I'm not 100% sure, but I think it should be fine. The GC thread currently also waits on all nodes to update the status in the state cache to avoid backend state queries. That would still need to be done in a separate thread.

pszi1ard:
> In theory the GC thread should not have high demands, and might be fine running on the same core. In practice there is also some locking going on between submission and GC thread (submission thread produces new nodes, and GC thread removes them again), so especially in the eager case it is likely that there is little concurrency between them anyway.

Not sure which thread is which but (after the optimization in this PR) we saw 20-30% and 50-60% CPU utilization for the two hipSYCL runtime threads, respectively; any double-digit value I'd consider relatively high demand since I have the feeling that these runtime threads interfere with each-other and cause performance loss (and/or at performance inconsistency).

Assigning both of these threads to the same core can be done as a manual hack at runtime (by setting affinities after application startup), but in general that's just a hack. 

> EDIT: I wonder what would happen if we dispatched submission and GC to the same thread. 

That's exactly what my thought was, purely based on the amount of work observed empirically, a single thread should be sufficient (especially if the GC trigger frequency is low). 

> A larger value means less frequent but larger GC jobs. Note that currently, when the application waits on something like a queue::wait() or event::wait() , GC is spawned in any case. The assumption is that when the code is waiting, a submission burst is over and that might be a good time to clean up. I think we might be able to loosen this if it causes too many GC jobs.

The issue is that at least with MPI we need to wait quite frequently (several time per iteration) because communication has to be initiated from the CPU upon the readiness of data on the device. While the current data Andrey presented is single-GPU hence no MPI and very infrequent CPU waiting on GPU, any multi-GPU run will suffer from the GC triggered by ` event::wait()`.

pszi1ard:
Something else: do you have any experience with the impact of the HSA worker thread's placement relative to the hipSYCL worker threads e.g. same cores, same CCX, same NUMA, different NUMA and at least in the first 2-3 cases with/without contention/competition for resources?

al42and:
> Do you have an idea what the HSA thread does when it spikes, or if there are some particular patterns that trigger its high CPU utilization that we should avoid?

It's not spiking, it's more or less stably at 60-90% (depending on how much we're hammering it). Randomly sampling backtraces with `gdb` was not enlightening since AMD stripped all debug info from their binary distributions: most time is spent deep in `libhsa-runtime64.so.1`, and sometimes in `ioctl`.

illuhad:
> That's exactly what my thought was, purely based on the amount of work observed empirically, a single thread should be sufficient (especially if the GC trigger frequency is low).

Thinking more about this, I'm not sure to what extent this is meaningfully possible as GC without knowing that events have completed makes little sense. So the current couple between waiting on events (which needs to be done asynchronously) and GC might be difficult to remove.

>   The issue is that at least with MPI we need to wait quite frequently (several time per iteration) because communication has to be initiated from the CPU upon the readiness of data on the device. While the current data Andrey presented is single-GPU hence no MPI and very infrequent CPU waiting on GPU, any multi-GPU run will suffer from the GC triggered by  event::wait().

Understood. We could potentially experiment with other heuristics for your use case, but for this we should first look at some data of how big the impact in the MPI case is in practice.

> Something else: do you have any experience with the impact of the HSA worker thread's placement relative to the hipSYCL worker threads e.g. same cores, same CCX, same NUMA, different NUMA and at least in the first 2-3 cases with/without contention/competition for resources?

Sorry, I don't know. I also don't know what the HSA workers do exactly and how they relate to input data from HIP or hipSYCL layers.

> It's not spiking, it's more or less stably at 60-90% (depending on how much we're hammering it). Randomly sampling backtraces with gdb was not enlightening since AMD stripped all debug info from their binary distributions: most time is spent deep in libhsa-runtime64.so.1, and sometimes in ioctl.

Ok, that's unfortunate that we know so little about this.

illuhad:
We should probably merge this for now, as it has been shown to improve things, and then figure out how to improve performance further from there.

pszi1ard:
> We should probably merge this for now, as it has been shown to improve things, and then figure out how to improve performance further from there.

I agree. We are looking into performance in detail, performance is consistently improved in fast-iterating cases with fully GPU-resident mode of GROMACS where tens of iterations are launched without CPU involvement. We are going to look next at cases where there is CPU involvement per-iteration, including synchronization prior to MPI and will report back.

PR #1063, opened 2023-06-09
Use `size_t` in decl and def of `createExitWithID`

### LINKED ISSUE #1060: The function createExitWithID is defined inconsistently in src/compiler/cbs/SubCfgFormation.cpp
Declaration and definition have different types [here](https://github.com/OpenSYCL/OpenSYCL/blob/develop/src/compiler/cbs/SubCfgFormation.cpp#L377).

fodinabor:
Fixes #1060


illuhad (approved):
Thanks!

Issue #1058, opened 2023-06-04
`RULE_LAUNCH_COMPILE`/`RULE_LAUNCH_LINK` overwritten by CMake integration


tom91136:
**Bug summary**

I was trying to compare compile time performance of different SYCL compilers so I've added custom rules to my CMake SYCL project to help with this:

```cmake

# ...
find_package(hipSYCL CONFIG REQUIRED)
# ...

set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "time -v")
set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK "time -v")
# ...
```
That works for DPCPP and normal host compilers but not hipSYCL.
Turns out, any existing COMPILE/LINK rules are overwritten:
https://github.com/OpenSYCL/OpenSYCL/blob/12fdcaedfa990ab58ddf8bce304fa8cf917e6182/cmake/hipsycl-config.cmake.in#L202-L203

Replacing the above with the following change resolves the issue:

```cmake
get_target_property(EXISTING_LAUNCH_COMPILE_RULE "${ADD_SYCL_TARGET}" RULE_LAUNCH_COMPILE)
if ("${EXISTING_LAUNCH_COMPILE_RULE}" STREQUAL "EXISTING_LAUNCH_COMPILE_RULE-NOTFOUND" )
  set(EXISTING_LAUNCH_COMPILE_RULE "")
endif ()
get_target_property(EXISTING_LAUNCH_LINK_RULE "${ADD_SYCL_TARGET}" RULE_LAUNCH_LINK)
if ("${EXISTING_LAUNCH_LINK_RULE}" STREQUAL "EXISTING_LAUNCH_LINK_RULE-NOTFOUND" )
    set(EXISTING_LAUNCH_LINK_RULE "")
endif ()
  
set_target_properties("${ADD_SYCL_TARGET}" PROPERTIES RULE_LAUNCH_COMPILE "${EXISTING_LAUNCH_COMPILE_RULE} ${OPENSYCL_SYCLCC_LAUNCH_RULE}")
set_target_properties("${ADD_SYCL_TARGET}" PROPERTIES RULE_LAUNCH_LINK "${EXISTING_LAUNCH_LINK_RULE} ${OPENSYCL_SYCLCC_LAUNCH_RULE}")
```
I'm happy to open a PR if needed.

**To Reproduce**
Steps to reproduce the behavior. 

[...]

PR #1065, opened 2023-06-11
Keep CMake target rules if defined

tom91136:
This addresses #1058.


illuhad (commented):
Thanks! Can you also apply the changes to `hipsycl-config.in` given that `opensycl-config.in` will eventually have to be renamed?

tom91136:
Done

illuhad (approved):
Thanks!

PR #1066, opened 2023-06-13
Update `sycl::exception` class to SYCL2020

nilsfriess:


illuhad:
FYI: https://github.com/KhronosGroup/SYCL-Docs/pull/431
This is likely to go in. I think that this PR here is not affected, right?

include/hipSYCL/sycl/exception.hpp
```diff
@@ -28,111 +28,121 @@
...
 
-  // Implementation in context.hpp
+  // Defined in context.hpp
   context get_context() const;
 
 private:
-  void set_error_code(){
-    error_code = std::error_code(_error_details.info().error_code().get_code(), std::system_category());
-  }
+  std::shared_ptr<context> _context;
```
illuhad, line 136: Why do we need this `shared_ptr` here? The SYCL spec already guarantees "common reference semantics" (which is effectively shared_ptr-like behavior) for context.

illuhad:
Seems I found the issue :D

PR #1067, opened 2023-06-13
fix HIPSYCL_LIBKERNEL_COMPILER_SUPPORTS_{CUDA,HIP} variables

garymm:
They were being set to 0 or 1 but being checked to see if they
were defined. Instead just define them rather than setting them to 1
and do not do anything rather than set them to 0.

### DIFF
--- a/include/hipSYCL/sycl/libkernel/cuda/cuda_backend.hpp
+++ b/include/hipSYCL/sycl/libkernel/cuda/cuda_backend.hpp
@@ -30,10 +30,10 @@
 #define HIPSYCL_LIBKERNEL_CUDA_BACKEND_HPP
 
 #if defined(__CUDACC__)
- #define HIPSYCL_LIBKERNEL_COMPILER_SUPPORTS_CUDA 1
+ #define HIPSYCL_LIBKERNEL_COMPILER_SUPPORTS_CUDA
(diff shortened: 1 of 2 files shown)


illuhad:
Thanks. We are using the pattern of being set to 0 or 1 massively also for all other-backend queries such as `HIPSYCL_LIBKERNEL_IS_DEVICE_PASS_CUDA/HIP/...`. I wonder if it might be better to just align the use of these variables with them being consistently set to 0 or 1, as with the other variables?

garymm:
@illuhad please just tell me what you want, or feel free to make the change yourself, should be as simple as modifying this to check for 1 rather than defined:

https://github.com/OpenSYCL/OpenSYCL/blob/27542ab71a72a0dfefdca58c2b0a42462a9c5300/include/hipSYCL/sycl/backend.hpp#L51

garymm:
Still valid but closing since it's been ignored for so long.

illuhad:
Apologies, nobody has the intent to ignore things, but sometimes other things can come in between and then things can be forgotten :(

If you still want this in, I'd prefer modifying the checks, not the definition of these macros.

garymm:
I'm not using this library at the moment, but it seems like a bug you might want to fix.

PR #1069, opened 2023-06-15
Let hipSYCL-rt be a non-transitive dependency.

### LINKED ISSUE #1068: `add_sycl_to_target()` creates unresolvable INTERFACE dependency for installed libraries.
**Bug summary**

For an installed library built with CMake and installing CMake "config" package support, the installed library conveys a transitive dependency on the `hipSYCL::hipSYCL-rt` (or `OpenSYCL::hipSYCL-rt`) CMake target because the PUBLIC qualifier is used for `target_link_libraries` in `hipsycl-config.cmake`|`opensycl-config.cmake`.

If client software of the accelerated library _should_ transitively link the `hipSYCL-rt` target when calling `target_link_libraries(myclient somelibrary)`, then the accelerated library could add `find_dependency(...)` to its installed `<package>-config.cmake` file. However
* it is not clear how best to help the client to find the same hipSYCL version with the same/compatible configuration (e.g. what should HIPSYCL_TARGETS be set to in order for `find_dependency()` to succeed?),
* it is not clear that this would be sufficient, since it seems like the client would want to independently call `add_sycl_to_target()` if and only if it also directly uses sycl acceleration, and
* it is not clear this dependency was intended to be transitively conveyed this way to _all_ client software of packages built using OpenSYCL

We suspect that the transitive dependency is unintentional, or at least not necessary, and that changing the dependency scope from PUBLIC to PRIVATE would be the appropriate mitigation.

**To Reproduce**

See, for example, https://gitlab.com/gromacs/gromacs/-/issues/4793

A minimal reproducer would look something like the following. (Ref https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#creating-packages)

In a library project:
```
add_library(mylibrary)
add_sycl_to_target(TARGET mylibrary SOURCES ${SYCL_SOURCES})
install(TARGETS mylibrary
            EXPORT mylibrary
            LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
)
install(EXPORT mylibrary
           DESTINATION share/cmake/mypackage
        )

include(CMakePackageConfigHelpers)
configure_package_config_file(
    mypackage-config.cmake.in mypackage-config.cmake
    INSTALL_DESTINATION share/cmake/mypackage/
)
write_basic_package_version_file(mypackage-config-version.cmake)

install(FILES
    mypackage-config-version.cmake
    mypackage-config.cmake
)
```

with the expectation that a client project does something like 
```
find_package(mypackage)
add_library(client)
target_link_libraries(client PRIVATE mylibrary)
```
However, the INTERFACE_LINK_LIBRARIES for `mylibrary` includes `OpenSYCL::hipSYCL-rt` and it is not clear how the client build system should deal with this.

**Expected behavior**

For an installed library built with OpenSYCL, CMake clients either should not depend on additional CMake targets, or it should be clear how clients of the library project should help to appropriately help client build systems.

**Describe your setup**

OpenSYCL is typically provided through an lmod or environment module on a high performance computing cluster. Multiple installations may be available, and none are expected to be discoverable by `find_package(hipSYCL)` on default paths.

eirrgang:
Switch from PUBLIC to PRIVATE in the CMake `target_link_libraries` call at the end of `add_sycl_to_target()`. Clients are assumed to call `add_sycl_to_target()` if they are using hipsycl on their own, and most libraries that are built with `add_sycl_to_target()` do not need to convey their dependence on `hipSYCL-rt` to their clients.

Fixes #1068

### DIFF
--- a/cmake/hipsycl-config.cmake.in
+++ b/cmake/hipsycl-config.cmake.in
@@ -202,5 +202,5 @@ function(add_sycl_to_target)
   set_target_properties("${ADD_SYCL_TARGET}" PROPERTIES RULE_LAUNCH_COMPILE "${OPENSYCL_SYCLCC_LAUNCH_RULE}")
   set_target_properties("${ADD_SYCL_TARGET}" PROPERTIES RULE_LAUNCH_LINK "${OPENSYCL_SYCLCC_LAUNCH_RULE}")
 
-  target_link_libraries(${ADD_SYCL_TARGET} PUBLIC hipSYCL::hipSYCL-rt)
+  target_link_libraries(${ADD_SYCL_TARGET} PRIVATE hipSYCL::hipSYCL-rt)
 endfunction()
--- a/cmake/opensycl-config.cmake.in
+++ b/cmake/opensycl-config.cmake.in
@@ -122,5 +122,5 @@ function(add_sycl_to_target)
   set_target_properties("${ADD_SYCL_TARGET}" PROPERTIES RULE_LAUNCH_COMPILE "${OPENSYCL_SYCLCC_LAUNCH_RULE}")
   set_target_properties("${ADD_SYCL_TARGET}" PROPERTIES RULE_LAUNCH_LINK "${OPENSYCL_SYCLCC_LAUNCH_RULE}")
 
-  target_link_libraries(${ADD_SYCL_TARGET} PUBLIC OpenSYCL::hipSYCL-rt)
+  target_link_libraries(${ADD_SYCL_TARGET} PRIVATE OpenSYCL::hipSYCL-rt)
 endfunction()


eirrgang:
This pull request assumes that there is _not_ a reason that a client would need to link the hipSYCL-rt binary. (It doesn't look like any compilation details are actually conveyed by the CMake target itself.)

If a hipsycl-consuming library exposes dependencies on the headers provided through the hipSYCL-rt CMake target, that library can either explicitly `target_link_libraries(... PUBLIC hipSYCL-rt)` or just add the header directory to its INTERFACE_INCLUDE_DIRS.

illuhad:
That seems reasonable to me. I vaguely remember we had some discussions in the past to what extent we need to expose `hipSYCL-rt` in the cmake infrastructure. Maybe @psalz or @fknorr know anything about the rationale of the current code? If not, I'm happy to merge.

psalz:
Sorry, I have no idea whether making this `PUBLIC` was a deliberate choice back then. In any case, this change doesn't seem to break anything for our current setup so I don't have any objections!

illuhad (approved):

PR #1072, opened 2023-06-21
Fix zero dimensional accessors

nilsfriess:
(This is the same as #916, but I can't add changes there because I deleted the base repository...)

Zero dimensional accessors can be used to access the first element of a 1D buffer. However, the methods in the accessor class expected the dimensions of the accessor and the buffer to match, causing a compilation error when trying to define a zero dimensional accessor.
To fix this, a compile time constant adjDimensions is introduced that is set to 1 if dimensions == 0 and set to the value of dimensions if dimensions > 0. This is then used instead of dimensions where necessary.


illuhad (approved):
Do you think this is still up-to-date and should go into 24.02?

nilsfriess:
CI is still happy, so I think we can merge this.

I wanted to test it with the CTS but the accessor tests there are currently completely disabled for AdaptiveCpp and since they consist of over 100 files, this might be a bigger endeavor... But since this PR also contains a unit test, I think we're good to go.

illuhad (approved):
Thanks!

PR #1084, opened 2023-07-14
Return device architecture in info::device::version

al42and:
[Per spec](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_device_information_descriptors), `sycl::device::get_info<sycl::info::device::version>` "Returns a backend-defined [device](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#device) version."

Currently, it returns "1.2 {info::platform::version}", which is redundant and (IMO) not particularly useful. Returning device architecture version seems spec-compliant and could have some uses.


illuhad:
Good idea! Yeah, the current return value comes from SYCL 1.2.1 days, where "1.2 <Something>" was referring to OpenCL profiles.

illuhad (approved):

PR #1100, opened 2023-08-05
sycl::minimum & maximum with sycl::vec

tdavidcl:
Hi,

I've added a switch to use sycl::min / fmin (or max) instead of standard comparison like `(x > y) ? x : y`, to allow sycl::minimum<T> and sycl::maximum<T> to work also with sycl::vec. 

Also added a test case for functionals in math.

It seems to work though i'm not sure whether it is the intended way to do this (should i loop over each coordinates instead ?).


illuhad:
I think I'd like to a void pulling in the builtins into the functionals. The functional header is also needed by stdpar, but not the builtin header. Since the builtin header is one of the most expensive ones to parse, this would increase compile time for stdpar noticably.

It might be more straight-forward to just continue using comparisons. That will also avoid the additional distinction between `min` and `fmin`.

For `vec`, you can use `operator<` too, however it will do the comparison component-wise and return a `vec` of an integer-type (so you'd have to cast the components to bool).
Since you'd then have to iterate over the components anyway, it might be easier to just iterate directly for vec.

It might also be interesting to ensure support for `marray`.

PR #1102, opened 2023-08-07
A few `marray` fixes

nilsfriess:
This PR introduces the following fixes:
- make `marray` constexpr-constructible as required by the spec by making `device_array` constexpr
- in the methods `count_elements` and `init_offset` (which are called from the constructor), replace a check if `T` is scalar with a check if `T` is convertible to `DataT`; here `T` is the type of an element passed to the constructor. This allows for slightly more general initialisation of marrays (the spec is a bit vague here, but the CTS has some tests that previously didn't compile and now do)
- Fix two copy-paste-errors/ typos


illuhad:
Thanks, it seems there's still an nvc++ issue. Can you perhaps try with a more recent version of nvc++? Perhaps we just need to update CI to resolve it.

include/hipSYCL/sycl/libkernel/marray.hpp
```diff
@@ -104,7 +92,9 @@ template <typename DataT, std::size_t NumElements> class marray {
   constexpr marray(marray<DataT, NumElements>&& rhs) = default;
 
   // Available only when: NumElements == 1
-  operator scalar_conversion_type() const {
+  template <std::size_t N = NumElements,
+            std::enable_if_t<N == 1, bool> = true>
+  constexpr operator DataT() const {
```
illuhad, line 98: This change might imply a semantic change. In the old implementation, the following should work:

```
void f(int x) {....}

marray<int, 1> x = ...;
f(x);
```
which I'm not sure will continue to work after the change because implicit conversion rules are different for templates.

I remember we had lots of issues getting the implicit conversion from `id<1>` to `size_t` to work, and this was the solution.

Here it might be worth checking what DPC++ does to ensure compatibility.

nilsfriess: IIRC I did check what DPC++ does and I just checked again, and they are doing the same as we do now. Your example seems to still work, but just to be sure, I'll add a test case.

nilsfriess: So the following does _not_ work anymore
```cpp
sycl::marray<int, 1> arr;
long val = arr;
```` 
But it did work before. So conversion to the same type works, implicit conversion to other types does not. I can change it back to how it was before if you prefer.

Comment (illuhad) [reply include/hipSYCL/sycl/libkernel/marray.hpp:98]: Ok, if DPC++ also handles it this way, that's fine.

**Comment by illuhad** [comment]
CI says there are still some nvc++ issues. Can you have a look?

**Comment by nilsfriess** [comment]
It seems that nvc++ is happy now (I think nvc++ was correctly complaining before the last commit; empty fold expressions are not allowed for the `+` operator).

File: install/include/hipSYCL/sycl/libkernel/marray.hpp [hunk 2]
```diff
@@ -0,0 +1,518 @@
+/*
```
illuhad: There seems something wrong here. This adds code to an `marray.hpp` in an `install` directory rather than modifying  the actual existing header. Have you by accident added the wrong file?

nilsfriess: Whoops, not sure how that happened. I put the changes in the correct file.

illuhad (approved):
Thanks!

PR #1106, opened 2023-08-15
[L0] Take EUs into account when calculating compute units

illuhad:
This change aligns us with what DPC++ returns in this case: https://github.com/intel/llvm/blob/6e524e25c422b616d37f13fe3e9fef218f1ac235/sycl/plugins/unified_runtime/ur/adapters/level_zero/device.cpp#L205


al42and:
Note: while the new code agrees with what DPC++ uses by returning the total number of EUs, but it violates the current wording of the spec (https://github.com/intel/llvm/issues/6313):

> The SYCL work-group (`sycl::group` class) is a representation of a collection of related work items that execute on a single compute unit.

SubSlices / Xe-Cores better fit for the current "compute unit" definition than EUs.

But Intel is pushing for a spec change (https://github.com/KhronosGroup/SYCL-Docs/issues/329), so not a big deal in the long term.

PR #1109, opened 2023-08-19
Add OpenCL backend

illuhad:
This adds a new runtime backend using OpenCL for SPIR-V devices. The target device must currently either support the Intel USM extensions (which are not only implemented by Intel, but also e.g. pocl ~~or rusticl~~[2]), or, as a fallback, fine-grained system SVM [1].
For SPIR-V generation, it relies on our generic single-pass compiler, i.e. `--hipsycl-targets=generic`.

The backend has been tested using the Intel OpenCL implementations for CPUs and GPUs. It runs fine on CPUs, iGPUs and PVC.

Other potential targets to experiment with could be pocl, rusticl or the oneAPI construction kit.

- [x] It needs some more testing, especially for multi-device setups.

[1] The requirement for fine-grained system SVM may appear very strong compared to the semantics of SYCL USM memory. However, system SVM is needed because all other forms of SVM require that pointers to all allocations that a kernel might use are passed to the runtime at submission time. In a SYCL USM scenario, we cannot however know this in general, because the compiler cannot infer all the pointers that might be used (think of pointer-based data structures like a linked list).

[2] As reported in the discussion of this PR, rusticl USM support is not there yet, but fine-grained system SVM is -- so it might be a target through the fallback path.


illuhad:
Works fine on multi-gpu Intel Max 1550 system. However, there seems to be a strange (driver?) issue with shared context across all the GPUs on the test system. On such systems, the environment variable `HIPSYCL_RT_OCL_NO_SHARED_CONTEXT=1` must be set. This may limit the ability to directly copy between devices.

biergaizi:
> The target device must currently either support the Intel USM extensions (which are not only implemented by Intel, but also e.g. pocl or rusticl)

Just a small note: This statement is not quite right. USM is currently unimplemented in Rusticl, and this is so far a major roadblock of SYCL support. Development is still ongoing. For the purpose of development, Intel OpenCL Intercept Layer is used to emulate USM on top of SVM. [1] 

On the other hand, fine-grained system SVM support has already landed in Mesa. [2] So this fallback in OpenSYCL is nice to see, as it would allow compatibility (better than DPC++) during the USM gap.

[1] https://gitlab.freedesktop.org/mesa/mesa/-/issues/9066

[2] https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/19712

illuhad:
@biergaizi Ah, I thought this was already being worked on - thanks for the clarification! Good to hear that the fallback is useful! :)

PR #1115, opened 2023-08-28
[OpenCL] Fix accessing build log

fxzjshm:
According to the [specification](https://man.opencl.org/clGetProgramBuildInfo.html), if return value of `_program.getBuildInfo(_dev, CL_PROGRAM_BUILD_LOG, &build_log);` is `CL_SUCCESS`, build log is available in `build_log`.


illuhad (approved):
Thanks for tracking down & fixing this typo! :)

PR #1124, opened 2023-09-05
Use `FindCUDAToolkit` for cmake versions >= 3.17

### LINKED ISSUE #1119: CMake: Replace deprecated FindCUDA with FindCUDAToolkit
**Bug summary**

CMake has [deprecated](https://cmake.org/cmake/help/latest/module/FindCUDA.html) FindCUDA module in favor of FindCUDAToolkit, with CMake 3.27 adding [a noisy warning](https://cmake.org/cmake/help/latest/policy/CMP0146.html#policy:CMP0146) whenever `find_package(CUDA)` is used.

**To Reproduce**

Try building hipSYCL with CUDA backend using CMake 3.27, observe the warning in the terminal:

```
CMake Warning (dev) at CMakeLists.txt:42 (find_package):
  Policy CMP0146 is not set: The FindCUDA module is removed.  Run "cmake
  --help-policy CMP0146" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

```

**Expected behavior**

No CMake warnings

**Describe your setup**

* hipSYCL `develop`
   * Clang 16
   * CUDA 12.2
   * ROCm 5.3.3 (unrelated, although it does add its own CMake warning)

**Additional context**

- Despite the message saying "The FindCUDA module is removed" it is not _really_ removed (yet). It works fine except for the warning, and even that can be silenced by setting the corresponding CMake policy. But relying on deprecated functionality does not look right for the project driving the future of heterogeneous/accelerated compute :)

nilsfriess:
As reported in #1119, `FindCUDA` is deprecated in recent cmake versions.
This PR adds our own `FindCUDA` that falls back to cmake's `FindCUDA` for cmake versions < 3.17. For cmake versions >= 3.17 it uses the replacement for `FindCUDA`, `FindCUDAToolkit`.

Fixes #1119

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,4 @@
-cmake_minimum_required (VERSION 3.5)
-if(NOT CMAKE_VERSION VERSION_LESS 3.12)
-  cmake_policy(SET CMP0074 NEW) # Don't complain about using BOOST_ROOT...
-endif()
+cmake_minimum_required (VERSION 3.17)
 
 set(HIPSYCL_VERSION_MAJOR 0)
 set(HIPSYCL_VERSION_MINOR 9)
@@ -39,7 +36,7 @@ if(NOT Boost_FOUND)
 endif()
 
 # Check for CUDA/ROCm and clang
-find_package(CUDA QUIET)
+find_package(CUDAToolkit QUIET)
 find_package(HIP QUIET HINTS ${ROCM_PATH} ${ROCM_PATH}/lib/cmake)
 # Check for OpenCL
 find_package(OpenCL QUIET)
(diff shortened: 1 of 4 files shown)


illuhad (commented):
Thank you, this is great to see!

I think nvc++ CI fails because it sets `-DCUDA_TOOLKIT_ROOT_DIR` to point it to nvc++'s CUDA installation.
I think it might make sense to continue to support `CUDA_TOOLKIT_ROOT_DIR` for backwards compatibility even when we use `CUDAToolkit`. Our documentation references it, and chances are user installations might also rely on it.

nilsfriess:
Now it checks if `CUDA_TOOLKIT_ROOT_DIR` is defined and sets `CUDAToolkit_ROOT` to the value in `CUDA_TOOLKIT_ROOT_DIR` (unless `CUDAToolkit_ROOT` is itself defined)

illuhad (approved):

RedSoutherly:
Hello, I think this has caused a problem for me. This seems to make building with a CUDA backend on aarch64 linux fail. Looks like because the Cmake code is looking for the install dirs specific to a x64 machine. Currently working around by cloning the previous merge commit. But this could be an issue for other people developing on aarch64 systems. (Namely NVIDIA Jetson Developer Kits if you're curious about the use case).

nilsfriess:
Thanks for reporting. 
Can you try if manually passing `-DCUDA_TOOLKIT_ROOT_DIR=/path/to/cuda` to cmake solves the issue?

tdavidcl:
> Hello, I think this has caused a problem for me. This seems to make building with a CUDA backend on aarch64 linux fail. Looks like because the Cmake code is looking for the install dirs specific to a x64 machine. Currently working around by cloning the previous merge commit. But this could be an issue for other people developing on aarch64 systems. (Namely NVIDIA Jetson Developer Kits if you're curious about the use case).

similar issue on debian (self-hosted runner) : 
```sh
cmake \
    -DCMAKE_CXX_COMPILER=/usr/bin/clang++-15 \
    -DCLANG_EXECUTABLE_PATH=/usr/bin/clang++-15 \
    -DLLVM_DIR=/usr/lib/llvm-15/cmake \
    -DWITH_CUDA_BACKEND=ON -DWITH_ROCM_BACKEND=ON \
    -DWITH_LEVEL_ZERO_BACKEND=OFF \
    -DCUDA_TOOLKIT_ROOT_DIR=~/opt/cuda \
    -DROCM_PATH=/opt/rocm \
    -DCMAKE_INSTALL_PREFIX=../OpenSYCL_comp .
```

when running make : 

```
Scanning dependencies of target rt-backend-cuda
make[2]: *** No rule to make target '/usr/lib/x86_64-linux-gnu/librt.so /home/docker/opt/cuda/lib64/stubs/libcuda.so', needed by 'src/runtime/librt-backend-cuda.so'.  Stop.
make[2]: *** Waiting for unfinished jobs....
```

in the docker image CUDA is installed like so : 
```docker
RUN mkdir -p ~/opt/cuda
RUN wget -q -O cuda.sh http://developer.download.nvidia.com/compute/cuda/11.0.2/local_installers/cuda_11.0.2_450.51.05_linux.
RUN sudo sh ./cuda.sh --override --silent --toolkit --no-man-page --no-drm --no-opengl-libs --installpath=~/opt/cuda || true
RUN echo "CUDA Version ${{matrix.cuda}}" | sudo tee ~/opt/cuda/version.txt
```

this bit is the same as the OpenSYCL CI beside the install path

nilsfriess:
Thanks, I was able to reproduce with cmake version 3.15. I thought we test with cmake version < 3.17 but maybe we actually don't...

Could you try if #1127 fixes it?

Issue #719, opened 2022-03-10
ROCTX/NVTC instrumentation for hipSYCL runtime


al42and:
**Describe the motivation for the feature request**

When profiling an application with `rocprof` or `nvprof`, it can be helpful to see not only low-level API calls, but also how they correspond to higher-level runtime activity. For that, optional instrumentation flags can be set in hipSYCL runtime using NVTX/ROCTX APIs.

**Describe the solution you'd like**

A compile-time option to enable runtime instrumentation with ROCTX or NVTX. 

**If applicable, describe alternatives you've considered**

- Not using any instrumentation. One can live without it, but reasoning about the performance of the whole stack is complicated.
- Instead of recompiling hipSYCL, an LD_PRELOAD-based solution can be concocted, similar to mpitrace. Hard to maintain, since hipSYCL is a quickly-moving target.

**Additional context**

In GROMACS, we currently achieved a performance match between native HIP and hipSYCL kernels. However, the total application performance differs by ~30%. We would like to get a better understanding of the behavior of hipSYCL runtime.

illuhad:
Thanks for the suggestion. I have some plans that go in that direction, although I don't think that NVTX or ROCTX are the right approach here. They are stack-specific and hence may not be available in all cases, and additionally I believe that any reasonable whole-stack SYCL profiling must be able to reason about the SYCL task graph, its dependencies and tightly integrate into all `wait()` calls in a way that is more similar to causal profiling.

al42and:
>  They are stack-specific and hence may not be available in all cases

That's a valid point. But I don't think that we have any cross-platform solution to get a unified trace of host and device work? So, we either only have a host-side view, or we anyway have to resort to stack-specific APIs, right?

PR #1128, opened 2023-09-08
Add instant submission mode

illuhad:
This adds a new task submission model: Instant submission. In instant submission, the entire existing scheduling and DAG construction logic is bypassed, and the node is instead immediately submitted to an `inorder_executor`.
This also allows us to circumvent the existing garbage collection and submission worker thread usage, which can make the behavior more predictable in scenarios with high CPU utilization due to user code.

The hope is that instant submission can reduce latencies in strong-scaling scenarios.

Instant submission is automatically activated when appropriate conditions are met:
- task uses no buffers (USM-only)
- task uses no SYCL 2020 reductions
- task has no running dependencies on non-instant tasks
- task is bound to a device
- task is hard-wired to an in-order `backend_executor`, which happens e.g. when using a SYCL in-order queue.

Note that stdpar code always satisfies the conditions for instant submission.

- [x] The current implementation is *NOT THREAD-SAFE*, as in instant mode, multiple user threads can in principle call directly into runtime internals, which were developed under the assumption that only the runtime worker thread would be using them. Before merging, we need to add all the mutexes.
- [x] coarse-grained events / queue completion events are not yet thread-safe when decaying to fine-grained events.
- [x] We should also consider whether we want to enable this automatically by default as it is now, or whether it should - at least initially - be opt-in. In that case the user would need to set a macro to enable this optimization, if tasks are eligible. Introducing an entire new submission model makes it likely that some issues might pop up, so for stability reasons, it might be a good idea to not immediately expose this to all the code.

CC @al42and @pszi1ard


al42and:
With [3056a88](https://github.com/OpenSYCL/OpenSYCL/pull/1128/commits/3056a8809501d4041b7edfdc141194e78081bb08), the following code hangs in `hipStreamSynchronize()`; log attached below; ROCm 5.3.3.

```cpp
#include <sycl/sycl.hpp>

using T = char;

int main()
{
    for (const auto& dev : sycl::device::get_devices())
    {
        sycl::queue q{ dev, { sycl::property::queue::in_order() } };
        char*       arr_d = sycl::malloc_device<T>(123, q);
        char*       arr_h = sycl::malloc_host<T>(123, q);

        std::iota(arr_h, arr_h + 123, 42);

        q.submit(sycl::property_list{ sycl::property::command_group::hipSYCL_coarse_grained_events() },
                 [&](sycl::handler& cgh) { cgh.memcpy(arr_d, arr_h, 123 * sizeof(T)); });

        q.submit(sycl::property_list{ sycl::property::command_group::hipSYCL_coarse_grained_events() },
                 [&](sycl::handler& cgh) { cgh.memcpy(arr_h, arr_d, 123 * sizeof(T)); });


        q.wait_and_throw();

        sycl::free(arr_d, q);
        sycl::free(arr_h, q);
    }
}
```

[verbose.log](https://github.com/OpenSYCL/OpenSYCL/files/12564011/verbose.log)

EDIT: fixed the code and the log.

illuhad:
Thanks, I've already seen this issue in CI. It seems to only happen for some backends. CUDA seems to be fine.

illuhad:
Seems like it incorrectly generates memcopies between the CPU device for some reason while dispatching the work to GPU, which cannot go well...

al42and:
The CPU copies are only there because the test is first executed on the CPU device. If I only run on a GPU, no CPU->CPU copies are executed.

illuhad:
@al42and This is not what I meant; the issue I observed was that in certain cases a CPU->CPU copy was dispatched to the *GPU* queue. However, I could since fix this issue. It was limited to OpenCL and caused by a bug in the OpenCL backend's USM pointer query implementation, causing it to not correctly recognizing the device attached to the device pointer (and thus assuming the pointer is a non-USM host pointer).
The hang is independent of this, and unfortunately I don't know what causes it yet.

illuhad:
@al42and I think I found it, I was being stupid and did not correctly configure the nodes for submission.. This has caused the `inorder_executor` to assume that every task is from some unknown backend, and thus needs special synchronization...
Please try again :)

illuhad:
With this PR, for Lulesh with C++ standard parallelism offloading to NVIDIA in the CPU-bound regime, runtime goes down from 86s to 70s. It's now only 13s slower than nvc++ in this case, and most of these 13s can likely be attributed to SSCP JIT and runtime startup.

illuhad:
Whether instant submission is allowed is now controlled via the `HIPSYCL_ALLOW_INSTANT_SUBMISSION` macro.

In SYCL mode, instant submission is now opt-in, requiring that `#define HIPSYCL_ALLOW_INSTANT_SUBMISSION 1` is present prior to including `sycl.hpp`. In stdpar mode, it is active by default.

Once things are more mature and tested, we should consider also enabling it by default in SYCL.

al42and:
The preliminary results look good. All tests pass, and, for a small system (24k atoms), in GPU-resident mode with, we get noticeable speed-up:

On my workstation, RTX3060: 383 (374) -> 397 ns/day.
On Dardel, MI250X: 542 (472) -> 585 ns/day.

We usually set  `HIPSYCL_RT_MAX_CACHED_NODES=0`, but the numbers in `()` show the results without it, using the default caching behavior. So, in the "default" mode without any environment variables, the difference is much more dramatic. The CPU usage is also reduced.

illuhad:
@al42and Excellent, thanks for the feedback! How do we compare to HIP?

al42and:
Looking at a more pathological case (6k atoms), the picture gets weirder.

For example, we have three kernels launched almost back-to-back (they are in different modules, so there is some host activity in-between, but not too much).

Here's how native HIP looks (the first hipLaunchKernel corresponds to the yellow kernel in the lower right corner); it takes ~13µs to launch all kernels:

image

If we look at the old code,  the version of hipSYCL without this patch with set `HIPSYCL_RT_MAX_CACHED_NODES=0`, it takes slightly longer, ~21µs:

image

Now, if we look at the instant submission case, things get much worse, 38µs:

image

Not sure what's causing the slowdown, but it also lowers the overall application performance, 1073.321 ns/day with HIP, 781.733 with Instant submission, 833.722 with old submission logic and no caching in this case (running 5 GROMACS threads bound to a 8c/16t CCX, no reserved cores, `HSA_OVERRIDE_CPU_AFFINITY_DEBUG=0`).

When the CPU is oversubscribed, though, this patch significantly improves performance (although, that's inadvisable way to run things to begin with).

illuhad:
Can you check with the following program? With this program, for me instant submission is always substantially faster than the old mechanism (30-50%) either with or without node caching:

```cpp

// Change to enable / disable instant submission
#define HIPSYCL_ALLOW_INSTANT_SUBMISSION 0
#include <sycl/sycl.hpp>
#include <iostream>
#include <chrono>

class profiler_clock {
public:
  using rep = uint64_t;
  using period = std::nano;
  using duration = std::chrono::duration<rep, period>;
  using time_point = std::chrono::time_point<profiler_clock>;

  constexpr static bool is_steady = true;

  static time_point now() {
    return time_point{
        duration{std::chrono::steady_clock::now().time_since_epoch()}};
  }

  static std::size_t ns_ticks(const time_point& tp) {
    return tp.time_since_epoch().count();
  }

  static double seconds(const time_point& tp) {
    auto ticks = ns_ticks(tp);
    return static_cast<double>(ticks) /
           1.e9;
  }
};


int main() {
  sycl::queue q{
    {sycl::property::queue::in_order{},
     sycl::property::queue::hipSYCL_coarse_grained_events{}}
  };
  int* data = sycl::malloc_device<int>(1, q);
  // Run once to trigger any potential JIT
  q.single_task([=](){*data += 1;});
  
  double start = profiler_clock::seconds(profiler_clock::now());
  for(int i = 0; i < 100000; ++i)
    q.single_task([=](){*data += 1;});
  q.wait();
  double stop = profiler_clock::seconds(profiler_clock::now());
  
  std::cout << stop-start << " seconds" << std::endl;
  
  sycl::free(data, q);
}

```

The only case where I can see the old mechanism being faster is if you can benefit from it submitting asynchronously (either because it JITs which is expensive, or because you can overlap with other work on the host).

al42and:
Normally, the "new" (instant) behavior is indeed faster:

```console
$ ROCR_VISIBLE_DEVICES=4 ./1128_instant_on
0.32079 seconds
$ ROCR_VISIBLE_DEVICES=4 ./1128_instant_off
0.395244 seconds
```

But if we bind all the cores to the CCX corresponding to the GPU (EDIT: or any single CCX), the "old" case gets much faster:

```console
$ ROCR_VISIBLE_DEVICES=4 hwloc-bind --cpubind core:1-7 ./1128_instant_on
0.318388 seconds
$ ROCR_VISIBLE_DEVICES=4 hwloc-bind --cpubind core:1-7 ./1128_instant_off
0.27572 seconds
```

The "old" case also sometimes crashes when run with `HIPSYCL_RT_MAX_CACHED_NODES=0`, but that's besides the point:

```console
$ HIPSYCL_RT_MAX_CACHED_NODES=0 ROCR_VISIBLE_DEVICES=4 ./1128_instant_off
double free or corruption (fasttop)free(): invalid pointer

Aborted
andreyal@nid002910:~$ HIPSYCL_RT_MAX_CACHED_NODES=0 ROCR_VISIBLE_DEVICES=4 gdb --args ./1128_instant_off
Reading symbols from ./1128_instant_off...
(gdb) r
(gdb) bt
#0  0x00000000006d51c0 in ?? ()
#1  0x00007ffff7fdebc9 in hipsycl::rt::dag_direct_scheduler::submit(std::shared_ptr<hipsycl::rt::dag_node>) () from /cfs/klemming/home/a/andreyal/OpenSYCL/build-5.3.3-instant-submission/install/lib/libhipSYCL-rt.so
#2  0x00007ffff7fe5725 in hipsycl::rt::dag_unbound_scheduler::submit(std::shared_ptr<hipsycl::rt::dag_node>) () from /cfs/klemming/home/a/andreyal/OpenSYCL/build-5.3.3-instant-submission/install/lib/libhipSYCL-rt.so
#3  0x00007ffff7fe6f65 in std::_Function_handler<void (), hipsycl::rt::dag_manager::flush_async()::$_0>::_M_invoke(std::_Any_data const&) () from /cfs/klemming/home/a/andreyal/OpenSYCL/build-5.3.3-instant-submission/install/lib/libhipSYCL-rt.so
#4  0x00007ffff7fecbd6 in hipsycl::rt::worker_thread::work() () from /cfs/klemming/home/a/andreyal/OpenSYCL/build-5.3.3-instant-submission/install/lib/libhipSYCL-rt.so
#5  0x00007ffff62773b4 in ?? () from /usr/lib64/libstdc++.so.6
#6  0x00007ffff54266ea in start_thread () from /lib64/libpthread.so.0
#7  0x00007ffff595aa8f in clone () from /lib64/libc.so.6
```

illuhad:
Ok... I cannot reproduce this behavior, even with `hwloc-bind`. What is the host CPU?

I also cannot reproduce the crash so far. Obviously, if something UB is going on there, then it might be difficult to compare performance. Is ROCm LLVM involved here? If so I'd suggest to try stock LLVM. Wouldn't be the first time ROCm LLVM miscompiles code.

al42and:
> What is the host CPU?

AMD EPYC 7A53

> Is ROCm LLVM involved here? 

Yes... Also, it's ROCm 5.3.3, which has some runtime latency issues that the newer versions do not.

> If so I'd suggest to try stock LLVM. Wouldn't be the first time ROCm LLVM miscompiles code.

I can try Cray LLVM, but last time I checked (~1 year ago), it had troubles compiling hipSYCL, so I wouldn't hold it as a gold standard.

I can take a stab at building stock LLVM from scratch in the next days.

illuhad:
>  AMD EPYC 7A53

Interesting, I can't find information on that one; is that a special LUMI CPU or did you mean 7453?
In any case, I don't have any Milan CPUs here for testing, but I can try to get access to some.

> I can try Cray LLVM, but last time I checked (~1 year ago), it had troubles compiling hipSYCL, so I wouldn't hold it as a gold standard.
> I can take a stab at building stock LLVM from scratch in the next days.

Is there perhaps any container solution available (e.g. singularity or charliecloud)? That would help with reproducability.

al42and:
> Interesting, I can't find information on that one; is that a special LUMI CPU or did you mean 7453?

It's the special LUMI (Frontier, Dardel) CPU. https://docs.olcf.ornl.gov/systems/frontier_user_guide.html#simplified-node-layout

I can speculate that, with 128 virtual CPUs available, the OS scheduler might decide to spread the threads over different L3 domains hurting the communications between hipSYCL runtime threads.

Deliberately binding the test app on the edge of two domains (two cores from the first domain 0-7, two from the second 8-15) also kills the performance:

```
$ ROCR_VISIBLE_DEVICES=4 hwloc-bind --cpubind core:0-3 ./1128_instant_off
0.276253 seconds
$ ROCR_VISIBLE_DEVICES=4 hwloc-bind --cpubind core:7-10 ./1128_instant_off
0.413205 seconds
$ ROCR_VISIBLE_DEVICES=4 hwloc-bind --cpubind core:8-12 ./1128_instant_off
0.272617 seconds
```

> Is there perhaps any container solution available (e.g. singularity or charliecloud)?

With a Cray system image? Not that I know of. We asked LUMI staff a while ago, and they could not suggest anything.

illuhad:
> It's the special LUMI (Frontier, Dardel) CPU. https://docs.olcf.ornl.gov/systems/frontier_user_guide.html#simplified-node-layout

I see, thanks!

> I can speculate that, with 128 virtual CPUs available, the OS scheduler might decide to spread the threads over different L3 domains hurting the communications between hipSYCL runtime threads.
> Deliberately binding the test app on the edge of two domains (two cores from the first domain 0-7, two from the second 8-15) also kills the performance:

I can totally see that this might be an issue for the old submission mechanism. Instant submission does not use any internal hipSYCL threads, so it should mostly be unaffected by binding (as your results confirm). But I'm surprised that the old mechanism outperforms instant submission. That would mean that for the little work that it still does, it still is beneficial to offload that to a worker thread.
We could try having a submission worker thread per queue that does nothing but submission for the instant submission case.

>  With a Cray system image? Not that I know of. We asked LUMI staff a while ago, and they could not suggest anything

That's unfortunate :(  I'll make sure that our university never purchases a Cray system...

al42and:
> But I'm surprised that the old mechanism outperforms instant submission. That would mean that for the little work that it still does, it still is beneficial to offload that to a worker thread.

I don't understand how offloading stuff to a worker thread helps here. If we were doing something compute-intensive in the main thread that can overlap with submission, sure. But here we just submit a ton of kernels and then wait. As far as I can tell, the only thing we can do concurrently is checking the node properties and calling the HIP API. Is the former really so expensive that it makes sense to overlap the two?

Would it be more lightweight to use a per-submission/per-queue flag enforcing "instant" mode, and letting the user shoot themselves in the foot if they so desire?

> I can totally see that this might be an issue for the old submission mechanism. 

BTW, I can see how the same problem might happen on consumer devices with multiple L3 caches, like 7950X3D. Non-HPC users are unlikely to bother with manual affinity setting, so I feel like getting rid of threads and associated problems when running in a latency-sensitive mode might be the best way forward.

include/hipSYCL/sycl/handler.hpp
```diff
@@ -933,7 +911,52 @@ class handler {
...
+    } else {
+      // instant submission
+      hints.add_hint(rt::make_execution_hint<rt::hints::instant_execution>());
+
+      rt::dag_node_ptr node = std::make_shared<rt::dag_node>(
+          hints, _requirements.get(), std::move(op), _rt);
+      node->assign_to_device(
+          hints.get_hint<rt::hints::bind_to_device>()->get_device_id());
+      node->assign_to_executor(executor);
+      executor->submit_directly(node, node->get_operation(), _requirements.get());
```
al42and, line 982: Can we pass and empty array instead instead of `_requirements` here? As far as I understand, `submit_directly` iterates over the requirements internally, but that's not necessary here?.

illuhad: No, there might be other requirements explicitly specified by the user with `depends_on()`.

illuhad:
> I don't understand how offloading stuff to a worker thread helps here. If we were doing something compute-intensive in the main thread that can overlap with submission, sure. But here we just submit a ton of kernels and then wait. As far as I can tell, the only thing we can do concurrently is checking the node properties and calling the HIP API. Is the former really so expensive that it makes sense to overlap the two?

It should not be that expensive, but I really don't know how the old mechanism might outperform instant submission otherwise. Instant submission just bypasses a lot of stuff; the only thing the old mechanism has that might change the equation is that it immediately offloads to another thread.

> Would it be more lightweight to use a per-submission/per-queue flag enforcing "instant" mode, and letting the user shoot themselves in the foot if they so desire?

To be clear, I really don't think that what happens in `handler` and `queue` itself is expensive so that it might benefit from a worker thread. But there are more layers below until we get to the kernel launch, which are certainly more impactful than the couple of lines in `handler`:  `inorder_executor`, `hip/cuda/ocl/..._queue`, `sscp/hiplike_kernel_launcher`. What they do is mostly trivial too, but I don't know how to explain this otherwise at the moment. So I don't think there's something to be gained from such a flag apart from making the behavior more risky.

illuhad:
> BTW, I can see how the same problem might happen on consumer devices with multiple L3 caches, like 7950X3D. Non-HPC users are unlikely to bother with manual affinity setting, so I feel like getting rid of threads and associated problems when running in a latency-sensitive mode might be the best way forward.

Do I understand correctly that you would prefer instant submission in its current form, even if in certain circumstances the old mechanism might be faster? Instant submission does not currently rely on any threads.

al42and:
> Do I understand correctly that you would prefer instant submission in its current form, even if in certain circumstances the old mechanism might be faster? Instant submission does not currently rely on any threads.

Well, it complicated. If we look at our PME grappa systems, the behavior depends a lot on how we're scheduling things. In the fully GPU-resident mode, we just schedule tons of operations in advance and then just wait for them to complete, while in the partially GPU-resident mode, we compute some forces on the CPU and thus have to do a synchronization with host every ~10 kernels to copy them to the GPU. The table below shows the performance (ns/day) of the fastest configuration among those tested. "Old" runs are with  `HIPSYCL_RT_MAX_CACHED_NODES=0`; bold highlights the best performance for each mode:


<table>
<thead>
  <tr>
    <th rowspan="2"> <br>Size, katoms</th>
    <th colspan="2">Fully GPU-resident</th>
    <th colspan="2">Partially GPU-resident</th>
  </tr>
  <tr>
    <th>old</th>
    <th>new</th>
    <th>old</th>
    <th>new</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td>6</td>
    <td><b>936.0</b></td>
    <td>916.4</td>
    <td><b>837.8</b></td>
    <td>781.7</td>
  </tr>
  <tr>
    <td>12</td>
    <td>796.2</td>
    <td><b>812.1</b></td>
    <td><b>751.7</b></td>
    <td>709.9</td>
  </tr>
  <tr>
    <td>24</td>
    <td>616.8</td>
    <td><b>624.4</b></td>
    <td><b>605.1</b></td>
    <td>597.0</td>
  </tr>
  <tr>
    <td>48</td>
    <td>456.7</td>
    <td><b>464.0</b></td>
    <td>448.1</td>
    <td><b>460.0</b></td>
  </tr>
  <tr>
    <td>96</td>
    <td>287.6</td>
    <td><b>291.7</b></td>
    <td>286.8</td>
    <td><b>292.8</b></td>
  </tr>
  <tr>
    <td>192</td>
    <td>160.0</td>
    <td><b>162.0</b></td>
    <td>158.1</td>
    <td><b>159.9</b></td>
  </tr>
  <tr>
    <td>384</td>
    <td>80.5</td>
    <td><b>80.8</b></td>
    <td>81.0</td>
    <td><b>81.6</b></td>
  </tr>
  <tr>
    <td>768</td>
    <td>41.1</td>
    <td><b>41.2</b></td>
    <td>41.1</td>
    <td><b>41.2</b></td>
  </tr>
</tbody>
</table>

So, the instant submission considerably improves the performance in many cases. As I mentioned, 6k is a pathological case, but it demonstrates an abnormal behavior in the runtime, so it might be worth examining further. As seen in the traces earlier, the submission is still much slower than native HIP.

How soon do you want a go/no-go decision on this PR? I'd rather spend some quality time with profiler here, but I don't think I'll have time for that until the last week of September at least.

BTW, here's where #719 could have helped (no judging, we're stalling on adding a proper support for tracing to GROMACS too).

illuhad:
> The table below shows the performance (ns/day) of the fastest configuration among those tested. "Old" runs are with HIPSYCL_RT_MAX_CACHED_NODES=0; bold highlights the best performance for each mode:

Wow, thanks for compiling all this data! This is indeed very interesting.

> So, the instant submission considerably improves the performance in many cases. As I mentioned, 6k is a pathological case, but it demonstrates an abnormal behavior in the runtime, so it might be worth examining further. As seen in the traces earlier, the submission is still much slower than native HIP.

I agree, I'd like to know what is going on too.

> How soon do you want a go/no-go decision on this PR? I'd rather spend some quality time with profiler here, but I don't think I'll have time for that until the last week of September at least.

I don't see that there is much at stake since the instant submission mode as of this PR is disabled by default in SYCL. I have some interest in it going in for stdpar. So, one option could be for it to go in knowing that it might need further tweaking in the future. This is unless you tell me that the approach is fundamentally not what you want. In that case I'd rather not merge it than rip it out again later :-)

Once the name change is over (at last), we should urgently do a release. Let's say optimistically we do name change within a couple of days, then a couple of weeks for testing/stabilization, release could be within a month.
I think this feature would be interesting enough to be part of the release, if it does not break things for you. I have a use case where it noticably speeds up things.

pszi1ard:
@illuhad Thanks for the effort, this work looks quite promising!

I have just done some testing with more recent ROCm versions (>=5.5 which are N/A on Dardel/LUMI) and I observe significantly improved runtime performance, so I think we need totest with new ROCm too and use both e.g. 5.3 and 5.6 as our baseline. In particular, efficiency of the HIP/HSA stack improves, hiding runtime inefficiencies e.g. by async submission in a hipSYCL worker thread and +/- caching nodes become less useful. That might suggest that the observed gain of instant submission could increase, but the opposite is possible (and the above data seems to have caught such cases). One reason could be that inefficiencies in the SYCL runtime can become more pronounced and exposed to the application.

To this feature, @al42and's trace where the instant submission is showing much increased time between back-to-back launches is somewhat concerning (along the crash he encountered) so I think it would be best to understand what is happening there.
There can be a complex interaction between the hipSYCL, HIP and HSA runtimes and to make sure that we get broadly applicable real-world benefits in use-cases we want to see improvements for GROMACS, I think we should evaluate this feature on hardware and SW stack we target. We will also need time to assess stability for production and compare to the earlier RT-OPT, all as a function of ROCm version.

Therefore, how we / whether we can allocate time for such testing depends also on the release plans for the next 4-6 months. An imminent new stable release would mean that there will not be much time left for in-depth testing instant submission feature in addition to the release itself.  That's OK if there will be another stable release between now and ~Feb 2024 which can improve on instant submission if needed.

pszi1ard:
PS: tangentially related, but here is some data on runtime overheads measured in the GROMACS application thread (i.e. total cost of invoking hipSYCL launches); unfortunately this is with pre RT-OPT I've not had time (nor hardware available) to re-test this with RT-OPT or instant submission.


|                                  | MAX_CACHED_NODES=100 |               | MAX_CACHED_NODES=5 |                  | MAX_CACHED_NODES=0        |                  |
|---------------------------------|------------------|-------------------------------|------------------|-------------------------------|------------------|-------------------|
| rocm version                  | wall t/iteration (μs) | GPU launch (μs)             | wall t/iteration | GPU launch wall t             | wall t/iteration | GPU launch wall t |
| 5.3.3                      | 305.03           | 101.19                        | 317.28           | 103.69                        | 384.88           | 158.59            |
| 5.4.3                      | 306.73           | 74.7                          | 291.29           | 160.89                        | 383.93           | 157.84            |
| 5.5.1                      | 299.34           | 74.75                         | 278.39           | 159.99                        | 412.08           | 155.09            |
| 5.6.1                      | 296.34           | 75.8                          | 276.89           | 169.74                        | 393.43           | 155.19            |

al42and:
That's interesting: with ROCm 5.3.3, the submission time increases as the kernels get longer. I suspect that when there is a lot of work in the GPU queue, the HIP API calls might take longer on average, but I don't have fresh traces to prove it (but that API latency likes to randomly spike is established).

That's the [same dataset as above](https://github.com/OpenSYCL/OpenSYCL/pull/1128#issuecomment-1720201865), but with more detailed breakdowns ([using Szilárd's script](https://github.com/OpenSYCL/OpenSYCL/pull/1128#issuecomment-1721414749)). "old" is run with `HIPSYCL_RT_MAX_CACHED_NODES=0`.

<table><thead><tr><th rowspan="2">Size, katoms</th><th colspan="9">Fully GPU-resident</th></tr><tr><th colspan="3">old</th><th colspan="3">new</th><th colspan="3">HIP</th></tr></thead><tbody><tr><td></td><td>perf, ns/day</td><td>wall t/iter, µs</td><td>GPU launch, µs</td><td>perf, ns/day</td><td>wall t/iter, µs</td><td>GPU launch, µs</td><td>perf, ns/day</td><td>wall t/iter, µs</td><td>GPU launch, µs</td></tr><tr><td>6</td><td>939.9</td><td>184.7</td><td>46.6</td><td>917.5</td><td>188.3</td><td>158.1</td><td>1120.8</td><td>154.2</td><td>118.1</td></tr><tr><td>12</td><td>792.3</td><td>218.3</td><td>48.0</td><td>804.4</td><td>215.3</td><td>170.1</td><td>983.5</td><td>175.8</td><td>127.1</td></tr><tr><td>24</td><td>609.7</td><td>283.5</td><td>48.1</td><td>618.8</td><td>279.6</td><td>208.5</td><td>771.9</td><td>224.2</td><td>154.5</td></tr><tr><td>48</td><td>447.4</td><td>386.6</td><td>49.6</td><td>450.8</td><td>383.4</td><td>210.4</td><td>540.0</td><td>320.1</td><td>153.8</td></tr><tr><td>96</td><td>281.6</td><td>613.7</td><td>49.0</td><td>284.1</td><td>608.4</td><td>210.5</td><td>339.5</td><td>509.0</td><td>154.6</td></tr><tr><td>192</td><td>156.9</td><td>1101.3</td><td>55.0</td><td>157.9</td><td>1094.4</td><td>213.5</td><td>185.3</td><td>932.7</td><td>158.4</td></tr><tr><td>384</td><td>78.6</td><td>2198.8</td><td>58.2</td><td>78.9</td><td>2191.8</td><td>218.1</td><td>92.4</td><td>1872.5</td><td>160.4</td></tr><tr><td>768</td><td>39.7</td><td>4357.2</td><td>62.3</td><td>39.7</td><td>4350.9</td><td>218.3</td><td>45.7</td><td>3784.5</td><td>162.4</td></tr></tbody></table>

<table><thead><tr><th rowspan="2">Size, katoms</th><th colspan="9">Partially GPU-resident</th></tr><tr><th colspan="3">old</th><th colspan="3">new</th><th colspan="3">HIP</th></tr></thead><tbody><tr><td></td><td>perf, ns/day</td><td>wall t/iter, µs</td><td>GPU launch, µs</td><td>perf, ns/day</td><td>wall t/iter, µs</td><td>GPU launch, µs</td><td>perf, ns/day</td><td>wall t/iter, µs</td><td>GPU launch, µs</td></tr><tr><td>6</td><td>840.8</td><td>205.9</td><td>65.7</td><td>778.7</td><td>222.1</td><td>187.0</td><td>1102.3</td><td>156.8</td><td>128.3</td></tr><tr><td>12</td><td>749.3</td><td>232.3</td><td>66.6</td><td>709.0</td><td>244.3</td><td>194.0</td><td>962.3</td><td>179.7</td><td>133.2</td></tr><tr><td>24</td><td>603.9</td><td>286.4</td><td>67.9</td><td>593.6</td><td>291.4</td><td>208.0</td><td>775.0</td><td>223.2</td><td>142.7</td></tr><tr><td>48</td><td>442.8</td><td>390.5</td><td>68.2</td><td>452.2</td><td>382.5</td><td>221.6</td><td>527.6</td><td>327.8</td><td>152.4</td></tr><tr><td>96</td><td>277.5</td><td>624.2</td><td>64.4</td><td>285.0</td><td>606.6</td><td>267.7</td><td>328.2</td><td>527.2</td><td>193.4</td></tr><tr><td>192</td><td>153.9</td><td>1122.9</td><td>64.9</td><td>156.1</td><td>1107.1</td><td>286.0</td><td>186.3</td><td>927.9</td><td>207.6</td></tr><tr><td>384</td><td>79.3</td><td>2179.9</td><td>70.6</td><td>79.6</td><td>2171.6</td><td>287.1</td><td>92.1</td><td>1877.7</td><td>214.7</td></tr><tr><td>768</td><td>39.9</td><td>4333.6</td><td>74.8</td><td>40.0</td><td>4319.9</td><td>283.9</td><td>45.9</td><td>3764.1</td><td>211.0</td></tr></tbody></table>

Note: the numbers are median over 5x8 runs with `-ntomp 6`; aggregation is done sliightly differently, so the application performance is slightly different from the one I posted earlier.

pszi1ard:
@al42and what would be relevant here is to also include a sensible `HIPSYCL_RT_MAX_CACHED_NODES` that gives close to best perf with "non-instant" submission.

al42and:
For the rt_opt version (#1054), the `HIPSYCL_RT_MAX_CACHED_NODES=0` is near optimal for the total performance (not for the GPU launch, of course). For a `-ntomp 6` run, fully GPU-resident (I don't have `-bonded cpu` data there):

**wall t/iter, µs** | cache |   |   |   |   |   |   |  
-- | -- | -- | -- | -- | -- | -- | -- | --
**size** | 0 | 1 | 2 | 5 | 10 | 50 | 100 | 200
6 | 202.93 | 196.54 | **194.99** | 196.1 | 197.71 | 210.15 | 219.54 | 240.99
12 | **224.54** | 227.88 | 227.77 | 226.6 | 229.43 | 239.65 | 259.37 | 272.48
24 | 290.59 | **285.43** | 294.09 | 292.98 | 295.48 | 307.21 | 322.26 | 338.15
48 | **402.48** | 404.03 | 404.7 | 404.31 | 402.87 | 412.14 | 425.31 | 444.03
96 | 632.24 | **628.02** | 637.08 | 632.91 | 633.91 | 647.41 | 665.13 | 677.07
192 | **1127.88** | 1129.99 | 1138.38 | 1135.94 | 1130.38 | 1140.55 | 1151.38 | 1183.27
384 | 2255.04 | 2256.37 | 2259.93 | 2255.15 | **2252.04** | 2264.6 | 2285.1 | 2311.54
768 | 4469.92 | 4468.2 | 4465.53 | 4459.53 | **4451.03** | 4472.31 | 4472.25 | 4502.36

GPU launch, µs | cache |   |   |   |   |   |   |  
-- | -- | -- | -- | -- | -- | -- | -- | --
**size** | 0 | 1 | 2 | 5 | 10 | 50 | 100 | 200
6 | 74.4 | 67.0 | 67.4 | 57.8 | 55.6 | 53.9 | 55.9 | 43.4
12 | 62.9 | 60.8 | 63.6 | 58.0 | 52.3 | 50.8 | 55.0 | 44.0
24 | 71.3 | 65.9 | 61.8 | 51.9 | 57.2 | 51.4 | 51.5 | 41.9
48 | 58.1 | 52.8 | 52.0 | 42.3 | 57.6 | 52.1 | 49.8 | 41.2
96 | 43.3 | 42.5 | 49.9 | 41.6 | 53.9 | 48.5 | 43.8 | 39.7
192 | 40.7 | 41.2 | 47.9 | 45.1 | 34.8 | 39.6 | 48.3 | 38.2
384 | 51.2 | 42.5 | 44.8 | 36.5 | 36.6 | 39.6 | 43.7 | 38.7
768 | 43.8 | 53.0 | 43.8 | 38.5 | 40.4 | 41.0 | 41.2 | 38.6

EDIT: This data is for a different binding settings (cores 1-7, not 0-7, and without `OMP_PLACES=cores`), so not directly comparable.

pszi1ard:
@al42and Interesting, of course this is near-ideal from the point of view of runtime costs since we leave 2 full cores and hardware threads of the rest of the cores free.

What is clear is that the non-instant submission is effectively able to half to two thirds of the launch overheads exposed to the GROMACS application thread, a cost that becomes relevant when there is either CPU work or MPI communication on the critical path (so perhaps we should be measuring that separately by measuring the runtime overheads that delay CPU work?).

That would explain the slowdowns in cases where the CPU is on the critical path, but not the case where the CPU only does scheduling, so I'd still think there is some added overhead in the hipSYCL runtime, either new with instant submission or newly exposed to the caller application thread, that we should track down.

Do you have any native HIP data from recent benchmarks? Comparing that to the instant submission would help better understanding how much of what is exposed by instant submission is the HIP runtime's overhead and how much of it is added by the SYCL runtime.

pszi1ard:
PS: something odd I don't get is why does the launch overhead decrease with "partially GPU resident mode" when using instant submission (we are launching *more* work at least 3-4 extra API calls) but increases with RT_OPT?

al42and:
> Do you have any native HIP data from recent benchmarks? Comparing that to the instant submission would help better understanding how much of what is exposed by instant submission is the HIP runtime's overhead and how much of it is added by the SYCL runtime.

Yes, updated [the table above](https://github.com/OpenSYCL/OpenSYCL/pull/1128#issuecomment-1721495615).

For small sizes, (walltime - GPU_launch) is approx. the same for "new" and "HIP", so it checks out that the main difference is just the submission cost.

pszi1ard:
> For small sizes, (walltime - GPU_launch) is approx. the same for "new" and "HIP", so it checks out that the main difference is just the submission cost.

OK, that's good to know. For small sizes probably compute kernel differences become small, and the run is API launch-bound. To have a more reliable comparison across the whole range we'd have to make the kernels run at roughly equal speed (ideally closing the perf gap ;)

al42and:
Aaand, updated [the table again](https://github.com/OpenSYCL/OpenSYCL/pull/1128#issuecomment-1721495615), because I copypasted the wrong numbers in the "new" column (swapped Fully GPU-resident and ~~CPU-resident~~ Partially GPU-resident) when exporting the data here.

pszi1ard:
>  (swapped GPU-resident and CPU-resident)

what's CPU resident, `-update cpu`?

al42and:
> what's CPU resident, -update cpu?

Yes. I meant "Patially GPU-resident". We need to come up with more convenient names.

pszi1ard:
> Yes. I meant "Patially GPU-resident". We need to come up with more convenient names.

Nit: FWIW these are all "GPU resident".

al42and:
A [quick-and-dirty annotation with rocTX](https://github.com/al42and/hipSYCL/commit/35a706cdcf11d8d850af5fc60a6a551d8eb91484) was not particularly enlightening, mostly due to 1µs resolution and, apparently, the fact that the overhead is spread between multiple locations. But perhaps @illuhad can gain some insights?

"Submit GPU Work" markers are right around the `queue.submit(...)` call, the others correspond to hipSYCL functions:

Screenshot_20230915_221921

illuhad:
Interesting...

The gap between the end of `hip_queue::submit_kernel` and the end of `inorder_executor::submit_directly` is likely the coarse-grained event construction as really nothing else happens after the kernel submission. Due to SYCL's automatic memory management semantics this has to be a `shared_ptr`, so my guess is that the memory allocation on the heap is the culprit.

There's also memory allocation before the launch (to construct the DAG node). That might play a role in what is going on prior to submission inside `create_task()`.

The queue will also copy the queue configuration into the handler (which is an `std::vector`, which then gets processed and copied into the DAG node). I've wanted to experiment with small vector optimizations for a long time here...

Apart from that there's really not much that is happening.

For most cases this might not play a large role, but if the comparison is direct calls to `hipLaunchKernel` with trivial kernels, you might see all that.

al42and:
> The gap between the end of hip_queue::submit_kernel and the end of inorder_executor::submit_directly is likely the coarse-grained event construction as really nothing else happens after the kernel submission. 

One "tick" along the X axis is 1µs = profiler resolution, so we should be careful when interpreting such single events. But there's clearly a lot happening _before_ the HIP API is called. Here's the full trace: 
[8390_instant_on.json.gz](https://github.com/OpenSYCL/OpenSYCL/files/12641056/8390_instant_on.json.gz)

> For most cases this might not play a large role, but if the comparison is direct calls to hipLaunchKernel with trivial kernels, you might see all that.

It will also be an issue if we have CPU work we want to overlap with the GPU. When we have a lot of cores, having a separate launcher thread would be a good solution, perhaps even better than native HIP. But for cases like LUMI, where we have 8c/8t per GPU, with one core reserved by the scheduler (presumably, for the HSA thread), the CPU compute tasks only have 7 cores, and yielding one of them to the worker thread would be not great; similarly, delaying the compute tasks by having an inefficient same-thread submit is bad, and delaying the communication can be even worse.

So, I agree that most application might not notice a difference. But it's not limited to 1µs-long kernels either.

illuhad:
> It will also be an issue if we have CPU work we want to overlap with the GPU. When we have a lot of cores, having a separate launcher thread would be a good solution, perhaps even better than native HIP. But for cases like LUMI, where we have 8c/8t per GPU, with one core reserved by the scheduler (presumably, for the HSA thread), the CPU compute tasks only have 7 cores, and yielding one of them to the worker thread would be not great; similarly, delaying the compute tasks by having an inefficient same-thread submit is bad, and delaying the communication can be even worse.

Agreed. It sounds like we may need two different modes:
- Synchronous instant submission, which requires optimizing the submission path (likely including moving to small vector classes instead of `std::vector` and similar optimizations)
- Asynchronous instant submission where a worker thread is involved (Or at least we should investigate the benefit of this - maybe not much is to be gained once the submission path is sufficiently optimized)

Depending on the use case, one or the other might be chosen by the client application and/or the runtime.

al42and:
> The "old" case also sometimes crashes when run with HIPSYCL_RT_MAX_CACHED_NODES=0, but that's besides the point:

That happens even without this PR, so not a blocker here. I'll submit another issues once I get a better idea what happens there. Valgrind points at a long stack of recursive `void hipsycl::rt::(anonymous namespace)::descend_requirement_tree` calls, and randomness hints at a data race.

illuhad:
>  That happens even without this PR, so not a blocker here. I'll submit another issues once I get a better idea what happens there. Valgrind points at a long stack of recursive void hipsycl::rt::(anonymous namespace)::descend_requirement_tree calls, and randomness hints at a data race.

Thanks for the confirmation, I had expected this. This PR does not change the non-instant submission code paths at all, so I would have been very surprised if this had changed things.
The randomness could also be due to the DAG descent ignoring nodes that have been marked as completed. So it could also depend on whether an operation has already finished and the GC has noticed this.

pszi1ard:
>     * Synchronous instant submission, which requires optimizing the submission path (likely including moving to small vector classes instead of `std::vector` and similar optimizations)
> 
>     * Asynchronous instant submission where a worker thread is involved (Or at least we should investigate the benefit of this - maybe not much is to be gained once the submission path is sufficiently optimized)
> 
> 
> Depending on the use case, one or the other might be chosen by the client application and/or the runtime.

The interesting questions is what is the CPU / GPU latency tradeoff with both schemes fully optimized, e.g. by how much are GPU task launches delayed by async instant submission, whether/when is this outweighed by the benefit of overlapping some GPU API overhead, whether this has an impact on the critical path of an application. I think making an automated choice will be very hard without knowing what the application does, but there may be cases easy to identify (e.g. pathological cases of high GPU API overhead).

illuhad:
I agree. I guess for simple cases we can have some autotuning code path where we make some measurements of the latency to offload to another thread versus just submitting synchronously, but there should also be some mechanism to let expert users enforce a particular behavior.
As I said, in any case I think we need to first examine how much performance we can extract from the synchronous submission code path to check whether it's even worth the effort to offload to another thread.

al42and:
I haven't analyzed it in depth, but what if, instead of analyzing all the dependencies of a task, we track the possibility of the submission via a flag on the queue? 

Task can be instantly submitted iff:

- task uses no buffers (USM-only)
- task uses no SYCL 2020 reductions
- ~~task has no running dependencies on non-instant tasks~~
- ~~task is bound to a device~~
- ~~task is hard-wired to an in-order backend_executor, which happens e.g. when using a SYCL in-order queue.~~
- task has no explicit dependencies (`handler::depends_on` directly or via shortcuts)
- task is submitted to an "instant" queue.

Conditions on a queue being "instant":

- An out-of-order queue is always "non-instant";
- An in-order queue starts "instant";
- An in-order queue becomes "non-instant" if a non-instant task is submitted into it (i.e., the one with buffers, reductions, or dependencies)
- An in-order queue becomes "instant" again after `queue::wait()` is called (there might be some edge-cases here, so perhaps it can never revert back to being instant).

This way, we spare ourselves the need to walk through all the dependencies before submission, hopefully cutting away some overhead.

illuhad:
I suspect that would not help much. For an in-order queue, the number of dependencies is typical 0 to 1 such that there won't be much difference to what you suggest. I really don't think that walking the dependencies itself is what is costing here.

al42and:
> A [quick-and-dirty annotation with rocTX](https://github.com/al42and/hipSYCL/commit/35a706cdcf11d8d850af5fc60a6a551d8eb91484) was not particularly enlightening, mostly due to 1µs resolution and, apparently, the fact that the overhead is spread between multiple locations.

Perfetto is to blame; the JSONs generated by roctrace have proper nanosecond resolution (note: profiler overhead is unknown): [8390_instant_on.json.gz](https://github.com/AdaptiveCpp/AdaptiveCpp/files/12729500/8390_instant_on.json.gz)

Below is not a proper analysis, but an average of three randomly-selected submission events. "Delay at start" is the delay between the start of this function and the start of the function below, in nanoseconds. "Delay at end" is the delay between the end of the function below and the start of this one. For "submit_kernel", the "function below" is the trio of HIP API calls. Times are measured by manually placed markers (see code linked above). A lot of variability in total submission times come from `hipLaunchKernel` (3-5 µs), but the overheads are more consistent.

Region | Delay at start | Delay at end
-- | -- | --
Submit GPU work | 150 | 1112
[queue::submit](https://github.com/al42and/hipSYCL/commit/35a706cdcf11d8d850af5fc60a6a551d8eb91484#diff-d60da68953254de8a41984f58bd446a9b9a986f140778b9e9d4a5c72baeed402R339) | 778 | 201
[queue::execute_submission](https://github.com/al42and/hipSYCL/commit/35a706cdcf11d8d850af5fc60a6a551d8eb91484#diff-d60da68953254de8a41984f58bd446a9b9a986f140778b9e9d4a5c72baeed402R910) | 1216 | 137
[handler::create_task](https://github.com/al42and/hipSYCL/commit/35a706cdcf11d8d850af5fc60a6a551d8eb91484#diff-4f098e568c780b37071ceecd6e75e664b4ff78f8af0085af21f29a7b0bc5f549R916) | 1757 | 93
[submit_directly](https://github.com/al42and/hipSYCL/commit/35a706cdcf11d8d850af5fc60a6a551d8eb91484#diff-3284bed84a5a4100db1f56198904339b39103b1bb7dd5058672f9cc2843dd062R111) | 547 | 969
[submit_kernel](https://github.com/al42and/hipSYCL/commit/35a706cdcf11d8d850af5fc60a6a551d8eb91484#diff-93f84fe9b862a679cc9b716e2ae78475450ad56e678685c46194c3f9a0291de8R345) | 1149 | 284

So the conclusion still holds, the overhead is annoyingly spread out between multiple places :(

illuhad:
Some things I'd like to try:
- Use bitmasks/bitmap in `execution_hints` for very efficiently querying whether a specific hint is there (very easy)
- Use a small vector implementation instead of `std::vector` for execution hints and `requirement_list` and dag node dependencies (easy, but a bit more involved)
- Spawn dag nodes and operations from object pools. (This is probably a larger change, so might be a project for later)

I'm officially on vacation until Oct 4, so my ability to do this at the moment is limited.

al42and:
> Spawn dag nodes and operations from object pools. (This is probably a larger change, so might be a project for later)

Perhaps using PMR with ([un](https://en.cppreference.com/w/cpp/memory/unsynchronized_pool_resource))[synchronized_pool_resource](https://en.cppreference.com/w/cpp/memory/synchronized_pool_resource) would be easier? It should still cut down the overhead of memory allocations significantly without a need to change code much besides just passing the allocator around.

illuhad:
Now depends on #1178. On my system with gfx906, the instant submission test program above is now ~10% faster (it was however always substantially faster than non-instant submission for me so YMMV).

One thing I noticed is that if Gromacs sets the `hipSYCL_coarse_grained_event` property per submission, I would recommend constructing the property list once and then reusing it, as constructing a `property_list` also requires dynamic memory allocation. This is harder to avoid, because the interface that SYCL has for `property_list` requires some form of type erasure. Once you have constructed it though, it can be passed to other submissions with little cost.

al42and:
> One thing I noticed is that if Gromacs sets the hipSYCL_coarse_grained_event property per submission, I would recommend constructing the property list once and then reusing it, as constructing a property_list also requires dynamic memory allocation. This is harder to avoid, because the interface that SYCL has for property_list requires some form of type erasure. Once you have constructed it though, it can be passed to other submissions with little cost.

Thanks for the suggestion, we'll look into that.

BTW, do you know any reason there is no `constexpr` constructor for `property_list` in the standard? Some properties, of course, might require runtime info, but currently the standard forbids constexpr construction of even empty property list. Here, we can live with `static const`, so I'm just curious.

illuhad:
> BTW, do you know any reason there is no constexpr constructor for property_list in the standard? Some properties, of course, might require runtime info, but currently the standard forbids constexpr construction of even empty property list. Here, we can live with static const, so I'm just curious.

`property_list` was already there in SYCL 1.2.1, so I suspect it was originally designed before `constexpr` was a priority. 
Performance-wise, an empty property list should not perform any operations either (or at least not more than default-constructing `std::vector`), but of course there are some additional semantic implications of `constexpr`.

Intel is working on adding support for compile-time properties to `property_list`, I assume `constexpr` support would also come as part of this.

illuhad:
Given that this is opt-in for now, I think we should merge given that for some use cases it will provide a performance-boosting option.

PR #1132, opened 2023-09-12
Allow for non-default-constructible iterators in `std::for_each_n`

nilsfriess:
When the macro `HIPSYCL_STDPAR_OFFLOAD` is used in `std::for_each_n`, the type `return_type` is `ForwardIt` (in the C++ standard it is called `InputIt`) and this iterator only has to meet the requirements of a `LegacyInputIterator` which does not include default-constructibility. By replacing the if/else with a ternary operator, the declaration that causes a call to the (possibly non-existent) default constructor can be removed.

Edit: ~I just realised that I accidentally removed a call to `q.wait()` in the second macro; ill fix that asap.~ Fixed it.


illuhad (approved):
Thanks!

PR #1133, opened 2023-09-13
[CI] Run self-hosted runners only for actions started from main repository

nilsfriess:
In forked repositories, these actions just hang for about a day and then time out, so we might as well just not run them at all.


illuhad:
Thanks! Is there some variable that we can use that points to the repository name of the fork's origin? Otherwise this will break once we rename ;)

nilsfriess:
I changed it to use the repo ID instead of the name which (hopefully) is the same even after changing the name.

illuhad (approved):
Thanks!

PR #1141, opened 2023-09-18
Add `get_backend` to Interop handle

### LINKED ISSUE #1103: `get_backend` missng in `interop_handle`
**Bug summary**
At some point SYCL2020 added the `get_backend`  method to `interop_handle` class.
Currently DPC++ implements this, would be great if hipSYCL can as well for conformance.

normallytangent:
Fixes #1103

### DIFF
--- a/tests/sycl/interop_handle.cpp
+++ b/tests/sycl/interop_handle.cpp
@@ -26,7 +26,6 @@
  */
 
 
-
 #include "sycl_test_suite.hpp"
 #include <boost/test/unit_test_suite.hpp>
 
--- a/tests/sycl/interop_handle.cpp
+++ b/tests/sycl/interop_handle.cpp
@@ -44,8 +44,6 @@ BOOST_AUTO_TEST_CASE(interop_handle_api) {
     d = sycl::device(sycl::cpu_selector_v);
   }
 
-  s::queue q(d);
-  
   rt::device_id assigned_device{rt::backend_descriptor{rt::hardware_platform::cpu,
                                 rt::api_platform::omp}, 12345};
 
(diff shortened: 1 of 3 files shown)


tests/sycl/interop_handle.cpp
```diff
@@ -0,0 +1,59 @@
...
+  s::queue q(d);
+  
+  rt::device_id assigned_device{rt::backend_descriptor{rt::hardware_platform::cpu,
+                                rt::api_platform::omp}, 12345};
+
+  rt::backend_executor *executor = nullptr;
+  s::interop_handle ih{assigned_device, executor};
+  s::backend b = ih.get_backend();
+
+  BOOST_CHECK(b == q.get_device().get_backend());
```
illuhad: This is not correct. `b` should always be the OpenMP backend because that's what you use for `assigned_device`. `q` may however be constructed using an arbitrary different backend.

Suggestion: You don't need `q`, just remove it and check whether `b` == s::backend::omp`.

tests/sycl/interop_handle.cpp
```diff
@@ -0,0 +1,57 @@
...
+  namespace s = cl::sycl;
+  using namespace hipsycl;
+
+  sycl::device d;
+
+  try {
+    d = sycl::device(sycl::gpu_selector_v);
+  }
+  catch (sycl::exception const &e) {
+    d = sycl::device(sycl::cpu_selector_v);
```
illuhad: `d` seems to be no longer in use, so maybe just remove that part?

illuhad (approved):
Thanks!

PR #1157, opened 2023-09-23
update build instructions - easy

jamesreinders:
Simple updates that align the build instructions for today - making cut and paste work when following these instructions


illuhad (approved):
Thanks James!

PR #1163, opened 2023-09-28
add parallel_for overloads with a number

bashbaug:
We have been updating our SYCL book samples for revision two of the book and we have been trying to test them with as many SYCL implementations and backends as we can.

Many of our samples use a form of `parallel_for` where the range is specified with just a number, for brevity.  For example, our [first sample](https://github.com/Apress/data-parallel-CPP/blob/main/samples/Ch01_introduction/fig_1_1_hello.cpp) has:

```c++
  q.parallel_for(sz, [=](auto& i) {
     result[i] -= 1;
   }).wait();
```

rather than:

```c++
  q.parallel_for(range<1>(sz), [=](auto& i) {
     result[i] -= 1;
   }).wait();
```

This syntax doesn't currently work in AdaptiveCpp though, and the explicit `range<1>` syntax is currently required.

The book syntax is explicitly allowed in the spec, see [parallel_for invoke](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_parallel_for_invoke):

> The parallel_for overload without an offset can be called with either a number or a braced-init-list with 1-3 elements. In that case the following calls are equivalent:
> * `parallel_for(N, some_kernel)` has same effect as `parallel_for(range<1>(N), some_kernel)`

In practice, I think this means that we need an explicit overload for at least this case.  This PR adds the explicit overloads.  We may need explicit overloads for the other multi-dimensional braced-init-list cases also, but that can be done in a separate PR if needed, and we do not use this syntax in our book samples currently.

There are still some failures in the book samples after this fix, which we are looking into, but many of the samples are building and running well with AdaptiveCpp with this change 👍 .


illuhad (dismissed):
This is awesome to see! Yeah, I had done some experiments to try to get the explicit conversion from `size_t` to `range<Dim>` to work, but that never worked so I agree that explicit overloads might be required (is this how DPC++ solves this?).

In any case I agree that just passing a plain number should be supported.

illuhad:
Ok, CI says that now calls with `range<1>` argument are ambiguous :( How does DPC++ solve that problem?

EDIT: Maybe `enable_if` so that the templated `range<Dim>` overload only participates if Dim > 1?

bashbaug:
> is this how DPC++ solves this?

Yes, I believe so, DPC++ has explicit overloads for the 1D, 2D, and 3D ranges: [link](https://github.com/intel/llvm/blob/aa195fd714ccd41742f86335e30d2dfd5108c4f4/sycl/include/sycl/queue.hpp#L2291).

I'll look into the CI failure.  I'm a little surprised by the ambiguous call though.  Here is what we were experimenting with, where we thought the one overload would be sufficient (for the scalar case, at least): https://godbolt.org/z/Kbz78deh6

illuhad:
The CI failure seems to be triggered by the `parallel_for` overload that has the additional deprecated `offset` argument. My guess is, since there is no explicit `range<1>, id<1>, ReductionsAndKernel` overload, the compiler does not know whether the `id<1>` argument should be part of the variadic template pack (in which case it could match `range<1>` exactly), or whether it should take the existing overload which still has the offset.

So probably it could be solved either with explicit overloads that include the `id<1>` offset, or perhaps by constraining reductions better such that `id<1>` is discarded for the variadic template argument.

bashbaug:
OK, the easiest fix seems to be to add the explicit overload for the offset case also.  Since the SYCL spec only describes parallel_for overloads for the cases without an offset (not sure why), debatably this will allow some cases to compile that really shouldn't.  For example, this will compile with acpp after these changes:

```c++
int size = 16;
int offset = 0;
h.parallel_for(size, id<1>(offset), ...);
```

Since the offset case was deprecated in SYCL 2020 maybe this is acceptable?  I've added this overload for now, and the tests that were failing previously are now building.

The other option to consider is to `enable_if` based on properties of the template parameter pack, specifically if it only contains reductions and a kernel.  This appears to be what DPC++ is doing ([link](https://github.com/intel/llvm/blob/aa195fd714ccd41742f86335e30d2dfd5108c4f4/sycl/include/sycl/handler.hpp#L2343)).  I haven't checked to see how hard it would be to do something similar in acpp.

illuhad:
> Since the offset case was deprecated in SYCL 2020 maybe this is acceptable? I've added this overload for now, and the tests that were failing previously are now building.

Yeah, we can just call it an extension ;) I think we shouldn't put more effort than necessary into deprecated features.

illuhad (approved):
Thanks Ben!

PR #1167, opened 2023-10-04
Print more detailed version information; allow custom version suffix

### LINKED ISSUE #1165: More detailed version reporting

`acpp --acpp-version` report the manually set version (e.g., "AdaptiveCpp version: 23.10.0"), which does not uniquely identify the actual version of the source code. E.g., all the versions between 7c63e9bcb44b8cae36971c611e689764cfb98f2e (Jan 2023) and 522bd1af96c91d240cd509fd644cfe4d79a7ebe7 (Sep 2023) are reported as "0.9.4", despite many big changes happening.

This makes it harder for user-developers, who might use intermediate versions between official releases, to correctly identify the version used when reporting bugs, publishing results, or just reviewing old logs.

1. Reporting the hash of the commit used to build the code, not just the hardcoded version.

2. Optionally, make it possible to set a custom suffix to the version (e.g., if some distribution ships AdpativeCpp 23.10.0 with a custom patch to fix a search path or something, the reported version could be "23.10.0-ubuntu"). Currently, that would require manually patching `CMakeLists.txt` or the installed `syclcc.json`.

- Users already can set `version-patch` in `syclcc.json` as they please manually.

illuhad:
* By default, the git commit hash is appended to the version as `+gitHASH`, with `HASH` being the git hash.
* Suffix can be overridden using `-DACPP_VERSION_SUFFIX` in cmake
* This also adds a new field to `syclcc.json` called `version-suffix`.
* Changes macros in code from `HIPSYCL_VERSION_<TYPE>` to `ACPP_VERSION_<TYPE>`. Whether macros should better use `ADAPTIVECPP_` prefix still needs to be discussed; I've used `ACPP_` for consistency with cmake.

The new version string is e.g. presented when invoking `acpp --acpp-version`, but might also be visible in the version number that the clang plugin presents itself, or the `version_string()` that our SYCL headers use.

Fixes #1165


CMakeLists.txt
```diff
@@ -3,11 +3,23 @@ if(NOT CMAKE_VERSION VERSION_LESS 3.12)
...
+if(NOT ACPP_VERSION_SUFFIX)
+  # Get the latest abbreviated commit hash of the working branch
+  execute_process(
+    COMMAND git log -1 --format=%h
+    WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+    OUTPUT_VARIABLE ACPP_GIT_COMMIT_HASH
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+  )
+
+  set(ACPP_VERSION_SUFFIX "+git${ACPP_GIT_COMMIT_HASH}")
```
al42and: When building from a tarball (or if git fails for any other reason), the `ACPP_GIT_COMMIT_HASH` will be just `+git`, which is quite misleading. I suggest checking the success of the git query:

```suggestion
    OUTPUT_VARIABLE ACPP_GIT_COMMIT_HASH
    RESULT_VARIABLE GIT_RETURN_CODE
    OUTPUT_STRIP_TRAILING_WHITESPACE
  )

  if (GIT_RETURN_CODE EQUAL 0)
    set(ACPP_VERSION_SUFFIX "+git${ACPP_GIT_COMMIT_HASH}")
  endif()
```

illuhad: Good point, thanks!

pszi1ard:
A few tips: 
- it can be very useful to have a date in the version string of a build from git since git hashes do not help understanding whether a specific build is from a day a month or a year ago. 
- as a developer, you may find it useful to know whether your build is from a dirty tree or not ;)

There are a few more cases you may want consider, as inspiration take a look at our versioning:
https://gitlab.com/gromacs/gromacs/-/blob/main/cmake/gmxVersionInfo.cmake?ref_type=heads#L167-193

and here is the code we use to generate the git-based version string: https://gitlab.com/gromacs/gromacs/-/blob/main/cmake/gmxGenerateVersionInfo.cmake

illuhad:
Thanks for the suggestions @pszi1ard :-) I have expanded the version string to include head date, git branch and dirty flag. Example string is now:
```
$ install/bin/acpp --acpp-version
acpp [AdaptiveCpp compilation driver], Copyright (C) 2018-2023 Aksel Alpay and the AdaptiveCpp project
  AdaptiveCpp version: 23.10.0+git.f19037f9.20231013.branch.feature/detailed-version-information.dirty
```

PR #1172, opened 2023-10-05
[SSCP][llvm-to-spirv] Do not use cmake INSTALL_DIR to install llvm-spirv translator

### LINKED ISSUE #1169: CMake error when building SSCP as non-root user with root-owned CMAKE_INSTALL_PREFIX
**Bug summary**

The idea is to install AdaptiveCpp system-wide (e.g., in `/usr/local`), while running as a non-root user for the majority of the build steps (e.g., `cmake .... && make && sudo make install`).

However, `cmake` tries to create a directory inside `CMAKE_INSTALL_PREFIX`, which fails due to permissions.

**To Reproduce**

```console
aland@tcbl16:~/AdaptiveCpp/build$ git describe --dirty --tags 
v0.9.3-291-g7c259679
$ cmake --version | head -1
cmake version 3.27.6

aland@tcbl16:~/AdaptiveCpp/build$ cmake ..
[...]
-- Performing Test CXX_FILESYSTEM_NO_LINK_NEEDED
-- Performing Test CXX_FILESYSTEM_NO_LINK_NEEDED - Success
CMake Error at build/src/compiler/llvm-to-backend/LLVMSpirvTranslator-prefix/tmp/LLVMSpirvTranslator-mkdirs.cmake:6 (file):
  file failed to create directory:

    /usr/local/lib/hipSYCL/ext/llvm-spirv

  because: No such file or directory
  /usr/share/cmake-3.27/Modules/ExternalProject.cmake:1744 (include)
  /usr/share/cmake-3.27/Modules/ExternalProject.cmake:4281 (_ep_set_directories)
  src/compiler/llvm-to-backend/CMakeLists.txt:121 (ExternalProject_Add)
[...]
```

AdaptiveCpp sets `INSTALL_DIR` for LLVMSpirvTranslator to  `${CMAKE_INSTALL_PREFIX}/lib/hipSYCL/ext/llvm-spirv`:

https://github.com/AdaptiveCpp/AdaptiveCpp/blob/7c259679e2ea4687980bceb0ac59a36f260efb4d/src/compiler/llvm-to-backend/CMakeLists.txt#L116-L127

But this is incorrect, according to https://cmake.org/pipermail/cmake/2011-March/043406.html:

> This is a bug in the project, not in CMake. It uses the
> ExternalProject.cmake module in a bogus way, probably by setting the
> INSTALL_DIR option of ExternalProject_Add() to somewhere in the
> CMAKE_INSTALL_PREFIX. This is wrong, it should be somewhere in the build
> tree, and then the client project should install the necessary files itself.

As I understand (but have not verified), `INSTALL_DIR` should not be set at all (or set to some temporary directory), and instead `-DCMAKE_INSTALL_PREFIX:PATH=${LLVMSPIRV_INSTALLDIR}` should be passed via `CMAKE_CACHE_ARGS`.

**Expected behavior**

`cmake` able to configure the project without needing root privileges.

**Describe your setup**
* g7c259679
* Describe the dependencies that AdaptiveCpp sits on top of in your setup: 
   * Clang-17 from LLVM APT repos
   * GCC 11.4.0 from Ubuntu repos

**Additional context**

Not a concern for GROMACS, just reporting a weird behavior I noticed.

illuhad:
Hopefully fixes #1169


al42and:
Can confirm that it fixes #1169.

al42and:
My bad, should have tested one step further.

Now it requires `root` access during build, which is better than during configuration stage, but still not ideal:

```console
$ cmake -S ~/AdaptiveCpp/ -B build
[...]
-- Configuring done (8.8s)
-- Generating done (0.0s)
-- Build files have been written to: /tmp/build
$ cmake --build build  -- -j12
[...]
[ 95%] Built target LLVMSPIRVLib
[100%] Linking CXX executable llvm-spirv
[100%] Built target llvm-spirv
Install the project...
-- Install configuration: ""
CMake Error at cmake_install.cmake:54 (file):
  file cannot create directory:
  /usr/local/lib/hipSYCL/ext/llvm-spirv/include/LLVMSPIRVLib.  Maybe need
  administrative privileges.
```

It also rebuilds the libSPIRV every time the build command is rung, even if the code and the cache are completely unchanged (this behavior is not specific to this PR).

illuhad:
> It also rebuilds the libSPIRV every time the build command is rung, even if the code and the cache are completely unchanged (this behavior is not specific to this PR).

This is because we need to patch llvm-spirv translator and cmake has no mechanism to figure out whether the patch is already applied, so it has to rebuild. It's massively annoying. Unfortunately it seems things get bleak when you want to include an external git dependency that needs patching with cmake.

> Now it requires root access during build, which is better than during configuration stage, but still not ideal:

Right. I'm not sure how to fix that, short of just manually copying `llvm-spirv` to the destination directory during install (which might be brittle if llvm-spirv uses rpath, or changes their libraries etc).

al42and:
Yaaay, packaging C++ dependencies is so fun. Well, I guess it can't be helped then. This MR already improves things: at least we don't need to run CMake as root, and we don't misuse `INSTALL_DIR` in `ExternalProject_Add`, so perhaps that's enough.

One solution would be to allow using an external `LLVMSpirvTranslator` installation. This would also play nicer with package managers, which might wish to handle dependencies themselves. Probably not terribly complicated to do, but a pain to test.

illuhad:
> One solution would be to allow using an external LLVMSpirvTranslator installation. This would also play nicer with package managers, which might wish to handle dependencies themselves. Probably not terribly complicated to do, but a pain to test.

How would this work, given that we need a patched llvm-spirv translator? If someone plugged in an external llvm-spirv translator that does not have the patch (or is even built against a different LLVM version) the consequence is incorrect code generation.

al42and:
> How would this work, given that we need a patched llvm-spirv translator? 

Ah, right, the indexing of address spaces needs to be harmonized with the rest. 

> I'm not sure how to fix that, short of just manually copying llvm-spirv to the destination directory during install (which might be brittle if llvm-spirv uses rpath, or changes their libraries etc).

I guess the "uses rpath" issues might be solved by manually setting `INSTALL_RPATH`, and "changes their libraries", IMO, is less likely than some internal reorganization that breaks the patching process. Still perhaps too much effort for a relatively niche use-case.

illuhad:
> I guess the "uses rpath" issues might be solved by manually setting INSTALL_RPATH, and "changes their libraries", IMO, is less likely than some internal reorganization that breaks the patching process. Still perhaps too much effort for a relatively niche use-case.

I do wonder whether there is a way to tell cmake to run the install target of external projects within our install target, not within the build target. Since we don't need llvm-spirv during the build itself, that would solve our problem..

illuhad:
Ok, I think I have found (and pushed to this branch) a solution that addresses both problems:
* `make && sudo make install` pattern works as expected
* `llvm-spirv` is not built for every single invocation of `make`.


The way it works feels a bit hacky, and I'd appreciate if someone with more cmake experience could have a look at it and tell whether this is an acceptable approach or totally UB:
* Use `FetchContent` instead of `ExternalProject` to fetch the sources
* Only if content is not yet populated apply the patch and run the cmake configuration. This happens with two `execute_process` calls (is there a better solution?).
* Then use a custom build target to run `cmake --build --config Release` for LLVM-SPIRV translator if the project is built
* Use a custom `install()` invocation to invoke another `execute_process()` to get get `cmake --install` to run when the project is installed.


From what I can tell:
* Patch is applied correctly
* llvm-spirv is rebuilt whenever the cmake configuration changes, but not for every single `make`
* `make && sudo make install` can be used to install to privileged locations.

illuhad:
I guess the best course of action might be to merge and see if any issues pop up.

PR #1195, opened 2023-10-17
[CI][CBS] Enable LLVM 17 in CI

fodinabor:
This enables LLVM 17 to CI.

HIP SMCP needed a bit of a workaround to fully disable any extra passes landing in the device pipeline from CBS.
(InstCombine triggers some invalid addrspace cast issue)


illuhad (approved):
Thanks!

PR #1196, opened 2023-10-17
[CI] Expand LIT infrastructure to SSCP and add some SSCP CI tests

illuhad:
* Expands the LIT infrastructure for SSCP and adds test cases for issues that were fixed recently.
* The existing CBS LIT github workflow is now run as part of a new "Linux LIT tests" workflow, which also runs SSCP tests.
* Adds SSCP LIT to github CI via the Intel CPU OpenCL implementation.

In the future we should also run the tests on our self-hosted runners to test the other JIT GPU targets. And of course add more test cases :)


fodinabor (approved):
LGTM :)

PR #1199, opened 2023-10-20
Fix compatibility with git pre-2.6

al42and:
CentOS/RHEL 7 with git 1.8 is still used (e.g., Karolina at Technical University of Ostrava), and using `--date=format:` syntax is not supported there, leading to `version-suffix` being empty.

See #1165


illuhad (approved):
Thanks!

PR #1201, opened 2023-10-20
Warn if using different LLVM then specified by LLVM_DIR

al42and:
Previously, if `LLVM_DIR` was set incorrectly, ACpp would silently use another LLVM installation when available. Now we issue a warning in this case.

It is easy to set `-DLLVM_DIR=/opt/rocm/llvm/lib/cmake/` instead of `-DLLVM_DIR=/opt/rocm/llvm/lib/cmake/llvm/`, and then miss the "Building hipSYCL against LLVM configured from" message in the long CMake output.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -196,7 +196,11 @@ if(WITH_OPENCL_BACKEND OR WITH_LEVEL_ZERO_BACKEND)
 endif()
 
 if(BUILD_CLANG_PLUGIN)
+  set(LLVM_DIR_OLD ${LLVM_DIR})
   find_package(LLVM CONFIG)
+  if(LLVM_DIR_OLD AND NOT ("${LLVM_DIR_OLD}" STREQUAL "${LLVM_DIR}"))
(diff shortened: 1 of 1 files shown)


illuhad (approved):
Great idea!

I have sent an invite to you to join the AdaptiveCpp github organization. The main effect is that this will make CI workflows run automatically for you, without needing approval first.
It will also allow you to approve workflow runs on other PRs; feel free to do this, but please have a glance at the PR first to ensure no obviously sketchy or malicious code is executed on our self-hosted CI infrastructure.

al42and:
> I have sent an invite to you to join the AdaptiveCpp github organization. The main effect is that this will make CI workflows run automatically for you, without needing approval first.

Got it, thanks!

As I understand, my approvals now count for the purpose of merging PRs?

illuhad:
> As I understand, my approvals now count for the purpose of merging PRs?

From github's perspective yes, although in practice so far my PRs usually go in bypassing github's review requirements because there is rarely someone who reviews them :(
If you want to review, by all means do it :-)

PR #1204, opened 2023-10-23
Set CMake version and CXX standard in examples

Calandracas606:
Explicitly set c++ 17

clang <= 15 will not default to -std=c++17, which may cause IDEs using clangd (CLion) to report errors.

This change will improve developer experience, and provide a more complete "example" of a CMakeLists.txt


illuhad (approved):
Thanks!

PR #1207, opened 2023-10-24
Remove use of deprecated getWithSamePointeeType

al42and:
llvm::PointerType::getWithSamePointeeType is deprecated in LLVM 17 and can be removed later.

Reference: https://reviews.llvm.org/D155585


fodinabor (approved):
Lgtm, thanks :)

illuhad (approved):
Thanks! We should consider factoring this ifdef branching out into a new function such that the logic can be shared, but that could also be done in a future commit I think.

PR #1208, opened 2023-10-24
CBS/canonicalizeExitBarriers: Don't return uninitialized value

al42and:


fodinabor (approved):
🙃 thx :)
Are you compiling with all the warnings activated? ^^

al42and:
> 🙃 thx :) Are you compiling with all the warnings activated? ^^

clang-tidy. Tons of warnings there, but this and the one with `std::move` are the straightforward ones. I can share the log (or you can run it yourself, `-DCMAKE_CXX_CLANG_TIDY=clang-tidy-17` is the only CMake flag necessary), but, unfortunately, I won't have time to look into the remaining issues.

illuhad (approved):

PR #1209, opened 2023-10-24
fix target includes destination to match new nested include dir

### LINKED ISSUE #1179: Link with AdaptiveCpp
Regarding #1069. so what's now a proper way to link with acpp and propagate include dirs to cmake targets which hawe a library using acpp as a dependency?

bsobol:
This should fix #1179 

Results in `adaptivecpp-targets.cmake` containing 
```cmake
set_target_properties(AdaptiveCpp::acpp-common PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include;${_IMPORT_PREFIX}/include/AdaptiveCpp"
)  
```
instead of original
```cmake
set_target_properties(AdaptiveCpp::acpp-common PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
)  
```
and similarily for `acpp-rt` and other namespaces.

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -46,6 +46,7 @@ endif()
 
 install(TARGETS acpp-common
         EXPORT install_exports
+        INCLUDES DESTINATION include/AdaptiveCpp
         RUNTIME DESTINATION bin
         LIBRARY DESTINATION lib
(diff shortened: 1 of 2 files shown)


illuhad (approved):
Thanks!

PR #1213, opened 2023-10-25
[stdpar] Do not set visibility of `operator delete` and friends to `HiddenVisibility`

nilsfriess:
This has caused some segfaults due to some standard library functions not being able to "see" our `delete` and therefore freeing USM-allocated memory using regular `delete`/`free`.

This was introduced to prevent our own runtime library to call the USM/stdpar `delete` but is not necessary anymore since the runtime library links with symbolic functions which causes `operator delete` to be resolved internally.


illuhad (commented):
Thanks, I think the documentation on stdpar also mentions the visibility. Can you change it there as well?

doc/stdpar.md
```diff
@@ -90,7 +90,7 @@ The backend used to perform USM allocations is the backend managing the executin
...
+Functions for memory allocation are only exchanged for USM variants within translation units compiled with `--acpp-stdpar`. Our USM functions for releasing memory are however overriding the standard functions within the entire linkage unit. This is motivated by the expectation that pointers may be shared within the application, and the place where they are released may not be the place where they are created. As our functions for freeing memory can handle both regular and USM allocations, making them more widely available seems like the safer choice.
```
illuhad:
```suggestion
Functions for memory allocation are only exchanged for USM variants within translation units compiled with `--acpp-stdpar`. Our USM functions for releasing memory are however overriding the standard functions within the entire application. This is motivated by the expectation that pointers may be shared within the application, and the place where they are released may not be the place where they are created. As our functions for freeing memory can handle both regular and USM allocations, making them more widely available seems like the safer choice.
```

PR #1216, opened 2023-10-26
CMake: Check availability of libLLVM when building SSCP

al42and:
Error if it cannot be found. Theoretically, in this case static linking might work, but it did not for me and LLVM from ROCm 5.7.0.

More elegant solutions are welcome, but I haven't found a way to robustly implement heuristics for automatically disabling SSCP in top-level CMakeLists.txt. But hey, at least it's not a weird linking error anymore.

Inspired by https://github.com/secure-software-engineering/phasar/blob/de6d50500f4958fbbafec3daba95f3d4a15ea85d/Config.cmake.in#L45C7-L52

Linking to libLLVM: https://github.com/llvm/llvm-project/blob/7f677fe3100131214386f9ce1fa308c235a595e9/llvm/cmake/modules/LLVM-Config.cmake#L95

### DIFF
--- a/src/compiler/llvm-to-backend/CMakeLists.txt
+++ b/src/compiler/llvm-to-backend/CMakeLists.txt
@@ -29,7 +29,7 @@ function(create_llvm_based_library)
   target_compile_definitions(${target} PRIVATE ${LLVM_DEFINITIONS} -DHIPSYCL_COMPILER_COMPONENT)
   find_library(LLVM_LIBRARY NAMES LLVM HINTS ${LLVM_LIBRARY_DIRS})
   if(NOT LLVM_LIBRARY)
-    message(FATAL_ERROR "LLVM at ${LLVM_DIR} does not have libLLVM.so. Please disable SSCP and OpenCL (-DWITH_SSCP_COMPILER=OFF -DWITH_OPENCL_BACKEND=OFF) or choose another LLVM installation")
+    message(FATAL_ERROR "LLVM at ${LLVM_DIR} does not have libLLVM.so. Please disable SSCP and related backends (-DWITH_SSCP_COMPILER=OFF -DWITH_OPENCL_BACKEND=OFF -DWITH_LEVEL_ZERO_BACKEND=OFF) or choose another LLVM installation")
   endif()
   llvm_config(${target} USE_SHARED core support irreader passes)
   # We need symbolic functions for stdpar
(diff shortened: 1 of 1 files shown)


src/compiler/llvm-to-backend/CMakeLists.txt
```diff
@@ -27,6 +27,10 @@ function(create_llvm_based_library)
     ${LLVM_TO_BACKEND_INCLUDE_DIRS})
   
   target_compile_definitions(${target} PRIVATE ${LLVM_DEFINITIONS} -DHIPSYCL_COMPILER_COMPONENT)
+  find_library(LLVM_LIBRARY NAMES LLVM HINTS ${LLVM_LIBRARY_DIRS})
+  if(NOT LLVM_LIBRARY)
+    message(FATAL_ERROR "LLVM at ${LLVM_DIR} does not have libLLVM.so. Please disable SSCP and OpenCL (-DWITH_SSCP_COMPILER=OFF -DWITH_OPENCL_BACKEND=OFF) or choose another LLVM installation")
```
illuhad: I think in principle you also need `-DWITH_LEVEL_ZERO_BACKEND=OFF`. The only reason why L0 does not cause this issue more frequently is because it is currently not auto-detected by cmake, unlike OpenCL. But if a user explicitly enables it, it would probably behave the same way.

(Independently of this PR, we should look into how we can resolve the dependency between SSCP <-> L0, OpenCL in a more intuitive, automatic way. Ideally `WITH_SSCP_COMPILER=OFF` would also disable depending backends)

illuhad (commented):
Thank you! I think this is the right approach. 

The reason why we currently link dynamically is because LLVM has some global objects like command line handlers, and it does not like if those are constructed multiple times (which might happen if multiple backends load their SSCP JIT compiler, and thus LLVM). Perhaps it's still possible with static linkage, but some more investigation is needed.

src/compiler/llvm-to-backend/CMakeLists.txt
al42and: > I think in principle you also need -DWITH_LEVEL_ZERO_BACKEND=OFF. T

Good point, done.

> (Independently of this PR, we should look into how we can resolve the dependency between SSCP <-> L0, OpenCL in a more intuitive, automatic way. Ideally WITH_SSCP_COMPILER=OFF would also disable depending backends)

I fully agree, but my attempts at it produced something obscenely ugly :( I feel that, with so many feature flags, a proper dependency handling must be implemented, rather than manually coding a forest of `if()`s to handle all the possible combinations of manually set flags, automatically set flags, and environment factors while also providing helpful error messages.

illuhad (approved):
Thanks!

PR #1217, opened 2023-10-26
acpp-info: Add --help (-h) and --list-devices (-l) flags

al42and:
Add `-l` option to only print devices and backends without further details (inspired by `clinfo -l`).
Since we now have options, also added `-h`.

### DIFF
--- a/src/tools/acpp-info/acpp-info.cpp
+++ b/src/tools/acpp-info/acpp-info.cpp
@@ -25,6 +25,8 @@
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <iostream>
+
 #include "hipSYCL/runtime/application.hpp"
 #include "hipSYCL/runtime/backend.hpp"
 #include "hipSYCL/runtime/runtime.hpp"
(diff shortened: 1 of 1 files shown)


src/tools/acpp-info/acpp-info.cpp
```diff
@@ -189,15 +191,41 @@ void list_devices(rt::runtime* rt) {
...
+    std::cout << "Usage: " << exe_name << " [options]\n\n";
+    std::cout << "Options:\n";
+    std::cout << "\t-h\t\tShow this message.\n";
+    std::cout << "\t-l\t\tOnly list backends and devices, without detailed info\n";
+}
+
+int main(int argc, char *argv[]) {
+  bool print_device_details = true;
+  for (int arg = 1; arg < argc; arg++) {
+    if (std::string(argv[arg]) == "-h") {
```
illuhad:
```suggestion
    if (std::string{argv[arg]} == "-h" || std::string{argv[arg]} == "--help") {
```

src/tools/acpp-info/acpp-info.cpp
```diff
@@ -189,15 +191,41 @@ void list_devices(rt::runtime* rt) {
...
+      print_help(argv[0]);
+      return 0;
+    }
+    else if (std::string(argv[arg]) == "-l") {
```
illuhad:
```suggestion
    else if (std::string{argv[arg]} == "-l" || std::string{"--list-devices"}) {
```

src/tools/acpp-info/acpp-info.cpp
```diff
@@ -189,15 +191,41 @@ void list_devices(rt::runtime* rt) {
   });
 }
 
-int main() {
+void print_help(const char* exe_name)
+{
+    std::cout << "Usage: " << exe_name << " [options]\n\n";
+    std::cout << "Options:\n";
+    std::cout << "\t-h\t\tShow this message.\n";
```
illuhad:
```suggestion
    std::cout << "\t-h,--help\t\tShow this message.\n";
```

src/tools/acpp-info/acpp-info.cpp
```diff
@@ -189,15 +191,41 @@ void list_devices(rt::runtime* rt) {
...
+    std::cout << "\t-l\t\tOnly list backends and devices, without detailed info\n";
```
illuhad:
```suggestion
    std::cout << "\t-l,--list-devices\t\tOnly list backends and devices, without detailed information\n";
```

illuhad (commented):
Great idea! I had until now always used something like `ACPP_DEBUG_LEVEL=3 ./acpp | head`, but this is definitely more convenient :)

Perhaps we can add spelled out versions of these flags? I usually spell `--help` out.. old habits ;)

al42and:
> Perhaps we can add spelled out versions of these flags? I usually spell --help out.. old habits ;)

Good point. Done in 0a87843bb0a94384ac2f35c8a2674c12a48f831c

illuhad (approved):
Thanks!

PR #1218, opened 2023-10-26
Hide incompatible OpenCL devices and platforms

### LINKED ISSUE #1203: Hide non-supported OpenCL devices

OpenCL backend of AdaptiveCpp requires "SPIR-V compatible OpenCL devices" (per README).

Notably, NVIDIA and AMD ROCm platforms don't support SPIR-V, and even correctly report that by returning empty `CL_DEVICE_IL_VERSION`.

Hiding such devices by default seems like a reasonable choice.

- If the device can't be used via OpenCL backend (e.g., it does not support SPIR-V), it should not be presented.

illuhad:
Fixes #1203 

Previously by default OpenCL devices were exposed even if there were potential incompatibilities. This PR hides such platforms and devices by default.

* Entire OpenCL platforms are hidden if the OpenCL version is < 3.0.
* Individual OpenCL devices are hidden if they don't support SPIR-V (makes kernel execution impossible), or support neither USM extensions nor fine-grained system SVM (makes memory allocations impossible).

The environment variable `ACPP_RT_OCL_SHOW_ALL_DEVICES=1` can be set to override this behavior and expose all devices.

This PR should cause AdaptiveCpp to not pick up by default Intel FPGA Emulator, AMD OpenCL and NVIDIA OpenCL.

Incompatible devices or platforms cause a warning to be emitted with further information why they were ignored. This should help in case a user is wondering why these devices don't show up. Possible output:

```
$ ACPP_VISIBILITY_MASK="omp;ocl" install/bin/acpp-info
[AdaptiveCpp Warning] backend_loader: Could not load backend plugin: /home/aksel/src/hipSYCL/master/build/install/lib/hipSYCL/librt-backend-hip.so
[AdaptiveCpp Warning] libamdhip64.so.5: cannot open shared object file: No such file or directory
[AdaptiveCpp Warning] ocl_hardware_manager: Platform 'Intel(R) FPGA Emulation Platform for OpenCL(TM)' does not support OpenCL 3.0; hiding platform. Set ACPP_RT_OCL_SHOW_ALL_DEVICES=1 to override. (OpenCL version string was: 'OpenCL 1.2 Intel(R) FPGA SDK for OpenCL(TM), Version 20.3')
[AdaptiveCpp Warning] ocl_hardware_manager: Platform 'Intel(R) FPGA Emulation Platform for OpenCL(TM)' does not support OpenCL 3.0; hiding platform. Set ACPP_RT_OCL_SHOW_ALL_DEVICES=1 to override. (OpenCL version string was: 'OpenCL 1.2 Intel(R) FPGA SDK for OpenCL(TM), Version 20.3')
[AdaptiveCpp Warning] ocl_hardware_manager: Platform 'Portable Computing Language' does not support OpenCL 3.0; hiding platform. Set ACPP_RT_OCL_SHOW_ALL_DEVICES=1 to override. (OpenCL version string was: 'OpenCL 2.0 pocl 1.8  Linux, None+Asserts, RELOC, LLVM 11.1.0, SLEEF, DISTRO, POCL_DEBUG')
[AdaptiveCpp Warning] ocl_hardware_manager: OpenCL device 'NVIDIA GeForce MX150' does not support SPIR-V; hiding device. Set ACPP_RT_OCL_SHOW_ALL_DEVICES=1 to override.
=================Backend information===================
[...]
```

### DIFF
--- a/doc/env_variables.md
+++ b/doc/env_variables.md
@@ -17,5 +17,5 @@
 * `ACPPL_RT_MAX_CACHED_NODES`: Maximum number of nodes that the runtime buffers before flushing work.
 * `ACPP_SSCP_FAILED_IR_DUMP_DIRECTORY`: If non-empty, hipSYCL will dump the IR of code that fails SSCP JIT into this directory.
 * `ACPP_RT_GC_TRIGGER_BATCH_SIZE`: Number of nodes in flight that trigger a garbage collection job to be spawned
-* `ACPP_RT_OCL_NO_SHARED_CONTEXT`: If set to a non-zero value, instructs the OpenCL backend to not attempt to construct a shared context across devices within a platform. This can be necessary on OpenCL implementations that do not support this. Note that if shared contexts are unavailable, support for data transfers between devices might be limited as the devices can no longer directly talk to each other.
-* `ACPP_RT_OCL_SHOW_ALL_DEVICES`: If set to a non-zero value, instructs the OpenCL backend to expose all found devices, even if those might be incompatible with AdaptiveCpp or unable to execute kernels.
+* `ACPP_RT_OCL_NO_SHARED_CONTEXT`: If set to `1`, instructs the OpenCL backend to not attempt to construct a shared context across devices within a platform. This can be necessary on OpenCL implementations that do not support this. Note that if shared contexts are unavailable, support for data transfers between devices might be limited as the devices can no longer directly talk to each other.
+* `ACPP_RT_OCL_SHOW_ALL_DEVICES`: If set to `1`, instructs the OpenCL backend to expose all found devices, even if those might be incompatible with AdaptiveCpp or unable to execute kernels.
--- a/doc/env_variables.md
+++ b/doc/env_variables.md
@@ -18,3 +18,4 @@
 * `ACPP_SSCP_FAILED_IR_DUMP_DIRECTORY`: If non-empty, hipSYCL will dump the IR of code that fails SSCP JIT into this directory.
 * `ACPP_RT_GC_TRIGGER_BATCH_SIZE`: Number of nodes in flight that trigger a garbage collection job to be spawned
 * `ACPP_RT_OCL_NO_SHARED_CONTEXT`: If set to a non-zero value, instructs the OpenCL backend to not attempt to construct a shared context across devices within a platform. This can be necessary on OpenCL implementations that do not support this. Note that if shared contexts are unavailable, support for data transfers between devices might be limited as the devices can no longer directly talk to each other.
+* `ACPP_RT_OCL_SHOW_ALL_DEVICES`: If set to a non-zero value, instructs the OpenCL backend to expose all found devices, even if those might be incompatible with AdaptiveCpp or unable to execute kernels.
--- a/src/runtime/ocl/ocl_hardware_manager.cpp
+++ b/src/runtime/ocl/ocl_hardware_manager.cpp
@@ -84,6 +84,7 @@ bool should_include_platform(const std::string& platform_name, const cl::Platfor
                              "version for platform " +
                                  platform_name,
                              error_code{"CL", err}});
+    return false;
   } else {
     int ocl_version_major = 0;
     if (!parse_ocl_version_string(ocl_version, ocl_version_major)) {
--- a/src/runtime/ocl/ocl_hardware_manager.cpp
+++ b/src/runtime/ocl/ocl_hardware_manager.cpp
@@ -36,6 +36,7 @@
 #include <CL/cl.h>
 #include <CL/opencl.hpp>
 #include <cstddef>
+#include <array>
 #include <optional>
 
 
@@ -43,6 +44,15 @@ namespace hipsycl {
 namespace rt {
 
 namespace {
+
+// These commonly encountered OpenCL platforms will be silently hidden due to
+// known incompatibilities. The hardware they support is eithers supported via other backends,
+// or not supported at all (e.g. FPGA). All other platforms will be hidden with a warning if
+// they fail feature support queries.
+constexpr std::array incompatible_ocl_platforms = {
+    "NVIDIA CUDA", "Intel(R) FPGA Emulation Platform for OpenCL(TM)",
+    "AMD Accelerated Parallel Processing"};
+
 template<int Query, class ResultT>
 ResultT info_query(const cl::Device& dev) {
   ResultT r{};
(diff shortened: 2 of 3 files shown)


src/runtime/ocl/ocl_hardware_manager.cpp
```diff
@@ -55,6 +56,97 @@ ResultT info_query(const cl::Device& dev) {
...
+          << "ocl_hardware_manager: Could not parse OpenCL version string "
+              "of platform '"
+          << platform_name
+          << "'; hiding platform. Set ACPP_RT_OCL_SHOW_ALL_DEVICES=1 "
+              "to override. (OpenCL version string was: '"
+          << ocl_version << "')" << std::endl;
+      return false;
+    } else {
+      if (ocl_version_major < 3) {
+        HIPSYCL_DEBUG_WARNING
```
al42and, line 122: I would argue this should be `HIPSYCL_DEBUG_LEVEL_INFO`. While there is no universally accepted definition of logging levels, e.g., [Python](https://docs.python.org/3/library/logging.html#logging-levels) says the following about warning:

> An indication that something unexpected happened, or that a problem might occur in the near future (e.g. ‘disk space low’). The software is still working as expected.

All AMD and Intel users will have at such a non-compatible platform in their default install, so such situation is hardly "unexpected" (unlike the "Could not retrieve/parse OpenCL version" messages above).

[Rust](https://docs.rs/log/latest/log/enum.Level.html#variant.Warn) defines warning as "hazardous situations", which is also not something happening here, IMHO.

src/runtime/ocl/ocl_hardware_manager.cpp
```diff
@@ -55,6 +56,97 @@ ResultT info_query(const cl::Device& dev) {
...
+}
+
+bool should_include_device(const std::string& dev_name, const cl::Device& dev) {
+  const bool show_all_devices = application::get_settings().get<setting::ocl_show_all_devices>();
+  if(show_all_devices)
+    return true;
+
+  if(info_query<CL_DEVICE_IL_VERSION, std::string>(dev).find("SPIR-V") ==
+           std::string::npos) {
+    HIPSYCL_DEBUG_WARNING
```
al42and, line 143: Similarly, I would argue that the presence of an OpenCL device from NVIDIA can hardly be considered "unexpected" or "hazardous", so this should be an INFO.

doc/env_variables.md
```diff
@@ -17,4 +17,5 @@
...
 * `ACPP_RT_GC_TRIGGER_BATCH_SIZE`: Number of nodes in flight that trigger a garbage collection job to be spawned
-* `ACPP_RT_OCL_NO_SHARED_CONTEXT`: If set to a non-zero value, instructs the OpenCL backend to not attempt to construct a shared context across devices within a platform. This can be necessary on OpenCL implementations that do not support this. Note that if shared contexts are unavailable, support for data transfers between devices might be limited as the devices can no longer directly talk to each other.
\ No newline at end of file
+* `ACPP_RT_OCL_NO_SHARED_CONTEXT`: If set to a non-zero value, instructs the OpenCL backend to not attempt to construct a shared context across devices within a platform. This can be necessary on OpenCL implementations that do not support this. Note that if shared contexts are unavailable, support for data transfers between devices might be limited as the devices can no longer directly talk to each other.
+* `ACPP_RT_OCL_SHOW_ALL_DEVICES`: If set to a non-zero value, instructs the OpenCL backend to expose all found devices, even if those might be incompatible with AdaptiveCpp or unable to execute kernels.
```
al42and: Nit: setting it to `2` (or `42`) does not work:

```suggestion
* `ACPP_RT_OCL_SHOW_ALL_DEVICES`: If set to `1`, instructs the OpenCL backend to expose all found devices, even if those might be incompatible with AdaptiveCpp or unable to execute kernels.
```

al42and (approved):

src/runtime/ocl/ocl_hardware_manager.cpp
al42and: I suggest offering a simple extension to check whether any devices were hidden (`bool sycl::backend::ext_adaptivecpp_has_hidden_devices()`?), so that `acpp-info` can display a warning in this case ("Note: some incompatible OpenCL devices hidden; set ACPP_RT_OCL_SHOW_ALL_DEVICES=1 to show them"), while the default behavior does not raise any repeated, unactionable, and usually irrelevant warnings.

al42and: Such extension can also take into account whether any devices were hidden by `ACPP_VISIBILITY_MASK` (or even by backend-specific means like `(CUDA|ROCR)_VISIBLE_DEVICES`).

Such messages are useless when setting the filters inline (`ACPP_VISIBILITY_MASK="omp;ocl" install/bin/acpp-info`), but SLURM or various scripts might set such variables unbeknownst to user. It's not polite to override them, but it might be helpful to know of them.

illuhad: >  All AMD and Intel users will have at such a non-compatible platform in their default install, so such situation is hardly "unexpected" (unlike the "Could not retrieve/parse OpenCL version" messages above).

I don't have AMD OpenCL in my ROCm, so it depends on which packages you've actually installed. But to the point: Isn't it a matter of perspective what is "unexpected"?
I think you are probably thinking of the unaware end user, who does not know the difference between OpenCL and CUDA and then might be confused by the warning. I see this point.
But I think there could also be a class of users where silently hiding devices might be the unexpected thing. E.g. a user tries to get AdaptiveCpp to work on OpenCL devices that we are not familiar with (e.g. on mobile platforms). Or they are from an OpenCL vendor, and are looking to extent hardware support to SYCL. In that case, they might expect that their expected device shows up, and then it can be very helpful to be warned that it won't due to reasons XYZ.

Perhaps we need to distinguish between "commonly known OpenCL implementation with commonly known issues and alternative backend available" and the other case of an unknown OpenCL implementation?

Otherwise yeah, limiting the warning to `acpp-info` could be an option, if users are aware enough to try it. But even in that case it might be difficult to forward the reason for hiding the device through such a query. I'm not sure we can easily do that at the moment.

al42and: >  But I think there could also be a class of users where silently hiding devices might be the unexpected thing. E.g. a user tries to get AdaptiveCpp to work on OpenCL devices that we are not familiar with (e.g. on mobile platforms). Or they are from an OpenCL vendor, and are looking to extent hardware support to SYCL. In that case, they might expect that their expected device shows up, and then it can be very helpful to be warned that it won't due to reasons XYZ.

I see the utility in this case, but such "advanced" users are (a) less numerous than the "casual" users (hopefully, as SYCL get more popular, much less numerous :slightly_smiling_face:); (b) likely will quickly try setting a higher `HIPSYCL_DEBUG_LEVEL`, where the message will be printed.

> Perhaps we need to distinguish between "commonly known OpenCL implementation with commonly known issues and alternative backend available" and the other case of an unknown OpenCL implementation?

Agree, that's perhaps the best solution. Rather straightforward and strikes the balance between the two groups.

doc/env_variables.md
illuhad: Right, this is because we use `istream` with `bool` which is only well-defined for `0` and `1`. But the new variable is not the only one affected since all share the same infrastructure; I have thus also changed the documentation for another boolean variable which had a similar description.

src/runtime/ocl/ocl_hardware_manager.cpp
illuhad: I have changed it now such that it
* first checks the platform name against commonly known problematic OpenCL implementations (NVIDIA, AMD, Intel FPGA Emulator). If there is a match, it rejects the platform with an `HIPSYCL_DEBUG_INFO` message.
* Otherwise it proceeds to check the platform and device capabilities as before. If it finds an issue there, it emits a warning.

al42and (approved):

PR #1221, opened 2023-10-29
[SSCP] Add initial sycl::stream support

illuhad:
Currently, using `sycl::stream` inside SSCP causes a JIT failure, because it directly calls `printf` which is not available in SSCP kernels.

This PR
* Adds a new SSCP builtin, `__hipsycl_sscp_print(const char*)`, which prints a charater string to the screen. Implementations for SPIR-V, PTX and amdgcn are also provided.
* Causes `stream` class to call `__hipsycl_sscp_print()` if `operator<<` is invoked with a  string argument. For other arguments, it currently prints that it does not support printing that type.

Of course in the long term, we will need to implement `stream` for other types as well. For this purpose we will have to come up with a device formatting library for ints, floats etc, which might be a larger job.

While this PR thus does not provide a complete implementation as only `char*` and `const char*` arguments are currently supported, it allows us to run the classic "Hello World" (which is also part of the SYCLAcademy curriculum).
And for more complex cases with different types, the user is now actually informed about the problem instead of confronting them with a JIT failure message which might be almost impossible to understand for the average user.

### DIFF
--- a/include/hipSYCL/sycl/libkernel/sscp/builtins.hpp
+++ b/include/hipSYCL/sycl/libkernel/sscp/builtins.hpp
@@ -25,32 +25,6 @@
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-/*
- * This file is part of hipSYCL, a SYCL implementation based on CUDA/HIP
- *
- * Copyright (c) 2021 Aksel Alpay
- * All rights reserved.
- *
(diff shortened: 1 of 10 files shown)


illuhad:
Tested on Intel, NVIDIA, and AMD.

On my AMD machine, it runs successfully but does not actually print anything. However, the exact same behavior is also observed with a "Hello World"-HIP program and hipcc. So my guess is that this is due to immature printf-support in ROCm, and not due to our code.

PR #1222, opened 2023-10-30
Add missing prefixes when checking if function is in `std` namespace

nilsfriess:
According to [the Itanium ABI spec](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangling-compression) certain classes in the `std` namespace have their own abbreviations and do not start with `St` so the current check would not mark them as `std` functions. However, I am not sure if these abbreviations are actually used by the LLVM Itanium mangler. I haven't looked at the source code yet but I looked at the symbols of some test applications using `nm` and all mangled names of `std::` functions started with `_ZSt`, `_ZNSt` or `_ZNKSt` (this last one is for `const` member functions and was missing previously), so maybe it is enough to just check for these three. 

Opening this as a draft for now until I'm sure which is the correct approach (hoping that  someone who knows more about this sees this and can comment on it).

### DIFF
--- a/src/compiler/stdpar/MallocToUSM.cpp
+++ b/src/compiler/stdpar/MallocToUSM.cpp
@@ -100,7 +100,7 @@ bool isRestrictedToRegularMalloc(llvm::Function* F) {
 }
 
 bool isStdFunction(llvm::Function* F) {
-  static const std::string StdPrefixes[21] =
+  static const std::string StdPrefixes[] =
     {
       "_ZSt", "_ZNSt", "_ZNKSt",
       "_ZSa", "_ZNSa", "_ZNKSa",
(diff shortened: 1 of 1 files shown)


src/compiler/stdpar/MallocToUSM.cpp
```diff
@@ -100,9 +100,21 @@ bool isRestrictedToRegularMalloc(llvm::Function* F) {
 }
 
 bool isStdFunction(llvm::Function* F) {
+  static const std::string StdPrefixes[21] =
```
illuhad:
```suggestion
  static const std::string StdPrefixes[] =
```
Let the compiler figure out the length?

illuhad (commented):
Good catch. Even if it is a draft now, I wonder if this should go into the 23.10 release as a fix. What do you think?

nilsfriess:
I opened #1224 which only contains the additional check for `const` member functions. I'm quite sure that this is actually missing, so I think we can safely include #1224 in the release but should do more research about the other checks included in this PR.

src/compiler/stdpar/MallocToUSM.cpp
nilsfriess: Whoops, I only put the length there explicitly it for me to be sure that I added all prefixes that I wanted to add

PR #1223, opened 2023-10-30
HIP: Support hipPointerAttribute_t from ROCm 6

al42and:
In ROCm 5.5, `hipPointerAttribute_t::memoryType` was deprecated in favor of `hipPointerAttribute_t::type`.

The old version will be removed in ROCm 6: https://rocm.docs.amd.com/en/docs-5.7.0/release.html#upcoming-changes-for-hip-in-rocm-6-0-release

Haven't actually tested ROCm 6, so there might be other issues.


illuhad (approved):
Thanks!

PR #1224, opened 2023-10-30
[stdpar] Add check for `const` std member functions

nilsfriess:
In the function that checks whether a method is in the `std::` namespace, a check for `const` member functions was missing.

In #1222 a few more checks are included but this one added here is the only one where I'm sure that it's really missing. So I think this one can safely go into the 23.10 release.


illuhad (approved):
Thanks!

Issue #1225, opened 2023-10-30
Distributing SYCL Applications with AdaptiveCpp (Windows)


blinkfrog:
Hello,

I've been working on a SYCL-based application using AdaptiveCpp, and I've run into challenges when distributing the compiled application. While the app runs perfectly on my development machine, it fails to initialize the CUDA backend on other systems, unless they have a full AdaptiveCpp installation. This requirement is a significant barrier to entry for end-users.

Specifics:
I've compiled my app with --acpp-targets="cuda:sm_75"
The application works fine on my machine.
Distributing the app with only hipSYCL-rt.dll and cudart64_12.dll is insufficient.
Paths to the AdaptiveCpp folder appear to be hardcoded in hipSYCL-rt.dll. If I rename AdaptiveCPP folder then the app stops to work on my machine too.

Questions:
Is there a way to make AdaptiveCpp look for its backend DLLs in the application's folder, or specify the path programmatically?
Are there options for static linking?
Are there any best practices or recommended approaches for distributing AdaptiveCpp-dependent applications in a user-friendly manner?
I'd appreciate any guidance or advice on how to distribute an AdaptiveCpp-dependent application more easily. Thank you!

illuhad:
> Is there a way to make AdaptiveCpp look for its backend DLLs in the application's folder, or specify the path programmatically?

It should already do something like this, and it seems there is also a dedicated code path for Windows:

https://github.com/AdaptiveCpp/AdaptiveCpp/blob/3952b468c9da89edad9dff953cdcab0a3c3bf78c/src/runtime/backend_loader.cpp#L158

The idea is that it should first try to find out where `libhipSYCL-rt` / `libacpp-rt` is located, and look for backend plugins relative to that path. If that fails, it should look in the installation directory that got hardcoded.
I don't know though why it doesn't seem to work in your case.

Overall, the deployment model of AdaptiveCpp, and SYCL in general, is poorly understood at the moment. It's not clear yet where in the stack between drivers and application the line should be drawn between components that should be provided by the system, and components that are provided by the application.

Runtime backends are plugins, so static linking, even if it worked, would defeat that design purpose. In fact, it would make deployment more difficult because if e.g. some backend dependency is unavailable (e.g. CUDA is not found), the entire application will fail instead of the current behavior where it can just skip that backend in this case.

For your use case, I think we should try to gain an understanding of why the logic I have linked fails.

[...]

PR #1227, opened 2023-11-01
[Win] MinGW generates dlls with lib prefix.

fodinabor:
Fixes the primary issue from #1225 where on Windows, the backend dlls were note found if they are not in the original `CMAKE_INSTALL_PREFIX` anymore, but were distributed somewhere else.


illuhad (approved):
Thanks! (We really should make `acpp-rt` use `acpp-common` for the backend loader filesystem operations, then I think all of this filesystem logic would be contained in one place)

PR #1228, opened 2023-11-01
Make CUDA detection error not fatal

al42and:
If a broken installation is present, a fatal error CMake error was throws, even if WITH_CUDA_BACKEND=OFF was set. We downgrade it to warning now and mark CUDA as not found.


illuhad:
Out of curiosity, where did you encounter a CUDA installation without nvvm subdirectory?

illuhad (approved):

al42and:
It was uninstalled, but not completely for some reason, and was still detected by CMake (albeit completely non-functional)

PR #1232, opened 2023-11-07
CMake Fails when building from release tarball

Calandracas606:
When building from a release tarball, CMake fails on `string(REPLACE "-" "" ACPP_GIT_DATE ${ACPP_GIT_DATE})`

This is because release tarballs do not have the .git directory, causing the $ACPP_GIT_DATE to be empty, and the CMake replace to fail

Without this change, ACPP_VERSION_SUFFIX will need to be set when building from release tarball

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,17 @@ endif()
 set(ACPP_VERSION_MAJOR 23)
 set(ACPP_VERSION_MINOR 10)
 set(ACPP_VERSION_PATCH 0)
+
+execute_process(
+        COMMAND git status
+        WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+        RESULT_VARIABLE GIT_STATUS
+)
+# With a release tarball, "git status" will fail (return non zero)
+if(GIT_STATUS)
+  set(ACPP_VERSION_SUFFIX "Release")
+endif()
+
 if(NOT ACPP_VERSION_SUFFIX)
   # Get the latest abbreviated commit hash of the working branch
   execute_process(
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,56 +13,57 @@ execute_process(
         WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
         RESULT_VARIABLE GIT_STATUS
 )
-# With a release tarball, "git status" will fail (return non zero)
-if(GIT_STATUS)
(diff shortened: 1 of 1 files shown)


al42and:
Thanks for reporting and the fix!

I'm not sure, though, that setting "Release" suffix is a great idea. `git status` might fail for another reason, and then a random version would get tagged as "release". Since we cannot reliably determine the version without a working `git`, leaving the suffix empty is safer, IMO.

(P.S.: I'm not the maintainer here, just a user / frequent contributor; so the above is just my thoughts, not a direct request for change)

Calandracas606:
> I'm not sure, though, that setting "Release" suffix is a great idea. `git status` might fail for another reason, and then a random version would get tagged as "release". Since we cannot reliably determine the version without a working `git`, leaving the suffix empty is safer, IMO.

I thought that this might be the case. 

Would leaving the suffix empty possible cause problems? As an alternative to leaving the suffix empty, maybe it should be set to "Unknown"?

I'll wait for further input

fodinabor:
I'd second that. `release` is not sth we would want for random versions.
One could have the `stable` branch carry a patch that sets this to release and otherwise have it say `develop` or so? Release tags should then be created from `stable` only.

al42and:
> As an alternative to leaving the suffix empty, maybe it should be set to "Unknown"?

I think this will have the opposite problem, being unfair to honest release versions build from a tarball :)

> Would leaving the suffix empty possible cause problems?

During the original discussion about git-based version tagging (https://github.com/AdaptiveCpp/AdaptiveCpp/pull/1167#discussion_r1347264044), the empty suffix was deliberately chosen for such a case (before being broken in a later PR), so it would not cause any technical issues.

That said, I did not spend too much time thinking about upsides and downsides of this approach, and I suspect @illuhad did not conduct a deep research either. So, while empty suffix definitely works, I don't want to claim that it is the best solution.

illuhad:
> That said, I did not spend too much time thinking about upsides and downsides of this approach, and I suspect @illuhad did not conduct a deep research either. So, while empty suffix definitely works, I don't want to claim that it is the best solution.

That's true. But I would say since this PR wants to fix a concrete problem, let us first have this PR resolve the problem by falling back to empty string if there's no `.git` directory. This would then recover the behavior that was originally intended. 

Independently, whether there's a better approach compared to empty string could be discussed in a new issue or subsequent PR.

illuhad (approved):
Thanks everyone! I've changed the PR to use empty string in case git fails as discussed.

PR #1234, opened 2023-11-08
[SYCL2020] Make vec class `constexpr`-constructible

### LINKED ISSUE #1189: Missing constexpr for sycl::vec
**Bug summary**
According to the latest sycl spec `sycl::vec` has constexpr constructors: https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:vector.type Implementation is missing `constexpr`: https://github.com/AdaptiveCpp/AdaptiveCpp/blob/f577603d89f3b65575b456de084bc0799c886e29/include/hipSYCL/sycl/libkernel/vec.hpp#L305

**To Reproduce**
Try to compile `constexpr sycl::vec<int, 3> vec = {1, 2, 3};`

**Expected behavior**
`sycl::vec` should have constexpr constructor - no compiler errors for the expression above.

nilsfriess:
Fixes #1189

### DIFF
--- a/.github/workflows/linux-lit.yml
+++ b/.github/workflows/linux-lit.yml
@@ -2,6 +2,9 @@ name: Linux LIT tests
 
 on: [push, pull_request]
 
+env:
+  BOOST_TEST_LOG_LEVEL: test_suite
+
 jobs:
   cbs:
     name: CBS, clang ${{ matrix.clang }}, ${{ matrix.os }}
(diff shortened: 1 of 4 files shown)


nilsfriess:
Note sure why tests are failing again, maybe it's something like in #1173, looking into it.

nilsfriess:
The segfault seems to occur in the group function tests. @DieGoldeneEnte I vaguely remember that you had some problems with weird segfaults at some point, maybe this is the same and is fixed in #966?

PR #1237, opened 2023-11-10
[stdpar] Add struct reflection support; only offload if no invalid device pointers are found

illuhad:
* Adds initial infrastructure for reflection builtins;
* Adds first reflection builtin: `introspect_flattened_struct`, which can provide information about all members contained in aggregate data types (like structs or classes), such as member offsets, sizes and whether it is an int, a floating type, pointer or something else. This also supports nested structs; in this case the struct is flattened and the information is provided as per the data layout of the struct in memory. This operates on the LLVM IR level, so there is no distinction between e.g. pointers or references. **This builtin can be used to e.g. find, validate or otherwise introspect all pointers that are stored in a kernel lambda!**
* The builtin is used to add support for detecting PSTL calls that might reference host memory, and not offload in that case. As a side effect, this adds functional support for capture-by-reference in PSTL: Previously, something like

```c++
int y = 2;
std::for_each(... [&](auto& x){ x += y; });
```
might have been offloaded, and then crashed on device due to the reference to host-memory for `y`. With this PR, we detect that an argument to `for_each` contains a pointer that is not valid on device, and execute the call on the host to be on the safe side.

Note that, since the pointer validation is ultimately done at runtime, we can distinguish these cases:

```c++
// We rewire heap allocations to device-accessible USM allocations,
// so this call would get offloaded
int* y = some_function_that_returns_a_heap_ptr();
std::for_each(... [=](auto& x){ x += *y; });

// We would detect that z now does not contain a valid device pointer,
// and not offload.
int* z = some_function_that_returns_a_stack_ptr();
std::for_each(... [=](auto& x){ x += *z; });

```

@tom91136 I believe you might have some use case too for the new reflection functionality ;)


tests/compiler/reflection/introspect_struct.cpp
```diff
@@ -0,0 +1,54 @@
+// RUN: %acpp %s -o %t --acpp-targets=generic
+// RUN: %t
```
nilsfriess:
```suggestion
// RUN: %t | FileCheck %s
```

tests/compiler/reflection/introspect_struct.cpp
```diff
@@ -0,0 +1,54 @@
...
+// RUN: %acpp %s -o %t --acpp-targets=generic -O3
+// RUN: %t
```
nilsfriess:
```suggestion
// RUN: %t | FileCheck %s
```

tests/compiler/reflection/introspect_struct.cpp
```diff
@@ -0,0 +1,54 @@
...
+// RUN: %acpp %s -o %t --acpp-targets=generic -g
+// RUN: %t
```
nilsfriess:
```suggestion
// RUN: %t | FileCheck %s
```

tom91136:
Thanks for letting me know! This is very interesting indeed and should allow users to inject custom heuristics for memory movement in a safe way. I'll play around with this in the meantime.

illuhad:
At least modifications for such use cases would be limited to playing with the headers, instead of fiddling around with compiler/runtime internals :) And it works seamlessly across all our clang-based compilation flows: SSCP, cuda, hip, omp.accelerated.

I'd like to look into adding some heuristics for automatic prefetch/memadvise next, but we first need an efficient data structure to map pointers to some allocation parameters - particularly allocation size.

PR #1248, opened 2023-11-18
[stdpar] Reimplement synchronization elision using CFG analysis

illuhad:
This PR completely reimplements the synchronization elision optimization for stdpar. While it started out by looking into a regression preventing synchronization elision following #1237, this PR introduces a completely new approach that has several benefits.

Previously, there was a builtin call at the beginning and end of stdpar functions. When these two builtins met in the IR without memory accesses, control flow or function calls in between, synchronization was elided. This approach was fairly limited. In particular, it could not elide synchronization between stdpar calls e.g. in a loop:

```
for(int i = 0; i < num_iterations; ++i) {
  std::for_each(....);
}
```

It also failed if the compiler inserts store operations into the lambda struct to assemble the function argument prior to invoking to call. This pattern happens reliably as soon as arguments exceed trivial size.

The new approach addresses both issues:

It works as follows:
1. Detect stdpar functions using the new `hipsycl_stdpar_entrypoint` annotation attribute.
2. Detect calls to `__hipsycl_stdpar_optional_barrier()` inside stdpar functions.
3. Remove `__hipsycl_stdpar_optional_barrier()` calls from within the stdpar function definition, and reinsert them after the call instruction to the stdpar function (i.e. move them out of the stdpar function, and to the callsite)
4. Move calls to `__hipsycl_stdpar_optional_barrier()` down the instruction flow, taking all routes through the control flow graph until a place is encountered where barriers must be present for correctness:
  * memory accesses such as loads/stores
  * calls to other functions that are not stdpar calls, since we cannot know what these functions do and our control flow analysis currently cannot continue in other functions. Stdpar calls are ignored, which is what allows us to avoid synchronization in between them.
  * exit of control flow from the current function
  * If a barrier is already present at one of the determined insertion points, no additional barrier is inserted.
  * If a position where synchronization is needed is found, or a present barrier, that path in the control flow graph is considered finished, and the other paths are investigated.

Even if synchronization could not be elided completely, this algorithm might still yield a speedup because it effectively causes synchronization to be as delayed as possible.

`__hipsycl_stdpar_optional_barrier()` accesses an internal counter of enqueued operations, which is incremented by each stdpar call. `__hipsycl_stdpar_optional_barrier()` only synchronizes if that counter is > 0, and after a synchronization resets it. This is because the exact requirement for synchronization now depend on the precise route through the control flow graph that the program has taken, and in some control flow graphs it can happen that it hits multiple calls to `__hipsycl_stdpar_optional_barrier()` without a new stdpar call in between.

Store instructions are tricky and obey additional logic: Stores are commonly inserted by clang before stdpar calls to assemble function arguments, such as lambda objects passed to the stdpar algorithm. This is problematic, because hitting these stores can prevent barriers from being moved beyond them, and ultimately prevent the SyncElision optimization from becoming beneficial in many cases. Thus, a way is needed to distinguish stores that purely relate to assembling stdpar arguments from stores to memory locations that might actually be used inside kernels.

In theory, we know that stores to addresses that are then exclusively used for by-value
arguments to stdpar calls can be considered to be part of stdpar argument handling, and can thus
be skipped. In practice, we cannot easily determine this due to opaque pointers and clang/LLVM's
inconsistent use of the `byval(T)` attribute. Currently, we use the following heuristic to detect "harmless" stores that do not require synchronization:
* Consider stores to memory locations that originate from an alloca
* The original alloca memory location and all its derived uses must only be in the form of `getelementptr` instructions, stores, and the stdpar call.
* The store must occur in the same block as the stdpar call, and precede it in the instruction order.

I believe that there might be pathological cases where this returns false positives, especially in the presence of system USM where stack memory might be used inside kernels too. In practice, for cases where this becomes relevant we should not offload anyway because the problem size would be way too small to be an efficient offload use case. It's also not yet clear to me whether a meaningful program could be written that triggers such a pathological case.

This PR also introduces a test infrastructure for synchronization elision. This should help us detect issues and regressions better.

With the new approach, we can elide far more synchronization. For example, in the case of Lulesh, we find that we can enqueue up to 14 kernels before a wait is needed:

```
[AdaptiveCpp Info] [stdpar] Initializing wait for 14 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 5 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 14 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 5 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 14 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 5 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 14 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 5 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 14 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 5 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 14 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 5 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 14 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 5 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 14 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 5 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 14 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 5 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 14 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 5 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 14 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 5 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 14 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 5 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 14 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 5 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 14 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 5 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 14 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 5 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 14 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 5 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 14 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 2 operations
[AdaptiveCpp Info] [stdpar] Initializing wait for 1 operations
```

CC @tom91136 - perhaps this might also be relevant for you :)

### DIFF
--- a/.github/workflows/linux-lit.yml
+++ b/.github/workflows/linux-lit.yml
@@ -71,7 +71,7 @@ jobs:
           cd ${GITHUB_WORKSPACE}/build/tests-cpu
           LD_LIBRARY_PATH=${GITHUB_WORKSPACE}/build/install/lib ./sycl_tests
   sscp:
-    name: SSCP/Reflection, clang ${{ matrix.clang }}, ${{ matrix.os }}
+    name: SSCP/Reflection/stdpar, clang ${{ matrix.clang }}, ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
@@ -122,3 +122,9 @@ jobs:
           cd ${GITHUB_WORKSPACE}/build/tests-reflection  
           cmake -DACPP_TARGETS=generic -DAdaptiveCpp_DIR=${GITHUB_WORKSPACE}/build/install/lib/cmake/AdaptiveCpp  ${GITHUB_WORKSPACE}/tests
           LD_LIBRARY_PATH=${GITHUB_WORKSPACE}/build/install/lib make check-reflection
+      - name: run stdpar LIT tests
+        run: |
+          mkdir ${GITHUB_WORKSPACE}/build/tests-stdpar
+          cd ${GITHUB_WORKSPACE}/build/tests-stdpar
+          cmake -DACPP_TARGETS=generic -DAdaptiveCpp_DIR=${GITHUB_WORKSPACE}/build/install/lib/cmake/AdaptiveCpp  ${GITHUB_WORKSPACE}/tests
+          LD_LIBRARY_PATH=${GITHUB_WORKSPACE}/build/install/lib make check-stdpar
--- a/include/hipSYCL/compiler/stdpar/SyncElision.hpp
+++ b/include/hipSYCL/compiler/stdpar/SyncElision.hpp
@@ -35,23 +35,68 @@
 
 namespace hipsycl {
 namespace compiler {
-// Detects stdpar calls, replaces noinline with always_inline attribute
-// and ensure __hipsycl_stdpar_consume_sync() is right at the start of the function.
-class SyncElisionEntrypointPreparationPass
-    : public llvm::PassInfoMixin<SyncElisionEntrypointPreparationPass> {
+
+/// Implements the main synchronization elision logic. The idea is as follows:
+/// 1.) Detect stdpar functions using "hipsycl_stdpar_entrypoint" annotation attribute.
+/// 2.) Detect calls to __hipsycl_stdpar_optional_barrier() inside stdpar functions.
+/// 3.) Remove __hipsycl_stdpar_optional_barrier() calls, and reinsert them after the call
+/// instruction to the stdpar function (i.e. move them out of the stdpar function, and to the
+/// callsite) 4.) Move calls to __hipsycl_stdpar_optional_barrier() down the instruction flow,
+/// taking all routes through the control flow graph until a place is encountered where barriers
+/// must be present for correctness:
+///  - memory accesses such as loads/stores
+///  - calls to other functions that are not stdpar calls, since we cannot know what these functions
+///    do and our control flow analysis currently cannot continue in other functions
+///  - exit of control flow from the current function
+///
+/// If a barrier is already present at one of the determined insertion points, no additional
+/// barrier is inserted.
+/// If a position where synchronization is needed is found, or a present barrier, that path in the
+/// control flow graph is considered finished, and the other paths are investigated.
+///
+/// This algorithm effectively causes synchronization to be as delayed as possible, potentially
+/// even removing synchronization entirely between two stdpar calls.
+///
+/// In order to properly function, __hipsycl_stdpar_optional_barrier() should have an internal
+/// counter of enqueued operations, and only synchronize if that counter is > 0. This is because
+/// in some control flow graphs it can happen that there is a path that crosses multiple
+/// synchronization points.
+///
+/// Additional logic is applied when hitting store instructions: These are commonly inserted
+/// by clang before stdpar calls to assemble function arguments, such as lambda objects passed
+/// to the stdpar algorithm. This is problematic, because hitting these stores can prevent barriers
+/// from moving beyond them, and ultimately prevent this optimization from becoming relevant in
+/// most cases. Thus, a way is needed to distinguish stores that purely relate to assembling stdpar
+/// arguments from stores to memory locations that might actually be used inside kernels.
+///
+/// In theory, we know that stores to addresses that are then exclusively used for by-value
+/// arguments to stdpar calls can be considered to be part of stdpar argument handling, and can thus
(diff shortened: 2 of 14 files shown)


illuhad:
Based on the CI failures, it looks like there is an issue with the oneAPI repository, which we need for Intel CPU OpenCL for our SSCP LIT tests :(

```

2s
Run wget -O- https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS.PUB | gpg --dearmor | sudo tee /usr/share/keyrings/oneapi-archive-keyring.gpg > /dev/null
--2023-11-18 02:28:57--  https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS.PUB
Resolving apt.repos.intel.com (apt.repos.intel.com)... 23.58.175.22, 2600:1408:ec00:28b::4b23, 2600:1408:ec00:28a::4b23
Connecting to apt.repos.intel.com (apt.repos.intel.com)|23.58.175.22|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 4738 (4.6K) [application/vnd.exstream-package]
Saving to: ‘STDOUT’

     0K ....                                                  100% 4.59G=0s

2023-11-18 02:28:57 (4.59 GB/s) - written to stdout [4738/4738]

deb [signed-by=/usr/share/keyrings/oneapi-archive-keyring.gpg] https://apt.repos.intel.com/oneapi all main

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

Get:1 file:/etc/apt/apt-mirrors.txt Mirrorlist [142 B]
Hit:2 http://azure.archive.ubuntu.com/ubuntu jammy InRelease
Hit:3 http://azure.archive.ubuntu.com/ubuntu jammy-updates InRelease
Get:6 https://packages.microsoft.com/ubuntu/22.04/prod jammy InRelease [3611 B]
Hit:4 http://azure.archive.ubuntu.com/ubuntu jammy-backports InRelease
Hit:5 http://azure.archive.ubuntu.com/ubuntu jammy-security InRelease
Get:7 https://apt.repos.intel.com/oneapi all InRelease [4451 B]
Hit:8 https://ppa.launchpadcontent.net/ubuntu-toolchain-r/test/ubuntu jammy InRelease
Get:10 https://apt.repos.intel.com/oneapi all/main amd64 Packages [346 kB]
Get:11 https://apt.repos.intel.com/oneapi all/main all Packages [93.7 kB]
Hit:9 https://apt.llvm.org/jammy llvm-toolchain-jammy-16 InRelease
Fetched 447 kB in 1s (450 kB/s)
Reading package lists...
Building dependency tree...
Reading state information...
58 packages can be upgraded. Run 'apt list --upgradable' to see them.

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

Reading package lists...
Building dependency tree...
Reading state information...
ocl-icd-libopencl1 is already the newest version (2.2.14-3).
ocl-icd-libopencl1 set to manually installed.
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 intel-oneapi-runtime-compilers : Depends: intel-oneapi-runtime-openmp (= 2023.2.2-47) but 2024.0.0-49819 is to be installed
E: Unable to correct problems, you have held broken packages.
Error: Process completed with exit code 100.
```

I don't know who at Intel can help here - perhaps @IgorVorobtsov knows though :)

Issue #1244, opened 2023-11-14
Building AdaptiveCpp with omp.accelerated compilation flow on Windows


blinkfrog:
Hello. I am trying to build AdaptiveCpp with `omp.accelerated` compilation flow on Windows.

This is the configuring command I use:

`cmake .. -G Ninja -DCMAKE_C_COMPILER=c:/llvm/bin/clang.exe -DCMAKE_CXX_COMPILER=c:/llvm/bin/clang++.exe -DBOOST_ROOT=c:/libraries/boost_1_83_clangw17 -DLLVM_DIR=c:/llvm/lib/cmake/llvm -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_INSTALL_PREFIX=c:/AdaptiveCpp/ -DWITH_SSCP_COMPILER=OFF -DWITH_OPENCL_BACKEND=OFF -DWITH_LEVEL_ZERO_BACKEND=OFF -DWITH_ACCELERATED_CPU=ON`

First, there was a minor error:

```
[12/77] Building CXX object src/compiler/CMakeFiles/acpp-clang.dir/cbs/IRUtils.cpp.obj
FAILED: src/compiler/CMakeFiles/acpp-clang.dir/cbs/IRUtils.cpp.obj
[...]

PR #1250, opened 2023-11-21
Set LLVM_LIBS to LLVM_AVAILABLE_LIBS on Windows if not defined

blinkfrog:
## Overview
This pull request introduces a workaround for an issue where `LLVM_LIBS` is not set in the CMake configuration on Windows environments. This issue was previously discussed in Issue #1244 

## Problem Description
On Windows, `LLVM_LIBS` remains unset after calling `find_package(LLVM ...)`, leading to build issues in environments that rely on this variable.

## Proposed Workaround
The workaround checks if `LLVM_LIBS` is empty and, if so, sets it to the value of `LLVM_AVAILABLE_LIBS`. This aims to ensure the necessary LLVM libraries are linked, improving compatibility with Windows.

## Implementation Details
- Added a conditional check in `CMakeLists.txt` to test if `LLVM_LIBS` is not set.
- If `LLVM_LIBS` is empty, it is set to the value of `LLVM_AVAILABLE_LIBS`.

## Considerations
- It's unclear if `LLVM_AVAILABLE_LIBS` is semantically equivalent to `LLVM_LIBS`. `LLVM_AVAILABLE_LIBS` might contain a broader set of libraries than what's needed.
- Further discussion or review might be needed to ensure this workaround does not introduce unintended side effects, especially in non-Windows environments.

## Request for Feedback
I've implemented this workaround to address the immediate build issues on Windows. I welcome any feedback or suggestions, especially regarding the potential implications of using `LLVM_AVAILABLE_LIBS` as a substitute for `LLVM_LIBS`.

## Next Steps
This PR is intended to let the workaround run through CI to identify any potential issues across different environments. Looking forward to the community's input and results from CI tests.

Thank you for considering this contribution!


nilsfriess:
Maybe we should rather use `llvm_map_components_to_libnames(LLVM_LIBS component1 component2 ... )` instead? Not sure which components are actually required, looking at the linker errors in #1244, maybe this should suffice `llvm_map_components_to_libnames(LLVM_LIBS support analysis passes)`?

blinkfrog:
Thank you for your suggestion. This indeed would be a better solution and would allow to not link unecessarily libraries from a such broad set.
However, when I tried suggested solution I got a lot of "duplicate symbol" linking errors.
After some thinking and trial and error, I've found that this set is necessarily and sufficient: `llvm_map_components_to_libnames(LLVM_LIBS analysis core support passes)`. I'll push the changes.

Interesting what is the value of `LLVM_LIBS` on the Linux systems set after `find_package(LLVM ...)`? I mean, if it is broader than these 4 libs, then may be it is worth to widen this set (may be this can be useful for other AdaptiveCpp build configurations). Or, the contrary, may be it s worth to make this solution (`llvm_map_components_to_libnames(LLVM_LIBS analysis core support passes)`) global and not only for Windows, as it allows to link less libraries.

Thank you.

blinkfrog:
I've pushed new commit with changes, based on @nilsfriess review and suggestion.

nilsfriess (approved):

PR #1260, opened 2023-11-27
[CI] Use preinstalled LLVM on Windows runners 

nilsfriess:
I accidentally discovered that the Windows runners come with a version of LLVM 16 that successfully compiles AdaptiveCpp itself as well as the tests :open_mouth: . It was only necessary to add quotes around the `CMAKE_CXX_COMPILER` variable when constructing the `syclcc_launcher` line in `adaptivecpp-config.cmake.in` due to the space in Windows's `Program Files`.

Running the tests still fails, no idea why. I'll see if I can figure out what's failing exactly but if that turns out to be more difficult, I think it might make sense to disable that temporarily and at least have a Github action that compiles AdaptiveCpp and the tests.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -8,7 +8,6 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        clang: [15]
         os: [windows-2022]
     steps:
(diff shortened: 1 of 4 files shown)


nilsfriess:
Superseeded by #1291.

PR #1266, opened 2023-11-29
Add `-D_USE_MATH_DEFINES` to compile flags on Windows

### LINKED ISSUE #1176: Windows Compilation Error: Use of undeclared identifier 'M_PI'
When attempting to compile a SYCL program using AdaptiveCPP on Windows, I encountered errors related to the undeclared identifier M_PI. The issue seems to be that M_PI is not defined in Windows C++ libraries by default, unlike Linux libraries.

Here is an example of one of the original errors:

```
C:\AdaptiveCPP\bin\..\include\sycl/../hipSYCL/sycl/libkernel/generic/hiplike/builtins.hpp:489:11: error: use of
      undeclared identifier 'M_PI'
  return (M_PI / 180.f) * x;
          ^
```
Other errors occur in other `builtins.hpp` files.

Workarounds:

1. Manually defining M_PI in the AdaptiveCPP headers where the error occurs.

```
#ifndef M_PI
    #define M_PI 3.14159265358979323846
#endif
```

2. Defining _USE_MATH_DEFINES in my source files before including <sycl/sycl.hpp>.

While these workarounds solve the issue, former requires manual editing of AdaptiveCPP files, and latter reduces the portability of SYCL programs, because of need to modify the sources that are compiled on Windows.

Suggested Fix
Could the headers be updated to define M_PI when not already defined? Alternatively, would it be possible to conditionally define it only on Windows platforms to improve compatibility?

nilsfriess:
MSVC only defines math constants like `M_PI` (which is used in the builtins) if `_USE_MATH_DEFINES` is set.

Fixes #1176


illuhad (approved):
Thanks!

illuhad:
Can you resolve merge conflicts?

PR #1267, opened 2023-11-29
Log debug output to `stderr`

### LINKED ISSUE #1116: Consider logging to stderr instead of stdout
Open SYCL currently [prints warnings and errors to stdout](https://github.com/OpenSYCL/OpenSYCL/blob/67e972bee890eafa3e29243ecac9cb987226a395/include/hipSYCL/common/debug.hpp#L64). These lines mix with any output the user decides to print as part of normal program execution, and any script that wants to ingest the output of an Open SYCL program must manually strip out any log lines first.

Even though logging can be disabled through `HIPSYCL_DEBUG_LEVEL`, I would suggest to write log output to stderr instead, so that warnings can be seen even while output is captured.

normallytangent:
AdaptiveCpp currently logs debug output to `stdout`. According to the [standard](https://en.cppreference.com/w/c/io/std_streams) `stderr` is preferred for logging diagnostic output. 

Fixes #1116


VaiTon:
Any news on this one? When running like 
```
./a.out > data.csv
```

I sometimes get the JIT warning inside the data.csv :/

illuhad:
Hm, I forgot about this PR. Let me rerun CI to see if it comes back clean now. By the way, you can suppress error/warning output using the `ACPP_DEBUG_LEVEL` environment variable, which might be a workaround for your problem.

illuhad (approved):

PR #1274, opened 2023-12-04
[stdpar] SyncElision: Ignore loads/store to stack locations if we don't have system USM

illuhad:
In general, loads/stores between stdpar calls prevent synchronization elision. We do have some heuristics to ignore some stores that are merely used to assemble stdpar function arguments.
However, if we know that we don't have system USM, we can be more aggressive and ignore any load/stores that reference stack memory. This PR implements this.


illuhad:
Superseded by stdpar-mqd branch.

PR #1276, opened 2023-12-05
CMake: Add support for ccache and other compiler launchers

### LINKED ISSUE #1045: Support for CCache
[CCache](https://ccache.dev/) is a valuable tool for speeding up software builds in CI or other cases when the timestamp-based heuristics of `make` & co do not work.

From the looks of it, replacing

https://github.com/OpenSYCL/OpenSYCL/blob/12fdcaedfa990ab58ddf8bce304fa8cf917e6182/cmake/syclcc-launcher#L71-L74

with

```cpp
  if os.path.samefile(cxx_compiler_exe, command_in):
    command_line = [*syclcc_exe, *syclcc_specific_args, *command_in_args]
  elif os.path.basename(command_in) == 'ccache' and os.path.samefile(cxx_compiler_exe, command_in_args[0]):
    command_line = [command_in, *syclcc_exe, *syclcc_specific_args, *command_in_args[1:]]
  else:
    command_line = [command_in, *command_in_args]
```

works okay'ish with CMake (`-DCMAKE_C_COMPILER_LAUNCHER=/usr/bin/ccache -DCMAKE_CXX_COMPILER_LAUNCHER=/usr/bin/ccache -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++`), but I haven't tested it exhaustively.

Caching builds would be great, especially in multi-pass mode when targeting multiple architectures, but it may require more testing or better insight into all the compilation processes than I have.

EDIT: Upon further testing, the device builds (at least for `TARGETS=cuda:`) are also cached.

psalz:
This adds support for passing additional `CMAKE_CXX_COMPILER_LAUNCHER`s into `syclcc-launcher`, which is required for running with `ccache`.

Additionally I had to change the optimization warning in `acpp` to print to stderr, as ccache interprets messages to stdout as a sign that something went very wrong (and thus does not populate the cache).

Closes #1045.

### DIFF
--- a/cmake/syclcc-launcher
+++ b/cmake/syclcc-launcher
@@ -63,14 +63,22 @@ if __name__ == '__main__':
           'command [command-args...]'.format(sys.argv[0], CXX_COMPILER_ARG, SYCLCC_ARG), file=sys.stderr)
     sys.exit(1)
 
-  command_in = sys.argv[command_offset]
-  command_in_args = sys.argv[command_offset + 1:]
+  # If this is a compilation step, attempt to find the expected compiler (e.g. clang++).
+  # This may not be the first argument, in case additional CMAKE_CXX_COMPILER_LAUNCHERs are set.
+  compiler_offset = command_offset
+  while compiler_offset < len(sys.argv) and (
+      not os.path.isfile(sys.argv[compiler_offset]) or
+      not os.path.samefile(cxx_compiler_exe, sys.argv[compiler_offset])):
+    compiler_offset += 1
+  is_compilation_step = compiler_offset < len(sys.argv)
 
-  # When invoked with a command line for expected compiler (e.g. clang++), replace with a syclcc invocation.
+  # When invoked with a command line for expected compiler, replace with a syclcc invocation.
+  if is_compilation_step:
+    launcher_commands = sys.argv[command_offset:compiler_offset]
+    compiler_args = sys.argv[compiler_offset + 1:]
+    command_line = [*launcher_commands, *syclcc_exe, *syclcc_specific_args, *compiler_args]
   # Otherwise, e.g. for invocations of `ar` for linking static libraries, just continue with the command as-is.
-  if os.path.samefile(cxx_compiler_exe, command_in):
-    command_line = [*syclcc_exe, *syclcc_specific_args, *command_in_args]
   else:
-    command_line = [command_in, *command_in_args]
+    command_line = sys.argv[command_offset:]
 
   sys.exit(subprocess.run(command_line).returncode)


psalz:
Not sure why CI is failing, but it seems unrelated? (`CMake Error: CMake was unable to find a build program corresponding to "Ninja"`)

nilsfriess:
Yes, Windows CI is currently not working properly.

flying-scotsman:
I created #1259 to solve this issue, I think it's a simpler solution.

psalz:
> I created #1259 to solve this issue, I think it's a simpler solution.

It doesn't solve it though, because the compiler executable (e.g. `clang++` or `c++`) still needs to be replaced with `acpp`. In it's current form the script assumes the first argument to be that compiler executable, and if its anything else it just gives up.

flying-scotsman:
You're right, thanks, my oversight. Then I'll close mine and give this one a thumbs up :slightly_smiling_face:

illuhad:
> Not sure why CI is failing, but it seems unrelated? (CMake Error: CMake was unable to find a build program corresponding to "Ninja")

Windows CI is well, Windows, and the SSCP LIT compiler tests currently suffer because the Intel oneAPI repository is broken. We try to install CPU OpenCL from there to test SPIR-V JIT in github runners. So looks fine :)

psalz:
> Windows CI is well, Windows, and the SSCP LIT compiler tests currently suffer because the Intel oneAPI repository is broken. We try to install CPU OpenCL from there to test SPIR-V JIT in github runners.

In Celerity we solve this by having a nightly CI run of Celerity master against upstream dependencies' (DPC++ and AdaptiveCpp) HEAD revision, to check whether anything broke. If all is good we tag those revisions as "latest" and use them for regular CI!

nilsfriess:
Should we test this at least with some configurations in CI?

illuhad:
Yeah, if we could have some sort of testing, that might be good.

bin/acpp
```diff
@@ -2032,7 +2032,7 @@ if __name__ == '__main__':
       if not config.has_optimization_flag():
         print("acpp warning: No optimization flag was given, optimizations are "
               "disabled by default. Performance may be degraded. Compile with e.g. -O2/-O3 to "
-              "enable optimizations.")
+              "enable optimizations.", file=sys.stderr)
```
illuhad: I think we print warnings to stdout in more places than just here (you might also want to rebase on current develop). Probably those places should be changed too. Or even better, have shared functions to print warnings and errors.

psalz: Done in #1340.

psalz:
> Yeah, if we could have some sort of testing, that might be good.

A test could be as simple as printing a string from a compiler launcher and then grepping for it in the compilation output. I don't have time to figure out your CI setup however, so some help here would be appreciated.

nilsfriess:
In some CI configurations there seems to be an issue caused by #1340:
`acpp error: fatal: print_warning() got an unexpected keyword argument 'file'`

psalz:
> In some CI configurations there seems to be an issue caused by #1340: `acpp error: fatal: print_warning() got an unexpected keyword argument 'file'`

No that's because you accidentally reintroduced the `file=sys.stderr` into the call to `print_warning` (NOT `print`) through the back-merge. I would recommend rebasing this PR onto #1340 and then rebasing yours onto this.

nilsfriess:
Whoops sorry. I rebased it onto develop

psalz:
> Whoops sorry. I rebased it onto develop

Since #1340 addressed the logging issue, the first commit in this PR ("Print optimization warning to stderr") could be dropped (which I just did)!

illuhad (approved):
Thanks!

PR #1279, opened 2023-12-06
Make kernel_cache singleton shared_ptr to fix use-after-free errors

nilsfriess:
In certain cases the static kernel_cache object created by the first call to `kernel_cache::get()` was destructed before the `backend_manager` which caused segmentation faults since `backend_manager::~backend_manager()` calls a method of this (then possibly already destroyed) object.

This PR fixes this by instead creating a `shared_ptr` for the kernel_cache singleton and storing a copy of it in each class that might need it. The reference counting then ensures that the kernel_cache object is destroyed after all other objects that use it.


include/hipSYCL/sycl/handler.hpp
```diff
@@ -1012,6 +1013,8 @@ class handler {
   rt::runtime* _rt;
 
   bool _operation_uses_reductions = false;
+
+  std::shared_ptr<rt::kernel_cache> _kernel_cache;
```
illuhad, line 1017: Rather than store an additional `shared_ptr` (which will need to be initialized for every kernel launch) could we not just store a copy of the `shared_ptr` inside the runtime, and then access that one in `handler`? After all, `handler` already stores a `runtime` pointer.

illuhad (approved):
Thanks!

PR #1281, opened 2023-12-07
[stdpar] Don't attempt to prefetch with system USM

illuhad:
In system USM mode, the memory tracking interfaces are unavailable. The current code still attempts to use them for prefetches, which causes compilation issues. This PR guards the prefetch code with macros such that it is only active in the non-system-USM case.


illuhad:
Superseded by #1424

PR #1290, opened 2023-12-13
Introduce default targets; compile for generic by default

illuhad:
Previously, `acpp` by default would error if `--acpp-targets` (or `ACPP_TARGETS` environment variable) was not given. There was the option to use a default targets specification in cmake, but it was not used by default, and judging by its name `-DDEFAULT_GPU_ARCH`, came from the very old days when we still had `--hipsycl-gpu-arch`.

This is a problem for cmake-based projects using the stdpar model. In this model, since it is just standard C++, generally the `CMAKE_CXX_COMPILER` needs to be set to `acpp`. However, `cmake` tests whether the compiler works, and then finds that it does not, because `acpp` is not given `--acpp-targets` during cmake validation and thus complains.
This is a major inconvenience for any user of stdpar and cmake. The only workaround is to use environment variables like `ACPP_TARGETS` which can however be confusing and hacky.

This PR:
* Deprecates `-DDEFAULT_GPU_ARCH` and introduces `-DDEFAULT_TARGETS` cmake option;
* For backwards compatibility, `DEFAULT_TARGETS` is filled with `DEFAULT_GPU_ARCH` if `DEFAULT_GPU_ARCH` is given, and `DEFAULT_TARGETS` is not.
* If neither `DEFAULT_TARGETS` nor `DEFAULT_GPU_ARCH` are given, `DEFAULT_TARGETS` is set to `generic` if the SSCP compiler is enabled, and to `omp` otherwise.

This means that every installation of AdaptiveCpp (assuming it has SSCP enabled, which is the default for any recently new LLVM) will be able to just `acpp -o test test.cpp` have a binary that can offload to anywhere.

That was always the game plan for SSCP, although originally I had planned to wait a little longer until at least group algorithms are ported over. However, the new stdpar use case merits a change of plans in my opinion.

It is unclear whether we should also include `omp` as default target instead of just `generic` (i.e. `omp;generic`). 
This would increase compile time, but would allow execution for OpenMP by default. On the other hand, execution on CPU is also possible through `generic` and OpenCL, and users could enable `omp` explicitly if they wanted to.

@al42and I recall that you run a configuration on LUMI with default target specified. Please verify that this does not break things.
@tom91136 This should make things a little more convenient for you :)


al42and:
I like this PR and how it makes the overall interface more uniform.

The new option works fine.

But there old approach is now broken. Previously, ` -DDEFAULT_PLATFORM=rocm -DDEFAULT_GPU_ARCH=gfx90a` is how it was supposed to be used (platform could be skipped if there's only one). Now, these flags are equivalent to `DEFAULT_TARGETS=gfx90a`, which does not work.

And we should also deprecate `DEFAULT_PLATFORM` and remove `default-platform` from syclcc.json

illuhad:
Thanks for checking, @al42and ! I totally forgot about `DEFAULT_PLATFORM` :( That stuff is truly ancient, from back when we could only target a single device at a time..

I have now changed the behavior for backwards compatibility, i.e. if `DEFAULT_TARGETS` is not set:
* If `DEFAULT_PLATFORM` is not given, but `DEFAULT_GPU_ARCH` is, then `DEFAULT_TARGETS=${DEFAULT_GPU_ARCH}`
* If `DEFAULT_PLATFORM` is `cpu`, then `DEFAULT_TARGETS=omp`.
* If both `DEFAULT_PLATFORM` and `DEFAULT_GPU_ARCH` are given, then:
   * For `DEFAULT_PLATFORM=="cuda"`, `DEFAULT_TARGETS="cuda:${DEFAULT_GPU_ARCH}"`
   * For `DEFAULT_PLATFORM=="rocm"`, `DEFAULT_TARGETS="hip:${DEFAULT_GPU_ARCH}"`

Can you check if it is now working correctly?

I have also deprecated `DEFAULT_PLATFORM` and removed it from `syclcc.json`.

Unfortunately, since `DEFAULT_PLATFORM` was set by default depending on available backends (in an incomplete way, since it only took host/NVIDIA/AMD into account), I had to remove the existing default values of `DEFAULT_PLATFORM` in order to not conflict with the new defaults for `omp` and `generic`.

This means that if the variables are not set, this PR can potentially change the default behavior. I don't see any way around that though, unless we want to abandon the idea of `generic` as default. This however does not seem like the right choice to me, since the old defaults do not allow to actually be compatible with all supported devices like `generic`. So `generic` does seem like the natural choice here.

CMakeLists.txt
```diff
@@ -429,6 +423,43 @@ else()
...
+      if(DEFAULT_PLATFORM STREQUAL "cuda")
+        set(DEFAULT_TARGETS "cuda:${DEFAULT_GPU_ARCH}")
+      elseif(DEFAULT_PLATFORM STREQUAL "rocm")
+        set(DEFAULT_TARGETS "hip:${DEFAULT_GPU_ARCH}")
+      endif()
+      
+    endif()
+  elseif(DEFAULT_GPU_ARCH)
+    message(DEPRECATION "DEFAULT_GPU_ARCH is deprecated; use DEFAULT_TARGETS instead.")
+    set(DEFAULT_TARGETS ${DEFAULT_GPU_ARCH})
```
al42and: Shall we instead throw an error? I don't think there are any valid values for `DEFAULT_GPU_ARCH` that are also valid as valid targets (without specifying `platform:` prefix)

CMakeLists.txt
```diff
@@ -429,6 +423,43 @@ else()
...
+      set(DEFAULT_TARGETS "omp")
+    endif()
+    
+    if(DEFAULT_GPU_ARCH)
+      message(DEPRECATION "DEFAULT_GPU_ARCH is deprecated; use DEFAULT_TARGETS instead.")
+
+      if(DEFAULT_PLATFORM STREQUAL "cuda")
+        set(DEFAULT_TARGETS "cuda:${DEFAULT_GPU_ARCH}")
+      elseif(DEFAULT_PLATFORM STREQUAL "rocm")
+        set(DEFAULT_TARGETS "hip:${DEFAULT_GPU_ARCH}")
```
al42and, line 448: If we're breaking the old behavior, let's at least be very clear about it:

```suggestion
        set(DEFAULT_TARGETS "hip:${DEFAULT_GPU_ARCH}")
      else()
        message(SEND_ERROR "Invalid value for DEFAULT_PLATFORM: \"${DEFAULT_PLATFORM}\". When DEFAULT_GPU_ARCH is specified, only \"cuda\" and \"hip\" are supported.")
```

CMakeLists.txt
al42and:
```suggestion
    message(SEND_ERROR "DEFAULT_GPU_ARCH cannot be used without specifying DEFAULT_PLATFORM.")
```

al42and:
> Can you check if it is now working correctly?

Works fine!

I made a few suggestions above to make the misconfigurations more apparent. Better to see the problem ASAP, and not only after installing and trying to use the compiler.

> This however does not seem like the right choice to me, since the old defaults do not allow to actually be compatible with all supported devices like generic. So generic does seem like the natural choice here.

I agree with you on `generic`. What I am not so sure is setting `omp` as the default target in absence of SSCP. If you ask me, this is a useful target to have, but it is unlikely to be the main target for people using ACpp, and defaulting to it can lead to confusion. A clear error asking to add `--acpp-target` flag would be better in this case, IMO

CMakeLists.txt
illuhad: We can do that. Is there any particular reason why the error message talks about `"cuda"` and `"hip"`? My understanding was that the `DEFAULT_PLATFORM` variable previously allowed "cuda" and "rocm", so shouldn't it be:

```suggestion
        set(DEFAULT_TARGETS "hip:${DEFAULT_GPU_ARCH}")
      else()
        message(SEND_ERROR "Invalid value for DEFAULT_PLATFORM: \"${DEFAULT_PLATFORM}\". When DEFAULT_GPU_ARCH is specified, only \"cuda\" and \"rocm\" are supported.")
```

illuhad:
> I made a few suggestions above to make the misconfigurations more apparent. Better to see the problem ASAP, and not only after installing and trying to use the compiler.

Thanks, good suggestions! (see my one comment)

> I agree with you on generic. What I am not so sure is setting omp as the default target in absence of SSCP. If you ask me, this is a useful target to have, but it is unlikely to be the main target for people using ACpp, and defaulting to it can lead to confusion. A clear error asking to add --acpp-target flag would be better in this case, IMO

I see your point, but I also think there are some counterpoints:

* The reason we have `generic` and `omp` is that these are the two targets that we know should work (mostly) independently of the end user's hardware. If we don't use `omp` as a fallback, we will not be able to have a default target in case SSCP is unavailable.
* If we don't set  a default target without SSCP, our communication regarding the `--acpp-targets` flag will be more confusing.  It would be useful and less confusing if we could say that `acpp -o test test.cpp` will always at least work, whether or not it will offload. I can imagine that might be similar tooling issues as the one described in the initial comment, and of course such a behavior will be better for teaching.
* There are users who are in fact using stdpar with the `omp` backend on CPU since it has a better NUMA behavior than e.g. regular TBB-based PSTL. Those users might not need the LLVM dependency, and thus not want SSCP. They will be stuck with the same cmake problem that motivated this PR in the first place (as described in the initial comment) if we don't set `omp` as default target.

How about we issue a warning, if we set the default target to `omp` saying that an invocation without `--acpp-targets` will not be able to offload?

If users still try to use a SYCL device from a GPU backend in this case they will still get an error, although one at runtime. I admit that this error is currently not very intuitive (I believe it will be something like `could not find requested kernel launcher`), but we could add to the message there that users should check their `--acpp-targets` flag.

CMakeLists.txt
al42and: Yes, you're right

al42and:
> If we don't use omp as a fallback, we will not be able to have a default target in case SSCP is unavailable.

I think that is good. Without SSCP, the user *must* have the target hardware in mind when compiling, so I think it is better to be explicit about this rather than to have a bad and counter-intuitive default. With SSCP, the choice is delayed till the runtime, so one does not have to make a choice when compiling.

> It would be useful and less confusing if we could say that acpp -o test test.cpp will always at least work, whether or not it will offload.

I would question the definition of "works" here. The whole reason people use PStl or SYCL is for performance. If they expect GPU offloading but get CPU multithreading instead -- that might not count as "working" in their books (if it does, they can set the target explicitly).

> If we don't set a default target without SSCP, our communication regarding the --acpp-targets flag will be more confusing.

I don't think it is more confusing than "without specifying --acpp-targets, either a CPU-only binary or a fully-portable (adaptive) binary with CPU and GPU support will be produced".

> There are users who are in fact using stdpar with the omp backend on CPU since it has a better NUMA behavior than e.g. regular TBB-based PSTL. Those users might not need the LLVM dependency, and thus not want SSCP. They will be stuck with the same cmake problem that motivated this PR in the first place (as described in the initial comment) if we don't set omp as default target.

They can set `ACPP_TARGETS` when building explicitly. Or change `syclcc.json`. Or use `CMAKE_CXX_FLAGS`.

That's an important case, and it's amazing how AdaptiveCpp is beating TBB. But AdaptiveCpp is also beating NVC++, if I recall correctly; so, GPU-offloading is an attractive use-case too, and we should avoid unexpected and potentially-undesirable behavior there.

> How about we issue a warning, if we set the default target to omp saying that an invocation without --acpp-targets will not be able to offload?

Agree, that's also fine. As long as it's clear what's happening.

Alternatively, we can issue a CMake-time error is SSCP is not available and `ACPP_TARGETS` is not set.

illuhad:
> They can set ACPP_TARGETS when building explicitly. Or change syclcc.json. Or use CMAKE_CXX_FLAGS.

The common pattern in existing cmake infrastructure for stdpar projects is already to rely on something like `-DCMAKE_CXX_COMPILER=acpp -DCMAKE_CXX_FLAGS="--acpp-stdpar --acpp-targets=...."`.  (See e.g. https://github.com/uob-hpc/cloverleaf)

However this won't be sufficient because when cmake validates `acpp` as C++ compiler, it will find that `acpp` does not run unless `--acpp-targets` is provided somehow. At the point when it does that, `CMAKE_CXX_FLAGS` is not yet applied.

You can circumvent that by also setting the environment variable `ACPP_TARGETS` at cmake time, but this is inconvenient because you need to set the target twice, and it might also be confusing. Modifying `syclcc.json` is similarly more of a hack, and perhaps `syclcc.json` should be regarded more of an implementation detail since it is only half-documented.

This cmake detail is why I think it is useful especially for stdpar to have `acpp` without arguments to be able to compile programs without error by default. For SYCL, it is less critical, because there you can `find_package(AdaptiveCpp)` and then `add_sycl_to_target()` such that no change of `CMAKE_CXX_COMPILER` is needed.

> That's an important case, and it's amazing how AdaptiveCpp is beating TBB. But AdaptiveCpp is also beating NVC++, if I recall correctly; so, GPU-offloading is an attractive use-case too, and we should avoid unexpected and potentially-undesirable behavior there.

Yes, we generally either closely match nvc++ or crush it, depending on the workload. This is assuming that all needed STL algorithms are of the subset for which we implement offloading. When it comes to raw quantity and the pure number of offloadable STL algorithms, nvc++ still wins.

I think we are in agreement that we want to ensure that nothing unexpected happens. I would propose that I add the warning, and then we can perhaps refine the defaults from there based on actual user feedback in the future, if needed.

al42and:
> However this won't be sufficient because when cmake validates acpp as C++ compiler, it will find that acpp does not run unless --acpp-targets is provided somehow. At the point when it does that, CMAKE_CXX_FLAGS is not yet applied.

Hm, I see a different behavior with CMake 3.28.1. I don't have the default target set and the test fails without CXX_FLAGS, but the following command configures CloverLeaf just fine, and all compiler sanity tests are run with `acpp-targets` flag set: `cmake .. -DCMAKE_C_COMPILER=clang-17 -DCMAKE_CXX_COMPILER=acpp -DCMAKE_CXX_FLAGS=--acpp-targets=generic`

> I think we are in agreement that we want to ensure that nothing unexpected happens. I would propose that I add the warning, and then we can perhaps refine the defaults from there based on actual user feedback in the future, if needed.

Yes, that sounds reasonable, I agree.

I would perhaps even suggest to always print a warning when no default target is set: "acpp warning: target not set, using XXX. Add --acpp-targets flag or add the default target to your configuration file to silence this warning". Best of both worlds: we gently encourage user to be explicit in their wishes to avoid surprises in the future, without impeding the simplest use of `acpp` now.

illuhad:
>  Hm, I see a different behavior with CMake 3.28.1. I don't have the default target set and the test fails without CXX_FLAGS, but the following command configures CloverLeaf just fine, and all compiler sanity tests are run with acpp-targets flag set: cmake .. -DCMAKE_C_COMPILER=clang-17 -DCMAKE_CXX_COMPILER=acpp -DCMAKE_CXX_FLAGS=--acpp-targets=generic

Thanks for checking, I've just tried again too and you are right, the issue is a little more nuanced: If `-DCMAKE_CXX_FLAGS` is set at command line, it works. If `CMAKE_CXX_FLAGS` is set programmatically inside `CMakeLists.txt`, it does not work.

The latter I think is quite important in practice to allow cmake-based apps to have flags for enabling stdpar offloading.

For example, cloverleaf (https://github.com/UoB-HPC/CloverLeaf/blob/main/CMakeLists.txt) has `CXX_EXTRA_FLAGS`, where you are supposed to provide compiler-specific flags to get it to offload. This seems like a reasonable pattern to me, as it separates common optimization flags (`-O3 -DNDEBUG` etc) from compiler-specific flags, which helps ensuring that all runs are done with proper optimization flags across compilers. However, it does not work with AdaptiveCpp without this PR.

Similarly, other apps which are more focused on stdpar might have a flag to enable stdpar offloading, as done here:
https://github.com/jbeisman/nbody-stdpar/blob/master/CMakeLists.txt
For AdaptiveCpp support, we would like to extend this with a `USE_ACPP` flag or similar, which currently probably would not work.

I think fundamentally, the issue is that currently `acpp` behaves differently from stdpar vendor compilers like `nvc++`, `icpx`, or `roc-stdpar`. All of those can generate a binary without special flags, but might require special flags to offload.
As stdpar as a model gets more popular, I think we should ensure that we are compatible with existing patterns such that users can support all compilers with the same cmake infrastructure.
Perhaps even after this PR there is still an even more nuanced issue: These vendor compilers are able to compile regular C++ code without extra flags (and stdpar code is always just regular C++), but will error when trying to compile explicitly heterogeneous code (SYCL, CUDA, HIP) without special flags. So to completely mirror the behavior, we would need to be able to compile C++ code without extra flags, and require extra flags for SYCL or stdpar support.
I'm not sure if there's a way to achieve this with the way `acpp` is currently designed. It's built around the assumption that if you use it, you want to use SYCL and/or stdpar, and we don't have an `--acpp-sycl` flag that would correspond to `-fsycl` or similar. We could perhaps abuse `--acpp-targets` for that: If it's not set and there's no default target, then just perform regular C++ compilation, which should error if we encounter SYCL code. For the latter some extra logic might be needed because otherwise it might just silently use the library-only host path in our SYCL headers.

I think that might perhaps be a good solution for the `DEFAULT_TARGETS=omp` discussion in the long term, but it might be a good idea to decouple that from this PR as some additional investigation might be needed. It might also be useful to have some user feedback on whether it is useful or appreciated to be able to unconditionally compile SYCL code just with `acpp` without additional flags.

> I would perhaps even suggest to always print a warning when no default target is set: "acpp warning: target not set, using XXX. Add --acpp-targets flag or add the default target to your configuration file to silence this warning". Best of both worlds: we gently encourage user to be explicit in their wishes to avoid surprises in the future, without impeding the simplest use of acpp now.

My concern would be that this would reinforce the belief that AdaptiveCpp needs to be custom-built for all users, even though we have `generic` which is supposed to work for all users if it is available.

al42and:
> My concern would be that this would reinforce the belief that AdaptiveCpp needs to be custom-built for all users, even though we have generic which is supposed to work for all users if it is available.

I don't see how. We're asking users to be explicit, since many compilation flows are supported and genuinely have their uses. Compare with `-O` flags: ACpp always complains when they are not set, because that could lead to potentially undesired behavior; but if the user elaborates their wishes by setting `-O0`, then no warning is necessary, even though the resulting behavior is not changed.

> I think fundamentally, the issue is that currently acpp behaves differently from stdpar vendor compilers like nvc++, icpx, or roc-stdpar. All of those can generate a binary without special flags, but might require special flags to offload.

That also sounds like a good solution. Maybe delegate checking of the target validity to the Clang plugin, so that it can see if offload is used in the source, and only then raises an error? And, as you said, that follows existing patterns -- always a good thing.

al42and:
> I think fundamentally, the issue is that currently acpp behaves differently from stdpar vendor compilers like nvc++, icpx, or roc-stdpar. All of those can generate a binary without special flags, but might require special flags to offload.

Curiously, even `acpp --acpp-version` does not work without specifying a target architecture.

illuhad:
This is strange indeed. `--acpp-version` does want to print all of the configuration, so it might access the `targets` value. However, it should handle that: https://github.com/AdaptiveCpp/AdaptiveCpp/blob/f2d41975ba3e5db44713abc5ae3dbe974f28abcf/bin/acpp#L1981

EDIT: For me `acpp --acpp-version` works without target architecture. Both with and without a configured default platform.

illuhad:
After some internal discussion, my proposed direction is:
* we do not deviate from the existing behavior of `acpp` of being an offloading C++17 compiler. In the future, if there is user demand/sufficient reason, we could introduce a new compiler driver in addition to `acpp` which would adopt the behavior (similarly to e.g. `icpx`) of being just a regular host compiler unless offload flags are provided explicitly. 
  * This new compiler driver could also solve issue #1300, although it is not clear to me whether this issue is something that AdaptiveCpp needs to solve.
  * The current `acpp` semantics are also not without precedent, as they mirror what e.g. `nvcc` or `hipcc` do, although these support older C++ standards than we do.
* Ideally, if we do not have SSCP, we should enforce that users explicitly provide a `DEFAULT_TARGETS` setting to cmake. However, I am concerned that this might break existing deployments of AdaptiveCpp (e.g. installation and packaging scripts maintained by third parties). I have thus expanded the warning in this case to say that setups without SSCP in the future may only be supported if `DEFAULT_TARGETS` is provided.
  * This hopefully causes users to migrate to providing `-DDEFAULT_TARGETS` so that we can enforce users providing `DEFAULT_TARGETS` in future versions.
* I have changed the default target to `omp;generic` from just `generic`. This is to avoid nasty, silent performance surprises if users run their applications with default targets on CPU. We can remove `omp` once the `omp` backend can also be targeted through the `generic` target.

al42and:
> EDIT: For me acpp --acpp-version works without target architecture. Both with and without a configured default platform.

I cannot reproduce it now either. Oh, well. 

>Ideally, if we do not have SSCP, we should enforce that users explicitly provide a DEFAULT_TARGETS setting to cmake.

I think the warning is fine. Forcing people to always choose a default can lead to them choosing a bad default just to get on with the installation, and a bad default is worse than no default when it comes to using the compiler. The warning text describes the limitations quite nicely, so they can make an informed decision if they want.

PR #1291, opened 2023-12-13
[Windows] Use LLVM MinGW in CI and fix two small bugs

nilsfriess:
This PR fixes the Windows CI by using a LLVM/Clang based MinGW toolchain to build AdaptiveCpp, boost, and the CPU tests. The prebuilt LLVM/Clang comes from [this repo](https://github.com/mstorsjo/llvm-mingw). 

I haven't tried to build the CUDA backend using this toolchain yet, maybe we have to build LLVM with different flags enabled but the repo also contains build scripts that should make this easy. 

I also included two Windows related fixes in this PR:
- Building the tests failed with some very weird linker errors which were caused by using the wrong boost `unit_test.hpp` include file (there is `boost/test/unit_test.hpp` and `boost/test/included/unit_test.hpp` and apparently you're supposed to use the latter in `.cpp` files)
- The `std::free` function of the Microsoft C Runtime Library cannot handle aligned memory, therefore `omp_allocator::free` always calls `_aligned_free` provided by the Microsoft CRT. This, however, can only free memory allocated with `_aligned_malloc`, but `_aligned_malloc` returns NULL when calling it zero alignment. Therefore we now set `min_alignment = 1` if `min_alignment == 0` in `omp_allocator::allocate` instead of calling malloc.


nilsfriess:
The changes in the omp_allocator seem to cause some issues on Linux, I‘ll look into it.

illuhad:
Great to see! Given the self-hosted CI failures, is it possible that some of the changes have affected Linux?

nilsfriess:
Found out what the problem was.

boost's unit test framework offers two options: a header-only version and a library version. For some reason we were using a mixture of both which was first causing issues on Windows, and then my changes caused some issues on Linux. I have now changed it to only use the header-only variant which means:
- Putting the following three lines in exactly one translation unit per test module (e.g. in `sycl_test_suite.cpp`) and including `<boost/test/unit_test.hpp>` in all other translation units of that test module:
    ```
    #define BOOST_TEST_MODULE name_of_the_test_module
    #define BOOST_TEST_DYN_LINK
    #include <boost/test/included/unit_test.hpp>
    ```
- Not linking against `${BOOST_LIBRARIES}` in CMake

illuhad (approved):
Thanks! What was the issue causing the crashes on Linux?

nilsfriess:
The problem was that I changed the header `boost/test/unit_test.hpp` to `boost/test/included/unit_test.hpp` in the tests, but the latter is only supposed to be used when using Boost.Test in header-only mode (see [here](https://www.boost.org/doc/libs/1_84_0/libs/test/doc/html/boost_test/usage_variants.html)). But we were additionally linking against boost's test library which then caused the issues.

PR #1292, opened 2023-12-13
[CI] Fix failing OpenCL install in Linux LIT tests

nilsfriess:
Installation of OpenCL from the oneAPI repos was broken, I found [this](https://community.intel.com/t5/oneAPI-Registration-Download/Cannot-install-package-quot-intel-oneapi-runtime-compilers-quot/m-p/1546419#M6471) bug report with the solution to use the `-2024` versions of the packages. Maybe this will break again if packages are renamed or so, but I guess for now this is fine.


illuhad (approved):
Ahhhhh, thank you! :heart: :heart: :heart: I was told by Intel that they are working on fixing it, and that we should just wait.. but that was a couple of weeks ago. Great that there's a workaround in the meantime! :)

PR #1293, opened 2023-12-18
HIP: Update platform macros for ROCm 6

al42and:
Old `__HIP_PLATFORM_NVCC__` and `__HIP_PLATFORM_HCC__` no longer work.
Now `__HIP_PLATFORM_AMD__` and `__HIP_PLATFORM_NVIDIA__` are used.

See https://github.com/ROCm/clr/commit/8fe1d9dda109e31acb7600e92f1697366c35590b

### DIFF
--- a/include/hipSYCL/runtime/hip/hip_target.hpp
+++ b/include/hipSYCL/runtime/hip/hip_target.hpp
@@ -37,11 +37,17 @@
 #endif
 
 #ifdef HIPSYCL_RT_HIP_TARGET_CUDA
-#define __HIP_PLATFORM_NVCC__
+#define __HIP_PLATFORM_NVCC__ // Only needed for ROCm <6.0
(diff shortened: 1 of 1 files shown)


illuhad (commented):
Thank you! Shouldn't the `__HIP_ROCclr__` define be outside the `__HIP_PLATFORM_HCC__/__HIP_PLATFORM_AMD__` ifdefs to avoid a potential redefinition if neither is defined?

al42and:
> Thank you! Shouldn't the __HIP_ROCclr__ define be outside the __HIP_PLATFORM_HCC__/__HIP_PLATFORM_AMD__ ifdefs to avoid a potential redefinition if neither is defined?

Good point. I just followed the existing pattern, but I guess putting it into its own `ifdef` makes more sense.

illuhad (approved):
Thanks!

PR #1305, opened 2024-01-07
Make `marray` constexpr compatible

stukowski:
Adding the `constexpr` specifier to methods of the `device_array` class template to make the following compile:

```c++
constexpr sycl::marray<float, 3> v{0.0f, 0.0f, 0.0f};
```

Currently, this fails to compile:

```
constexpr_test.cpp: error: constexpr variable 'v' must be initialized by a constant expression
        constexpr sycl::marray<float, 3> v{0.0f, 0.0f, 0.0f};
                                         ^~~~~~~~~~~~~~~~~~~
/bin/../include/AdaptiveCpp/CL/../hipSYCL/sycl/libkernel/marray.hpp:56:7: note: non-constexpr function 'operator[]' cannot be used in a constant expression
      _data[offset] = x;
      ^
/bin/../include/AdaptiveCpp/CL/../hipSYCL/sycl/libkernel/marray.hpp:65:5: note: in call to '&v->init_offset(0, 0.F)'
    init_offset(current_offset, arg);
    ^
/bin/../include/AdaptiveCpp/CL/../hipSYCL/sycl/libkernel/marray.hpp:100:6: note: in call to '&v->initialize_from_arg(current_offset, 0.F)'
    (initialize_from_arg(current_offset, args), ...);
     ^
constexpr_test.cpp: note: in call to 'marray(0.F, 0.F, 0.F)'
        constexpr sycl::marray<float, 3> v{0.0f, 0.0f, 0.0f};
                                         ^
/bin/../include/AdaptiveCpp/CL/../hipSYCL/sycl/libkernel/detail/device_array.hpp:46:6: note: declared here
  T& operator[] (size_t i) noexcept
```

### DIFF
--- a/include/hipSYCL/sycl/libkernel/detail/device_array.hpp
+++ b/include/hipSYCL/sycl/libkernel/detail/device_array.hpp
@@ -43,25 +43,25 @@ struct device_array
   using const_iterator = const T*;
 
   HIPSYCL_UNIVERSAL_TARGET
-  T& operator[] (size_t i) noexcept
+  constexpr T& operator[] (size_t i) noexcept
   {
     return _data[i];
   }
 
   HIPSYCL_UNIVERSAL_TARGET
-  const T& operator[] (size_t i) const noexcept
+  constexpr const T& operator[] (size_t i) const noexcept
   {
     return _data[i];
   }
 
   HIPSYCL_UNIVERSAL_TARGET
-  size_t size() const noexcept
+  constexpr size_t size() const noexcept
   {
     return N;
   }
 
   HIPSYCL_UNIVERSAL_TARGET
-  bool operator== (const device_array& other) const noexcept
+  constexpr bool operator== (const device_array& other) const noexcept
   {
     for(size_t i = 0; i < N; ++i)
       if(_data[i] != other._data[i])
@@ -70,7 +70,7 @@ struct device_array
   }
 
   HIPSYCL_UNIVERSAL_TARGET
-  bool operator!= (const device_array& other) const noexcept
+  constexpr bool operator!= (const device_array& other) const noexcept
   {
     return !(*this == other);
   }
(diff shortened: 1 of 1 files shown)


nilsfriess:
Thanks for the PR.
I tried to compile your example (`constexpr sycl::marray<float, 3> v{0.0f, 0.0f, 0.0f};`) with AdaptiveCpp built from your branch but I still get a compile error since the marray constructors are not marked `constexpr`.

I opened #1102 a while ago which basically is this PR + a few additional changes in marray.hpp. I actually thought it was already merged but there seems to be one small issue in #1102 that needs to be fixed, I'll have a look at that right away and then we can merge #1102 and probably close this PR. Sorry for the unnecessary work...

stukowski:
I am not sure why you are not seeing it, but my pull request is based on the current `develop` branch. Here, `marray` constructors are already marked as `constexpr`:

https://github.com/AdaptiveCpp/AdaptiveCpp/blob/7041cdf31521571d1cc7aad4de6fe7dc27593317/include/hipSYCL/sycl/libkernel/marray.hpp#L95C43-L95C43

nilsfriess:
You're absolutely right, I just looked at the error message that said something about non constexpr constructors and thought that it referred to the marray constructors. I actually complains about the device_array constructors. The class `device_array<T,N>` is specialised for `N=1,2,3,4`. And the (implictly defined) default constructor for these specialised classes are not automatically constexpr. That's why I zero-initialised all the entries for these cases (see e.g. [here](https://github.com/AdaptiveCpp/AdaptiveCpp/pull/1102/files#diff-2c89fad3016d951965db1fb4acda03cbf95a1c9c59a197b852845e10447e107cR258)).

stukowski:
Oh, I see. My Apple clang compiler on macOS did not complain about this. Sorry, so far I haven't tested my solution on any other platform.

nilsfriess:
I you open other PRs in the future (which we are always happy to see!), then you can simply add a test case (probably there is already a corresponding test file somewhere in the `test/sycl` directory) and CI will then run on various platform/compiler combinations :)

stukowski:
As a new user of AdaptiveCpp, I wasn't familiar with the test suite. But let me take a look and add a test case to my pull request. Thanks.

I expect the test will fail on most platforms due to the issue you mentioned. I will then make the necessary code changes to fix it.

nilsfriess:
As I said earlier, the missing changes are already included in #1102 (which is just not merged yet). So there's no need for you to do the extra work

stukowski:
I see. Thanks for the clarification. Then I am closing this PR.

PR #1314, opened 2024-01-17
[SSCP][llvm-to-amdgpu] Work around hipRTC no longer correctly linking device libs for newer ROCm versions

illuhad:
ROCm versions starting at least with 5.7 no longer correctly link the ROCm AMD device bitcode libraries when we give our SSCP bitcode libraries to hipRTC. This causes a JIT failure, since builtins remain undefined.

To me it looks like the issue is that hipRTC *first* links the AMD device bitcode libraries, and only afterwards links our user-provided bitcode. Because at the point when the device libraries are linked none of their symbols are actually required yet, nothing happens.

More worryingly, it looks like the underlying action in comgr, which powers hipRTC, is deprecated and supposed to be removed (was?) in ROCm 6.0: https://github.com/ROCm/llvm-project/blob/197f5bc2fccbeac4a6386fe93329358ef45ed630/amd/comgr/include/amd_comgr.h.in#L1585
It seems that device bitcode library linking is now supposed to happen only when the clang driver processes source code, instead of the IR level where we are.

As a hotfix, this PR introduces manually linking the required bitcode libraries. The logic to determine which bitcode libraries are exactly needed depends on the target device, compiler flags, and ROCm/LLVM version. It is thus non-trivial. For this reason, we invoke `hipcc -###` and parse the output to find the list of device libraries that `hipcc` would link against.

This is a bit unfortunate, because it might impact the deployment process of SSCP binaries: Previously the only direct dependency was `hipRTC`, but now we need to have ROCm compilers and bitcode libraries available when a program is executed. I don't see a way around that at the moment. And of course since this issue is critical as it blocks using more recent ROCm versions with SSCP altogether, a quick fix is desirable, even if we might perhaps find a better solution with more effort.

This PR also removes a fallback codepath for ROCm version prior to ROCm 5.2 that did not have hipRTC. However, this code path was broken and we nowadays say in our documentation that SSCP requires ROCm 5.3+.

@Calandracas606 @breyerml This might fix your problem with `generic` target on at least ROCm 5.7. Please let me know if that is the case, and if it also helps on ROCm 6.0 if you have a ROCm 6.0 installation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,7 +96,7 @@ find_package(OpenCL QUIET)
 # In case HIP is not found via the cmake pkgs we fall back to the legacy rocm discovery:
 if (NOT HIP_FOUND)
   message(STATUS "Could not find HIP cmake integration, falling back to finding hipcc")
-  find_program(HIPCC_COMPILER NAMES hipcc HINTS ${ROCM_PATH})
+  find_program(HIPCC_COMPILER NAMES hipcc HINTS ${ROCM_PATH} ${ROCM_PATH}/bin)
   set(ROCM_PATH /opt/rocm CACHE PATH "Path to ROCm installation")
   if(HIPCC_COMPILER MATCHES "-NOTFOUND")
     message(STATUS "Could not find ROCm installation by looking for hipcc. ROCm support unavailable.")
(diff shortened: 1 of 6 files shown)


breyerml:
Thanks for the quick hotfix.
For me, SSCP works now with ROCm 5.7.0.
However, I had to remove the include https://github.com/AdaptiveCpp/AdaptiveCpp/blob/0da2ecea0ce98d2a08df9d2f4e71d25517f2869a/src/compiler/llvm-to-backend/amdgpu/LLVMToAmdgpu.cpp#L35
to be able to build AdaptiveCpp with SSCP using upstream clang 17.0.4. It also seems that this include is unused: this LLVM header only defines a `NoneType` enum and a `None` variable both of which aren't used in the `LLVMToAmdgpu.cpp` file.

illuhad:
@breyerml Thanks for the feedback. Looks like that include was pulled in during development and is no longer needed. I will remove it. Thanks!

PR #1316, opened 2024-01-17
[stdpar][Reflection] Make struct introspection more robust

### LINKED ISSUE #1282: StdPar causes compiler segfault with `-fsanitize=undefined`
**Bug summary**
Compiler crashes when combining `-acpp-stdpar` with `-fsanitize=undefined`

**To Reproduce**

```
> cat acpp_stdpar_sigsegv.cpp
#include <iostream>
#include <execution>
#include <algorithm>

int main(void)
{
   int xs[4096] = {};
   std::for_each(std::execution::par_unseq, xs, xs+4096, [](auto &x){ });
   return 0;
}
> /home/tom/software/x86_64/acpp/367cb7a/bin/syclcc acpp_stdpar_sigsegv.cpp -std=c++17 -O3 -Wall --acpp-stdpar --acpp-targets=omp.accelerated -fsanitize=undefined -ltbb
[AdaptiveCpp Warning] IntrospectStructPass: _Z37__hipsycl_introspect_flattened_structIPiEvPS0_S1_S1_S1_ instantiation could not be connected to type information, ignoring.
[AdaptiveCpp Warning] IntrospectStructPass: _Z37__hipsycl_introspect_flattened_structIN7hipsycl6stdpar14decorated_typeIPiJNS1_11decorations21no_pointer_validationEEEEEvPS3_S7_S7_S7_ instantiation could not be connected to type information, ignoring.
[AdaptiveCpp Warning] IntrospectStructPass: _Z37__hipsycl_introspect_flattened_structIZ4mainE3$_0EvPPiS2_S2_S2_ instantiation could not be connected to type information, ignoring.
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace, preprocessed source, and associated run script.
Stack dump:
0.      Program arguments: /usr/bin/clang++ -cc1 -triple x86_64-redhat-linux-gnu -emit-obj -dumpdir a- -disable-free -clear-ast-before-backend -disable-llvm-verifier -discard-value-names -main-file-name acpp_stdpar_sigsegv.cpp -mrelocation-model static -mframe-pointer=none -fmath-errno -ffp-contract=on -fno-rounding-math -mconstructor-aliases -funwind-tables=2 -target-cpu x86-64 -tune-cpu generic -debugger-tuning=gdb -fcoverage-compilation-dir=/home/tom/Desktop -resource-dir /usr/bin/../lib/clang/17 -isystem /home/tom/software/x86_64/acpp/367cb7a/bin/../include/AdaptiveCpp -isystem /home/tom/software/x86_64/acpp/367cb7a/bin/../include/AdaptiveCpp/hipSYCL/std/stdpar -include /home/tom/software/x86_64/acpp/367cb7a/bin/../include/AdaptiveCpp/hipSYCL/std/stdpar/detail/sycl_glue.hpp -D __OPENSYCL__ -D __HIPSYCL__ -D __ADAPTIVECPP__ -D __ACPP__ -D __OPENSYCL_STDPAR__ -D __HIPSYCL_STDPAR__ -D __ADAPTIVECPP_STDPAR__ -D __ACPP_STDPAR__ -D HIPSYCL_ALLOW_INSTANT_SUBMISSION=1 -D __HIPSYCL_ENABLE_OMPHOST_TARGET__ -D __HIPSYCL_USE_ACCELERATED_CPU__ -I /usr/include -D _ENABLE_EXTENDED_ALIGNED_STORAGE -internal-isystem /usr/bin/../lib/gcc/x86_64-redhat-linux/13/../../../../include/c++/13 -internal-isystem /usr/bin/../lib/gcc/x86_64-redhat-linux/13/../../../../include/c++/13/x86_64-redhat-linux -internal-isystem /usr/bin/../lib/gcc/x86_64-redhat-linux/13/../../../../include/c++/13/backward -internal-isystem /usr/bin/../lib/clang/17/include -internal-isystem /usr/local/include -internal-isystem /usr/bin/../lib/gcc/x86_64-redhat-linux/13/../../../../x86_64-redhat-linux/include -internal-externc-isystem /include -internal-externc-isystem /usr/include -O3 -Wall -std=c++17 -fdeprecated-macro -fdebug-compilation-dir=/home/tom/Desktop -ferror-limit 19 -fopenmp -fsanitize=alignment,array-bounds,bool,builtin,enum,float-cast-overflow,function,integer-divide-by-zero,nonnull-attribute,null,object-size,pointer-overflow,return,returns-nonnull-attribute,shift-base,shift-exponent,signed-integer-overflow,unreachable,vla-bound,vptr -fsanitize-recover=alignment,array-bounds,bool,builtin,enum,float-cast-overflow,function,integer-divide-by-zero,nonnull-attribute,null,object-size,pointer-overflow,returns-nonnull-attribute,shift-base,shift-exponent,signed-integer-overflow,vla-bound,vptr -fno-sanitize-memory-param-retval -fno-sanitize-address-use-odr-indicator -fgnuc-version=4.2.1 -fcxx-exceptions -fexceptions -fcolor-diagnostics -vectorize-loops -vectorize-slp -load /home/tom/software/x86_64/acpp/367cb7a/bin/../lib/libacpp-clang.so -fpass-plugin=/home/tom/software/x86_64/acpp/367cb7a/bin/../lib/libacpp-clang.so -mllvm -hipsycl-stdpar -faddrsig -D__GCC_HAVE_DWARF2_CFI_ASM=1 -o /tmp/acpp_stdpar_sigsegv-fd01f0.o -x c++ acpp_stdpar_sigsegv.cpp
1.      <eof> parser at end of file
2.      Optimizer
 #0 0x00007f7a1d53df6a llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (/lib64/libLLVM-17.so+0x93df6a)
 #1 0x00007f7a1d53b924 llvm::sys::RunSignalHandlers() (/lib64/libLLVM-17.so+0x93b924)
 #2 0x00007f7a1d53e7db (/lib64/libLLVM-17.so+0x93e7db)
 #3 0x00007f7a1c65c9a0 __restore_rt (/lib64/libc.so.6+0x3e9a0)
 #4 0x00007f7a1cbd2737 llvm::Constant* hipsycl::compiler::(anonymous namespace)::createConstantIntArray<8u>(llvm::SmallVector<int, 8u> const&, llvm::Module&) IntrospectStructPass.cpp:0:0
 #5 0x00007f7a1cbd17d4 hipsycl::compiler::IntrospectStructPass::run(llvm::Module&, llvm::AnalysisManager<llvm::Module>&) (/home/tom/software/x86_64/acpp/367cb7a/bin/../lib/libacpp-clang.so+0x1287d4)
 #6 0x00007f7a1cb1cdfd llvm::detail::PassModel<llvm::Module, hipsycl::compiler::IntrospectStructPass, llvm::PreservedAnalyses, llvm::AnalysisManager<llvm::Module>>::run(llvm::Module&, llvm::AnalysisManager<llvm::Module>&) (/home/tom/software/x86_64/acpp/367cb7a/bin/../lib/libacpp-clang.so+0x73dfd)
 #7 0x00007f7a1d6f094b llvm::PassManager<llvm::Module, llvm::AnalysisManager<llvm::Module>>::run(llvm::Module&, llvm::AnalysisManager<llvm::Module>&) (/lib64/libLLVM-17.so+0xaf094b)
 #8 0x00007f7a25d6b21e (/lib64/libclang-cpp.so.17+0x1b6b21e)
 #9 0x00007f7a25d625e2 clang::EmitBackendOutput(clang::DiagnosticsEngine&, clang::HeaderSearchOptions const&, clang::CodeGenOptions const&, clang::TargetOptions const&, clang::LangOptions const&, llvm::StringRef, llvm::Module*, clang::BackendAction, llvm::IntrusiveRefCntPtr<llvm::vfs::FileSystem>, std::unique_ptr<llvm::raw_pwrite_stream, std::default_delete<llvm::raw_pwrite_stream>>) (/lib64/libclang-cpp.so.17+0x1b625e2)
#10 0x00007f7a2615fba8 (/lib64/libclang-cpp.so.17+0x1f5fba8)
#11 0x00007f7a26d59c1c clang::MultiplexConsumer::HandleTranslationUnit(clang::ASTContext&) (/lib64/libclang-cpp.so.17+0x2b59c1c)
#12 0x00007f7a24b17946 clang::ParseAST(clang::Sema&, bool, bool) (/lib64/libclang-cpp.so.17+0x917946)
#13 0x00007f7a26d23566 clang::FrontendAction::Execute() (/lib64/libclang-cpp.so.17+0x2b23566)
#14 0x00007f7a26c90390 clang::CompilerInstance::ExecuteAction(clang::FrontendAction&) (/lib64/libclang-cpp.so.17+0x2a90390)
#15 0x00007f7a26da3b64 clang::ExecuteCompilerInvocation(clang::CompilerInstance*) (/lib64/libclang-cpp.so.17+0x2ba3b64)
#16 0x0000000000413f5b cc1_main(llvm::ArrayRef<char const*>, char const*, void*) (/usr/bin/clang+++0x413f5b)
#17 0x00000000004112b5 (/usr/bin/clang+++0x4112b5)
#18 0x000000000040ffd9 clang_main(int, char**, llvm::ToolContext const&) (/usr/bin/clang+++0x40ffd9)
#19 0x000000000041d636 main (/usr/bin/clang+++0x41d636)
#20 0x00007f7a1c64614a __libc_start_call_main (/lib64/libc.so.6+0x2814a)
#21 0x00007f7a1c64620b __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x2820b)
#22 0x000000000040d9a5 _start (/usr/bin/clang+++0x40d9a5)
clang++-17: error: unable to execute command: Segmentation fault (core dumped)
clang++-17: error: clang frontend command failed due to signal (use -v to see invocation)
clang version 17.0.4 (Fedora 17.0.4-1.fc39)
Target: x86_64-redhat-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
clang++-17: note: diagnostic msg: 
********************

PLEASE ATTACH THE FOLLOWING FILES TO THE BUG REPORT:
Preprocessed source(s) and associated run script(s) are located at:
clang++-17: note: diagnostic msg: /tmp/acpp_stdpar_sigsegv-6f21fe.cpp
clang++-17: note: diagnostic msg: /tmp/acpp_stdpar_sigsegv-6f21fe.sh
clang++-17: note: diagnostic msg: 

********************


```

**Expected behavior**
Program compiles, as is the case without `-fsanitize=undefined`
```
> /home/tom/software/x86_64/acpp/367cb7a/bin/syclcc acpp_stdpar_sigsegv.cpp -std=c++17 -O3 -Wall --acpp-stdpar --acpp-targets=omp.accelerated -ltbb
> ./a.out  # exit code is 0
```

**Describe your setup**
```
clang version 17.0.4 (Fedora 17.0.4-1.fc39)
Target: x86_64-redhat-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
Found candidate GCC installation: /usr/bin/../lib/gcc/i686-redhat-linux/13
Found candidate GCC installation: /usr/bin/../lib/gcc/x86_64-redhat-linux/13
Selected GCC installation: /usr/bin/../lib/gcc/x86_64-redhat-linux/13
Candidate multilib: .;@m64
Candidate multilib: 32;@m32
Selected multilib: .;@m64
Found CUDA installation: /usr/local/cuda, version 11.8
Found HIP installation: /opt/rocm, version 5.5.30202
``` 

acpp commit is 367cb7a

nilsfriess:
In #1282 it was reported that the compiler segfaults when trying to compile with `--acpp-stdpar -fsanitize=undefined`.

The problem was that the `IntrospectStructPass` assumed that calls to the builtin `__hipsycl_introspect_flattened_struct` are always preceded by a alloca instruction in LLVM IR. However, it seems that when compiling with `-fsanitize=undefined`, the compiler adds some `__ubsan...` calls after [this line](https://github.com/AdaptiveCpp/AdaptiveCpp/blob/4e6608fe7ad9cc0a045d56de43705ff37d8c62e0/include/hipSYCL/glue/reflection.hpp#L49). Even worse, in LLVM IR this then ends up as a separate block, so that just traversing the instructions before the call to `__hipsycl_introspect_flattened_struct` does not work anymore.

This PR makes the introspect pass a bit more robust by checking the other blocks in the function for the alloca instruction (in case it wasn't found by the first approach). This is done, however, only if there is exactly one BasicBlock containing alloca instructions; within this block it takes the last alloca instruction.

Fixes #1282

### DIFF
--- a/include/hipSYCL/glue/reflection.hpp
+++ b/include/hipSYCL/glue/reflection.hpp
@@ -29,9 +29,11 @@
 #define HIPSYCL_REFLECTION_HPP
 
 template<class StructT>
-void __hipsycl_introspect_flattened_struct(int **num_flattened_members,
-                                      int **member_offsets, int **member_sizes,
-                                      int **member_kinds);
+void __hipsycl_introspect_flattened_struct(StructT *s,
+                                           int **num_flattened_members,
+                                           int **member_offsets,
+                                           int **member_sizes,
+                                           int **member_kinds) {}
 
 namespace hipsycl::glue::reflection {
 
@@ -44,12 +46,12 @@ public:
   : _num_members{nullptr} {
     // Compiler needs this copy to figure out
     // the struct type in the presence of opaque pointers.
-    // Currently it looks for alloca instructions preceding the
-    // builtin call for this purpose.
+    // Currently it checks if the first operand of the call
+    // to the builtin comes from an alloca instruction.
     StructT s_copy = s;
-    __hipsycl_introspect_flattened_struct<StructT>(
-        &_num_members, &_member_offsets, &_member_sizes,
-        reinterpret_cast<int **>(&_member_kinds));
+    __hipsycl_introspect_flattened_struct(&s_copy, &_num_members,
+                                          &_member_offsets, &_member_sizes,
+                                          reinterpret_cast<int **>(&_member_kinds));
   }
 
   int get_num_members() const {
--- a/include/hipSYCL/glue/reflection.hpp
+++ b/include/hipSYCL/glue/reflection.hpp
@@ -29,11 +29,11 @@
 #define HIPSYCL_REFLECTION_HPP
 
 template<class StructT>
-void __hipsycl_introspect_flattened_struct(StructT *s,
+void __hipsycl_introspect_flattened_struct(void *s,
                                            int **num_flattened_members,
                                            int **member_offsets,
                                            int **member_sizes,
-                                           int **member_kinds) {}
+                                           int **member_kinds);
 
 namespace hipsycl::glue::reflection {
 
@@ -49,9 +49,9 @@ public:
     // Currently it checks if the first operand of the call
     // to the builtin comes from an alloca instruction.
     StructT s_copy = s;
-    __hipsycl_introspect_flattened_struct(&s_copy, &_num_members,
-                                          &_member_offsets, &_member_sizes,
-                                          reinterpret_cast<int **>(&_member_kinds));
+    __hipsycl_introspect_flattened_struct<StructT>(&s_copy, &_num_members,
+                                                   &_member_offsets, &_member_sizes,
+                                                   reinterpret_cast<int **>(&_member_kinds));
   }
 
   int get_num_members() const {
--- a/src/compiler/reflection/IntrospectStructPass.cpp
+++ b/src/compiler/reflection/IntrospectStructPass.cpp
@@ -182,61 +182,92 @@ llvm::PreservedAnalyses IntrospectStructPass::run(llvm::Module& M, llvm::ModuleA
     }
   }
 
-  
   llvm::SmallDenseMap<llvm::Function*, TypeInformation> StructTypeInfoForBuiltin;
   llvm::SmallVector<llvm::CallInst*, 16> Calls;
   for(auto* F: BuiltinInstantiations) {
-    for(auto* U : F->users()) {
-      if(llvm::CallInst* CI = llvm::dyn_cast<llvm::CallInst>(U)) {
+    for (auto *U : F->users()) {
+      if (llvm::CallInst *CI = llvm::dyn_cast<llvm::CallInst>(U)) {
         Calls.push_back(CI);
+
         // Obtain type referenced by builtin calls. To workaround opaque pointers, this
         // assumes that an alloca instruction with the type is present in the preceding code.
-        auto* AI = getPrevAllocaInst(CI);
-        if(AI) {
-          llvm::Type* T = AI->getAllocatedType();
+        auto *AI = getPrevAllocaInst(CI);
+        if (AI == nullptr) {
+          /* In some cases it might happen that the alloca instruction is in a
+           * different BasicBlock (e.g., when compiling with -fsanitize=undefined).
+           * We therefore traverse all the BasicBlocks in the function and check
+           * if there is exactly one BasicBlock containing
+           * alloca instructions. Within this block, we take the last alloca
+           * instruction. */
+          llvm::BasicBlock *AllocaBlock;
+          std::size_t NumAllocaBlocks = 0;
+
+          // Instruction::getParent() gives the parent BasicBlock,
+          // BasicBlock::getParent the parent Function
+          for (auto &Block : *CI->getParent()->getParent()) {
+            for (auto &Inst : Block) {
+              if (const auto *NewAI = llvm::dyn_cast<llvm::AllocaInst>(&Inst)) {
+                AllocaBlock = &Block;
+                NumAllocaBlocks++;
+                break;
+              }
+            }
+          }
+
+          if (NumAllocaBlocks == 1) {
+            for (auto &Inst : *AllocaBlock) {
+              if (auto *NextAI = llvm::dyn_cast<llvm::AllocaInst>(&Inst)) {
+                AI = NextAI;
+              }
+            }
+          }
+        }
+
+        if (AI) {
+          llvm::Type *T = AI->getAllocatedType();
           auto TI = getTypeInformation(T, M);
           StructTypeInfoForBuiltin[F] = TI;
         } else {
           HIPSYCL_DEBUG_WARNING
-            << "IntrospectStructPass: " << F->getName().str()
-            << " instantiation could not be connected to type information, ignoring.\n";
+              << "IntrospectStructPass: " << F->getName().str()
+              << " instantiation could not be connected to type information, ignoring.\n";
         }
       }
     }
   }
 
-  for(auto* CI : Calls) {
-    llvm::Function* F = CI->getCalledFunction();
-    if(F) {
-      auto It = StructTypeInfoForBuiltin.find(F);
+  if (StructTypeInfoForBuiltin.size() > 0)
+    for(auto* CI : Calls) {
+      llvm::Function* F = CI->getCalledFunction();
+      if(F) {
+        auto It = StructTypeInfoForBuiltin.find(F);
 
-      auto GVs = storeTypeInformationAsGlobals(F, M, It->getSecond());
+        auto GVs = storeTypeInformationAsGlobals(F, M, It->getSecond());
 
-      if(It != StructTypeInfoForBuiltin.end()) {
-        if(F->getFunctionType()->getNumParams() != 4) {
-          HIPSYCL_DEBUG_WARNING << "IntrospectStructPass: Call to " << F->getName().str()
-                                << " has the wrong number of parameters, ignoring call\n";
-        } else {
-          auto createStoreOp = [&](int ArgIndex, llvm::GlobalVariable* GV) {
-            if(!CI->getArgOperand(ArgIndex)->getType()->isPointerTy()) {
-              HIPSYCL_DEBUG_WARNING
-                  << "IntrospectStructPass: Call to " << F->getName().str()
-                  << " is invalid; argument is not a pointer type. Ingoring call.\n";
-              return;
-            }
-            [[maybe_unused]] llvm::StoreInst *S =
+        if(It != StructTypeInfoForBuiltin.end()) {
+          if(F->getFunctionType()->getNumParams() != 4) {
+            HIPSYCL_DEBUG_WARNING << "IntrospectStructPass: Call to " << F->getName().str()
+            << " has the wrong number of parameters, ignoring call\n";
+          } else {
+            auto createStoreOp = [&](int ArgIndex, llvm::GlobalVariable* GV) {
+              if(!CI->getArgOperand(ArgIndex)->getType()->isPointerTy()) {
+                HIPSYCL_DEBUG_WARNING
+                << "IntrospectStructPass: Call to " << F->getName().str()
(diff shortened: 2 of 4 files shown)


illuhad:
Wow, thanks for looking into this!

I know of at least one other place where we also abuse allocas for this purpose: https://github.com/AdaptiveCpp/AdaptiveCpp/blob/08ec1a0cad71af6e11d3ed7601bae6760f481579/src/compiler/sscp/KernelOutliningPass.cpp#L97
We should also check whether this is affected too.

I wonder if we could not make it even better by going backwards in the user-tree: Find out where the argument to the function comes from. If it is from an alloca, great, we have found our alloca. If it is from a `getelementptr` instruction that's also okay because we can extract the type from there too. Any other instructions likely indicate that there is some additonal indirection which is not expected.

include/hipSYCL/glue/reflection.hpp
```diff
@@ -29,9 +29,11 @@
 #define HIPSYCL_REFLECTION_HPP
 
 template<class StructT>
-void __hipsycl_introspect_flattened_struct(int **num_flattened_members,
-                                      int **member_offsets, int **member_sizes,
-                                      int **member_kinds);
+void __hipsycl_introspect_flattened_struct(StructT *s,
```
nilsfriess: I've now added a pointer to StructT here and pass a pointer to the `s_copy` when calling the builtin. This seems to work properly. I first tried to make `s` here just of type StructT (no pointer) and pass `s` below directly. Then `getOperand(0)` on the builtin CallInst is a GetElementPtrInst but I couldn't get the type of the StructT this way, not sure what I did wrong. 

Also for some reason clang complains that the builtin is used but not defined as soon as you add the `StructT *` argument (I don't understand 100% why, it seems that for some stdpar calls the type `StructT` is a lambda and certain lambda types cannot  be shared across translation unit boundaries, see e.g. here https://stackoverflow.com/questions/76582942/lambda-variable-shared-between-translation-units). Therefore I added the empty definition `{}`.

Edit: I guess this might not work since the call to the builtin could just be optimized away?

nilsfriess: I think it can be even more simplified if we pass the `StructT` parameter by value to the builtin because we can then simply use something like
```cpp
llvm::Type *T = nullptr;
if (F->hasParamAttribute(0, llvm::Attribute::ByVal))
  T = F->getParamAttribute(0, llvm::Attribute::ByVal).getValueAsType();
else
  T = F->getArg(0)->getType();
StructTypeInfoForBuiltin[F] = getTypeInformation(T, M);
```
This has the additional benefit that we don't need the extra copy in the constructor of `introspect_flattened_struct`.
However, we'd still have the problem that the builtin has to be defined and not just declared (due to the issue with lambdas mentioned above). Although we could probably add `__attribute__((optnone))` to ensure that it's not optimized away.

illuhad: > Edit: I guess this might not work since the call to the builtin could just be optimized away?

Yes.

> I think it can be even more simplified if we pass the StructT parameter by value 

I've experimented a lot with passing things by value. In my experience, it is inherently unreliable because there's no guarantee how it gets passed, since that depends on ABI. For example, on x86, if the struct is smaller than 16 bytes, the struct might be decomposed into its individual members and those might be passed as separate arguments. If it is passed as one argument, it may or may not be passed as pointer with LLVM byval attribute depending on platform etc.

nilsfriess:
> I know of at least one other place where we also abuse allocas for this purpose:

By the way: I don't think this is affected, since this function [here](https://github.com/AdaptiveCpp/AdaptiveCpp/blob/08ec1a0cad71af6e11d3ed7601bae6760f481579/src/compiler/sscp/KernelOutliningPass.cpp#L133) just traverses all basic blocks in the function and looks for allocas instead of starting from some specific instruction and trying to go backwards.

src/compiler/reflection/IntrospectStructPass.cpp
```diff
@@ -182,61 +171,61 @@ llvm::PreservedAnalyses IntrospectStructPass::run(llvm::Module& M, llvm::ModuleA
...
-        if(AI) {
-          llvm::Type* T = AI->getAllocatedType();
-          auto TI = getTypeInformation(T, M);
-          StructTypeInfoForBuiltin[F] = TI;
+
+        if (auto *AI = llvm::dyn_cast<llvm::AllocaInst>(CI->getOperand(0))) {
+          llvm::Type *T = AI->getAllocatedType();
+          StructTypeInfoForBuiltin[F] = getTypeInformation(T, M);
+        } else if (auto *BI = llvm::dyn_cast<llvm::BitCastInst>(CI->getOperand(0))) {
+          llvm::Type *T = BI->getSrcTy()->getNonOpaquePointerElementType();
```
illuhad: The documentation on this says:
> Only use this method in code that is not reachable with opaque pointers, or part of deprecated methods that will be removed as part of the opaque pointers transition. 

So it might be a good idea to add an `isOpaquePointerTy()` check before invoking this API.
What would be even better is to have a recursive algorithm, which goes backwards through bitcasts (and perhaps getelementptr) until it hits an `AllocaInst`.

illuhad (approved):
Thanks!

PR #1318, opened 2024-01-18
Lazily construct multi-queue executors on all backends

### LINKED ISSUE #1246: Overhead of unused streams on AMD devices
**Bug summary**

At runtime startup, multi_queue_executor creates 4 streams for each device. If the application only uses in-order queues on a single device, all those streams are unused. At least with ROCm 5.3.3 and 8 GPUs, those background streams have a negative effect on the application performance when there are multiple devices in the system.

**To Reproduce**

- Get access to a machine with many AMD GPUs.
- Get https://github.com/zjin-lcf/HeCBench/tree/master/src/nbody-sycl.
- Apply the following patch to build code with AdaptiveCpp and allow setting the device via an env. variable: [acpp-nbody.txt](https://github.com/AdaptiveCpp/AdaptiveCpp/files/13393558/acpp-nbody.txt)
- Build the code.
- Try running 8 copies, one per device, ensuring the correct CPU-GPU pinning:

```bash
GPUS=(4 5 2 3 6 7 0 1)
for i in `seq 0 7`; do
    C0=$(($i * 8)); C1=$(($C0 + 7)); G=${GPUS[${i}]};
    DEVICE=$G hwloc-bind --cpubind core:${C0}-${C1} ./main 100000 100 &
done
```

The "Average Performance" varies a lot: from "2645.19 +- 696.754" to "3473.29 +- 157.254"

- Run the same, except limit GPU visibility on HIP level:

```bash
for i in `seq 0 7`; do
    C0=$(($i * 8)); C1=$(($C0 + 7)); G=${GPUS[${i}]};
    ROCR_VISIBLE_DEVICES=$G DEVICE=0 hwloc-bind --cpubind core:${C0}-${C1} ./main 100000 100 &
done
```

The performance is now higher and more stable: from "3573.92 +- 92.1017" to "3608.98 +- 83.5057"

- Hack `src/runtime/multi_queue_executor.cpp` to have `memcpy_concurrency` and `kernel_concurrency` set to 0, install it, rebuild the `nbody-sycl` application. Run it with all devices visible:

```bash
for i in `seq 0 7`; do
    C0=$(($i * 8)); C1=$(($C0 + 7)); G=${GPUS[${i}]};
    DEVICE=$G hwloc-bind --cpubind core:${C0}-${C1} ./main 100000 100 &
done
```

Performance is again good and stable, between "3574.66 +- 87.9561" and "3606.06 +- 84.6698"

**Expected behavior**

There should not be any overhead from unused devices just because they are available.

**Describe your setup**
* AdaptiveCpp 23.10.0
* Describe the dependencies that AdaptiveCpp sits on top of in your setup: 
   * ROCm Clang 15 from ROCm 5.3.3, on Dardel, but the behavior initially noticed with GROMACS on LUMI by a colleague.
       * When running on LUMI, different `hwloc-bind` flags are required to handle reserved cores.
   * 8 x MI250X

**Additional context**

- All these extra streams are also quite annoying in CUDA profiling traces.

- It makes enumerating devices slow: `acpp-info -l` takes ~9 seconds with AdaptiveCpp 23.10 on the machine; with the hack described above, it takes ~0.2 seconds.

- The hack described above is only useful for hypothesis testing, and is definitely not a proposed solution.

- We can blame ROCm for not handling idle queues, but having `multi_queue_executor` behave lazily and not create streams for out-of-order queues unless actually needed would be better.

illuhad:
Hopefully fixes #1246 

This PR causes multi-queue executors to only be constructed once they are actually needed. This should solve the issue for apps which only use in-order queues. However it does not solve the issue when out-of-order queues only use a subset of devices. That would require lazy construction inside `multi_queue_executor`. I suspect this might be more difficult.

CC @al42and

### DIFF
--- a/include/hipSYCL/runtime/multi_queue_executor.hpp
+++ b/include/hipSYCL/runtime/multi_queue_executor.hpp
@@ -189,8 +189,10 @@ public:
       return _ptr.get();
     else {
       std::lock_guard<std::mutex> lock{_mutex};
-      _ptr = _factory();
-      _is_initialized.store(true, std::memory_order_release);
+      if(!_is_initialized.load(std::memory_order_acquire)) {
+        _ptr = _factory();
+        _is_initialized.store(true, std::memory_order_release);
+      }
       return _ptr.get();
     }
   }
--- a/include/hipSYCL/runtime/multi_queue_executor.hpp
+++ b/include/hipSYCL/runtime/multi_queue_executor.hpp
@@ -182,7 +182,7 @@ class lazily_constructed_executor {
 public:
   template<class Factory>
   lazily_constructed_executor(Factory&& F)
-  : _factory{F}, _is_initialized{false} {}
+  : _factory{std::forward<Factory>(F)}, _is_initialized{false} {}
 
   Executor* get() {
     if(_is_initialized.load(std::memory_order_acquire))
--- a/include/hipSYCL/runtime/multi_queue_executor.hpp
+++ b/include/hipSYCL/runtime/multi_queue_executor.hpp
@@ -31,6 +31,8 @@
 #include <cmath>
 #include <cassert>
 #include <functional>
+#include <atomic>
+#include <mutex>
 
 #include "backend.hpp"
 #include "device_id.hpp"
@@ -173,6 +175,31 @@ private:
   std::size_t _num_submitted_operations;
   std::vector<inorder_queue*> _managed_queues;
   backend_id _backend;
+};
+
+template<class Executor>
+class lazily_constructed_executor {
+public:
+  template<class Factory>
+  lazily_constructed_executor(Factory&& F)
+  : _factory{F}, _is_initialized{false} {}
+
+  Executor* get() {
+    if(_is_initialized.load(std::memory_order_acquire))
+      return _ptr.get();
+    else {
+      std::lock_guard<std::mutex> lock{_mutex};
+      _ptr = _factory();
+      _is_initialized.store(true, std::memory_order_release);
+      return _ptr.get();
+    }
+  }
+
+private:
+  std::atomic<bool> _is_initialized;
+  std::mutex _mutex;
+  std::function<std::unique_ptr<Executor>()> _factory;
+  std::unique_ptr<Executor> _ptr = nullptr;
 };
 
 }
--- a/include/hipSYCL/runtime/cuda/cuda_backend.hpp
+++ b/include/hipSYCL/runtime/cuda/cuda_backend.hpp
@@ -64,7 +64,7 @@ public:
   create_inorder_executor(device_id dev, int priority) override;
 private:
   mutable cuda_hardware_manager _hw_manager;
-  mutable multi_queue_executor _executor;
+  mutable lazily_constructed_executor<multi_queue_executor> _executor;
 };
 
 }
--- a/include/hipSYCL/runtime/hip/hip_backend.hpp
+++ b/include/hipSYCL/runtime/hip/hip_backend.hpp
@@ -64,7 +64,7 @@ public:
   hip_event_pool* get_event_pool(device_id dev) const;
 private:
   mutable hip_hardware_manager _hw_manager;
-  mutable multi_queue_executor _executor;
+  mutable lazily_constructed_executor<multi_queue_executor> _executor;
 };
 
 }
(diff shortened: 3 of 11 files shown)


include/hipSYCL/runtime/multi_queue_executor.hpp
```diff
@@ -175,6 +177,31 @@ class multi_queue_executor : public backend_executor
...
+  template<class Factory>
+  lazily_constructed_executor(Factory&& F)
+  : _factory{std::forward<Factory>(F)}, _is_initialized{false} {}
+
+  Executor* get() {
+    if(_is_initialized.load(std::memory_order_acquire))
+      return _ptr.get();
+    else {
+      std::lock_guard<std::mutex> lock{_mutex};
+      _ptr = _factory();
```
al42and: Won't there be a race condition here?

- T1 checks the flag on line 188, it's false
- T2 checks the flag on line 188, it's false
- T1 enters this branch, acquires the mutex, calls the factory, sets the flag to `true`.
- T2 enter this branch, waits for the mutex and acquires it, calls the factory again.

al42and (approved):
I can confirm that it solves the problem as described. Thanks!

Apparently, at least with ROCm 5.3.3, creating a stream "taints" the GPU even after the stream is destroyed, so our device detection code (which iterates over devices, creating a stream on each one) still causes the performance fluctuations for GROMACS with multiple visible devices. But there's nothing ACpp can do here.

> Hopefully fixes https://github.com/AdaptiveCpp/AdaptiveCpp/issues/1246

Not sure if we want to close the issue completely even if this is resolved: I'd consider it strange when runtime allocates *any* resources on the devices unused, even if it causes no performance regression. Why even touch an integrated GPU when running on a discreet one? Even if the solution is not possible in the near future, that perhaps is still worth keeping track of.

include/hipSYCL/runtime/multi_queue_executor.hpp
illuhad: Yes... I forgot an additional `if`. I even had that in an earlier version of the code -.- Thanks for spotting this!

illuhad:
> Apparently, at least with ROCm 5.3.3, creating a stream "taints" the GPU even after the stream is destroyed, so our device detection code (which iterates over devices, creating a stream on each one) still causes the performance fluctuations for GROMACS with multiple visible devices. But there's nothing ACpp can do here.

Seems you are guilty of the same crime as I am :-) Out of curiosity, what is the background here for this device detection code?

>  Not sure if we want to close the issue completely even if this is resolved: I'd consider it strange when runtime allocates any resources on the devices unused, even if it causes no performance regression. Why even touch an integrated GPU when running on a discreet one? Even if the solution is not possible in the near future, that perhaps is still worth keeping track of.

Isn't this primarily a problem of the OpenCL backend, where many different devices may be handled by one `multi_queue_executor`? On HIP or CUDA backends, only very little resources will be used unless you use out-of-order queues. Basically it only queries device information.

As the original problem was solved, I think I'd prefer a new issue for the general problem.

al42and:
> Seems you are guilty of the same crime as I am :-)

I'd say not quite, we destroy the streams that are not used :) But for ROCm, that's equally bad.

> Out of curiosity, what is the background here for this device detection code?

- Report all devices in the log together with their status (usable or not),
- Provide early error-reporting in case the device is unusable (e.g., for native CUDA, if we explicitly compiled for an incompatible sm; or whatever broken driver issues might be, like, e.g., limitations of ACpp OpenCL backend on NVIDIA/AMD due to lack of SPIR-V ingestion (before those platforms were masked)).
- In case of multiple devices, avoid automatically selecting broken ones.

> Isn't this primarily a problem of the OpenCL backend, where many different devices may be handled by one multi_queue_executor? On HIP or CUDA backends, only very little resources will be used unless you use out-of-order queues. Basically it only queries device information.

Well, if an application uses out-of-order queues on a single device, the problem described in #1246 will be back. I don't want to test it, but I suspect there might be issues in situations with, e.g., a discreet GPU and an integrated GPU, and several processes (using out-of-order queues) running on the discreet GPU would affect the performance of the integrated device, despite not actually using it for anything.

> As the original problem was solved, I think I'd prefer a new issue for the general problem.

Sure, whatever works best for you!

pszi1ard:
@illuhad @al42and Thanks for working this out, hopefully with this resolved we can stop requiring `ROCR_VISIBLE_DEVICES` just to avoid performance falling off of a cliff.

> I'd consider it strange when runtime allocates any resources on the devices unused, even if it causes no performance regression.

I agree, ideally we should at least be able to tell the runtime to avoid doing so.

However, I guess we can work around this on our side in GROMACS by skipping sanity checks that require resources (e.g. to run a test kernel).

PR #1322, opened 2024-01-21
Declare project before cmake_minimum_required in tests

sbalint98:
Newer cmake versions (tested with 3.28) complain about calling the top-level project call before `cmake_minimum_required`. This PR fixes that.


illuhad (approved):
Thanks!

PR #1323, opened 2024-01-21
Emmit error in find_package() if no targets specified

sbalint98:
According to the [documentation](https://github.com/AdaptiveCpp/AdaptiveCpp/blob/develop/doc/using-hipsycl.md#using-the-cmake-integration) 

> A typical configure command line looks like this: `cmake .. -DAdaptiveCpp_DIR=/acpp/install/dir/lib/cmake/AdaptiveCpp -DACPP_TARGETS="<targets>"`.
> `ACPP_TARGETS` has to be set either as environment variable or on the command line for the `find_package` call to succeed. See the documentation of this flag above.

This PR emits a Fatal error from the CMake integration in case no targets have been found. Prior to this change, the configuration succeeded even though there were no targets specified.

To consider: Maybe emitting a warning and setting the package not found would be more elegant in this case?


illuhad:
hm... this might clash with #1290, depending on the direction we take there.

sbalint98:
superseded by #1290

PR #1326, opened 2024-01-24
Deprecate syclcc and syclcc-clang

illuhad:
Adds a warning if `syclcc` and `syclcc-clang` are used, instructing users to use `acpp` instead.

### DIFF
--- a/bin/acpp
+++ b/bin/acpp
@@ -2012,9 +2012,9 @@ if __name__ == '__main__':
   
   filename = os.path.basename(os.path.realpath(__file__))
   if filename == "syclcc":
-    print("acpp warning: syclcc is deprecated; please use acpp instead.")
+    print_warning("syclcc is deprecated; please use acpp instead.")
   if filename == "syclcc-clang":
-    print("acpp warning: syclcc-clang is deprecated; please use acpp instead.")
+    print_warning("syclcc-clang is deprecated; please use acpp instead.")
 
   args = sys.argv[1:]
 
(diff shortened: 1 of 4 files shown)


bin/acpp
```diff
@@ -2010,6 +2010,12 @@ if __name__ == '__main__':
     print("acpp requires python 3.")
     sys.exit(-1)
   
+  filename = os.path.basename(os.path.realpath(__file__))
+  if os.path.basename(os.path.realpath(__file__)) == "syclcc":
+    print("acpp warning: syclcc is deprecated; please use acpp instead.")
+  if os.path.basename(os.path.realpath(__file__)) == "syclcc-clang":
```
al42and: Looks like you left this change half-way :)

```suggestion
  if filename == "syclcc":
    print("acpp warning: syclcc is deprecated; please use acpp instead.")
  if filename == "syclcc-clang":
```

illuhad: :rofl: :cry: In my defense, I tested it and it worked correctly :P

bin/acpp
```diff
@@ -2010,6 +2010,12 @@ if __name__ == '__main__':
     print("acpp requires python 3.")
     sys.exit(-1)
   
+  filename = os.path.basename(os.path.realpath(__file__))
+  if filename == "syclcc":
+    print("acpp warning: syclcc is deprecated; please use acpp instead.")
```
fknorr: Consider printing to stderr instead (see #1276)

illuhad: Yep, will rebase on current develop which includes #1340

PR #1330, opened 2024-01-26
[SSCP] Add initial SSCP group algorithm builtin APIs excluding scans

illuhad:
This PR adds
* SSCP builtin interface API
* Top-level implementation on top of SSCP builtins

for all group algorithms except for scans (which need some more thought).
Note that this PR does not yet include the backend implementations of those SSCP group algorithm builtins, so the algorithms cannot yet actually be used by users. However, this PR is a first step towards exposing this functionality. And it might to better errors during JIT when group algorithms are used.

CC @sbalint98


illuhad:
@al42and I am currently moving ahead with adding extra optimizations in the SSCP compiler around the JIT compilation of specialized kernels. In the mid term, I would like to see if Gromacs can benefit from that, so I would like to ensure that SSCP works with Gromacs. Because of this I'd like to prioritize implementing those group algorithms that Gromacs needs in SSCP. Which group algorithms do you need? From a quick look at the code, I only see subgroup `reduce_over_group`. Do you need something else?

al42and:
Hi! We have the following, all over sub-groups: `sycl::any_of_group`, `sycl::reduce_over_group`, `sycl::select_from_group`, `sycl::shift_group_(left|right)`.

Having StdPar support for `std::exclusive_scan` (over a USM device buffer) would also be of interest soon-ish, but not a requirement.

illuhad:
Thanks, I believe we should be able to implement these fairly easily using ockl builtins on AMD.

> Having StdPar support for std::exclusive_scan (over a USM device buffer) would also be of interest soon-ish, but not a requirement.

Do you really want stdpar support and `std::exclusive_scan`, or do you want `hipsycl::algorithms::exclusive_scan`? The `hipsycl::algorithms` library is what sits below stdpar and has a more SYCL-like interface with `queue` etc. It can thus be used to express more control over asynchronous execution.
For stdpar you get automatic synchronization (which may or may not be optimized away by the compiler), and you pull in the all-is-shared-USM memory management layer, although that can be disabled if you don't want it.

al42and:
> Do you really want stdpar support and std::exclusive_scan, or do you want hipsycl::algorithms::exclusive_scan? The hipsycl::algorithms library is what sits below stdpar and has a more SYCL-like interface with queue etc. It can thus be used to express more control over asynchronous execution.

Thanks for asking. `hipsycl::algorithms::exclusive_scan` is a better fit.

illuhad:
I think we should merge this even though this PR does not yet provide implementations for the group algorithms, but it will allow other PRs to be built against the group algorithm API and provide implementations.

PR #1332, opened 2024-01-28
Add runpath $ORIGIN by default for runtime libraries

### LINKED ISSUE #1331: Loading shared libraries fails for execuatbles compiled with acpp (glibc 2.38)
**Bug summary**
The compiled executable by acpp does not find all dynamic library dependencies without setting `LD_LIBRARY_PATH`

**To Reproduce**
using test.cpp:
```cpp
#include<iostream>
int main()
{
        std::cout << "Hello World" << std::endl;
}
```
Compiling with acpp:
`acpp test.cpp`

The resulting executable returns the error:
```
[soproni@archlinux sycl_presentation]$ ./a.out 
./a.out: error while loading shared libraries: libacpp-common.so: cannot open shared object file: No such file or directory
```

**Expected behavior**

Executable runs without setting `LD_LIBRARY_PATH`

**Describe your setup**

AdaptiveCPP current head: 5a466f115446888b7265e9c4ed977a9f16547c58
`cmake  ..`
`make`
`cmake --install . --prefix=./install/`

glibc/ld version 2.38
Archlinux, updated at 2024-01-28

ELF dynamic section for the generated `./a.out`
```
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libacpp-rt.so]
 0x0000000000000001 (NEEDED)             Shared library: [libboost_context.so.1.83.0]
 0x0000000000000001 (NEEDED)             Shared library: [libboost_fiber.so.1.83.0]
 0x0000000000000001 (NEEDED)             Shared library: [libstdc++.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libomp.so]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x000000000000001d (RUNPATH)            Library runpath: [/home/soproni/Projects/AdaptiveCpp/build-repro/install/bin/../lib/:/usr/lib]
...
```
for `libacpp-rt.so`
```
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libacpp-common.so]
 0x0000000000000001 (NEEDED)             Shared library: [libstdc++.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x000000000000000e (SONAME)             Library soname: [libacpp-rt.so]
 0x000000000000001d (RUNPATH)            Library runpath: [/usr/local/lib:/usr/local/lib/hipSYCL/llvm-to-backend]
```

The `ld` documentation states:
```
       (3)  Using the directories specified in the DT_RUNPATH dynamic
            section attribute of the binary if present.  Such
            directories are searched only to find those objects required
            by DT_NEEDED (direct dependencies) entries and do not apply
            to those objects' children, which must themselves have their
            own DT_RUNPATH entries.  This is unlike DT_RPATH, which is
            applied to searches for all children in the dependency tree.
```

Therefore, based on the documentation, it makes sense that `libacpp-common.so` is not found by the executable since it is a transitive dependency, and its path should be included in the `RUNPATH` of `libacpp-rt.so`
**Additional context**
This issue is not present on older systems (tested with glibc 2.31). Probably something has been changed in the upstream linker.

sbalint98:
Resolve #1331 

Based on this presentation https://crascit.com/wp-content/uploads/2019/09/Deep-CMake-For-Library-Authors-Craig-Scott-CppCon-2019.pdf slide 99-110  this seems to be the canonical way to deal with the issue mentioned above

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -8,15 +8,13 @@ if(NOT APPLE AND NOT WIN32)
   # parallelism to bleed into code that is executed within the runtime, where
   # USM is a big no-no.
   set(HIPSYCL_STDPAR_RT_LINKER_FLAGS "-Wl,-Bsymbolic-functions")
+  set(base $ORIGIN)
 else()
   set(HIPSYCL_STDPAR_RT_LINKER_FLAGS "")
+  set(base @loader_path)
 endif()
 
 add_subdirectory(common)
-
-list( APPEND CMAKE_INSTALL_RPATH
-  ${CMAKE_INSTALL_PREFIX}/lib
-  ${CMAKE_INSTALL_PREFIX}/lib/hipSYCL/llvm-to-backend)
 
 set(WITH_LLVM_TO_AMDGPU_AMDHSA false)
 set(WITH_LLVM_TO_PTX false)
--- a/src/compiler/llvm-to-backend/CMakeLists.txt
+++ b/src/compiler/llvm-to-backend/CMakeLists.txt
@@ -4,6 +4,7 @@ set(LLVM_TO_BACKEND_INCLUDE_DIRS
   ${CMAKE_CURRENT_SOURCE_DIR}
   ${CMAKE_BINARY_DIR}/include)
 
+set(CMAKE_INSTALL_RPATH ${base} ${base}/../../)
 
 function(create_llvm_based_library)
   set(options)
--- a/src/runtime/CMakeLists.txt
+++ b/src/runtime/CMakeLists.txt
@@ -16,6 +16,8 @@ set(HIPSYCL_RT_EXTRA_LINKER_FLAGS ${ACPP_RT_SANITIZE_FLAGS})
 
 set(HIPSYCL_RT_EXTRA_LINKER_FLAGS ${HIPSYCL_RT_EXTRA_LINKER_FLAGS} ${HIPSYCL_STDPAR_RT_LINKER_FLAGS})
 
+set(CMAKE_INSTALL_RPATH ${base} ${base}/hipSYCL)
+
 add_library(acpp-rt SHARED
   application.cpp
   runtime.cpp
@@ -86,6 +88,7 @@ install(TARGETS acpp-rt
         LIBRARY DESTINATION lib
         ARCHIVE DESTINATION lib)
 
+set(CMAKE_INSTALL_RPATH ${base}/../ ${base}/llvm-to-backend)
 
 if(WITH_CUDA_BACKEND)
   add_library(rt-backend-cuda SHARED
@@ -251,7 +254,7 @@ if(WITH_CPU_BACKEND)
     omp/omp_queue.cpp)
 
     find_package(OpenMP REQUIRED)
-
+    
     target_include_directories(rt-backend-omp PRIVATE ${HIPSYCL_SOURCE_DIR}/include)
     if(APPLE)
       if(CMAKE_VERSION VERSION_LESS "3.16")
--- a/src/tools/acpp-info/CMakeLists.txt
+++ b/src/tools/acpp-info/CMakeLists.txt
@@ -11,5 +11,6 @@ target_link_libraries(acpp-info PRIVATE acpp-rt)
 # Make sure that acpp-info uses compatible sanitizer flags for sanitized runtime builds
 target_link_libraries(acpp-info PRIVATE ${ACPP_RT_SANITIZE_FLAGS})
 target_compile_options(acpp-info PRIVATE ${ACPP_RT_SANITIZE_FLAGS})
+set_target_properties(acpp-info PROPERTIES INSTALL_RPATH ${base}/../lib/)
 
 install(TARGETS acpp-info DESTINATION bin)


illuhad:
Interesting. I also don't see the issue on my recent Arch Linux. So I'm not sure if this is really a question of "how old is glibc".

I have a couple of questions:
* In the longer term, since `acpp-common` is an internal library, would it make more sense to just move it into `lib/hipSYCL`, which might already be rpathed? We might then have to move the runtime backend plugins into another subdirectory though.
* Can it break the other libraries that we built if we overwrite `CMAKE_INSTALL_RPATH` globally as in this PR?
* Why is this here not sufficient: https://github.com/AdaptiveCpp/AdaptiveCpp/blob/5a466f115446888b7265e9c4ed977a9f16547c58/src/CMakeLists.txt#L18 ? `acpp-common` lives in the lib directory, we add that path to the rpath paths, and `acpp-rt` is explicitly linked to the `acpp-common` target.

sbalint98:
> Interesting. I also don't see the issue on my recent Arch Linux. So I'm not sure if this is really a question of "how old is glibc".

Mhm, I will try to reproduce this in a container to be sure.

> Why is this here not sufficient: 

Ahh, I think you just need to move the `add_subdirectory(common)` call after this declaration, and it should work.  So I think lets just stick with this option.

> In the longer term, since acpp-common is an internal library, would it make more sense to just move it into lib/hipSYCL, which might already be rpathed?

I think structurally that would make more sense, but I don't think that this is too important, at least I am not aware any issue that would require this restructuring apart from aesthetics

illuhad:
> Ahh, I think you just need to move the add_subdirectory(common) call after this declaration, and it should work. So I think lets just stick with this option.

Sounds good. Are you going to update this PR?

> I think structurally that would make more sense, but I don't think that this is too important, at least I am not aware any issue that would require this restructuring apart from aesthetics

I agree, it's a low-priority issue. I just mentioned it because I think one of the Fedora packaging guidelines for rpath usage says that it should be only used for internal libraries, that then should be in some subdirectory. The question of packaging for Fedora was brought up previously by users.

sbalint98:
I made this draft because I realized that the underlying issue is different: Even though the `RPATH` is being set currently using `CMAKE_INSTALL_PREFIX` the problem is, when installing using `cmake --install . --prefix <install_path>` the `CMAKE_ISNTALL_PREFIX` path is not replaced with the new prefix.

Some people also find this practice bad on the internet: https://discourse.cmake.org/t/cmake-install-prefix-not-work/5040

However I am not sure how this should be done in a better way apart from using `$ORIGIN`  which has its own problems. At this point I think the best course of action would be to discourage the user from using `cmake --install . --prefix`   

EDIT: `$ORIGIN` should work well on _most UNIX based systems_ whatever that should mean. On Apple some other measures are needed, while on Windows RPATH is not supported at all...

sbalint98:
Okay, this is the best solution I could come up with. It is based on the _Craig Scott - Professional CMake_ A Practical Guide_  section 27.2.2

src/CMakeLists.txt
```diff
@@ -8,16 +8,14 @@ if(NOT APPLE AND NOT WIN32)
   # parallelism to bleed into code that is executed within the runtime, where
   # USM is a big no-no.
   set(HIPSYCL_STDPAR_RT_LINKER_FLAGS "-Wl,-Bsymbolic-functions")
+  set(base $ORIGIN)
 else()
   set(HIPSYCL_STDPAR_RT_LINKER_FLAGS "")
+  set(base @loader_path)
```
illuhad, line 14: What does this do?

illuhad (commented):
Looks good as far as I can see. Have you actually tried relocating an installtion? Is that working now?

src/CMakeLists.txt
sbalint98: This is needed for apple. I think it wont have an effect on windows

sbalint98:
> Looks good as far as I can see. Have you actually tried relocating an installtion? Is that working now?

Yes, I did, and it works well now. I noticed an issue with rocm libraries not being found with rocm 6.0, but that seems to be an issue with the rocm cmake integration. 

The failing CI check looks strange, any idea what might be causing that? To me, it looks unrelated to this PR.

src/CMakeLists.txt
illuhad: I understand that it's needed for Apple, I was just curious what `@loader_path` means :D

illuhad:
>  Yes, I did, and it works well now. I noticed an issue with rocm libraries not being found with rocm 6.0, but that seems to be an issue with the rocm cmake integration.

Excellent :)

> The failing CI check looks strange, any idea what might be causing that? To me, it looks unrelated to this PR.

Yep, unrelated to this PR. Happens occasionally. Probably some bug in the buffer-accessor model. Perhaps related to #1376 (if we're lucky, that PR fixes it).

illuhad (approved):

PR #1333, opened 2024-01-30
Fix warnings when compiling with `-acpp-stdpar -Wall -Wextra`

nilsfriess:
When compiling stdpar applications with `-Wall -Wextra` users will get a few warnings (specifically: comparison of int and size_t, unused parameters and constructor initialisier list order being different from member declaration order). This fixes those warnings.

The only one I wasn't sure about is `n` being unused in `get_operation_hash(AlgorithmType arg, Size n, Args...)`. I've now just removed it from the function arguments, maybe the first idea was to include the problem size in the hash? But I think the problem size is handled in offload.hpp separately, so I guess it can be removed here.

### DIFF
--- a/include/hipSYCL/std/stdpar/detail/offload.hpp
+++ b/include/hipSYCL/std/stdpar/detail/offload.hpp
@@ -424,7 +424,7 @@ bool should_offload(AlgorithmType type, Size n, const Args &...args) {
       state.get_configuration().get_min_ops_per_offload_decision();
 
   int num_predicted_ops = 0;
-  uint64_t op_hash = get_operation_hash(type, args...);
+  uint64_t op_hash = get_operation_hash(type, n, args...);
   // Identify ops using a combination of hash and problem size
   using op_id = offload_heuristic_state::op_id;
 
@@ -585,7 +585,7 @@ private:
 template<class AlgorithmType, class Size, class F, typename... Args>
 auto host_instrumentation(F&& f, AlgorithmType t, Size n, Args... args) {
 #ifndef __HIPSYCL_STDPAR_UNCONDITIONAL_OFFLOAD__
-  uint64_t hash = get_operation_hash(t, args...);
+  uint64_t hash = get_operation_hash(t, n, args...);
   host_invocation_measurement m{hash, n};
   return m(f);
 #else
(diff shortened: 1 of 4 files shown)


include/hipSYCL/std/stdpar/detail/offload.hpp
```diff
@@ -424,7 +424,7 @@ bool should_offload(AlgorithmType type, Size n, const Args &...args) {
       state.get_configuration().get_min_ops_per_offload_decision();
 
   int num_predicted_ops = 0;
-  uint64_t op_hash = get_operation_hash(type, n, args...);
+  uint64_t op_hash = get_operation_hash(type, args...);
```
illuhad: Yes, the idea was indeed to allow for the hash to include the problem size. We are not currently doing this because in the current implementation, we want to interpolate between problem sizes of one operation. Because of this, it is useful to store the operation data based on a hash without problem size, and then within the operation data store one entry per problem size.
However, I'm not sure if this is going to be the final way of doing things, so it might be a good idea to keep the problem size in the interface.

include/hipSYCL/std/stdpar/detail/offload.hpp
```diff
@@ -585,7 +585,7 @@ struct device_invocation_measurement {
 template<class AlgorithmType, class Size, class F, typename... Args>
 auto host_instrumentation(F&& f, AlgorithmType t, Size n, Args... args) {
 #ifndef __HIPSYCL_STDPAR_UNCONDITIONAL_OFFLOAD__
-  uint64_t hash = get_operation_hash(t, n, args...);
+  uint64_t hash = get_operation_hash(t, args...);
```
illuhad: Same here

include/hipSYCL/std/stdpar/detail/offload.hpp
```diff
@@ -596,7 +596,7 @@ auto host_instrumentation(F&& f, AlgorithmType t, Size n, Args... args) {
 template<class AlgorithmType, class Size, class F, typename... Args>
 auto device_instrumentation(F&& f, AlgorithmType t, Size n, Args... args) {
 #ifndef __HIPSYCL_STDPAR_UNCONDITIONAL_OFFLOAD__
-  uint64_t hash = get_operation_hash(t, n, args...);
+  uint64_t hash = get_operation_hash(t, args...);
```
illuhad: Same here

illuhad (commented):
Thanks!

include/hipSYCL/std/stdpar/detail/offload.hpp
nilsfriess: I've added it back

illuhad (approved):
Thanks!

PR #1335, opened 2024-01-30
NFC: Remove trailing whitespace

hdelan:
Non functional change. Removing trailing whitespace.


illuhad:
Wow, thank you! This PR touches pretty much every file. Would it be possible to split it up into smaller PRs for the different components of the project? Otherwise handling all the merge conflicts might be difficult.

hdelan:
Yep sure no problem. Let me know how I should split this up into smaller PRs

illuhad:
Perhaps one simple heuristic might be by subdirectory? Given that all of this is NFC, we can group the changes arbitrarily, so I don't have a strong opinion.

hdelan:
OK sure so is it enough to subdivide into `include`, `src` and `test`?

PR #1170, opened 2023-10-05
Update the path to include directory in AdaptiveCpp's installation

normallytangent:
Fixes #1121

PR #1337, opened 2024-01-31
Fix various warnings when building AdaptiveCpp with `-Wall -Wextra -Wpedantic -pedantic`

nilsfriess:
When building AdaptiveCpp with `-Wall -Wextra -Wpedantic -pedantic`, a lot of warnings are emitted (such as: order of initialisation in constructor initialisier lists, unused lambda captures, unused variables, not defining assignment operators even though a copy constructor was explicitly defined etc.).
This PR introduces some changes that silence _some_ warnings. There are still a lot of warnings concerning comparison of integers of different signs and unused parameters. Most of the unused parameters are in methods that override some virtual method of a base class and thus need to have certain arguments. We could consider adding `[[maybe_unused]]` or removing the parameter's name for them. Regarding the sign-compare warnings: I think they should also be fixed at some point but there are quite a lot of them (~200), so I didn't feel like going through and fixing them all yet. 

Now, compiling with `-Wall -Wextra -Wpedantic -pedantic -Wno-sign-compare -Wno-unused-parameter` compiles ~_almost_~ without warnings.~, there is one warning left about an unused variable where I'm not sure if it can just be removed or if there's some additional logic missing. 
The unused variable is [here](https://github.com/AdaptiveCpp/AdaptiveCpp/blob/5a466f115446888b7265e9c4ed977a9f16547c58/src/compiler/reflection/IntrospectStructPass.cpp#L227), and I think it can be removed (actually the whole branch starting [here](https://github.com/AdaptiveCpp/AdaptiveCpp/blob/5a466f115446888b7265e9c4ed977a9f16547c58/src/compiler/reflection/IntrospectStructPass.cpp#L215)).~

Just ignore this :D I guess the variable can be removed, but the constructor will already insert the created instruction  somewhere so that no additional `instructions->AddInst()` or something similar has to be called?

### DIFF
--- a/src/compiler/CMakeLists.txt
+++ b/src/compiler/CMakeLists.txt
@@ -79,9 +79,9 @@ add_library(acpp-clang SHARED
 
 target_include_directories(acpp-clang PRIVATE
   ../../include
-  ${LLVM_INCLUDE_DIRS}
   ${CMAKE_CURRENT_SOURCE_DIR}
   ${CMAKE_BINARY_DIR}/include)
+target_include_directories(acpp-clang SYSTEM PRIVATE ${LLVM_INCLUDE_DIRS})
 
 target_compile_definitions(acpp-clang PRIVATE
   ${LLVM_DEFINITIONS} -DHIPSYCL_COMPILER_COMPONENT)
--- a/src/runtime/multi_queue_executor.cpp
+++ b/src/runtime/multi_queue_executor.cpp
@@ -115,7 +115,7 @@ std::size_t determine_target_lane(dag_node_ptr node,
 
 multi_queue_executor::multi_queue_executor(
     const backend &b, queue_factory_function queue_factory)
-    : _num_submitted_operations{0}, _backend{b.get_unique_backend_id()} {
+    : _backend{b.get_unique_backend_id()} {
   std::size_t num_devices = b.get_hardware_manager()->get_num_devices();
 
 
--- a/include/hipSYCL/compiler/llvm-to-backend/Utils.hpp
+++ b/include/hipSYCL/compiler/llvm-to-backend/Utils.hpp
@@ -287,6 +287,7 @@ public:
         assert(CallArgs[i]->getType() == F->getFunctionType()->getParamType(i));
       }
 
+      llvm::CallInst::Create(llvm::FunctionCallee(F), CallArgs, "", BB);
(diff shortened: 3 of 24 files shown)


src/compiler/CMakeLists.txt
```diff
@@ -79,9 +79,9 @@ add_library(acpp-clang SHARED
 
 target_include_directories(acpp-clang PRIVATE
   ../../include
-  ${LLVM_INCLUDE_DIRS}
   ${CMAKE_CURRENT_SOURCE_DIR}
   ${CMAKE_BINARY_DIR}/include)
+target_include_directories(acpp-clang SYSTEM PRIVATE ${LLVM_INCLUDE_DIRS})
```
illuhad, line 84: I assume this means that the LLVM headers trigger a lot of warnings too? Is `SYSTEM` the common solution for these purposes? I am a bit concerned that `SYSTEM` might cause `cmake` to use `-isystem`, which can change the include order of system header search paths. We've had some problems with this previously AFAIR.

src/runtime/multi_queue_executor.cpp
```diff
@@ -115,7 +115,7 @@ std::size_t determine_target_lane(dag_node_ptr node,
 
 multi_queue_executor::multi_queue_executor(
     const backend &b, queue_factory_function queue_factory)
-    : _num_submitted_operations{0}, _backend{b.get_unique_backend_id()} {
+    : _backend{b.get_unique_backend_id()} {
```
illuhad, line 118: Where did the initialization go? Is it already handled somewhere else?

illuhad (commented):
Thanks!

src/runtime/multi_queue_executor.cpp
nilsfriess: This variable is not used anywhere, I also removed its declaration in multi_queue_executor.hpp

src/compiler/CMakeLists.txt
nilsfriess: Yes I added it to make cmake add `-isystem` to suppress warnings in LLVM headers. I remember there were some issues with the order of some system headers, but I thought this was related to how/where AdaptiveCpp was installed, see #1170 and the issues linked there. 

I don't think that this will cause issues here, I thought this was standard practice to suppress warnings from library headers.

illuhad (approved):
Thanks!

PR #1343, opened 2024-02-07
[CI] Add benchmark workflow that automatically reports performance regressions

nilsfriess:
This PR adds a new Github CI workflow that performs benchmarks and reports (in the form of a comment to the respective PR) potential performance regressions. 

This is still WIP but I had to open a PR to be able to actually run the workflow on the self-hosted runners.


nilsfriess:
/run-benchmark

In theory, a comment like this should trigger the benchmark run (I think this is a bit better than adding and removing labels) but according to the Github docs this only works if the workflow file is already present in the branch that the PR is trying to be merged into. 

Maybe it might also make sense to make this a bit more restrictive, e.g. only allowing some people (maybe members of the AdaptiveCpp GitHub organisation) to run benchmarks using a comment like this.

nilsfriess:
The github benchmark action [recently added](https://github.com/benchmark-action/github-action-benchmark/pull/223) a feature that comments benchmark results on pull requests (and updates the results if new commits are added to the PR), so maybe we can merge this and test if it works as expected and then as a follow-up extend the actual benchmarks that are run.

illuhad (approved):

PR #1344, opened 2024-02-07
[Packaging] Publish AdaptiveCpp package for Ubuntu

normallytangent:
This PR adds a Github workflow for publishing AdaptiveCpp packages on every `git-push`. 

- [x] Currently the `debian` directory is  handled manually: ideally I would want to automate this
- [x] Publish the resulting artifact on Github
- [x] Package for architecture `any`
- [x] Use custom compiler
- [x] Separate packages for cpu/sscp and cpu/cuda/rocm/ze backends


DhruvDh:
I tried adding a job for a `rocm` backend, but it runs out of storage space when installing the `rocm` ubuntu package.
In theory, it should work with a runner that has more storage.

Here it is, in case it helps -

```
  packaging-acpp-hip-backend:
    name: Packaging AdaptiveCpp, clang ${{matrix.clang}}, ${{matrix.os}}, ROCm ${{matrix.rocm}}
    runs-on: ${{matrix.os}}
    strategy:
      matrix:
        clang: ["17"]
        os: [ubuntu-22.04]
        rocm: [6.0.2]
    steps:
      - uses: actions/checkout@v4
        with:
          submodules: recursive
      - name: install rocm
        run: |
          wget https://repo.radeon.com/rocm/rocm.gpg.key -O - | \
              gpg --dearmor | sudo tee /etc/apt/keyrings/rocm.gpg > /dev/null
          echo "deb [arch=amd64 signed-by=/etc/apt/keyrings/rocm.gpg] https://repo.radeon.com/rocm/apt/6.0.2 jammy main" \
              | sudo tee --append /etc/apt/sources.list.d/rocm.list
          echo -e 'Package: *\nPin: release o=repo.radeon.com\nPin-Priority: 600' \
              | sudo tee /etc/apt/preferences.d/rocm-pin-600
          sudo apt update
          sudo apt install -y rocm${{matrix.rocm}}
      - name: install LLVM
        run: |
          wget https://apt.llvm.org/llvm.sh
          chmod +x llvm.sh
          sudo ./llvm.sh ${{matrix.clang}}
          sudo apt install -y libclang-${{matrix.clang}}-dev clang-tools-${{matrix.clang}} libomp-${{matrix.clang}}-dev llvm-${{matrix.clang}}-dev liblld-${{matrix.clang}}-dev
      - name: install dependencies
        run: |
          sudo apt install -y cmake libboost-all-dev wget gpg curl
      - name: setup build environment
        run: |
          export CXXFLAGS="$CXXFLAGS"
          if [[ "${{matrix.clang}}" != "11" && "${{matrix.clang}}" -lt "16" ]]; then
            export OMP_CXX_FLAGS="$CXXFLAGS -fexperimental-new-pass-manager"
            export CC=clang-${{matrix.clang}}
            export CXX=clang++-${{matrix.clang}}
          fi
          echo "CC=${CC}" >> $GITHUB_ENV
          echo "CXX=${CXX}" >> $GITHUB_ENV
          echo "OMP_CXX_FLAGS=${OMP_CXX_FLAGS}" >> $GITHUB_ENV
      - name: install packaging helpers
        run: |
          sudo  apt install -y build-essential
          sudo  apt install -y devscripts
          sudo  apt install -y debhelper
      - name: package AdaptiveCpp
        run: |
          cd ..
          cp -R AdaptiveCpp acpp-23.10.0
          tar -cf acpp_23.10.0.orig.tar.gz acpp-23.10.0/
          cd acpp-23.10.0/
          sed -i "s/-DWITH_ROCM_BACKEND=.*/-DWITH_ROCM_BACKEND=ON \\\/" debian/rules
          debuild -us -uc
      - name: Install the package
        run: |
          cd ${GITHUB_WORKSPACE}/../
          sudo dpkg -i acpp_23.10.0-1_amd64.deb
      - name: Release Packages
        uses: actions/upload-artifact@v4
        with:
          name: rocm-backend
          path: /home/runner/work/AdaptiveCpp/acpp_23.10.0-1_amd64.deb
      - name: build rocm tests
        run: |
          mkdir ${GITHUB_WORKSPACE}/build/tests-rocm
          cd ${GITHUB_WORKSPACE}/build/tests-rocm
          cmake -DAdaptiveCpp_DIR=${GITHUB_WORKSPACE}/build/install/lib/cmake/AdaptiveCpp ${GITHUB_WORKSPACE}/tests -DCMAKE_EXE_LINKER_FLAGS="-L/usr/lib/llvm-${{matrix.clang}}/lib"
          make -j2
```

illuhad:
@DhruvDh Thanks. I think the space problem can be solved by thinning out the default runner image a little, as we already do here: https://github.com/AdaptiveCpp/AdaptiveCpp/blob/b9b9d32226c88ea6cf67c59dd3d7dcbd8a11d019/.github/workflows/linux.yml#L145
However, I don't think that we need a separate ROCm package. We can just have one that supports everything :) Then what it can offload to only depends on whether users have CUDA/ROCm/OpenCL etc installed. I think the lack of ROCm support in the current version of this PR is just due to it being WIP.
(and probably we shouldn't target ROCm 6.0, as it has some issues)

.github/workflows/package-ubuntu.yml
```diff
@@ -0,0 +1,207 @@
...
+          LD_LIBRARY_PATH=/usr/lib ./sycl_tests
+  packaging-acpp-all-backends:
+    name: Packaging AdaptiveCpp, clang ${{matrix.clang}}, ${{matrix.os}}, CUDA ${{matrix.cuda}}, ROCm ${{matrix.rocm}}
+    runs-on: ${{matrix.os}}
+    strategy:
+      matrix:
+        clang: ['17']
+        os: [ubuntu-22.04]
+        cuda: [11.0.2]
+        rocm: [5.6.1]
```
illuhad, line 15: clang 17 is likely too new for ROCm 5.6.
SSCP requires that clang <= ROCm clang.
The old compilation flow requires that clang >= ROCm clang.

So best bet is probably to use the same clang/LLVM version that ROCm uses.

.github/workflows/package-ubuntu.yml
```diff
@@ -0,0 +1,207 @@
...
+            export CC=clang-${{matrix.clang}}
+            export CXX=clang++-${{matrix.clang}}
+          fi
+          echo "CC=${CC}" >> $GITHUB_ENV
+          echo "CXX=${CXX}" >> $GITHUB_ENV
+          echo "OMP_CXX_FLAGS=${OMP_CXX_FLAGS}" >> $GITHUB_ENV
+      - name: install CUDA
+        run : |
+          mkdir -p /opt/AdaptiveCpp/cuda
+          wget -q -O cuda.sh http://developer.download.nvidia.com/compute/cuda/11.0.2/local_installers/cuda_11.0.2_450.51.05_linux.run
```
illuhad: If we download CUDA from NVIDIA anyway, we might want to pick a newer one. CUDA 11.0 is already pretty old.

.github/workflows/package-ubuntu.yml
```diff
@@ -0,0 +1,207 @@
...
+      - name: install packaging helpers
+        run: |
+          sudo  apt install build-essential
+          sudo  apt install devscripts
+          sudo  apt install debhelper
+      - name: package AdaptiveCpp
+        run: |
+          cd ..
+          cp -R AdaptiveCpp acpp-24.02.0
+          tar -cf acpp_24.02.0.orig.tar.gz acpp-24.02.0/
```
illuhad: Does it actually need to be versioned here? Ultimately we'd like to have new packages for every push anyway, and it would be easier to maintain if we didn't have to always update version numbers.

normallytangent: yes, the tar file needs to carry this name-version scheme. Part of the initialization process is to create an upstream tarball `<name>_<version>-<package_revision>.tar.xz`

.github/workflows/package-ubuntu.yml
normallytangent: `libcudart12.0` is not available through aptitude, this creates a problem when the Debian helper `shlibdeps` is resolving CUDA dependency information (path to dynamic library and version) during the installation phase. So to circumvent that I also downgraded CUDA Toolkit to 11.0.

.github/workflows/package-ubuntu.yml
normallytangent: Okay, then I'll downgrade clang to 16 since that's what's shipped in ROCm 5.6 👍

illuhad:
Do I understand correctly that the packages are now ready for testing? If so, I call on everybody reading here to give them a try and see if they work :)

normallytangent:
The packages at this stage in this branch are ready. Would be great to hear some feedback, especially if the configuration even works. 
I am currently working on a way to introduce a testing-friendly naming convention for the binaries. Which is on a separate branch on my fork for reasons that I keep needing to make multiple commits with tiny changes to make the CI happy.

PR #1346, opened 2024-02-09
[SSCP][AMDGPU] Add missing write-back of result to argument in `frexp`.

fodinabor:
This seems to be missing..


illuhad (approved):
Thanks!

PR #1351, opened 2024-02-11
Align compilation flows by using -ffp-contract=fast by default

illuhad:
The clang CUDA/HIP toolchains, and by extension, our CUDA and HIP targets, by default compile with `-ffp-contract=fast`. This allows clang to emit FMA operations even for operations that are not part of the same statement.

`omp.accelerated` and `generic` however currently just do what clang does by default, which is `-ffp-contract=on`, which only allows FMA operations within the same statement.

These different defaults can cause noticable performance deltas for compute-intensive code. For the average user, it is very intransparent that this is just due to different defaults, and they might wrongly conclude e.g. that `--acpp-targets=generic` is slower than `--acpp-targets=cuda`, even when that is not necessarily the case.

This PR thus proposes to universally adopt `-ffp-contract=fast` if
* optimization level (via `-O`) is greater than 1
* We are compiling with clang. Other compilers may have different compiler flags, and different conventions about what is in included in `-O` flags. We could consider to also special-casing gcc and enabling it there too.

In other words, this PR aligns the default behavior of all clang-based compilation flows.
**Note that currently, with omp.library-only whether the flag is used depends on whether the host compiler is clang or gcc/some other compiler!**

Alternatively, we could also consider going the other route and disabling `-ffp-contract=fast` on CUDA/HIP to align them with the other flows. However, I believe that users might be surprised by a sudden performance drop when they upgrade AdaptiveCpp, or misinterpret AdaptiveCpp performance numbers when comparing to clang CUDA or clang HIP. So I believe it is better to have `-ffp-contract=fast` everywhere.

In an ideal world, none of this would be necessary as all of clang's toolchains would ideally just use the regular clang defaults and thus avoid surprises. Alas, clang CUDA and HIP have decided that they want to be different, so we have to deal with it.

Note that a similar discussion could also be had about `-ffast-math`, which the oneAPI SYCL compiler enables by default (and we don't, like every other sane compiler). I hate having `-ffast-math` enabled by default, but I'm also pretty certain that this is the explanation behind many cases where oneAPI leads AdaptiveCpp in benchmarks. Generally, these differences in default optimization flags are not widely known and rarely taken into account in my experience.

**For `omp.accelerated` and `generic` users, this PR might lead to substantial performance improvements as well as changes in the default rounding behavior.**

**For users of `hip` and `cuda` targets, this PR does not change anything compared to the current defaults.**

### DIFF
--- a/bin/acpp
+++ b/bin/acpp
@@ -865,6 +865,31 @@ class acpp_config:
   def common_compiler_args(self):
     return self._common_compiler_args
 
+
+  @property
+  def default_clang_optimization_args(self):
+    has_ffp_contract = False
+    opt_level = 0
+    opt_args = []
+
+    for arg in self._args:
+      if arg.startswith("-ffp-contract="):
+        has_ffp_contract = True
+      if arg.startswith("-O"):
+        ending = arg.replace("-O", "", 1)
+        if ending.isnumeric():
+          opt_level = int(ending)
+        elif ending == "fast":
+          opt_level = 3
+        else:
+          opt_level = 1
+
+    if opt_level > 1:
+      if not has_ffp_contract:
+        opt_args.append("-ffp-contract=fast")
+
+    return opt_args
+
   @property
   def acpp_include_path(self):
     return os.path.join(self.acpp_installation_path, "include", "AdaptiveCpp")
@@ -1639,6 +1664,9 @@ class compiler:
     self._is_stdpar = config.is_stdpar
     self._is_stdpar_system_usm = config.is_stdpar_system_usm
     self._is_stdpar_unconditional_offload = config.is_stdpar_unconditional_offload
+    self._clang_opt_args = config.default_clang_optimization_args
+    self._clang_path = config.clang_path
+
     try:
       self._stdpar_prefetch_mode = config.stdpar_prefetch_mode
     except OptionNotSet:
@@ -1941,6 +1969,9 @@ class compiler:
     cxx_flags = self.common_cxx_flags
     ld_flags = self.common_linker_flags
     compiler_executable = self._host_compiler
+
+    if compiler_executable == self._clang_path:
+      cxx_flags += self._clang_opt_args
 
     for backend_args in self._backends:
       cxx_flags += backend_args.get_cxx_flags()
--- a/bin/acpp
+++ b/bin/acpp
@@ -1665,7 +1665,10 @@ class compiler:
     self._is_stdpar_system_usm = config.is_stdpar_system_usm
     self._is_stdpar_unconditional_offload = config.is_stdpar_unconditional_offload
     self._clang_opt_args = config.default_clang_optimization_args
-    self._clang_path = config.clang_path
+    try:
+      self._clang_path = config.clang_path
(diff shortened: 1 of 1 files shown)


fxzjshm:
[The clang user manual](https://clang.llvm.org/docs/UsersManual.html#cmdoption-ffp-contract) says HIP's default is `fast-honor-pragmas` not `fast`, so should `acpp` set default to `fast-honor-pragmas` when compiling for HIP, or further, for all targets?
Note: `fast-honor-pragmas` wasn't supported in Clang driver until Clang 18 ([this commit](https://github.com/llvm/llvm-project/commit/e94c171ddb036e9672681d74725069352dd04d0e))

illuhad:
LLVM 18 is not even released, so I don't think adopting `-ffp-contract=fast-honor-pragmas` makes sense at the moment. Until clang 18, the default is `-ffp-contract=fast` for HIP too as far as I know.

It's saddening to see even clang CUDA and HIP not agreeing either, with HIP explicitly diverging from the CUDA default behavior.
In this case I think we have no choice to force a different default behavior on either CUDA or HIP. I'm inclined to stick to `-ffp-contract=fast` because
* I don't think we particularly guarantee the functionality of any pragmas in device code. If users want to use pragmas, they can still use `-ffp-contract=fast-honor-pragmas` explicitly
* Moving to `-ffp-contract=fast-honor-pragmas` requires additional effort because we also need to support older LLVM versions, which will continue to need `-ffp-contract=fast`. So we would have diverging behavior between AdaptiveCpp installations based on different LLVM Versions. This could be confusing.
* We could also adopt `-ffp-contract=fast` for CUDA and `-ffp-contract=fast-honor-pragmas` for HIP, but that would also be confusing because then even a single AdaptiveCpp will not behave consistently. And it's still unclear what the other targets (like `generic`) should follow.
* Aligning with HIP is not more important than aligning with CUDA

fxzjshm:
`-ffp-contract=fast-honor-pragmas` is default for HIP at least since LLVM 12 ([this commit](https://github.com/llvm/llvm-project/commit/cb08558caa3bad69213b08e6361586491232c745)), and can be verified [like this](https://godbolt.org/z/9Mc5ocKzW).

As said in that commit,
> What is worse is that since backend with 'Fast' fp fuse
option does not respect contract flag, there is no way to use #pragma clang fp contract
directive to enforce fp contract requirements.

To be honest, I think `fast` is nonsense because it violates/ignores the pragma the coder gives, even when it can.

I still think `fast-honor-pragmas` better than just `fast` since it still default to `fast`, thus has same performance; and only those who encounters numeric accuracy problems may change this flag, they either use `-ffp-contract=on` or `-ffp-contract=fast-honor-pragmas` & `#pragma STDC FP_CONTRACT ON` in required scope. 
So changing to `fast-honor-pragmas` doesn't change behavior in most cases, and those affected by numeric accuracy will not need to tune this flag manually, just add the pragma and let it work.


As to the problem that this flag is not recognized in driver until LLVM 18, it can be bypassed using
```
-Xclang -ffp-contract=fast-honor-pragmas
```
compilation verified in Clang CUDA 14.

PR #1360, opened 2024-02-14
Use absolute paths for cmake DEPENDS for bitcode libraries

sbalint98:
Before this change, the bitcode libraries were not re-generated upon a change in the source. Tested with cmake (3.22.1)


illuhad (approved):
Thanks!

PR #1361, opened 2024-02-14
[CBS] Fix potential miscompiles for accelerated CPU due to uniformity analysis

fodinabor:
This fixes two issues that could have led to miscompilations on the CPU due to the uniformity analysis failing or being misinterpreted.


illuhad (approved):
Thanks!

Issue #1345, opened 2024-02-09
Linking issue when building GROMACS


acmnpv:
Hello, I recently ran into a strange linking issue when building GROMACS 2023.3 with AdaptiveCpp 23.10.0

I build ACpp using gcc-12 against the HIP backend (using ROCm 6.0.0), and then followed up building GROMACS using the same GCC.
As expected from the build requirements from ACpp, it pulled in the context and fiber libraries from boost. But when I later tried to run the GROMACS binaries I saw that they also linked against the boost filesystem libraries, which as far as I understand shouldn't be the case as neither GROMACS nor ACpp have those as a requirement.

This actually caused linking errors on one system I tried it on, where it couldn't find and link against boost filesystem automatically, and crashed when forcing the link.

Any ideas about what could be the cause would be welcome. I wasn't able to reproduce the crash on a different system, but the link against boost filesystem was still there.

I already checked with @al42and but he also wasn't sure what the issue could be.

Thanks for your advice!

illuhad:
Interesting, I think @tom91136 previously mentioned some `boost.filesystem` dependency that should not be there. IIRC in the discussion with him the consensus was that `boost.filesystem` is probably transitively pulled in by some dependency of AdaptiveCpp.

Can you check whether boost context/fiber or other libraries that AdaptiveCpp depends on have a dependency on boost.filesystem?

acmnpv:
I just checked the fiber and context libraries, and neither of them have filesystem as a dependency themselves.

I think the problem is that there is an explicit search for boost filesystem here: 

`src/common/CMakeLists.txt:17:find_package(Filesystem REQUIRED Final Experimental Boost)`

Which means that it will (as far as I understand the FindFilesystem CMake script) prepare to link to boost filesystem, even though it is not an actual project dependency. I think it shouldn't search for boost filesystem in that line, because the project anyhow requires a C++17 compliant compiler and std library, and that should come with std::filesystem :)

illuhad:
I had assumed this too, but alas, there are many toolchains which in principle have C++17 support, but don't have the filesystem library yet (or need different linking). So the current cmake integration supports `std::filesystem`, `std::experimental::filesystem` and `boost::filesystem`. `boost.filesystem` should only be required if `std::filesystem`is not available.

See e.g. https://github.com/AdaptiveCpp/AdaptiveCpp/pull/667

acmnpv:
[...]

PR #1365, opened 2024-02-19
Change configuration to only search for boost filesystem if std::filesystem has not been found

acmnpv:
Small fix for configuration to only try to detect and use boost filesystem if a previous search for it in the stdlib hasn't been successful.

After applying this fix the linking issue I have been observing are no longer present, but I'm not sure if it might break something else.

Refs #1345


acmnpv:
I had another look through the FindFilesystem.cmake file and think that a different kind of approach is needed to prevent picking up boost filesystem, because it looks like it shouldn't search for it under the conditions I've been looking at. I'll close this for now and maybe open it again if I see a proper solution for the problem

PR #1368, opened 2024-02-20
Change two cases of directly using std::filesystem

acmnpv:
Two points in the code had still been using std::filesystem directly instead of using the fs:: prefix that makes sure the correct version is in use.


illuhad (approved):
Thanks!

acmnpv:
> Thanks!

thanks for the quick response :)

Still looking into the issue behind #1345, it wasn't related to this by-catch here

PR #1369, opened 2024-02-20
Use device_selector in group_functions test

sbalint98:


illuhad (commented):
Can we please separate the "dropping `cl::`" change? It's very difficult to see otherwise what exactly this PR changes w.r.t device selection.

PR #1372, opened 2024-02-21
[SSCP][Host] Making `generic` more generic: bringing SSCP to the CPU.

fodinabor:
This PR adds a missing piece to make SSCP even more "generic".
Until now, running an AdaptiveCpp binary on CPU & on any supported offloading device could be achieved in two ways:
- Setting `--acpp-targets="omp;generic"` built the kernels for the CPU (`omp`) and supported offloading devices (`generic`).
- Setting `--acpp-targets="generic"` would also support execution on CPUs through OpenCL CPU runtimes.

With this PR `--acpp-targets="generic"` supports execution on CPUs without any added dependencies.

This is achieved by extending the OpenMP backend to also provide SSCP capabilities.
A new `llvm-to-host` tool is added that uses the same transformations as the `omp.accelerated` compilation-flow to achieve decent performance across all kernel types (including `nd_range` `parallel_for`).

This has similar restrictions as the other SSCP targets: hierarchical and scoped parallelism is not yet supported.

This PR enables this per default (`omp;generic` will only generate kernels for `generic`, see https://github.com/AdaptiveCpp/AdaptiveCpp/commit/26c8460057fcf56effbdc8cd082e039fe6e90255). However, that commit should probably not land in `develop`. Instead, we need a bit more testing to be done (please report good/bad experiences here, if you give it a shot!) and should discuss if there is a smoother migration path.
One possibility I see would be to make the OpenMP `get_backend_score` environment variable configurable -> `ACPP_OMP_BACKEND_SCORE=0` would mean "prefer SSCP", `ACPP_OMP_BACKEND_SCORE=2` would mean "prefer old OpenMP flow".
I'd be happy to get further input on the migration path.

This is also in part due to the implementation currently relying on `clang++` to find the right libraries to link with, as the math lib is implemented in terms of the main 

Finally, some benchmark data:
SYCL-Bench, ran on 2x Epyc 7702, i.e. 128 cores.
sycl-bench

SYCL-Bench shows that most benchmarks do perform in a reasonable range with outliers in both directions notably being Polybench patterns, which in sycl-bench use basic `parallel_for` without specifying a group size.
In the `omp` backend, the basic `parallel_for` completely offloaded the decision on how to execute the loop (nest) to the OpenMP runtime. With SSCP the basic `parallel_for` currently has hard-coded work-group sizes that apply to all devices. We need to decide how to deal with this.

Other todos:
- [x] Deal with CPUs not supporting `-march=native` but `-mcpu=native` instead.
- [x] Decide how to handle `parallel_for` for different device types.
- [x] Decide how to test this in CI... we could replace / extend the current SSCP tests using Intel OpenCL CPU runtime..


illuhad:
Awesome!!

> This PR enables this per default (omp;generic will only generate kernels for generic, see https://github.com/AdaptiveCpp/AdaptiveCpp/commit/26c8460057fcf56effbdc8cd082e039fe6e90255). However, that commit should probably not land in develop. Instead, we need a bit more testing to be done (please report good/bad experiences here, if you give it a shot!) and should discuss if there is a smoother migration path.
One possibility I see would be to make the OpenMP get_backend_score environment variable configurable -> ACPP_OMP_BACKEND_SCORE=0 would mean "prefer SSCP", ACPP_OMP_BACKEND_SCORE=2 would mean "prefer old OpenMP flow".
I'd be happy to get further input on the migration path.

I think there is a very simple solution here. The philosophy with the other compilation flows is that if another compilation flow is explicitly provided by the user, prefer this over generic SSCP. The exact same logic can be applied here.

I think the only reason why omp currently behaves differently is because `acpp` always appends at least the sequential host target, which in turn enables the `omp_kernel_launcher`. 
That was needed so that kernels always can be executed at least "somewhere". With SSCP host support, this should no longer be needed.
We should change `acpp` such that the sequential host target is no longer silently enabled if `generic` is enabled too. That would be a simple change here:
https://github.com/AdaptiveCpp/AdaptiveCpp/blob/9981882348691f4b1494f9d978cd05a0e604c897/bin/acpp#L1764

(~Probably we should remove the `sequential` compilation flow stuff entirely in `CMakeLists.txt` and `acpp` as I don't think it would be needed anymore~ Might still be needed if generic is not targeted, e.g. `--acpp-targets=cuda:...` ? Although we could always argue that this targets specification requests a pure interop scenario and thus does not need to "always work" even if there is no device of the requested backend)

That would mean that 
* targets=`omp;generic` runs on host device through `omp_kernel_launcher` (it expresses: Target everything, but for `omp` please use the explicit OpenMP-interop paths)
* targets=`generic` runs on host device through SSCP
* 
This is aligned with the behavior for other compilation flows, e.g.
* targets=`cuda:...;generic` runs on NVIDIA through the cuda kernel launcher (it expresses: Target everything, but for `cuda` please use the explicit CUDA-interop paths)
* targets=`generic` runs on NVIDIA through SSCP.

The default targets should probably be changed from `omp;generic` to just `generic`, at least once we are confident that the new path is stable.
 This will result in reduced compilation times and avoids some potential problems that we sometimes see with the old kernel launcher and some OpenMP compilers (e.g. `omp.h` not found, runtime uses `libgomp` while app is linked against `libomp` and so on).

include/hipSYCL/compiler/cbs/IRUtils.hpp
```diff
@@ -55,6 +55,29 @@ static constexpr const char LocalIdGlobalNameZ[] = "__hipsycl_local_id_z";
...
+static constexpr const char GroupIdGlobalNameX[] = "__hipsycl_group_id_x";
+static constexpr const char GroupIdGlobalNameY[] = "__hipsycl_group_id_y";
+static constexpr const char GroupIdGlobalNameZ[] = "__hipsycl_group_id_z";
+static const std::array<const char *, 3> GroupIdGlobalNames{GroupIdGlobalNameX, GroupIdGlobalNameY,
+                                                            GroupIdGlobalNameZ};
+
+static constexpr const char NumGroupsGlobalNameX[] = "__hipsycl_num_groups_x";
+static constexpr const char NumGroupsGlobalNameY[] = "__hipsycl_num_groups_y";
+static constexpr const char NumGroupsGlobalNameZ[] = "__hipsycl_num_groups_z";
+static const std::array<const char *, 3> NumGroupsGlobalNames{
```
illuhad, line 74: We should probably really consider namespacing these identifiers, e.g. `__hipsycl_cbs_num_groups_x`. I guess this comes from a time before we had many other compiler builtins, but this is no longer the case.

include/hipSYCL/compiler/cbs/IRUtils.hpp
```diff
@@ -55,6 +55,29 @@ static constexpr const char LocalIdGlobalNameZ[] = "__hipsycl_local_id_z";
...
+    NumGroupsGlobalNameX, NumGroupsGlobalNameY, NumGroupsGlobalNameZ};
+
+static constexpr const char SscpDynamicLocalMemoryPtrName[] = "__hipsycl_sscp_dynamic_local_memory";
```
illuhad: Same here. Is it cbs specific?

include/hipSYCL/runtime/omp/omp_code_object.hpp
```diff
@@ -0,0 +1,84 @@
...
+#include "hipSYCL/runtime/device_id.hpp"
+#include "hipSYCL/runtime/kernel_cache.hpp"
+
+struct CUmod_st;
+
+namespace hipsycl {
+namespace rt {
+
+
+class omp_sscp_executable_object : public code_object {
```
illuhad, line 44: It might be worth a thought how much of this is actually "OpenMP" rather than just "host", in case we want to e.g. have our own thread pool at some point. But this is not urgent and also affects existing functionality like `omp_allocator`.

src/compiler/llvm-to-backend/host/HostKernelWrapperPass.cpp
```diff
@@ -0,0 +1,185 @@
...
+  }
+  ReplaceUsesOfGVWith(*Wrapper, SscpDynamicLocalMemoryPtrName, LocalMemPtr);
+
+  return Wrapper;
+}
+
+} // namespace
+
+llvm::PreservedAnalyses HostKernelWrapperPass::run(llvm::Function &F,
+                                                   llvm::FunctionAnalysisManager &AM) {
```
illuhad, line 194: What is this transformation that this pass does and why is it needed? It's not entirely clear to me. Perhaps a good candidate for documentation? :-)

src/compiler/llvm-to-backend/host/HostKernelWrapperPass.cpp
```diff
@@ -0,0 +1,185 @@
...
+  auto VoidPtrT = llvm::PointerType::getUnqual(Bld.getVoidTy());
+  auto UserArgsT = llvm::PointerType::getUnqual(VoidPtrT);
+
+  llvm::SmallVector<llvm::Type *> ArgTypes;
+  ArgTypes.push_back(llvm::PointerType::getUnqual(WorkGroupInfoT));
+  ArgTypes.push_back(UserArgsT);
+
+  auto WrapperT = llvm::FunctionType::get(Bld.getVoidTy(), ArgTypes, false);
+  auto Wrapper = llvm::cast<llvm::Function>(
+      M->getOrInsertFunction(("__sscp_wrapper_" + F.getName()).str(), WrapperT, F.getAttributes())
```
illuhad: If this is creating a new kernel entry point, can we rename the old kernel instead and give the wrapper the name of the kernel? That way, the symbol would rename demangleable, and consistent with the other backends.
Or is this an issue for ODR because on host, the JITted binary is loaded into the same process space as the one that already contains the SSCP kernel stub from the kernel launcher?

src/compiler/llvm-to-backend/host/LLVMToHost.cpp
```diff
@@ -0,0 +1,229 @@
...
+    : LLVMToBackendTranslator{sycl::sscp::backend::host, KN}, KernelNames{KN} {}
+
+bool LLVMToHostTranslator::toBackendFlavor(llvm::Module &M, PassHandler &PH) {
+
+  std::string Triple = llvm::sys::getProcessTriple();
+  // Fixme:
+  // std::string DataLayout =
+  // "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128";
+
+  M.setTargetTriple(Triple);
```
illuhad: Do we actually need to set the triple? It probably will already be set to the host from stage 1 compilation.

src/compiler/llvm-to-backend/host/LLVMToHost.cpp
```diff
@@ -0,0 +1,229 @@
...
+  std::string BuiltinBitcodeFile =
+      common::filesystem::join_path(common::filesystem::get_install_directory(),
+                                    {"lib", "hipSYCL", "bitcode", "libkernel-sscp-host-full.bc"});
+
+  if (!this->linkBitcodeFile(M, BuiltinBitcodeFile))
+    return false;
+
+  llvm::ModulePassManager MPM;
+
+  MPM.addPass(AddressSpaceInferencePass{ASMap});
```
illuhad: Do we really need the address space inference pass? Stage 1 compilation should give us IR where everything is already in AS 0 / follows host conventions.

src/libkernel/sscp/host/atomic.cpp
```diff
@@ -0,0 +1,588 @@
...
+#ifndef __HIP_MEMORY_SCOPE_WORKGROUP
+ #define __HIP_MEMORY_SCOPE_WORKGROUP 3
+#endif
+
+#ifndef __HIP_MEMORY_SCOPE_AGENT
+ #define __HIP_MEMORY_SCOPE_AGENT 4
+#endif
+
+#ifndef __HIP_MEMORY_SCOPE_SYSTEM
+ #define __HIP_MEMORY_SCOPE_SYSTEM 5
```
illuhad: Too much copy & paste? :D

src/libkernel/sscp/host/atomic.cpp
```diff
@@ -0,0 +1,588 @@
...
+    case __hipsycl_sscp_memory_scope::work_item:
+      return __HIP_MEMORY_SCOPE_SINGLETHREAD;
+    case __hipsycl_sscp_memory_scope::sub_group:
+      return __HIP_MEMORY_SCOPE_WAVEFRONT;
+    case __hipsycl_sscp_memory_scope::work_group:
+      return __HIP_MEMORY_SCOPE_WORKGROUP;
+    case __hipsycl_sscp_memory_scope::device:
+      return __HIP_MEMORY_SCOPE_AGENT;
+    case __hipsycl_sscp_memory_scope::system:
+      return __HIP_MEMORY_SCOPE_SYSTEM;
```
illuhad: Same here?

src/libkernel/sscp/host/barrier.cpp
```diff
@@ -0,0 +1,56 @@
...
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "hipSYCL/sycl/libkernel/sscp/builtins/barrier.hpp"
+
+extern "C" [[clang::convergent]] void __hipsycl_barrier();
```
illuhad: As I said, we should consider changing this to `__hipsycl_cbs_barrier()` or similar

src/libkernel/sscp/host/print.cpp
```diff
@@ -0,0 +1,35 @@
...
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+
+#include "hipSYCL/sycl/libkernel/sscp/builtins/print.hpp"
+
+#include <iostream>
```
illuhad: Better `stdio.h`. No need for fancy C++ in the builtins :P

src/libkernel/sscp/host/relational.cpp
```diff
@@ -0,0 +1,49 @@
...
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "hipSYCL/sycl/libkernel/sscp/builtins/relational.hpp"
+#include "hipSYCL/sycl/libkernel/host/builtins.hpp"
```
illuhad: I think this back-dependency on SYCL headers is not necessary and only adds complexity. Can we just call libc functionality?

src/runtime/kernel_cache.cpp
```diff
@@ -418,7 +418,7 @@ const code_object* kernel_cache::get_code_object_impl(code_object_id id) const {
   return it->second.get();
 }
 
-std::string kernel_cache::get_persistent_cache_file(code_object_id id_of_binary) const {
+std::string kernel_cache::get_persistent_cache_file(code_object_id id_of_binary) {
```
illuhad, line 421: Why did the const have to go when the implementation of this function was not changed?

src/runtime/omp/omp_code_object.cpp
```diff
@@ -0,0 +1,157 @@
...
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <algorithm>
+#include <cassert>
+#include <cctype>
+#include <sstream>
+#include <string>
+
+#include <unistd.h>
```
illuhad: Do we actually need this dependency?

src/runtime/omp/omp_code_object.cpp
```diff
@@ -0,0 +1,157 @@
...
+      _kernel_cache_path(kernel_cache::get_persistent_cache_file(_id) + "-" +
+                         std::to_string(getpid()) + ".so") {
+  _build_result = build(binary);
+}
+
+omp_sscp_executable_object::~omp_sscp_executable_object() {
+  if (_module)
+    detail::close_library(_module, "omp_sscp_executable");
+  try {
+    if (!fs::remove(_kernel_cache_path)) {
```
illuhad: Would it make sense to move `remove` support into `common::filesystem`, and thus isolate the filesystem dependencies there?

src/runtime/omp/omp_code_object.cpp
```diff
@@ -0,0 +1,157 @@
...
+result omp_sscp_executable_object::get_build_result() const {
+  return _build_result;
+}
+
+code_object_state omp_sscp_executable_object::state() const {
+  return _module ? code_object_state::executable : code_object_state::invalid;
+}
+
+code_format omp_sscp_executable_object::format() const {
+  return code_format::ptx;
```
illuhad: What? :D

src/runtime/omp/omp_code_object.cpp
```diff
@@ -0,0 +1,157 @@
...
+}
+
+backend_id omp_sscp_executable_object::managing_backend() const {
+  return backend_id::omp;
+}
+
+hcf_object_id omp_sscp_executable_object::hcf_source() const { return _hcf; }
+
+std::string omp_sscp_executable_object::target_arch() const {
+  return "cpu"; // fixme..?
```
illuhad:
```suggestion
  return "native-host";
```
Perhaps something like this?

src/runtime/omp/omp_queue.cpp
```diff
@@ -7,38 +7,52 @@
...
 #include "hipSYCL/runtime/queue_completion_event.hpp"
 #include "hipSYCL/runtime/signal_channel.hpp"
+#include "hipSYCL/runtime/util.hpp"
+
+#ifdef HIPSYCL_WITH_SSCP_COMPILER
+#include "hipSYCL/compiler/llvm-to-backend/host/LLVMToHostFactory.hpp"
+#include "hipSYCL/glue/generic/host/iterate_range.hpp"
+#include "hipSYCL/glue/llvm-sscp/jit.hpp"
+#include "hipSYCL/runtime/dylib_loader.hpp"
+#include "hipSYCL/sycl/libkernel/detail/local_memory_allocator.hpp"
```
illuhad: Similar here... `local_memory_allocator` potentially contains even worse bits :(

src/runtime/omp/omp_queue.cpp
```diff
@@ -7,38 +7,52 @@
...
+#include "hipSYCL/runtime/omp/omp_code_object.hpp"
+#include "hipSYCL/runtime/omp/omp_event.hpp"
 #include "hipSYCL/runtime/operations.hpp"
 #include "hipSYCL/runtime/queue_completion_event.hpp"
 #include "hipSYCL/runtime/signal_channel.hpp"
+#include "hipSYCL/runtime/util.hpp"
+
+#ifdef HIPSYCL_WITH_SSCP_COMPILER
+#include "hipSYCL/compiler/llvm-to-backend/host/LLVMToHostFactory.hpp"
+#include "hipSYCL/glue/generic/host/iterate_range.hpp"
```
illuhad: This currently pulls in `libkernel` code into the runtime which I'm not sure is a good idea (and also contradicts some of our documentation which states that this should not happen).

To share this code, we could template it so that it works with both `sycl::range` and `rt::range`?

src/runtime/omp/omp_queue.cpp
```diff
@@ -169,25 +185,67 @@ class omp_instrumentation_setup {
...
 
+using omp_sscp_kernel = void(const work_group_info *, void **);
 
-omp_queue::omp_queue(backend_id id)
-: _backend_id(id) {}
+result launch_kernel_from_so(void *handle, const std::string &kernel_name,
+                             const rt::range<3> &num_groups,
+                             const rt::range<3> &local_size,
+                             unsigned shared_memory, void **kernel_args) {
+  auto kernel = (omp_sscp_kernel *)detail::get_symbol_from_library(
```
illuhad: Could address finding not be done by `omp_sscp_code_object` constructor once for each kernel, instead of doing it for every kernel launch?

src/runtime/omp/omp_queue.cpp
```diff
@@ -169,25 +185,67 @@ class omp_instrumentation_setup {
...
-omp_queue::~omp_queue() {
-  _worker.halt();
+#ifdef _OPENMP
+#pragma omp parallel for collapse(3)
+#endif
+  for (std::size_t k = 0; k < num_groups.get(2); ++k) {
+    for (std::size_t j = 0; j < num_groups.get(1); ++j) {
+      for (std::size_t i = 0; i < num_groups.get(0); ++i) {
+        sycl::detail::host_local_memory::request_from_threadprivate_pool(
+            shared_memory);
```
illuhad: Don't see why we need this? The local memory allocator stuff solves for you alignment and memory management for multiple local accessors in one block of local memory.
At this point, all we care about is one block of local memory. So why can't we just ditch this dependency on libkernel and just have `static thread_local std::vector`? 

Perhaps it might be even better to just have a JIT-time statically sized global array like `LLVMToSpirv` that is handled on the compiler side. Then you don't have the complexity of having to pass in another parameter, and the compiler knows more about the allocation and can perhaps do better optimizations.

src/runtime/omp/omp_queue.cpp
```diff
@@ -169,25 +185,67 @@ class omp_instrumentation_setup {
...
+        work_group_info info{
+            num_groups, rt::id<3>{i, j, k}, local_size,
+            hipsycl::sycl::detail::host_local_memory::get_ptr()};
+        kernel(&info, kernel_args);
```
illuhad: Not something we need to solve in this PR, but something to keep in mind: Passing in `void**` might limit llvm-to-backend's capability to perform optimizations like dead-argument-elimination. I think hardwiring arguments to constants might still work if that optimization is run prior to backend flavoring in `llvm-to-host`...

illuhad:
> SYCL-Bench shows that most benchmarks do perform in a reasonable range with outliers in both directions notably being Polybench patterns, which in sycl-bench use basic parallel_for without specifying a group size.

I'm not sure I understand. I see polybench benchmarks at both extremes of the spectrum (syrk on the left and covariance on the right), and in both cases they use the normal range-based parallel for. Are you saying that range-based parallel for is the culprit here?

src/compiler/llvm-to-backend/host/HostKernelWrapperPass.cpp
fodinabor: added docs. it creates a new function with a generic signature that we can easily call from the host code.
The old kernel is inlined immediately and can then be thrown away.

src/compiler/llvm-to-backend/host/HostKernelWrapperPass.cpp
fodinabor: nah, just needed a new name, was changed back to the original kernel name, which should be demangleable anyways.
I changed it to rename the old function first and then create the wrapper with the actual kernel name directly.

src/runtime/kernel_cache.cpp
fodinabor: It's static now..

fodinabor:
> I'm not sure I understand. I see polybench benchmarks at both extremes of the spectrum (syrk on the left and covariance on the right), and in both cases they use the normal range-based parallel for. Are you saying that range-based parallel for is the culprit here?

Yes, that is what I was trying to say. I believe that depending on the work-load the work-group pattern used by basic `parallel_for` in SSCP vs. just plain OpenMP might be better or worse, so it might be difficult to find "optimal" defaults here?

src/runtime/omp/omp_queue.cpp
fodinabor: Yeah, I was considering to do that. 
Currently, it isn't even used, as the iteration order is swapped anyways.. and making that generic seems a bit unnecessary considering how simple the actual implementation is... I just forgot to cleanup the include.

One could consider to run-time specialize the for loop dimension again, in case `range[1] == 1 && range[2] == 1` or `range[2] == 1`, maybe that again would be sth, we can move to the (then generic) header... 🤷🏼

src/runtime/omp/omp_queue.cpp
fodinabor: I'd argue, we'd want local mem size to be only hard-wired on the same adaptivity level as the work-group size, since when the wg size changes, it's very likely the local mem size does as well...

Moved to a dynamically-sized alloca and added a build option to make the size statically known, but not activating the static version for now...

src/runtime/omp/omp_queue.cpp
fodinabor: Hmm... I would assume that DAE should happen rather early on, before any backend flavoring..?

In the end, we need to find a way to pass the arguments to the kernel, and since we don't have compile-time information about this, we need some runtime "variadic" way to do this. the `void**` thing seems pretty simple.. but haven't put much thought into alternatives, yet tbh. :)

illuhad:
> Yes, that is what I was trying to say. I believe that depending on the work-load the work-group pattern used by basic parallel_for in SSCP vs. just plain OpenMP might be better or worse, so it might be difficult to find "optimal" defaults here?

I suspect there is more at play than just work group size. For example, the `iterate_range_omp` that is now shared between the `omp_queue` for SSCP and the old `omp_kernel_launcher` might also need to have its loop iteration order inverted for SSCP. My guess is that currently the work groups are striding through memory which is causing NUMA and other problems.

EDIT: Nevermind, as you say the old functionality is not even used and the one in `omp_queue` already does the inversion. It might still be worth validating the iteration order of work groups if you haven't done so yet. All this flipping can be pretty confusing :D

src/runtime/omp/omp_queue.cpp
illuhad: Nah, if we don't actually need it we can just drop the include.

src/runtime/omp/omp_queue.cpp
illuhad: Ideally DAE happens after the optimization pipeline (without optimizations, we might not easily know whether an argument is actually used), which has to be run after backend flavoring. If we ran flavoring after optimizations, we wouldn't optimize the bitcode libraries that we link.
But I can see how we need some kind of type erasure in order to be able to define a function pointer in the runtime.

illuhad: Perhaps one solution might be to introduce another step between `prepareIR()` which does flavoring and optimization, and `translateToBackendFormat`. We could then move the the generation of the wrapper kernel interface to that new stage. DAE can then be run after optimizations at the end of `prepareIR`, but before wrapping.

src/runtime/omp/omp_queue.cpp
illuhad: > I'd argue, we'd want local mem size to be only hard-wired on the same adaptivity level as the work-group size, since when the wg size changes, it's very likely the local mem size does as well...

That's true. llvm-to-spirv already behaves this way though, so it wouldn't be a new problem. I agree that a dynamically sized `alloca` is a better solution. (I don't think this can work for llvm-to-spirv unfortunately - I think llvm-spirv translator would explode if we tried an alloca in AS 3 :D).

One thing to keep in mind though is that `alloca` lives in stack memory, so I suppose users can break their kernels if they request large enough amounts of local memory to cause a stack overflow? And it might be difficult to guarantee that the local size is small enough, since it depends on the amount of stack used by the rest of the program. An `std::vector` would solve such issues.

fodinabor: > One thing to keep in mind though is that alloca lives in stack memory, so I suppose users can break their kernels if they request large enough amounts of local memory to cause a stack overflow? 

Yes. This is a theoretical issue, but since local memory is generally limited as coming from other backends, such as CUDA/HIP, I don't see that any sane use cases stumbling over this...

A vector again makes it hard to support the "known array size" case..

illuhad: > Yes. This is a theoretical issue, but since local memory is generally limited as coming from other backends, such as CUDA/HIP, I don't see that any sane use cases stumbling over this...

True. The difference is though that these backends can return an error that says "insufficient local memory" or similar. So we should probably enforce some reasonable limits so that we can print an error instead of just crashing (even if it's not bullet proof, as it also depends on the code as mentioned)

> A vector again makes it hard to support the "known array size" case..

True, fair point :)

illuhad:
Would you think it perhaps makes sense to silence `-Wpass-failed=transform-warning`, at least by default, when JITting? The warning is expected for many kinds of code, and thus not necessarily something that users need to be alarmed about.

illuhad:
Compile time for babelstream

| Target | Compile time (develop) | Compile time (sscp-cpu-backend) |
|----|----|----|
| omp | 11.845 | 11.831 |
| omp;generic | 13.718 | 13.892 |
| generic | 13.612 | 12.053 |


So basically targeting `generic` is now roughly as fast as just targeting `omp`, while supporting more hardware.
And once the default target is changed from `omp;generic` to `generic` we can expect a compile time reduction of ~15% for the default configuration (at least for Babelstream, other codes may have different characteristics). Nice.

include/hipSYCL/glue/llvm-sscp/s2_ir_constants.hpp
```diff
@@ -57,6 +57,7 @@ namespace backend {
 inline constexpr int spirv = 0;
 inline constexpr int ptx = 1;
 inline constexpr int amdgpu = 2;
+inline constexpr int host = 2;
```
illuhad:
```suggestion
inline constexpr int host = 3;
```

include/hipSYCL/runtime/omp/omp_code_object.hpp
```diff
@@ -0,0 +1,100 @@
...
+#include <string>
+#include <vector>
+
+#include "hipSYCL/glue/kernel_configuration.hpp"
+#include "hipSYCL/runtime/device_id.hpp"
+#include "hipSYCL/runtime/error.hpp"
+#include "hipSYCL/runtime/kernel_cache.hpp"
+#include "hipSYCL/runtime/util.hpp"
+
+struct CUmod_st;
```
illuhad:
```suggestion
```

src/compiler/llvm-to-backend/CMakeLists.txt
```diff
@@ -190,4 +190,33 @@ if(WITH_SSCP_COMPILER)
...
+    endif()
+
+    add_hipsycl_llvm_backend(
+      BACKEND host
+      LIBRARY host/LLVMToHost.cpp host/HostKernelWrapperPass.cpp
+      TOOL host/LLVMToHostTool.cpp)
+
+    target_compile_definitions(llvm-to-host PRIVATE
+      -DHIPSYCL_CLANG_PATH="${CLANG_EXECUTABLE_PATH}" 
+      -DHIPSYCL_CUDA_PATH="${CUDA_TOOLKIT_ROOT_DIR}"
```
illuhad:
```suggestion
```
?

src/runtime/omp/omp_queue.cpp
```diff
@@ -169,25 +189,65 @@ class omp_instrumentation_setup {
...
 }
 
+result
+launch_kernel_from_so(omp_sscp_executable_object::omp_sscp_kernel *kernel,
+                      const rt::range<3> &num_groups,
+                      const rt::range<3> &local_size, unsigned shared_memory,
+                      void **kernel_args) {
+#ifdef _OPENMP
+#pragma omp parallel
+#endif
```
illuhad, line 225: Do you think it makes sense to emit a warning if this code path is invoked, but `_OPENMP` was not defined? I'm currently also considering adding a similar warning to the old `omp_kernel_launcher`.

include/hipSYCL/runtime/code_object_invoker.hpp
```diff
@@ -61,6 +61,20 @@ class sscp_code_object_invoker {
                                std::size_t *arg_sizes, std::size_t num_args,
                                const std::string &kernel_name,
                                const glue::kernel_configuration& config) = 0;
+
+  virtual rt::range<3> select_group_size(const rt::range<3> &num_groups,
```
illuhad: hm.. I'm not sure this is the best place to have this logic. The code object invoker knows basically nothing about the kernel, so if we want to have some more advanced calculation later on (e.g. based on register usage of the JIT compiled kernel or so), we cannot do that here.

My guess is that the adaptivity engine should probably handle that, especially if we want it to do group size auto-tuning later on. Although it's quite possible we need a different adaptivity engine (as the current one is focused on compiling the kernel once parameters are already known).

hm.. so perhaps this requires more thought, in which case we could do it the way you have done it as a temporary solution for now. What do you think?

src/runtime/omp/omp_queue.cpp
```diff
@@ -420,9 +595,35 @@ device_id omp_queue::get_device() const {
...
+
+rt::range<3> omp_sscp_code_object_invoker::select_group_size(
+    const rt::range<3> &num_groups, const rt::range<3> &group_size) const {
+  rt::range<3> selected_group_size = group_size;
+#ifdef _OPENMP
+  const int max_threads = omp_get_max_threads();
+#else
+  const int max_threads = 1;
+#endif
+  constexpr auto divisor = 1;
```
illuhad, line 631: This is in case we want to have some fraction <1 in the future, I assume?

include/hipSYCL/runtime/code_object_invoker.hpp
illuhad: It seems the way that this function is used, actually the first parameter is the global size, not the number of groups. This confused me for a while when I tried to understand the min/max logic that the OpenMP backend implements this function with. Perhaps rename the parameter?

src/runtime/omp/omp_queue.cpp
fodinabor: jip... the `k` from our discussion.

include/hipSYCL/runtime/code_object_invoker.hpp
fodinabor: jup.. it also doesn't know much about the device it targets...
so I'm not super bound to this, but it was a relatively non-intrusive location, which for the current use case also makes sense-ish.

MarkusBuettner:
I just tested this with our code, and performance is within 10% of the ahead-of-time compilation flow. I'm just wondering, is `-ffast-math` propagated to the JIT compiler in this branch? The first time I run the code I get warnings
```[AdaptiveCpp Warning] [KernelFlattening] failed to inline function <sqrtf>: 'external or indirect'```
which are not there when I compile with `--acpp-targets="omp" -ffast-math`.

**Comment by illuhad** [comment]
> I just tested this with our code, and performance is within 10% of the ahead-of-time compilation flow.

Actually we find that in many cases it is faster :-)

>  I'm just wondering, is -ffast-math propagated to the JIT compiler in this branch?

It looks like it already has #1353 merged in. However, I'm not sure whether the generated code is already 100% on par in the JIT flow for host. For host specifically, we might additionally have to rewrite function calls to e.g. `::sqrtf` to approximate builtins if fast-math is used. This is not needed for the existing GPU backends.

**Comment by MarkusBuettner** [comment]
It actually is faster also for us, but only marginally. And there is one case where it is a bit slower (about 2%), but that might be caused by the approximate builtins.

**Comment by illuhad** [comment]
I think what we need to do for the approximate builtins is to also build the SSCP host bitcode libraries using -ffast-math. Then at JIT time, if `-ffast-math` was set, link the fast-math variant of the bitcode libraries.
@fodinabor What do you think?

File: include/hipSYCL/runtime/code_object_invoker.hpp (hunk 27 above)
Comment (illuhad) [reply include/hipSYCL/runtime/code_object_invoker.hpp]: Hm.. I suppose we might as well have done this in the `sscp_kernel_launcher` where it was done previously. It's given the `dag_node`, which allows access to the assigned device, which we could check for the backend. But I'm not sure if that would have been any easier. We'll have to change it anyway, so I don't mind the way it is now.

File: src/compiler/llvm-to-backend/LLVMToBackend.cpp [hunk 29]
```diff
@@ -127,6 +127,8 @@ bool LLVMToBackendTranslator::setBuildOption(const std::string &Option, const st
   } else if (Option == "known-group-size-z") {
     KnownGroupSizeZ = std::stoi(Value);
     return true;
+  } else if (Option == "known-local-mem-size") {
+    KnownLocalMemSize = std::stoi(Value);
```
illuhad, line 131: If we're adding this option globally, we should probably also use it in `LLVMToSpirvTranslator` which currently already has its own flag exactly for this (which is already used by `ocl_queue` and `ze_queue`)

fodinabor: Won't, as this is an "optimization" flag, vs Spir-v actually needs it for correctness.

fodinabor:
@MarkusBuettner: Thanks for testing and reporting!
I think @illuhad already took care of the most relevant change, I just wanted to comment on the following:
> [AdaptiveCpp Warning] [KernelFlattening] failed to inline function <sqrtf>: 'external or indirect'

You likely should've seen this warning while building with the "old" flow, as this warns that if this external function contained a barrier, we wouldn't be able to see it and therefore handle it correctly. However, this might be a bit too verbose, as it also warns about well known, barrier-free functions as math functions.. not sure if the warning is overall superfluous or if there's a (not over the board) way to filter this... (e.g. a list of whitelist function names, but that might drive compilation times up unnecessarily..?)

illuhad:
> You likely should've seen this warning while building with the "old" flow, as this warns that if this external function contained a barrier, we wouldn't be able to see it and therefore handle it correctly. However, this might be a bit too verbose, as it also warns about well known, barrier-free functions as math functions.. not sure if the warning is overall superfluous or if there's a (not over the board) way to filter this... (e.g. a list of whitelist function names, but that might drive compilation times up unnecessarily..?)

In case it also happens with SSCP, we could probably address this by ignoring unresolved calls in the call graph of `__hipsycl_sscp_*`. I don't know though whether it's worth the effort compared to just disabling the warning. The SSCP behavior for the other backends in case of unresolved non-builtin functions is JIT error, not "nice warning" ;)

illuhad:
I noticed that `-DWITH_ACCELERATED_CPU=off` no longer works, since then cmake complains that `acpp-clang-cbs` no longer has any sources associated with it.
I think the fundamental problem is that we may have to decide what the interplay between accelerated CPU support and host support in SSCP is going to be.

Are they two independent features? Are they the same, and if so, how do they behave w.r.t `WITH_ACCELERATED_CPU` and `WITH_SSCP_COMPILER`?

Are there use cases when one might want to have one, but not the other?


EDIT: Having thought a little about that, I'm inclined to believe that it might be a good strategy to combine `WITH_STDPAR_COMPILER`, `WITH_ACCELERATED_CPU` and `WITH_SSCP_COMPILER` to a single flag. That would make it easier to reason about whether we can assume these features to be present, and it would also make the installation easier to follow for users, since there's fewer configurations we need to describe. It would imply that the minimum requirement for `omp.accelerated` would have to be bumped from LLVM 11 to LLVM 14 to align with SSCP. But I don't think many users still use LLVM <= 13.
I don't know however how large the scope of that change would be (as it might also require things like backwards compatibility to be handled), so I'm unsure whether it's too large for this PR.

fodinabor:
> In case it also happens with SSCP, we could probably address this by ignoring unresolved calls in the call graph of __hipsycl_sscp_*. I don't know though whether it's worth the effort compared to just disabling the warning. The SSCP behavior for the other backends in case of unresolved non-builtin functions is JIT error, not "nice warning" ;)

Yeah, it does happen with SSCP.
I changed it, that it generally only emits an info in the `KernelFlattening` and if inside an SSCP intrinsic, no warning in the `LoopSplitterInlining`.

> Are they two independent features? Are they the same, and if so, how do they behave w.r.t WITH_ACCELERATED_CPU and WITH_SSCP_COMPILER?

My intuition for now would be, and I changed the build config to support this:
`WITH_SSCP_COMPILER=ON`: build for all backends what is needed for SSCP.
`WITH_ACCELERATED_CPU`: if enabled, use accelerated CPU passes for the traditional compilation flow. if disabled, but SSCP enabled, don't enable the passes for the traditional compilation flow, but enable whatever is needed for SSCP

illuhad (approved):
Awesome, a massive thank you! :D

PR #1374, opened 2024-02-22
Only check for OpenCL if user has not explicitly set WITH_SSCP_COMPILER to False

nilsfriess:
If `-DWITH_SSCP_COMPILER=False` is passed during CMake config but OpenCL is found, then `WITH_SSCP_COMPILER` is later set to true, which is unexpected.


illuhad (approved):
Thanks!

PR #1377, opened 2024-02-26
Bump version to 24.02.0

illuhad:


illuhad:
Let's merge this so that at least this PR is still merged in February ;) We can then schedule the release for next week.

PR #1380, opened 2024-02-27
Add option to specify which filesystem implementation to search for

acmnpv:
This helps working around systems with broken std::filesystem where you just want to specify to use boost instead.

Addresses #1345


CMakeLists.txt
```diff
@@ -479,6 +479,8 @@ if(NOT DEFAULT_TARGETS)
   endif()
 endif()
 
+set(ACPP_FILESYSTEM_SEARCH_OPTIONS "Final Experimental Boost" CACHE STRING "Which filesystem libraries should be searched for")
```
illuhad: The `src` subdirectory, where this variable is used, is included already above this line. Could there be problems for the first cmake invocation, when the cache is empty, since this line has not yet been executed at the point when it the variable is used?

illuhad (commented):
Thanks! `find_package(Filesystem)` is also used in `src/runtime/CMakeLists.txt`.  Do we need a change there as well to use the new variable?

CMakeLists.txt
acmnpv: completely missed the include order there and the other use in `src/runtime/CMakeLists.txt`, fixing now

acmnpv: another question, would it make sense to search for the filesystem library in one single place instead of doing it separately in both common and runtime?

illuhad: > another question, would it make sense to search for the filesystem library in one single place instead of doing it separately in both common and runtime?

In theory yes, in practice it's not really a priority because the filesystem usage in `acpp-rt` is more of a temporary thing anyway. I think ultimately `acpp-rt` should just use functionality from `acpp-common`, so that filesystem usage is limited to this library. Then we'd only have a single `find_package` for `acpp-common` anyway.

If you want to address this feel free, but I would consider it low priority since the solution of the fundamental problem would make this superfluous.

CMakeLists.txt
```diff
@@ -339,6 +339,7 @@ endif()
 
 
 set(ROCM_LIBS "-L${ROCM_PATH}/lib -L${ROCM_PATH}/hip/lib -lamdhip64" CACHE STRING "Necessary libraries for ROCm")
+set(ACPP_FILESYSTEM_SEARCH_OPTIONS "Final Experimental Boost" CACHE STRING "Which filesystem libraries should be searched for")
```
nilsfriess: Defining this as a single string seems to make the FindFilesystem module think that we're looking for a filesystem library called "Final Experimental Boost". Putting semicolons between the names makes CMake interpret it as a list. 

```suggestion
set(ACPP_FILESYSTEM_SEARCH_OPTIONS "Final;Experimental;Boost" CACHE STRING "Which filesystem libraries should be searched for")
```

acmnpv: thanks for checking! changing now

CMakeLists.txt
acmnpv: I might get around to that to change the code in runtime to use the methods from common, but need to work on actual job first :)

illuhad (approved):
Thanks!

CMakeLists.txt
illuhad: No worries, it's not something that is urgent, and also not something that needs to be done in the context of this PR :)

PR #1382, opened 2024-02-27
[SSCP][llvm-to-host] If -ffast-math was set, link with bitcode libraries built with -ffast-math

illuhad:
@fodinabor @MarkusBuettner This should cause SSCP host JIT to emit approximate math builtins if `-ffast-math` was used during compilation.


fodinabor (approved):
LGTM :)

Issue #1381, opened 2024-02-27
Compiler warning with GCC: -Wfree-nonheap-object on sbo::small_vector


al42and:
**Bug summary**

The current tip of the `develop` branch, as well as v23.10.0, raise a compiler warning about `delete` misuse.

**To Reproduce**

```console
$ git checkout v23.10.0
$ cmake .. -DCMAKE_C_COMPILER=gcc-12 -DCMAKE_CXX_COMPILER=g++-12 -DLLVM_DIR=/usr/lib/llvm-17/lib/cmake/ -DWITH_SSCP_COMPILER=OFF -DWITH_OPENCL_BACKEND=OFF -DWITH_LEVEL_ZERO_BACKEND=OFF -DWITH_CUDA_BACKEND=OFF -DWITH_ROCM_BACKEND=OFF
$ make
....
In file included from /usr/include/x86_64-linux-gnu/c++/12/bits/c++allocator.h:33,
                 from /usr/include/c++/12/bits/allocator.h:46,
                 from /usr/include/c++/12/string:41,
                 from /usr/include/c++/12/stdexcept:39,
                 from /usr/include/c++/12/system_error:41,
                 from /usr/include/c++/12/mutex:41,
                 from /home/aland/AdaptiveCpp/include/hipSYCL/runtime/data.hpp:32,
                 from /home/aland/AdaptiveCpp/src/runtime/dag_builder.cpp:29:
In member function ‘void std::__new_allocator<_Tp>::deallocate(_Tp*, size_type) [with _Tp = std::shared_ptr<hipsycl::rt::dag_node>]’,
    inlined from ‘constexpr void sbo::small_buffer_vector_allocator<T, MaxSize, NonReboundT>::deallocate(void*, size_t) [with T = std::shared_ptr<hipsycl::rt::dag_node>; long unsigned int MaxSize = 8; NonReboundT = std::shared_ptr<hipsycl::rt::dag_node>]’ at /home/aland/AdaptiveCpp/include/hipSYCL/common/sbo/small_vector.hpp:53:33,
    inlined from ‘static void std::allocator_traits< <template-parameter-1-1> >::deallocate(_Alloc&, pointer, size_type) [with _Alloc = sbo::small_buffer_vector_allocator<std::shared_ptr<hipsycl::rt::dag_node>, 8, std::shared_ptr<hipsycl::rt::dag_node> >]’ at /usr/include/c++/12/bits/alloc_traits.h:345:23,
[...]

PR #1383, opened 2024-02-27
dag_builder: Use std::exchange to avoid sbo free() warning

illuhad:
@al42and For me this avoids the warning from #1381. My guess is that something in the auto-generated move assignment triggers  the warning. Perhaps we don't see it in other places because this is the only place where auto-generated move assignment was used with `small_vector` or similar.

### DIFF
--- a/src/runtime/dag_builder.cpp
+++ b/src/runtime/dag_builder.cpp
@@ -34,6 +34,7 @@
 #include "hipSYCL/sycl/access.hpp"
 
 #include <mutex>
+#include <utility>
 
 // TODO: Implement the following optimization:
 // - Reorder requirements such that larger accesses come first. This will cause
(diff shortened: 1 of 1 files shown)


al42and:
I don't think this is a good solution. Turning both assignments in `dag_builder.cpp` into copies instead of moves indeed fixes the warning, but at the potential performance cost (although, I suspect, a very minor one).

What do you think about using `dag final_dag = std::exchange(_current_dag, {});` in `dag_builder::finish_and_reset` instead? For me, it fixes the warning, and still, as far as I understand, should use moves instead of copies.

illuhad:
I think performance-wise, it should not make much of a difference, since the assumption is that the majority of cases will be covered by the small vector optimization. In this case, it will probably boil down to a `memcpy` or `memset` anyway.

But I agree that your suggestion is better, because it is also more idiomatic than what we currently have. I have changed the PR to use std::exchange instead. As you say, it silences the warning too :)

al42and (approved):

PR #1390, opened 2024-03-04
Silence LLVM warnings about opaque pointers

al42and:
Note to reviewers: The tests pass on my machine when built with Clang 17.0.6, and I double-checked the logic, but I haven't verified it with other LLVM versions.


fodinabor (approved):
LGTM, thanks!

illuhad (approved):
Thanks!

PR #1391, opened 2024-03-04
[stdpar] specialize memory streaming for host sscp

illuhad:
Previously, host stdpar memory streaming (e.g. used for `transform_reduce`) has assumed that whenever `__hipsycl_if_target_device()` is active, the GPU-optimized code path should be taken.
This assumption is no longer correct with the host SSCP backend. Here, it can happen that we run on the host even though we are in a device code path.

This RP adds a host/device specialization within SSCP JIT, which improves performance on Milan for babelstream dot (std-data model) from 50GB/s to 380GBs.
This means that SSCP on host now substantially outperforms the old compiler in this case, which only achieves ~280GB/s.

I'd like to work on the code specialization API for SSCP a bit more, so it's quite possible that this is not the final API that we can advertise to users. But it's something that works today which we can use to fix the mentioned performance regression.

### DIFF
--- a/include/hipSYCL/algorithms/util/memory_streaming.hpp
+++ b/include/hipSYCL/algorithms/util/memory_streaming.hpp
@@ -77,11 +77,16 @@ public:
     
     // TODO: This actually should be something like if_device_cpu, which
     //  we cannot express yet.
-    __hipsycl_if_target_device(
-      const std::size_t gid = idx.get_global_id(0);
-      for (std::size_t i = gid; i < problem_size; i += idx.get_global_range(0)) {
-        f(sycl::id<1>{i});
+    __hipsycl_if_target_sscp(
+      if(sycl::jit::introspect<sycl::jit::current_backend, int>() == sycl::jit::backend::host) {
+        run_host(problem_size, idx, f);
+      } else {
+        run_device(problem_size, idx, f);
       }
+      return;
+    );
+    __hipsycl_if_target_device(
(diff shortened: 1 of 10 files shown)


include/hipSYCL/algorithms/util/memory_streaming.hpp
```diff
@@ -77,11 +77,16 @@ class data_streamer {
     
     // TODO: This actually should be something like if_device_cpu, which
```
fodinabor: Is this comment obsolete now?

fodinabor (commented):
Just a cleanup nit, otherwise go brrr!

include/hipSYCL/algorithms/util/memory_streaming.hpp
illuhad: Yep, I've just removed it, thank you :)

fodinabor (approved):
LGTM now :)

illuhad:
Looks like there is some issue with the S2 IR constant mechanism, so that this PR does not always work reliably. Looking into it now.

illuhad:
Ok, it works now. Looks like LLVM GlobalOpt pass has optimized away S2 IR constants as globals without initializers. I've now added to this PR the functionality to make S2 IR constants have external linkage which fixes the issue for now.

I think in the long term I'd like to reimplement IR constants on top of functions instead of global variables. That seems more robust to me.

PR #1392, opened 2024-03-06
[NFC] Performance guide: Mention that AdaptiveCpp can target CPUs through OpenCL too

illuhad:


MarkusBuettner:
Maybe also link the Intel OpenCL documentation? (especially the vectorizer knobs: https://www.intel.com/content/www/us/en/docs/opencl-sdk/developer-guide-core-xeon/2018/vectorizer-knobs.html)

When I run our code with AdaptiveCpp and Intel OpenCL on CPUs, I have to explicitly set `CL_CONFIG_CPU_VECTORIZER_MODE=8` to achieve similar performance to oneAPI.

illuhad:
Interesting. Does DPC++ set those vectorization hints by itself? Usually the generated SPIR-V should be pretty similar between DPC++ and AdaptiveCpp.

MarkusBuettner:
> Does DPC++ set those vectorization hints by itself?

Maybe, but it looks like it does. I'm not sure what exactly icpx is doing with the code, and I haven't tested it with the open source DPC++ yet.

PR #1394, opened 2024-03-08
[NFC] Document DPC++ not correctly rounding sqrt on AMD by default

illuhad:


pranav-sivaraman:
Does this also apply to the `clang++` from https://github.com/intel/llvm?

illuhad:
@pranav-sivaraman I have not tried, but most likely yes. This is probably a bug, not some misguided optimization. You can check whether your compiler is affected as described here: https://github.com/intel/llvm/issues/12961

pranav-sivaraman:
@illuhad I have verified this this also affects the `clang++` on `nightly-2024-01-20` from https://github.com/intel/llvm. I am going to investigate if this is simple to patch.

PR #1410, opened 2024-03-17
[SSCP] Report JIT compiler errors from CUDA

feltech:
Use `CU_JIT_ERROR_LOG_BUFFER` when JIT compiling CUDA code, and report the error log in case of compilation failure.

For example, the following kernel I might (did) naively expect to work
```c++
using Allocator = sycl::usm_allocator<char, sycl::usm::alloc::shared>;
std::vector<char, Allocator> vals(Allocator{q});
static constexpr std::size_t buff_len = 20;
vals.resize(buff_len);

sycl::range<1> work_items{vals.size()};
q.submit(
	[&](sycl::handler & cgh)
	{
		cgh.parallel_for<class vector_double>(
			work_items,
			[buff = vals.data()](sycl::id<1> tid) {
				std::format_to_n(
					buff, buff_len, "Hello from thread {}", static_cast<int>(tid));
			});
	});
q.wait_and_throw();
```

Without this patch the output is not too helpful
```
1: [AdaptiveCpp Error] from /home/dave/workspace/conan_recipe_staging/adaptivecpp/all/src/src/runtime/cuda/cuda_code_object.cpp:96 @ build_cuda_module_from_ptx(): cuda_executable_object: could not load module (error code = CU:218)
1: [AdaptiveCpp Error] from /home/dave/workspace/conan_recipe_staging/adaptivecpp/all/src/src/runtime/cuda/cuda_queue.cpp:692 @ submit_sscp_kernel_from_code_object(): cuda_queue: Code object construction failed
```

With this patch the output tells you why it failed
```
1: [AdaptiveCpp Error] from /home/dave/workspace/conan_recipe_staging/adaptivecpp/all/src/src/runtime/cuda/cuda_code_object.cpp:112 @ build_cuda_module_from_ptx(): ptxas fatal   : Unresolved extern function 'memchr' (error code = CU:218)
1: [AdaptiveCpp Error] from /home/dave/workspace/conan_recipe_staging/adaptivecpp/all/src/src/runtime/cuda/cuda_queue.cpp:692 @ submit_sscp_kernel_from_code_object(): cuda_queue: Code object construction failed
```

I had a look at the unit tests and couldn't see anywhere a test for CUDA-specific failure modes could fit. Happy to add one if someone can point the way.

### DIFF
--- a/src/runtime/cuda/cuda_code_object.cpp
+++ b/src/runtime/cuda/cuda_code_object.cpp
@@ -87,14 +87,30 @@ result build_cuda_module_from_ptx(CUmod_st *&module, int device,
   // context on that device. This is important for the subsequent driver
   // API calls which assume that CUDA context has been created.
   cudaFree(0);
-  
+
+  static constexpr std::size_t num_options = 2;
+  std::array<CUjit_option, num_options> option_names{};
+  std::array<void*, num_options> option_vals{};
+
+  // set up size of compilation log buffer
+  option_names[0] = CU_JIT_ERROR_LOG_BUFFER_SIZE_BYTES;
+  static constexpr std::size_t error_log_buffer_size = 1024*1024;
+  option_vals[0] = reinterpret_cast<void*>(error_log_buffer_size);
+
+  // set up pointer to the compilation log buffer
+  option_names[1] = CU_JIT_ERROR_LOG_BUFFER;
+  std::string error_log_buffer(error_log_buffer_size, '\0');
+  option_vals[1] = error_log_buffer.data();
+
   auto err = cuModuleLoadDataEx(
-      &module, static_cast<void *>(const_cast<char *>(source.c_str())),
-      0, nullptr, nullptr);
+      &module, source.data(),
+      num_options, option_names.data(), option_vals.data());
 
   if (err != CUDA_SUCCESS) {
+    const auto error_log_size = reinterpret_cast<std::size_t>(option_vals[0]);
+    error_log_buffer.resize(error_log_size);
     return make_error(__hipsycl_here(),
-                      error_info{"cuda_executable_object: could not load module",
+                      error_info{error_log_buffer,
                                 error_code{"CU", static_cast<int>(err)}});
   }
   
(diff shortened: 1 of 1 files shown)


src/runtime/cuda/cuda_code_object.cpp
```diff
@@ -87,14 +87,30 @@ result build_cuda_module_from_ptx(CUmod_st *&module, int device,
...
   cudaFree(0);
-  
+
+  static constexpr std::size_t num_options = 2;
+  std::array<CUjit_option, num_options> option_names{};
+  std::array<void*, num_options> option_vals{};
+
+  // set up size of compilation log buffer
+  option_names[0] = CU_JIT_ERROR_LOG_BUFFER_SIZE_BYTES;
+  static constexpr std::size_t error_log_buffer_size = 1024*1024;
```
feltech: [question] Not sure on the best number here?  Another idea, is it worth making the log buffer `thread_local`? - then the size is less important.

illuhad: Probably it's not worth making it thread-local. JIT, when it happens, is always going to be expensive.
I would suggest though to reduce the size a bit. In case of an error, I'm skeptical that people would find it helpful to read through 1M of error messages :-) My guess is that some KB are enough to be helpful.

feltech: :+1: reduced to 10KiB in aeae66b5

src/runtime/cuda/cuda_code_object.cpp
```diff
@@ -87,14 +87,30 @@ result build_cuda_module_from_ptx(CUmod_st *&module, int device,
...
-      0, nullptr, nullptr);
+      &module, source.data(),
+      num_options, option_names.data(), option_vals.data());
 
   if (err != CUDA_SUCCESS) {
+    const auto error_log_size = reinterpret_cast<std::size_t>(option_vals[0]);
+    error_log_buffer.resize(error_log_size);
     return make_error(__hipsycl_here(),
-                      error_info{"cuda_executable_object: could not load module",
+                      error_info{error_log_buffer,
```
illuhad:
```suggestion
                      error_info{"cuda_executable_object: Could not load module, CUDA JIT log: "+error_log_buffer,
```
We should probably still emit the message that CUDA module loading failed in case the error log is empty for whatever reason (perhaps there was a failure prior to the JIT process starting or so).

feltech: Updated as suggested in 12778d76

illuhad (approved):
Thank you! :)

feltech:
Thanks! Though the intention of the `fixup!` commits was that they be squashed in before merge ;)

PR #1420, opened 2024-03-25
Specializations that make sense: Add support for sycl::specialized<T>

illuhad:
SYCL 2020 introduces specialization constants. Specialization constants allow turning runtime values into constants at JIT-time and can thus lead to substantial performance improvements.

The SYCL 2020 specialization constant API however suffers from two main issues:
- they are very cumbersome to use
- in case they need to be emulated (which happens whenever an ahead-of-time compiler is used instead of a JIT compiler) they add additional overhead due to the API design. This is especially unfortunate since due to their use case as an advanced kernel optimization feature, they are likely to appear the hottest parts of the code.

In summary, SYCL 2020 specialization constants require effort to use and then can be actively detrimental for performance. Because of this, we do not not want to recommend using them and do not support them in AdaptiveCpp - they are just not useful in their current form.

This PR introduces an alternative mechanism for the same problem. The basic reasoning is that if specialization constants potentially have to be emulated, it is perhaps better to express them as optimization hints to the runtime.
This is done in this PR: If a kernel argument is wrapped inside `sycl::specialized<T>` when passed inside the kernel, if possible, the runtime will introduce a constant based on the runtime value of the argument into the kernel at JIT time.

This makes specialization constants trivial to use, and, since `sycl::specialized` is a trivial wrapper, also a zero-cost abstraction when they need to be emulated.

The `sycl::specialized` hint only affects code generation for the SSCP compiler. For the other compilation flows, code still compiles and works if it uses `sycl::specialized`, but no performance impact is expected compared to code not using `sycl::specialized` (in fact, it is expected that the exact same code is generated).

Example:

```c++

sycl::queue q;

sycl::specialized<float> scaling_factor = // some runtime value
float* data = ...

  // The JIT compiler will treat the value of scaling_factor as a constant at JIT-time.
  // E.g, if scaling_factor is 1 at runtime, the compiler may generate an empty kernel.
  data *= scaling_factor;
});

```

@Luigi-Crisci fyi, if somebody asks about this at your IWOCL talk - this is the AdaptiveCpp answer to specialization constants.

### DIFF
--- a/doc/extensions.md
+++ b/doc/extensions.md
@@ -4,6 +4,33 @@ AdaptiveCpp implements several extensions that are not defined by the specificat
 
 ## Supported extensions
 
+### `HIPSYCL_EXT_SPECIALIZED`
+
+This extension adds a mechanism to hint to the SSCP JIT compiler that a kernel specialization should be generated. That is, when `sycl::specialized<T>` is passed as a kernel argument, the compiler will generate a kernel with the value of the object stored in the `specialized` wrapper hardcoded as a constant. This addresses the same problem as SYCL 2020 specialization constants, however it provides two major benefits:
+
+* It is much, much easier to use for users due to a far more intuitive API
+* It is a true zero-cost abstraction that does not add any overhead when the code is not compiled with a JIT compiler. This is in contrast to SYCL 2020 specialization constants which can actively *hurt* performance when they need to be emulated (e.g. if the compiler uses ahead-of-time compilation).
+
+As a consequence, unlike SYCL 2020 specialization constants, the `sycl::specialized` mechanism can be used as an optimization hint that has no drawbacks when the compiler cannot capitalize on it.
+
+Example:
+
+```c++
+
+sycl::queue q;
+
+sycl::specialized<float> scaling_factor = // some runtime value
+float* data = ...
+
+q.parallel_for(range, [=](auto idx){
+  // The JIT compiler will treat the value of scaling_factor as a constant at JIT-time.
+  // E.g, if scaling_factor is 1 at runtime, the compiler may generate an empty kernel.
+  data *= scaling_factor;
+});
+
+```
+
+`sycl::specialized` currently only affects the code generation of the SSCP JIT compiler (`--acpp-targets=generic`), and only if `ACPP_ADAPTIVITY_LEVEL` is set to any value larger than 0 (the default is 1).
 
 ### `HIPSYCL_EXT_SCOPED_PARALLELISM_V2`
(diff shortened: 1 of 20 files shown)


Luigi-Crisci:
Thanks @illuhad for the tag! I will surely mention this during the talk.
A question out of curiosity: apart from (I suppose) increased compile time, are there other reasons for which one should not use this feature? 
From what I see, every time one has a large enough kernel with no/few recompilation phases where the compile time/runtime tradeoff is positive you should always make parameters constants, but maybe there are other cases where this assumption does not hold

illuhad:
@Luigi-Crisci Every different value of `sycl::specialized` arguments will trigger a new JIT compilation. If you can offset that additional overhead with the performance increase, it's generally going to be worth it.
Theoretically, there is also some (tiny - we are talking one additional for loop when calculating the kernel hash) additional  overhead associated with the management of the different kernel variants. It's unlikely to be measurable in real world applications, but you could perhaps construct some pathological cases where it can matter.
The next step for us will be to have the runtime determine automatically when things are worth specializing :)

Luigi-Crisci:
That's indeed interesting, I will try it and check if I get any benefit.
The automatic specialization would be pretty neat combined with the stdpar offloading, looking forward to check it out when it's ready :)

illuhad:
I would be curious to hear your feedback :)
Note that `sycl::specialized` only causes kernel specialization for `--acpp-targets=generic` and if `ACPP_ADAPTIVITIY_LEVEL >= 1` (default is 1). In all other cases, the compiler is going to generate the same code as if `sycl::specialized` was not used.

Luigi-Crisci:
I launched the SYCL-Bench spec constant benchmark using the `sycl::specialized` class and it indeed works. 
This is on NVIDIA V100S
acpp_spec_constants_constructor

Question: does it make sense to add a copy assignment operator for the specialized class? It makes the writing process way easier (as you can use the specialized variable as a normal variable), instead of relying only on the constructor.
I re-run the SYCLBench spec bench using an adaptiveCPP fork with the copy assignment operator for `sycl::specialized` and it does not seem to harm performance (results are still on par with constexpr)
acpp_spec_constants_copyassignment
I already have the changes ready in that case :))

illuhad:
Awesome results, thanks for sharing :)

Yes, the intent is that the `specialized` class can be coveniently used, similarly to a regular variable. So it might make sense to add copy assignment operator.
We should also consider default constructor, but I suspect there might be complications if the wrapped type is not default-constructible.

illuhad:
@Luigi-Crisci I will merge this for now, please submit a PR with your desired changes to the interface :)

PR #1422, opened 2024-03-28
Fix HIP library linking on Windows

blinkfrog:
This PR addresses a linking issue encountered on Windows when building AdaptiveCpp applications targeting HIP backend. Previously, the CMake configuration used `-rpath` flags for linking against ROCm libraries, which are not supported on Windows. This led to warnings about unknown arguments and subsequent linking errors unless `amdhip64.lib` was manually placed in the build directory.

I've modified `CMakeLists.txt` to conditionally set `ROCM_LINK_LINE` without using `-rpath` on Windows. Instead, it directly specifies the path to `amdhip64.lib`. This approach is more aligned with Windows development practices and resolves the aforementioned warnings and errors.

The changes have been tested on a Windows development environment, ensuring that builds no longer emit `-rpath` warnings and can link against `amdhip64.lib` when integrated multipass compile flow is used without requiring manual file placement.

These changes streamline the build process on Windows by eliminating the need to manually manage `amdhip64.lib` placement and avoiding unnecessary compiler warnings. It should have no adverse effects on non-Windows builds, as the modified behavior is conditional on the WIN32 flag.


illuhad (approved):
Thanks!

PR #1423, opened 2024-03-28
Extend Clang/MSVC compatibility fix to support HIP on Windows with C++20

blinkfrog:
Following up on the previously merged PR #1264, which introduced a workaround for a compatibility issue when targeting CUDA backend on Windows using MSVC's standard library and C++20, it was discovered that targeting the HIP backend under the same conditions still resulted in compilation failures. The issue was traced back to the need for a `__attribute__((host))` alongside the `__attribute__((device))` for the `std::fma` long double overload declared in the custom cmath header.

This PR extends the workaround introduced in PR #1264 to address the identified gap for HIP backend targets. It adds a `__attribute__((host))` attribute to the `std::fma` declaration, ensuring compatibility when compiling with the HIP backend on Windows using the C++20 standard. This update maintains the integrity of the original solution for CUDA targets while extending support to HIP, thereby broadening the workaround's applicability.

This change was tested to ensure that it does not interfere with existing functionalities for both CUDA and HIP backends. The dual attributes `__attribute__((host, device))` ensure that the `std::fma` workaround is correctly applied in both host and device contexts, reflecting the diverse compilation environments supported by AdaptiveCpp.

This solution remains a workaround pending a more permanent resolution of the underlying compatibility issues between Clang and MSVC. Continued dialogue with the LLVM project may yield a more holistic fix, eliminating the need for such targeted adjustments.


illuhad (approved):
Thanks!

PR #1427, opened 2024-04-04
[MacOS][Build][OpenMP] Add OpenMP_ROOT to CMakeLists 

### LINKED ISSUE #1355: [MacOS] [OpenMP] openmp include path should be in default cxx flags

Hello, 
as in the MacOS CI, the include path to openMP must be passed down as a CXX flags when compiling with acpp. It would be better if the such include instead would be in the config for romp cxx flags.

https://github.com/AdaptiveCpp/AdaptiveCpp/blob/c8d5224bf994aee577a02dc513c6563333021005/.github/workflows/macos.yml#L36

on MacOS by default add the openmp (the outputs of `echo -I$OMP_ROOT/include` in the CI would do) path to `default-omp-cxx-flags`

**Additional context**

For newcomers it make adoption easier on MacOS

badumbatish:
This helps create an easier and more robust build for beginners, fixes #1355

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -339,10 +339,20 @@ endif()
 
 if(APPLE)
   set(DEFAULT_OMP_FLAG "-Xclang -fopenmp")
-  execute_process(COMMAND brew --prefix
-                OUTPUT_VARIABLE BREW_PREFIX
-                OUTPUT_STRIP_TRAILING_WHITESPACE)              
-  set(OpenMP_ROOT "${BREW_PREFIX}/opt/libomp")
+  
+  if (NOT DEFINED OpenMP_ROOT) 
+    # Workflow provided by
+    # AdaptiveCpp/.github/workflows/macos.yml       
+    execute_process(COMMAND brew list libomp
+                    COMMAND grep libomp.a 
+                    COMMAND sed -E "s/\\/lib\\/.*//"
+                    OUTPUT_VARIABLE DefaultOMP_ROOT
+                    OUTPUT_STRIP_TRAILING_WHITESPACE)   
+    # If not found, we are still able to set 
+    # since CMake will error out if no other platform is found
+    # but will not error out if another platform is found.
+    set(OpenMP_ROOT ${DefaultOMP_ROOT})
+  endif()
   
   if(Boost_FIBER_LIBRARY_DEBUG)
     set(DEFAULT_BOOST_LIBRARIES "${Boost_CONTEXT_LIBRARY_DEBUG} ${Boost_FIBER_LIBRARY_DEBUG} -Wl,-rpath ${Boost_LIBRARY_DIR}")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -340,20 +340,6 @@ endif()
 if(APPLE)
   set(DEFAULT_OMP_FLAG "-Xclang -fopenmp")
   
-  if (NOT DEFINED OpenMP_ROOT) 
-    # Workflow provided by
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -339,6 +339,10 @@ endif()
 
 if(APPLE)
   set(DEFAULT_OMP_FLAG "-Xclang -fopenmp")
+  execute_process(COMMAND brew --prefix
```
fodinabor: This won't allow for anyone to build OpenMP themselves...
So maybe only do all of this if `OpenMP_ROOT` is undefined?

Also, I'm not much of a Mac person, so I'm not sure if it makes a difference - in the CI we do the following to identify the full `OpenMP_ROOT` - is `prefix/opt/libomp` always going to be the correct path?
https://github.com/AdaptiveCpp/AdaptiveCpp/blob/0359cac9e08d44c48c586ef3ec88e95ac64a4d44/.github/workflows/macos.yml#L23

fodinabor (commented):
Thanks for looking into this!

CMakeLists.txt
badumbatish: Hi there, thanks for the review, I haven't looked into the yml yesterday but I've modified the `execute_process`  cmake command to have the same effect as `brew list libomp | grep libomp.a | sed -E "s/\/lib\/.*//"`.

Also add an IF(NOT DEFINED OpenMP_ROOT) to make sure other people can build it

fodinabor: Hi, thanks!
I just realized that still will override manually specified hints such as the environment variable `OMP_ROOT`.
So, maybe it makes sense to move this logic to where we're actually looking for OpenMP and first doing an `find_package(OpenMP QUIET)`, and if that fails, add the `OpenMP_ROOT` and search with `REQUIRED`.

The relevant location will be: https://github.com/AdaptiveCpp/AdaptiveCpp/blob/0359cac9e08d44c48c586ef3ec88e95ac64a4d44/src/runtime/CMakeLists.txt#L280

To show this working in CI, maybe remove the line I referenced earlier from the Action yml... The CI should the still work with your changes.

badumbatish: hmm i think on MacOS, CMake can only find OpenMP if either OMP_ROOT or OpenMP_ROOT is defined. When I did what you suggested first with find_package(OpenMP QUIET) before searching for OpenMP_ROOT, CMake can't seem to find the OpenMP package.

I also removed the old checks in CMakelists and remove the OMP_ROOT on yml

badumbatish:
It seems that automation of build/ is good but failed on actual linkage, maybe we didn't actually define an include path that is based on OpenMP_ROOT inside of CMakeLists.txt, do you want me to add that directly to CXXFlags just as the yml is doing?

badumbatish:
hmm i see the problem now, I don't think the CMakeLists.txt in tests/ gets any information about the OpenMP_ROOT variable in CMakeLists.txt in runtime/

badumbatish:
This should run fine now on github actions, I just want to make sure there's a way to actually use another variables like OPENMP_MACOS_INCLUDE_DIR instead of coupling the thing into CMakeLists.txt just like the yml.

I've refactored the appending of CMAKE_CXX_FLAGS into different target_include_dir of different targets in tests/ now.

I'd love some feedbacks @fodinabor

illuhad:
@fodinabor Do you still have some concerns?

fodinabor (approved):
Looks good now!
Thanks, and sorry for not responding earlier, didn't really monitor Github due to too many other things happening.

badumbatish:
No worries! I'm just happy every build passes now :)

PR #1430, opened 2024-04-04
Add constructors and copy assignment operators to specialized.hpp

Luigi-Crisci:
This PR includes some utility construction methods to the `specialized` class interface to make it more flexible.
It includes:
- Two copy assignment operators: the usual one and one using the `T` value explicitly
- Default, copy, and move constructors.

It does not include any tests as I am not very familiar with the AdaptiveCpp/LLVM test infrastructure (I only found tests for `sycl::specialized` [here](https://github.com/AdaptiveCpp/AdaptiveCpp/blob/develop/tests/compiler/sscp/specialized.cpp), but I would love to add some if you point me where to add them

### DIFF
--- a/include/hipSYCL/sycl/specialized.hpp
+++ b/include/hipSYCL/sycl/specialized.hpp
@@ -7,54 +7,77 @@
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
  *
- * 1. Redistributions of source code must retain the above copyright notice, this
- *    list of conditions and the following disclaimer.
+ * 1. Redistributions of source code must retain the above copyright notice,
+ * this list of conditions and the following disclaimer.
  * 2. Redistributions in binary form must reproduce the above copyright notice,
  *    this list of conditions and the following disclaimer in the documentation
  *    and/or other materials provided with the distribution.
  *
- * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
- * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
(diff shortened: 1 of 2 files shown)


include/hipSYCL/sycl/specialized.hpp
```diff
@@ -7,54 +7,77 @@
...
+
+  specialized<T> &operator=(const T &value) {
+    sycl::specialized<T> tmp{value};
+    swap(*this, tmp);
+    return *this;
+  }
 
-  operator T () const {
-    return _value.value;
+  specialized<T> &operator=(specialized<T> other) {
```
illuhad, line 64: Does this not create ambiguous overloads with the `const T&`  overload? Or was this supposed to be `specialized&&`?

illuhad:
Thank you, as for the tests: The file you have found is part of the LLVM LIT tests, which focuses more on compiler tests. For tests for the SYCL interface itself we also have the `sycl_tests`.
To test the C++ interface, you could probably either expand the existing LLVM LIT test cases, or you could add a new SYCL test in `tests/sycl/extensions.cpp`.

include/hipSYCL/sycl/specialized.hpp
Luigi-Crisci: The idea was to copy the parameter directly in the function parameter rather than using the const reference and manually copying it. 
From my tests, C++ falls back to the implicit conversion to `sycl::specialized` if it cannot find the more specific operator. Maybe declaring the `specialized(const T &value)` as explicit would be clearer, I don't think the user wants to construct a `sycl::specialized` implicitly anyway

Luigi-Crisci:
I added a simple test under `tests/sycl/extensions.cpp` using both copy assignment operators

include/hipSYCL/sycl/specialized.hpp
illuhad: Ohhhh... now I get it... the one overload is for `T`, the other for `specialized<T>`. I missed that distinction :)

illuhad (approved):
Thanks!

PR #1436, opened 2024-04-14
Replace deprecated APIs in the src and tests

ouankou:
While building ACPP from source and running the built-in tests, I noticed a few warnings caused by deprecated APIs.

1. bool llvm::StringRef::startswith(llvm::StringRef) const (in the source)
```
warning: ‘bool llvm::StringRef::startswith(llvm::StringRef) const’ is deprecated: Use starts_with instead [-Wdeprecated-declarations]
```
2. Atomic accessors in SYCL 2020 (in the tests)
```
warning: 'operator[]' is deprecated: Atomic accessors are deprecated as of SYCL 2020 [-Wdeprecated-declarations]
```
3. 'get_count' in SYCL 2020 (in the tests)
```
warning: 'get_count' is deprecated: renamed to 'size' in SYCL 2020 Specification [-Wdeprecated-declarations]
```

I'm still new to SYCL, so please let me know if the proposed change contains any errors. Thanks!

### DIFF
--- a/src/compiler/cbs/LoopSplitterInlining.cpp
+++ b/src/compiler/cbs/LoopSplitterInlining.cpp
@@ -137,7 +137,7 @@ bool fillTransitiveSplitterCallers(llvm::Function &F,
   std::transform(F.begin(), F.end(), std::back_inserter(Blocks), [](auto &BB) { return &BB; });
 
   if (fillTransitiveSplitterCallers(Blocks, SAA, FuncsWSplitter,
-                                    InIntrinsic || F.getName().startswith("__hipsycl_sscp"))) {
+                                    InIntrinsic || F.getName().starts_with("__hipsycl_sscp"))) {
     FuncsWSplitter.insert(&F);
     return true;
   }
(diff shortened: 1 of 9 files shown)


fodinabor (commented):
Re `starts_with`: the issue is that the new form was introduced only quite recently, so using the new form exclusively breaks any LLVM version older than the one doing the introduction. (See the github actions)
This is why I opted to stay with the old version when I saw this warning.
Longer term, we will have to use macros that select the function name based on `LLVM_VERSION_MAJOR`.
Two options for this: checking at every use, or create a macro `ACPP_LLVM_STARTS_WITH` or so that holds the correct function name.

nilsfriess:
Re the deprecated SYCL functions: Since these are only deprecated and not (yet) removed from the SYCL spec, I think we should still include them in the tests. Ideally we should test both the deprecated function and the "replacement". Maybe we can consider disable deprecation warnings for the tests to reduce the noise

illuhad:
> Re the deprecated SYCL functions: Since these are only deprecated and not (yet) removed from the SYCL spec, I think we should still include them in the tests. Ideally we should test both the deprecated function and the "replacement".

Yes. They are officially still part of SYCL, and SYCL CTS also tests for deprecated APIs.

ouankou:
Thanks for the clarifications! I'll close the PR since those deprecated APIs must still be kept and tested.

Issue #1433, opened 2024-04-07
Failed to 'make install' in macbook pro m1 : 


adi-lb-phoenix:
Thanks for the amazing repo , it has been a great learning experience trying to  build software's such as adaptive cpp ,  it introduced me to  LLVM JIT compiler ,  building LLVM from source , made me curious about sycl from its mentions in  documentation  .
So I have a macbook pro m1  , following are the steps I followed in trying to build adaptiveCpp .

Following the installation guide on https://github.com/AdaptiveCpp/AdaptiveCpp/blob/develop/doc/installing.md and trying to build on an macbook pro m1 with 
Chip : Apple M1
macOS : Ventura 13.1
Memory : 16GB
.

git clone https://github.com/AdaptiveCpp/AdaptiveCpp
cd AdaptiveCpp
mkdir build && cd build
cmake -DCMAKE_INSTALL_PREFIX=/Users/adithyalbhat/github/adaptive_cpp/AdaptiveCpp/build ..
make install
The following was the error output of make install .

 
```/Users/adithyalbhat/github/adaptive_cpp/AdaptiveCpp/src/libkernel/sscp/host/math.cpp:109:1: error: use of undeclared identifier 'exp10f'
  109 | HIPSYCL_SSCP_MAP_HOST_FLOAT_BUILTIN(exp10)
      | ^
/Users/adithyalbhat/github/adaptive_cpp/AdaptiveCpp/src/libkernel/sscp/host/math.cpp:36:12: note: expanded from macro 'HIPSYCL_SSCP_MAP_HOST_FLOAT_BUILTIN'
   36 |     return name##f(x);                                                 \
      |            ^
<scratch space>:267:1: note: expanded from here
  267 | exp10f
      | ^
/Users/adithyalbhat/github/adaptive_cpp/AdaptiveCpp/src/libkernel/sscp/host/math.cpp:109:37: error: use of undeclared identifier 'exp10'
  109 | HIPSYCL_SSCP_MAP_HOST_FLOAT_BUILTIN(exp10)
      |                                     ^
/Users/adithyalbhat/github/adaptive_cpp/AdaptiveCpp/src/libkernel/sscp/host/math.cpp:199:10: error: use of undeclared identifier 'lgammaf_r'
  199 |   return lgammaf_r(x,y);
      |          ^
[...]

PR #1437, opened 2024-04-15
Documentation for installing AdaptiveCpp on macbook Pro M1

adi-lb-phoenix:
Documentation for installing AdaptiveCpp on macbook Pro M1 with reference to  https://github.com/AdaptiveCpp/AdaptiveCpp/issues/1433

### DIFF
--- a/doc/install_AdaptiveCpp_on_macbook pro_M1.txt
+++ b/doc/install_AdaptiveCpp_on_macbook pro_M1.txt
@@ -1,0 +1,46 @@
+Bug reference https://github.com/AdaptiveCpp/AdaptiveCpp/issues/1433
+
+For better installation procedure I followed the instructions on this page https://github.com/AdaptiveCpp/AdaptiveCpp/blob/b61a18683cb66734b3d6a1d02ab3e3cb6f1d7d8d/.github/workflows/macos.yml
+
+The pre required packages and software  executed were 
+
+```
+set +e
+brew update
+brew install cmake
+brew install libomp
+brew install boost
+set -e
+```
+
+${GITHUB_WORKSPACE} is the root directory of the AdaptiveCpp
+```
+git clone https://github.com/AdaptiveCpp/AdaptiveCpp
+cd AdaptiveCpp
+mkdir build && cd build
+OMP_ROOT=`brew list libomp | grep libomp.a | sed -E "s/\/lib\/.*//"`
+cmake .. \
+-DOpenMP_ROOT=$OMP_ROOT \
+-DWITH_OPENCL_BACKEND=OFF \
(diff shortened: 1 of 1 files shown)


illuhad:
Thank you. Perhaps some of our users with Mac experience (@normallytangent ?) can comment on the documented steps.

It might be a good idea to also enable the SSCP compiler and document how to set things up with LLVM. The SSCP JIT compiler for host CPU I think should work on Mac.

At a higher level, I think the most important question is how this content would fit into the existing documentation. This PR just provides a text file that is referenced from nowhere, which makes it questionable whether it would be found by users. We could add some structure to cover different OS. Or, we do the same what we do currently with Windows and document the steps for Mac in a Wiki-page (https://github.com/AdaptiveCpp/AdaptiveCpp/wiki/Using-AdaptiveCpp-on-Windows).

adi-lb-phoenix:
@illuhad  Thank you for the reply  . 
I think its better if documentation related to macOS  can be replicated for how its done with windows as you suggested , because installing and running the application should not be cumbersome . Also the documentation related to macOS that's already existing was not complete enough . I will push the changes soon and keep you posted .
why do you suggest  the idea "it might be a good idea to also enable the SSCP compiler " ?

illuhad:
> why do you suggest the idea "it might be a good idea to also enable the SSCP compiler " ?

Because the SSCP JIT compiler (`--acpp-targets=generic`) is our default and most powerful compiler. It can also target the host CPU and in principle it should be able to run on Mac. So ideally, we can verify that it works and then also recommend people to use it on Mac. We are already heavily pushing Linux users towards using the SSCP compiler.

We should first enable it in Ma CI though; @fodinabor recently tried this and discovered some hiccups.

fodinabor:
I still would hope that somehow we could get CI running... On a local MacOS install, I was able to use SSCP with this branch: https://github.com/AdaptiveCpp/AdaptiveCpp/tree/feature/enable-sscp-macos However, I had to make really sure that I used the correct libc++ (i.e. the one from brew, I think).
However, as the relatively long line of attempts to debug CI on https://github.com/AdaptiveCpp/AdaptiveCpp/tree/macos-ci-dbg2 shows, I'm not so sure how to make this right..

The issue is: recent brew LLVM is built against its own libc++, therefore, we should also build AdaptiveCpp (runtime & tests) against that libc++. However, the current failure in the CI was that tests that test if a certain exception is thrown fail, as the exception is not captured by boost - my intuition is, that this is because boost & AdaptiveCpp are built against other libc++(-abi) versions, therefore not being able to capture each other's exceptions...
I'm not entirely sure how to deal with this: correctly + conveniently.

Issue #1431, opened 2024-04-06
On Arm platform tests for group functions failed when targeting omp.accelerated


ouankou:
**Bug summary**
Acpp can be built without issue on Oracle A1 VPS (4-core Ampere Altra chip). However, the built-in tests for group functions all failed.
I saw this issue related to Arm: https://github.com/AdaptiveCpp/AdaptiveCpp/issues/1044
It seems acpp has been tested on Arm. I'm wondering if I didn't set up acpp correctly. 

**To Reproduce**
```bash
# build acpp and builtin tests
./sycl_tests --run_test=group_functions_tests
```
**Expected behavior**

All tests should pass.

**Describe your setup**

CPU: Ampere Altra (Arm Neoverse-N1, 4 cores)
RAM: 24 GB
OS: Debian 12
Software: Boost 1.74, Python 3.11

* how you have installed AdaptiveCpp, and which version/branch/git commit:
  Build from source in GitHub repo develop branch
* Describe the dependencies that AdaptiveCpp sits on top of in your setup: 
   * clang version and how clang was installed: 
     Clang/LLVM 17.0.6 from official LLVM apt repo.
   * host compiler:
     GCC 12.2 from official Debian repo
   * CUDA version (if applicable)
   * ROCm version and how ROCm was installed (if applicable)
* How you have compiled your application and which arguments you have passed to `acpp`. In particular, which backends and hardware you have compiled for: 
```
cmake -DWITH_SSCP_COMPILER=OFF -DWITH_ACCELERATED_CPU=ON -DWITH_OPENCL_BACKEND=OFF -DCMAKE_INSTALL_PREFIX=$HOME/Projects/AdaptiveCpp/install -DWITH_CUDA_BACKEND=OFF -DWITH_ROCM_BACKEND=OFF -DWITH_LEVEL_ZERO_BACKEND=OFF -DDEFAULT_TARGETS='omp.accelerated' ..
```

**Optional additional diagnostic information**
The following information can potentially help us:
* For compilation/toolchain or setup-related issues: Setting the environment variable `HIPSYCL_DRYRUN=1` during compilation will cause `syclcc` to print the commands it would execute instead of actually executing. This can help verify the sanity of your setup.
* `syclcc --hipsycl-version` prints information about the AdaptiveCpp configuration that may be interesting for setup-related issues.
```
acpp warning: syclcc is deprecated; please use acpp instead.
[...]

PR #1438, opened 2024-04-17
Fix UB in some test cases

nilsfriess:
Compiling the tests with `-fsanitize=address,undefined` revealed some problems:
- In the atomic tests, reductions were performed on pointer types that were UB (it's UB to add to a non-zero offset to nullptr; but the reductions always started with the value 0). For the pointer types, the reductions now start either from 1 (when the reduction is adding values) or from `std::numeric_limits<int>::max()` (when the reduction is subtracting values).
- In the half tests, some variables had wrong types that could not represent the stored value (e.g., the result of `(unsigned int)1 - sycl::half{2}` was stored in a `unsigned int`)
- Two other tests contain helper functions that take a `sycl::vec` and return certain swizzles. However, the functions were called like `foo({1,2,3})` which causes the vec to be destroyed immediately after the function call, causing use-after-free errors (since the returned swizzles are views into the - now destroyed - vec). Constructing the vecs before the function call (and passing them by const ref) fixes this.

The tests still don't run clean when compiled with `-fsanitize=address,undefined`, there is an integer overflow somewhere in the group tests. I had a quick look at it, but I think the group tests are a bit too complex for a "quick look". We should probably look into this though, maybe it's the reason for the errors reported in #1431.

### DIFF
--- a/tests/sycl/atomic.cpp
+++ b/tests/sycl/atomic.cpp
@@ -29,6 +29,7 @@
 
 #include "sycl_test_suite.hpp"
 #include <cstddef>
+#include <limits>
 #include <type_traits>
 
 using namespace cl;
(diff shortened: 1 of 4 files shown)


illuhad (approved):
Thanks!

PR #1451, opened 2024-05-08
Add `buffer(shared_ptr<T[]>, ...)` constructor and add some buffer tests cases

nilsfriess:
This PR adds a missing constructor to the buffer class and introduces a few test cases, and makes a small change in the buffer `shared_ptr` constructors to handle the case when the `shared_ptr`s are empty.


tests/sycl/buffer.cpp
```diff
@@ -278,4 +278,161 @@ BOOST_AUTO_TEST_CASE(buffer_container_constructor_no_def_constr) {
...
+    for (auto val : hostdata)
+      BOOST_CHECK(val == testVal);
+  }
+
+  // Constructor that takes a unique ptr
+  {
+    std::unique_ptr<int, std::default_delete<int[]>> hostptr{
+        new int[size], std::default_delete<int[]>{}};
+
+    s::buffer<int> buf{std::move(hostptr), size};
```
fodinabor, line 339: @nilsfriess which overload is supposed to actually handle this?
Some MacOS build with old-ish libc++ isn't able to resolve to a specific constructor....

nilsfriess: I would have thought it chooses one of the constructors that takes a `std::shared_ptr<T>` as first argument (in this case the one without the `AllocatorT` argument).

fodinabor: https://github.com/AdaptiveCpp/AdaptiveCpp/pull/1609/commits/4041e9cd8763cabdd60e3692fd3cbd0b9a364bb2 helps for me.. Seems libc++ doesn't have this participate in overload resolution.

nilsfriess: But this is already tested (in the test starting in line 288), I added this particular test because the SYCL spec states the following in Sec. 4.7.2.3.:

> Note that since there is an implicit conversion of a std::unique_ptr to a std::shared_ptr, a std::unique_ptr can also be used to pass the ownership to the SYCL runtime. 

Maybe you could try something like this

```cpp
    const auto make_test_ptr = [size]() {
        return std::unique_ptr<int, std::default_delete<int[]>>{
            new int[size], std::default_delete<int[]>{}};
    };

    s::buffer<int> buf{make_test_ptr(), size};
```

fodinabor: hmm.. it seems like this issue is messing this up:
https://github.com/llvm/llvm-project/commit/a38a4654ce4b1d2ae8a03797d2e520e415150492

I.e. shared_ptr in libc++ has issues with being constructed from `T` vs `T[]` `unique_ptr`s.

Added a workaround: https://github.com/AdaptiveCpp/AdaptiveCpp/pull/1609/commits/db522727b94a71c6a93a7d8d44b29becf2ada18b

PR #1458, opened 2024-05-13
Add ACPP_APPDB_DIR env variable to override appdb location

illuhad:
This introduces `ACPP_APPDB_DIR` environment variable to override the location of the app db (by default `~/.acpp`) which can be useful e.g. in containerized environments where there might be no writable home directory by default.

@gonzalobg


gonzalobg:
Thank you Aksel!

PR #1471, opened 2024-06-06
Drop support for LLVM < 14

illuhad:


fodinabor:
while you're at it.. LLVM 18 is released - we might want to add that to the CI?

fodinabor (approved):
LGTM :)

illuhad:
> while you're at it.. LLVM 18 is released - we might want to add that to the CI?

Yep, we should do that, but perhaps in a separate PR :)

PR #1478, opened 2024-06-11
Reworking the includes in the hipSYCL include folder to relative includes.

keldu:
The move of the include headers to the AdaptiveCpp subfolder breaks some previously working behaviour if you compiled in a mixed compiler setup.  

Also might break on mixed installations of AdaptiveCpp with versions < 1.x.x if you're missing the include hint (-I option) .  
For a possible CI test I'm not too sure what to include exactly. I can write one though if necessary.


.gitignore
nilsfriess, line 1: Is this change supposed to be in this PR?

.nix/derivation.nix
nilsfriess, line 1: Same question here

default.nix
nilsfriess, line 1: And here

keldu: no. I thought I removed it tbh. will remove it and I'll just squash these changes.

keldu: Probably slipped my mind due to lunch break

.nix/derivation.nix
keldu: As mentioned in the default.nix discussion

.gitignore
keldu: As mentioned in the default.nix discussion

default.nix
keldu: Removed my build setup files.

nilsfriess:
> Also might break on mixed installations of AdaptiveCpp with versions < 1.x.x if you're missing the include hint (-I option) .
For a possible CI test I'm not too sure what to include exactly. I can write one though if necessary.

I'm not sure if we need an extra CI test. In each CI workflow, AdaptiveCpp is installed to some local directory, so that case is covered, and we will soon-ish have new CI workflow that packages AdaptiveCpp as a debian package, installs it system wide and runs the tests. So the case of a system wide installation would then also be covered. What do you think @illuhad ? I remember you two discussed this a bit on discord, did you have a specific test case in mind?

keldu:
I'm basically already packaging AdaptiveCpp locally. Not Debian though. Since I've forgot to remove the nix files yesterday I can also just say that it's nixOS. My example would be no problem on nix, but the combination of installing old AdaptiveCpp <1.0.0 and the new one plus using a mixed compiler setup would cause issues on a Debian system with the `#include "hipSYCL/....."` includes when not being careful.  

For me on nix this fix helps to avoid more compiler flags.  
Currently I avoid the flags by moving the subdirs to the system wide level when packaging AdaptiveCpp.  
  
The extra CI test would be for mixed compiler setups like mine, because I support multiple backends at once.  

And I'd personally like to avoid hiding every sycl mention in source files, that requires some workarounds.  
Kernel launches etc are all compiled with acpp, but explicitly copying data with SYCL 2020 USM is being compiled with clang++16 on my side.  

If you don't think this is necessary, then it's fine by me.

PR #1480, opened 2024-06-12
[Renaming] Update copyright notice: change project name, year, and shorten it

nilsfriess:
This PR updates the copyright notices of all files that previously contained copyright notices using the old name.

The new copyright notice reads:
```c
/*
 * This file is part of AdaptiveCpp (formerly hipSYCL, Open SYCL), an
 * implementation of SYCL and C++ standard parallelism for CPUs and GPUs
 *
 * Copyright (c) 2019-2024 Aksel Alpay and contributors
 * All rights reserved.
 *
 * AdaptiveCpp is released under the BSD 2-Clause "Simplified" License.
 * See file LICENSE in the project root for full license details.
 */
```
The year range depends on the previous copyright notice: if it already contains a range of years, it just updates the end year. If it contains only one year, then it changes it to {old_year} - 2024.

I took the liberty of shortening the whole thing by just referring to the LICENSE file instead of putting it in each file, but I can also copy the complete license back in. Of course I didn't do this by hand, I wrote a Python script to do it, so I can just change the new copyright notice and run it again.

PR #1485, opened 2024-06-20
Update the year listed by the acpp compilation driver

marcosolanki:
As part of my recent packaging efforts, I have been running `acpp --acpp-version` quite a lot. I could not help but notice that something was ever so slightly off... 🙂

### DIFF
--- a/bin/acpp
+++ b/bin/acpp
@@ -1925,7 +1925,7 @@ def print_config(config):
 
 def print_version(config):
   version = config.version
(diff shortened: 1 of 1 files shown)


illuhad (approved):
Thanks!

fyi @nilsfriess - perhaps you would like to align `acpp` copyright output with the changes to copyright notices in PR #1480 ? In the meantime I think we can merge this as it already improves the situation.

PR #1489, opened 2024-06-22
[Doc] Fix typos throughout

nmnobre:
Hi Aksel,

Long time ~no see~ no word.
I think that might change on Monday though, I registered as soon as you were a confirmed speaker. :)

Cheers,
-Nuno


illuhad (approved):
Thanks & nice to hear from you again :)

Unfortunately, I will only present remotely on monday - which reminds me, I still need to work on my presentation ;)

tdavidcl:
I've started using this to check for typos recently, it works fairly well
https://github.com/codespell-project/codespell

nmnobre:
> Unfortunately, I will only present remotely on monday - which reminds me, I still need to work on my presentation ;)

Ah, alas! Maybe on a future opportunity!

PR #1490, opened 2024-06-22
[SSCP] Add support for automatically detecting invariant kernel arguments for ACPP_ADAPTIVITY_LEVEL=2

illuhad:
This PR adds support for automatically detecting kernel arguments which remain mostly invariant, and then automatically specializing those arguments (i.e. hardwire them as constants during JIT-time).

For some applications, this can yield noticable speedups, e.g. +10% for miniBUDE when the application is run multiple times (@tom91136)

In order to enable this functionality, statistical information about kernel arguments is stored on disk in the "app db" for the current application. This data structure can be inspected using the new `acpp-appdb-tool` and is stored in messagepack format.

Note that the default adaptivity level is 1, so this feature is not enabled by default - for now users explicitly need to set `ACPP_ADAPTIVITY_LEVEL=2`. Typically applications need to be run 3-4 times for best performance.


doc/performance.md
```diff
@@ -54,6 +54,8 @@ This optimization process is complete when the following warning is no longer pr
 
 The extent of this can be controlled using the environment variable `ACPP_ADAPTIITY_LEVEL`. A value of 0 disables the feature. The default is 1. Higher levels are expected to result in higher peak performance, but may require more application runs to converge to this performance. The default level of 1 usually guarantees peak performance for the second application run.
 
+The maximum adaptivity level currently implemented is 2. At this level, AdaptiveCpp will attempt to detect invariant kernel arguments, and hardwire those as constants during JIT time. In some cases, this can result in substantial performance increases. It is thus advisable to try setting `ACPP_ADAPTIVITY_LEVEL=2` and running the application a couple of times.
+
 **For peak performance, you should not disable adaptivity, and run the application until the warning above is no longer printed.**
 
 *Note: Adaptivity levels higher than 1 are currently not implemented.*
```
nilsfriess:
```suggestion
*Note: Adaptivity levels higher than 2 are currently not implemented.*
```

doc/performance.md
```diff
@@ -54,6 +54,8 @@ This optimization process is complete when the following warning is no longer pr
 
 The extent of this can be controlled using the environment variable `ACPP_ADAPTIITY_LEVEL`. A value of 0 disables the feature. The default is 1. Higher levels are expected to result in higher peak performance, but may require more application runs to converge to this performance. The default level of 1 usually guarantees peak performance for the second application run.
 
+The maximum adaptivity level currently implemented is 2. At this level, AdaptiveCpp will attempt to detect invariant kernel arguments, and hardwire those as constants during JIT time. In some cases, this can result in substantial performance increases. It is thus advisable to try setting `ACPP_ADAPTIVITY_LEVEL=2` and running the application a couple of times.
```
nilsfriess: Maybe leave the first sentence out (it's mentioned below anyways) and instead say "At Adaptivity level > 1, AdaptiveCpp will attempt...". This way you don't have to remember to change this paragraph again when some day you introduce Adaptivity Levels > 2

illuhad: Thanks, I have slightly reformulated that section.

doc/performance.md
illuhad: Thanks, added :)

PR #1495, opened 2024-06-28
[CI] Fix benchmark action

nilsfriess:
This (hopefully) makes the benchmark action useful.

The action is enabled for PRs that are labeled with the `benchmark` label and is run when such a PR is opened, edited, or when new commits are added to the PR. The label can either be added directly when opening the PR or also later on (adding the label also triggers the action). For now, this always compares new commits against the base commit.


nilsfriess:
The action fails because it tries to comment on the PR but doesn't have the permission to do so.
I think this is only something you can change @illuhad (should be somewhere in the repo's settings under Actions > General).

PR #1499, opened 2024-06-30
[SSCP] Add automatic kernel dead argument elimination

illuhad:
Some backends allocate registers to store kernel arguments. We should thus reduce the number of kernel arguments as much as possible.

This PR does this by running a dead argument elination pass in the SSCP compiler *after* the optimization pipeline during JIT. The indices describing which arguments were not elided need to be stored persistently, since we also have a persistent cache for JIT binaries. The appdb introduced in #1490 is used for this purpose.
At runtime, the retained indices are then stored as part of the kernel code object.
Dead argument elimination only takes place for adaptivity level >= 1, or if there's only a single kernel in the binary. This allows us to save us an additional lookup at runtime to find such kernel-specific information.

Dead argument elimination is supported for every backend except CPU (`omp`). This is because the `omp` backend passes in kernel arguments differently, and it's also quite likely that dead argument elimination would not be very important there (@fodinabor please correct me if my assessment is wrong :) ).


fodinabor:
Omp also converged to using the argmapper, so I'm not sure as to what level the args are handled differently.
However, I will agree, that register pressure on CPUs is not as much a concern, as CPU ABI generally dictates how many arguments are given via registers, everything else is stack values anyways.

illuhad:
@fodinabor doesn't the generated kernel function have a different signature from device backends? I think it just takes a `void**`. This means that the compiler-side dead argument elimination will have a different impact on the function signature that the host side needs to match.

PR #1507, opened 2024-07-08
More kernel submission latency optimizations

illuhad:
This PR introduces more kernel submission latency optimizations. SSCP kernel submission latency is down 20-30% with this PR. The SMCP compilation flows may also benefit, but to a lesser extent.

- If a dedicated in-order executor is used, we no longer add dependencies on the same queue. Those would be removed later on anyway, but that adds a bit of latencies. 
   - This change breaks the `AdaptiveCpp_retarget` extension, which relies on in-order semantics emulation, so this extension now requires that the queue is explicitly constructed with a new `AdaptiveCpp_retargetable` queue property. The queue property again enables adding these in-order events as dependencies.
- We now use an unordered dense hash map instead of `std::unordered_map` in a couple of key places, like kernel cache and HCF cache
- Some optimizations around how things are passed into the `handler` when it is constructed
- A more efficient hash algorithm when constructing the kernel configuration hash. Instead of processing on byte at a time, we now process 64 bit words.
- Replace `thread_local` in the SSCP submission path (this affects kernel configuration and arg mapper) with class members and spin locks, which it turns out is much faster than thread local storage.
- Simplify instrumentation guards so that they don't interfere so much with submissions that ultimately don't request instrumentation
- HCF kernel info is now only queried a single time per kernel instantiation, and then cached. This avoids having to query the hash map for every submission for the kernel info.

@al42and not sure how large the improvements for SMCP are going to be, but this might be worth a try still! Are you still using `get_wait_list`? If so there might be an issue since this PR optimizes away events for the in-order case.

### DIFF
--- a/doc/extensions.md
+++ b/doc/extensions.md
@@ -172,6 +172,10 @@ struct AdaptiveCpp_retarget {
 };
 
 }
+
+namespace sycl::property::queue {
+struct AdaptiveCpp_retargetable {};
+}
 ```
 
 ##### Description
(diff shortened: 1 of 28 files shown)


al42and:
> @al42and not sure how large the improvements for SMCP are going to be, but this might be worth a try still!

There seems to be around 2-3% improvement (ROCm 6.1.2, MI250X, instant submission mode) in kernel launch timings with this PR (in 12k-768k atoms range). :+1: 

Notably, for really small kernels (6k atoms), it appears that the effect is the opposite (~3% slowdown), but could be just noise or some other unknown effect; we have seen some irregularities in this regime before.

> Are you still using get_wait_list?

No, we just do `queue.hipSYCL_enqueue_custom_operation([=](sycl::interop_handle&) {});` when we need an event.

illuhad:
Ok, thanks for the feedback :)

PR #1510, opened 2024-07-10
[CI] Add more benchmarks to the benchmark Github action

nilsfriess:
Add more benchmarks of the SYCL-Bench benchmark to the benchmark action. Note sure if they will be run if I add the benchmark label to this PR or if we have to merge this first.

### DIFF
--- a/.github/workflows/benchmark-self-hosted.yml
+++ b/.github/workflows/benchmark-self-hosted.yml
@@ -5,11 +5,11 @@ on:
     types: [opened,edited,synchronize,labeled]
 
 env:
-  SYCL_BENCH_BENCHMARKS: scalar_prod 2DConvolution
+  SYCL_BENCH_BENCHMARKS: 2DConvolution 2mm 3DConvolution 3mm CMakeDetermineCompilerABI_C.bin CMakeDetermineCompilerABI_CXX.bin DRAM arith atax atomic_reduction bicg correlation covariance fdtd2d gemm gesummv gramschmidt host_device_bandwidth kmeans lin_reg_coeff lin_reg_error local_mem matmulchain median mol_dyn mvt nbody pattern_L2 reduction scalar_prod segmentedreduction sf syr2k syrk usm_accessors_latency usm_allocation_latency usm_instr_mix usm_pinned_overhead vec_add
 
 jobs:
-  run-benchmark:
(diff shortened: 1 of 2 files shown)


.github/workflows/benchmark-self-hosted.yml
```diff
@@ -29,7 +29,7 @@ jobs:
         git clone https://github.com/bcosenza/sycl-bench.git
```
illuhad, line 29: The repo is now at https://github.com/unisa-hpc/sycl-bench/
Also, it might perhaps be good to fix the commit that we are using, so that we don't get surprises if the upstream repo changes?

.github/workflows/benchmark-self-hosted.yml
```diff
@@ -5,11 +5,11 @@ on:
     types: [opened,edited,synchronize,labeled]
 
 env:
-  SYCL_BENCH_BENCHMARKS: scalar_prod 2DConvolution
+  SYCL_BENCH_BENCHMARKS: 2DConvolution 2mm 3DConvolution 3mm CMakeDetermineCompilerABI_C.bin CMakeDetermineCompilerABI_CXX.bin DRAM arith atax atomic_reduction bicg correlation covariance fdtd2d gemm gesummv gramschmidt host_device_bandwidth kmeans lin_reg_coeff lin_reg_error local_mem matmulchain median mol_dyn mvt nbody pattern_L2 reduction scalar_prod segmentedreduction sf syr2k syrk usm_accessors_latency usm_allocation_latency usm_instr_mix usm_pinned_overhead vec_add
```
illuhad: I'd also be interested in the runtime tests, in particular task throughput! There are also some interesting SYCL 2020 benchmarks, like specialization constants.

.github/workflows/benchmark-self-hosted.yml
```diff
@@ -38,9 +38,9 @@ jobs:
         mkdir benchmarks
         for benchmark in $SYCL_BENCH_BENCHMARKS; do cp ../build/$benchmark ./benchmarks/$benchmark; done
         ./run-suite gpu-warmup
```
illuhad, line 40: Is this already what is used in the results that have been posted? If so we may want to increase the problem size a bit, as the runtimes are still pretty small.

PR #1514, opened 2024-07-11
Fix legacy multi_ptr ctor

al42and:
In SYCL2020, the legacy interface for `multi_ptr<ElementType, Space, access::decorated::legacy` has non-explicit constructor from a raw pointer.

See the listing below Table 92 in SYCL2020 spec rev. 8


illuhad (approved):
Thanks!

PR #1533, opened 2024-07-17
[CI] Release AdaptiveCpp Packages

normallytangent:


ThijsWithaar:
Note that this seems to have the same goal as what I'm after in https://github.com/AdaptiveCpp/AdaptiveCpp/pull/1094.
Surprisingly, they're quite complementary in terms of changes.
Would it make sense to combiner these PR's?

PR #1536, opened 2024-07-18
[cmake] Introduce ACPP_COMPILER_FEATURE_PROFILE, deprecate WITH_SSCP_COMPILER etc

illuhad:
Introduces `ACPP_COMPILER_FEATURE_PROFILE` and deprecates `WITH_SSCP_COMPILER`, `WITH_STDPAR_COMPILER`, `WITH_ACCELERATED_CPU`, `WITH_CUDA_NVCXX_ONLY`.

The idea is that by providing a couple of simple feature profiles instead of fine-grained feature switches, we can simplify the installation for users and make it easier for them to express what kind of setup they want, while also simplifying maintenance.

The following compiler feature profiles are available:
* `full` - everything enabled (default)
* `minimal` - as little compiler integration as possible. Only `hip` and `cuda` compiler-based compilation flows are available.
* `none` - library-only, only `omp.library-only` or potentially `cuda-nvcxx` is available.

We still have to support the old flags for backwards compatibility for at least one release cycle. This PR issues deprecation warnings if they are used, and if so, switches to an additional profile `deprecated-custom` which allows this fine-grained customization.

Because we still have to support the old options, I found that we still have to retain all of the old cmake infrastructure. This makes benefitting from the expressivity of the new model difficult for now. So the new `ACPP_COMPILER_FEATURE_PROFILE` currently effectively acts as a frontend that sets the old variables on behalf of users.

This PR does not do that yet (because we're still stuck in the old infrastructure) but the new approach could also allow us to generate better errors, or better comply with the user's intent. For example, if the feature profile is `none` or `minimal`, we would not enable the OpenCL backend which needs `full`.


CMakeLists.txt
```diff
@@ -200,8 +255,10 @@ else()
...
 
-if(WITH_CUDA_BACKEND AND NOT WITH_CUDA_NVCXX_ONLY)
-  set(BUILD_CLANG_PLUGIN true)
+if(WITH_CUDA_BACKEND)
+  if(NOT WITH_CUDA_NVCXX_ONLY AND NOT ACPP_COMPILER_FEATURE_PROFILE STREQUAL "none")
+    set(BUILD_CLANG_PLUGIN true)
+  endif()
 endif()
 if(WITH_ROCM_BACKEND)
   set(BUILD_CLANG_PLUGIN true)
```
al42and, line 264: If I follow the logic correctly, if the machine has ROCm installed, the `cmake .. -DACPP_COMPILER_FEATURE_PROFILE=none` will end up building the plugin. That does not look correct.

Same for `WITH_OPENCL_BACKEND` below.

illuhad: > If I follow the logic correctly, if the machine has ROCm installed, the cmake .. -DACPP_COMPILER_FEATURE_PROFILE=none will end up building the plugin. That does not look correct.

No, the plugin is built if `feature profile != none && !WITH_CUDA_NVCXX_ONLY`. So these two settings prevent the plugin to be built.

What does no longer exactly work as before is the `WITH_CUDA_NVCXX_ONLY` switch, since at the time where we initialize this in the custom-deprecated profile we don't know yet which backends need to be supported. But I think the interaction of this switch with SSCP and other backends and so on was poorly defined from the beginning.

For OpenCL/L0 you are right, but that is the same behavior as currently. The difficulty is to retain both the functionality of the old flags, as well as the new one.
We could enforce that when `feature profile == none` an error is emitted or so, but on the whole I think the full semantics of the new flag can really only shine once the old stuff can be removed.

If `feature profile == none` it does not *search* for OpenCL, which is also what is currently happening. 

I think that generally, we currently have the issue that infrastructure is centered around backends.
Backends are detected/enabled by users, and that then enables the compiler features needed.
For `feature profile == none` to work well, we'd need to reverse this. This requires some more involved refactoring which probably makes it difficult to retain the old behavior for backwards compatibility.

illuhad: I've changed it so that OpenCL and Level Zero don't pull in the clang plugin if the feature profile is "none" or "minimal".

PR #1539, opened 2024-07-19
Fix non-instant tasks not interfacing correctly with instant tasks for non-emulated in-order queue

### LINKED ISSUE #1525: Regression around reductions on instant-submission
e8717464 / #1507 introduces a regression around reductions on in-order queues when instant submission is active.

Found via bisecting AdaptiveCpp, it reliably causes Celerity test suite to fail. Creating a minimal reproducer turned out to be somewhat difficult, the following sometimes reproduces the error, but not very reliably. On my machine it appears to only trigger on the first run (CUDA device image cache miss?), so after one run the parameter to `combine()` has to be changed to make it trigger repeatedly.

Reproducer:
```c++
sycl::queue q(sycl::gpu_selector_v, sycl::property::queue::in_order{});
const auto buf = sycl::malloc_device<int>(1, q);
q.parallel_for(sycl::range(32), sycl::reduction(buf, sycl::plus<int>{},
	sycl::property::reduction::initialize_to_identity{}), [](auto, auto& r) { r.combine(1); });
auto e = q.parallel_for(sycl::range(32), [](auto){});
e.wait();
sycl::free(buf, q);
```
Build with
```
-DHIPSYCL_ALLOW_INSTANT_SUBMISSION=1 -DHIPSYCL_TARGETS=cuda:sm_75
```

Output:
```
[AdaptiveCpp Error] from /software-local/fabian/code/hipSYCL/src/runtime/cuda/cuda_event.cpp:46 @ wait(): cuda_node_event: cudaEventSynchronize() failed (error code = CUDA:700)
[AdaptiveCpp Error] from /software-local/fabian/code/hipSYCL/src/runtime/cuda/cuda_queue.cpp:201 @ ~cuda_queue(): cuda_queue: Couldn't destroy stream (error code = CUDA:700)
[AdaptiveCpp Error] from /software-local/fabian/code/hipSYCL/src/runtime/cuda/cuda_event_pool.cpp:41 @ destroy(): cuda_event_factory: Couldn't destroy event (error code = CUDA:700)
[AdaptiveCpp Error] from /software-local/fabian/code/hipSYCL/src/runtime/cuda/cuda_event_pool.cpp:41 @ destroy(): cuda_event_factory: Couldn't destroy event (error code = CUDA:700)
```

Acpp was built with:
```
-DWITH_CUDA_BACKEND=YES -DWITH_OPENCL_BACKEND=NO -DWITH_SSCP_COMPILER=NO
```

System: 4x RTX 2070, Linux 5.15 x86_64, CUDA 12.5, Clang 17

illuhad:
When submitting non-instant tasks to a non-emulated in-order queue and instant submission is enabled, we need to `flush_sync` so that subsequent instant tasks synchronize correctly.

This is not ideal, as it can have negative performance impact when `ACPP_ALLOW_INSTANT_SUBMISSION` is set to 1, but occasionally non-instant tasks (e.g. using a buffer) are submitted.
So we should look into whether there is a better approach in the future.

Fixes #1525 

CC @fknorr

### DIFF
--- a/include/hipSYCL/sycl/handler.hpp
+++ b/include/hipSYCL/sycl/handler.hpp
@@ -1112,6 +1112,10 @@ private:
   const rt::node_list_t& get_cg_nodes() const
   { return _command_group_nodes; }
 
+  bool contains_non_instant_nodes() const {
+    return _contains_non_instant_nodes;
+  }
+
   
   handler(const context &ctx, async_handler handler,
           const rt::execution_hints &hints, rt::runtime* rt,
@@ -1182,6 +1186,8 @@ private:
         op->is_requirement()) {
       // traditional submission
       rt::dag_build_guard build{_rt->dag()};
+      _contains_non_instant_nodes = true;
+      
       return build.builder()->add_command_group(std::move(op), requirements, hints);
     } else {
 
@@ -1220,6 +1226,7 @@ private:
   rt::runtime* _rt;
 
   bool _operation_uses_reductions = false;
+  bool _contains_non_instant_nodes = false;
 
   algorithms::util::allocation_cache* _allocation_cache;
 
--- a/include/hipSYCL/sycl/queue.hpp
+++ b/include/hipSYCL/sycl/queue.hpp
@@ -1043,9 +1043,20 @@ private:
     cgf(cgh);
 
     rt::dag_node_ptr node = this->extract_dag_node(cgh);
-    if (is_in_order() && _impl->needs_in_order_emulation) {
-      _impl->previous_submission = node;
+    if (is_in_order()) {
+      if(_impl->needs_in_order_emulation) {
+        _impl->previous_submission = node;
+      } else if(cgh.contains_non_instant_nodes()) {
+        // If we have instant submission enabled, non-emulated in-order queue
+        // but non-instant tasks, we need to flush the dag, otherwise future instant
+        // tasks might not wait on the tasks that have been cached in the dag
+        // builder.
+#if ACPP_ALLOW_INSTANT_SUBMISSION
+        _impl->requires_runtime.get()->dag().flush_sync();
+#endif
+      }
     }
+
     return node;
   }
       


fknorr:
This fixes the original issue in Celerity, although I don't quite understand why, because we explicitly call `dag::flush_async()` after every submission already 🤔 

Aside from this I like the fix because as far as I understand, this now guarantees forward progress of every submission onto an in-order queue even without calling `event::wait` or `queue::wait`, as long as instant submission is enabled.

illuhad:
> This fixes the original issue in Celerity, although I don't quite understand why, because we explicitly call dag::flush_async() after every submission already 🤔

I think `flush_async` won't be sufficient here, you'd need `flush_sync`. The latter blocks and waits until the task is really submitted, whereas the former just starts the asynchronous submission process.

The core issue is that when a subsequent instant kernel depends on a non-instant kernel, the instant kernel will be submitted right away whereas the non-instant kernel might still be queued up for submission in the non-instant path.
If you use `flush_async` it could still happen that the non-instant kernel has not completed submission yet by the time the instant kernel is submitted.

> Aside from this I like the fix because as far as I understand, this now guarantees forward progress of every submission onto an in-order queue even without calling event::wait or queue::wait, as long as instant submission is enabled.

Yes, although there might be negative performance implications because of the blocking `flush_sync`. So I'm quite sure that there are cases where this causes regressions :( But then, if you use a lot of buffers, you wouldn't use instant submission anyway.

Ideally we could find a way to simplify and reconverge all the different submission models that we have by now :(

fknorr:
Ah yes, `flush_async` doesn't actually do anything after an instant submission. That does make sense.

> Yes, although there might be negative performance implications because of the blocking flush_sync.

Then I guess we shouldn't depend on this remaining true in the future, right?

illuhad:
> Then I guess we shouldn't depend on this remaining true in the future, right?

I think for now this is an implementation detail. I think it might make sense to consider whether we can fold buffer operations completely into the instant submission model as well in the future though.

PR #1550, opened 2024-07-31
Fix deprecation warnings when using LLVM 18 & 19

nilsfriess:
In newest LLVM two functions in the StringRef class were deprecated. This PR adds a simple function to use the new one if compiling with LLVM >= 18.

### DIFF
--- a/.github/workflows/linux-lit.yml
+++ b/.github/workflows/linux-lit.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        clang: [14, 15, 16, 17]
+        clang: [14, 15, 16, 17, 18]
         os: [ubuntu-22.04]
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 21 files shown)


fodinabor:
while at it.. since LLVM 19 just branched (and you fixed one of the blocking issues already - the starts_with func), can you add that to CI and check if there's more issues?

al42and:
I happened to have that lying around: https://github.com/AdaptiveCpp/AdaptiveCpp/commit/8e13b26d92e2e33d18c0fa61a1b3eb86ba239c86

There's been some pruning of headers, so we cannot rely on transitive includes much more; `equals` method has been deprecated on some object, and, annoyingly, DiagnosticHandler::DiagnosticHandlerTy changed its first argument type from reference to a pointer.

illuhad:
You will probably also need this PR: https://github.com/AdaptiveCpp/AdaptiveCpp/pull/1551

nilsfriess:
Should we add LLVM 19 already to CI?

illuhad:
> Should we add LLVM 19 already to CI?

If it works, why not :)

illuhad (approved):

PR #1566, opened 2024-09-02
removed depracated `startswith`

Arpan3323:
I have updated the usage of `inline bool llvm::StringRef::startswith(llvm::StringRef Prefix) const` with `inline bool llvm::StringRef::starts_with(llvm::StringRef Prefix) const` as `llvm-18/llvm/ADT/StringRef.h` has marked `startswith` as deprecated.

Please let me know if you have any questions or if this pull request violates any contribution rules.

### DIFF
--- a/src/compiler/stdpar/SyncElision.cpp
+++ b/src/compiler/stdpar/SyncElision.cpp
@@ -82,7 +82,7 @@ void identifyStoresPotentiallyForStdparArgHandling(
                 if (StdparFunctions.contains(CB->getCalledFunction())) {
                   Users.push_back(Current);
                   return true;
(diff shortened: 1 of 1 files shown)


illuhad:
Thanks for the PR! I think this change is already included in #1550.
Note that AdaptiveCpp currently supports LLVM versions back to 14, so in general such fixes also require backwards-compatibility `#ifdef`s

Arpan3323:
Thanks for letting me know, I will close this request. Also, thanks for telling me how this project implements such changes by taking backwards-compatibility into account. Backwards-compatibility was definitely one of those things that I was worried my proposed changes would break but I was not confident about the cleanest way (for this project) to take it into account.

PR #1570, opened 2024-09-20
Replace spurious references to hipSYCL in `acpp --help`

marcosolanki:
I would think that at this point, we can probably safely replace these two remaining explicit references to hipSYCL without confusing anyone. 🙂

*(It also appears that [Kate](https://kate-editor.org) has automatically removed quite a lot of spurious whitespace. So, I guess that is a bonus?)*


illuhad (approved):
Thanks!

PR #1582, opened 2024-10-09
Rework index.md

marcosolanki:
This pull request reworks the [landing page for the MkDocs-based documentation pages](https://adaptivecpp.github.io/AdaptiveCpp/).

In particular, it includes fixes which address a copy/paste error (unless I missed the Rust logo advertised by the Markdown source), some dubious punctuation, and some potentially false advertising (claiming AdaptiveCpp can run on \"any\", i.e. all, hardware seems a bit much, eh? 😉).

### DIFF
--- a/doc/index.md
+++ b/doc/index.md
@@ -13,7 +13,7 @@ Welcome to the documentation of AdaptiveCpp!
 
     [:octicons-arrow-right-24: Installation](./installing.md)
 
--   :fontawesome-solid-gears:{ .lg .middle } __Can run on most hardware__
+-   :fontawesome-solid-gears:{ .lg .middle } __Can run on hardware from all major vendors__
 
     ---
 
--- a/doc/index.md
+++ b/doc/index.md
@@ -1,7 +1,7 @@
 #
-![The Rust Logo](img/logo/logo-color.png)
+![The AdaptiveCpp Logo](img/logo/logo-color.png)
 
-Welcome to the documentation of the AdaptiveCpp !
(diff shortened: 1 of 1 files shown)


doc/index.md
```diff
@@ -1,24 +1,23 @@
...
 
     ---
 
-    Configure and install AdaptiveCpp
+    Configure and install AdaptiveCpp.
 
     [:octicons-arrow-right-24: Installation](./installing.md)
 
--   :fontawesome-solid-gears:{ .lg .middle } __Can run on any hardware__
+-   :fontawesome-solid-gears:{ .lg .middle } __Can run on most hardware__
```
illuhad:
```suggestion
-   :fontawesome-solid-gears:{ .lg .middle } __Can run on hardware from all major vendors__
```
Perhaps this is more in line with what we write further down? If we write "most", then this raises the question where we don't run, and users might be more uncertain whether AdaptiveCpp works on their hardware.

marcosolanki: > If we write "most", then this raises the question where we don't run...

Apple GPUs and other unholy concoctions. 😉

But, fair enough. I suppose this would allow us to designate any vendor we do not explicitly target (and whose hardware cannot be targeted via a vendor-agnostic backend) as simply \"not a major vendor\".

tdavidcl (approved):
Looks good ! (Yeah there were a few copy paste from one of my project when i made this page :) )

illuhad (approved):

Issue #1584, opened 2024-10-24
Passing arguments to JIT/OpenCL compiler (handling of denormal numbers)


MarkusBuettner:
**Describe the motivation for the feature request**
Today we figured out why, while using Intels OpenCL runtime, we would see initially performance drops in our simulation. For the first hour of simulation time, performance would gradually decrease until it increases again at then stays at a (roughly) constant level:
![grafik](https://github.com/user-attachments/assets/ebca212e-1eb1-4507-9e0f-27466d2c01fa)

This is caused by denormal numbers present in that initial simulation period. Intels OpenCL JIT compiler has an option `-cl-denorms-are-zero`, which can be set during compilation with `-Xsycl-target-backend=spir64 "-cl-denorms-are-zero"`. Setting this results in high performance right from the start.

With AdaptiveCpp we used OpenMP ahead-of-time compilation with `-ffast-math`, which on Clang also causes denormals to be flushed. Hence we don't see this performance behaviour with OpenMP.
To my knowledge, there is currently no option in AdaptiveCpp to pass flags to the OpenCL/JIT compiler.

**Describe the solution you'd like**
In my opinion the "best" solution would be to add a flag similar to icpx's `-Xsycl-target-backend` (or maybe an environment variable?) to be able to pass compiler flags to the OpenCL/JIT compiler. One could then also control CUDA/HIP compilation similar to the ahead-of-time compilation flow.

**If applicable, describe alternatives you've considered**
One could, similar to Clangs approach, also enable flushing denormals if `-ffast-math` is set by default. Apparently this is already done for JIT PTX compilation  https://github.com/AdaptiveCpp/AdaptiveCpp/blob/49544a4828da7383ae481a3fc5c5e480c8a4f142/src/compiler/llvm-to-backend/ptx/LLVMToPtx.cpp#L172

I'm not sure whether this is a good idea, I don't know whether some codes rely on that... (I hope not). 

**Additional context**
Intels oneAPI implementation (tested version 2023.2):
- `-fsycl -O3` -- denormals are not flushed (with or without `-fp-model=precise`)
- `-fsycl -O3 -Xsycl-target-backend=spir64 "-cl-denorms-are-zero"` -- denormals are flushed

Clang:
- `-ffast-math` links `crtfastmath.o`, setting FTZ/DAZ bits (at link time), unless `-mno-daz-ftz` or `-shared` is present

AdaptiveCpp:
[...]

PR #1587, opened 2024-10-25
Pass -cl-denorms-are-zero to OpenCL

MarkusBuettner:
As mentioned in #1584, denormal numbers can sometimes cause performance degradation.
With `-ffast-math`, denormals should be flushed (similar to enabling `-ffast-math` on regular clang), but this has to be passed as an OpenCL build flag.

This works with Intels OpenCL runtime and PoCL, but I cannot test at the moment whether PoCL is actually affected by this.


illuhad (approved):
Thanks! I will merge once CI comes back with green check marks :)

PR #1593, opened 2024-11-12
[SSCP][Adaptivity] Add support for AdaptiveCpp_restrict_ptr and automatic inference of noalias semantics

illuhad:
This PR adds two things:

* New `AdaptiveCpp_restrict_ptr` extension, which can be used to make a kernel argument behave like a `restrict` pointer for the optimizer
* A new feature that can automatically detect pointers which qualify for `noalias`/restrict semantics. This is achieved by tracking memory allocations. Currently, this is enabled at adaptivity level 2, but we could consider enabling it at level 1, adding an environment variable to control memory tracking, and enabling it if memory tracking is enabled.


al42and:
Hi! Just curious, syntax preferences aside, will it provide more optimizations than using `__restrict__` keyword directly? E.g.,

```cpp
void f(sycl::queue q, sycl::range<1> range, float* __restrict__ data) {
    // Converts implicitly to the underlying pointer type - float* in this
    // example.
    data[idx] *= 1.5f;
  });
}
```

illuhad:
> Hi! Just curious, syntax preferences aside, will it provide more optimizations than using __restrict__ keyword directly? 

Yes. `__restrict__` does not work for SYCL kernel arguments at all - at least in AdaptiveCpp, no idea about DPC++. My understanding is that this is due to SYCL's use of lambdas/function objects as kernel parameters, and because `__restrict__` and the associated `noalias` LLVM IR attribute attach to *values* (e.g. variables or function arguments), not *types*.

Let's look at what happens in your example from the compiler point of view:


```c++


void f(sycl::queue q, sycl::range<1> range, float* __restrict__ data) {
  // Compiler generates anonymous class type for the kernel lambda which holds the pointer:  
  struct kernel {
    // Note that there is no __restrict__ here.
    // This is because __restrict__ does not apply to types, but to values.
    // Even if this were not the case, the compiler would need to prove
    // that every use of this class is exclusively initialized with restrict pointers.
    // This is not possible in general, so the compiler might have to generate different variants of the kernel
    // (restrict/non-restrict) which, even if possible, might not always be desired.
    float* data;

    void operator()(auto idx){ /* kernel code*/ }
  };
  kernel k;
  // restrict information lost here :(
  k.data = data;
  invoke_kernel(q, range, k);
}
```

The `AdaptiveCpp_restrict_ptr` extension avoids this issue by pulling the restrict information into the type system.

With that said, I'm having a bit of a hard time finding benchmarks where this really makes a difference (if you have some, please let me know). My guess is that since SYCL is a SIMT model and enforces SIMT semantics, many cases where code normally would benefit from restrict/noalias cannot occur since they will constitute a data race condition. In theory, the compiler backend could figure this out... But I don't know whether this actually happens. So far I've also only mainly looked at CPU and NVIDIA hardware. Maybe there's more of an effect on AMD due to the more explicit vector ISA compared to NVIDIA.

PR #1595, opened 2024-11-12
[SSCP][libkernel][llvm-to-host] Assume that local memory is always aligned to 512 byte boundaries

illuhad:
The local memory pointer for the host is constructed to be aligned to page sizes in `omp_queue.cpp`. So far we haven't exploited this.
This PR asserts to the optimizer that the local memory segment on host is at least aligned to 512 bytes, which seems a safe assumption given that the page size is larger on all relevant systems.


fodinabor (commented):
For all systems, that I know, this should be okay, and as far as I understand should help the compiler.

We could maybe really enforce this alignment by returning `max(512, system_specifc_value)` from `get_page_size` in `omp_queue.cpp`, in case there's some weird systems out there...?

illuhad:
@fodinabor Good idea, done!

fodinabor (approved):
LGTM :)

PR #1596, opened 2024-11-12
changed the arch repo add script to do full system upgrade instead of partial system upgrade

JuTheDragon:
The shell script to add the repo on arch contains 'pacman -Sy', which is an unsupported partial upgrade (see https://wiki.archlinux.org/title/System_maintenance#Partial_upgrades_are_unsupported).
This pr changes that command to 'pacman -Syu', which is a supprted full system upgrade.


illuhad:
Thanks, we don't use this script anymore for repositories. And pretty much all of the files in the `scripts` directory are deprecated. @sbalint98, should we perhaps just remove the directory?

illuhad:
Superseded by #1635

PR #1599, opened 2024-11-16
[SSCP] Add JIT reflection infrastructure

illuhad:
This adds a new JIT reflection infrastructure that supersedes the old S2 IR constant infrastructure. The new infrastructure allows querying things like target architecture, target vendor etc at JIT time and specializing code paths based on that.

The way it works is that `LLVMToBackend` is configured with a *reflection map* that maps string keys to values.
For each string key, the JIT compiler replaces calls to `__acpp_sscp_jit_reflect_<key>()` with the value.
The presence of arbitrary string keys can be queried using `__acpp_sscp_jit_reflect_knows_<key>()`.

Current available keys are:

* `target_vendor_id` - returns an integer denoting the vendor id. Some common vendor ids for comparison are in `sycl::jit::vendor_id`.  Currently 0 on OpenMP.
* `target_arch` - returns a numeric identifier for the target architecture. On NVIDIA, this is the compute capability. On AMD, this is the amdgcn architecture in hexadecimal base, e.g. `0x90c` for `gfx90c`. On other backends, this is currently 0.
* `is_cpu`/`is_gpu` - whether the target device is cpu/gpu
* `has_independent_forward_progress` - whether work items have independent forward progress (NVIDIA Volta and newer)
* `runtime_backend` - the value in the `sycl::backend` enum for the runtime backend that manages this kernel.

Currently all keys are configured for all backends universally, although some are not applicable to some backends. In the future there could also be backend-specific keys.

@sbalint98 - this could be interesting for you, as these reflection queries should also work in builtin bitcode files. Simply declare the function you wish to query :)

TODO:
- [x] Add documentation
- [x] Add pretty C++ API


illuhad:
We now also have a nice C++ API for JIT-time reflection:

```c++
namespace jit = sycl::AdaptiveCpp_jit;
__acpp_if_target_sscp(
  jit::compile_if(jit::reflect<jit::reflection_query::vendor_id>() == jit::vendor_id::nvidia,
    [](){
      jit::compile_if(jit::reflect<jit::reflection_query::target_arch>()  >= 70, []() {
        // Will only be codegen'ed for Volta and newer
      });
    });
);
```

PR #1607, opened 2024-11-22
Add --acpp-dryrun-only-std-flags flag to acpp executable

### LINKED ISSUE #1605: clang-tidy issue since #1599 merge
Hi, 
clang-tidy raise some error since the merge of #1599

In a Ubuntu 20 docker image (even though the actual version is irrelevant i think to this case)
```bash
docker run --rm -it --entrypoint bash ghcr.io/shamrock-code/ci-ubuntu20:main
```

clang tidy on a cpp file having just a single include to sycl/sycl.hpp create the following error : 
```
root@885939f463c4:/home/docker# clang-tidy-17 main.cpp  -- -Iacpp/include/AdaptiveCpp
1 warning and 1 error generated.
Error while processing /home/docker/main.cpp.
acpp/include/AdaptiveCpp/hipSYCL/algorithms/util/memory_streaming.hpp:64:27: error: expected namespace name [clang-diagnostic-error]
   64 |     namespace jit = sycl::AdaptiveCpp_jit;
      |                     ~~~~~~^
Suppressed 1 warnings (1 in non-user code).
Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.
Found compiler error(s).
```

The following commands create a clang-tidy error :
```bash
wget https://apt.llvm.org/llvm.sh
chmod +x llvm.sh
sudo ./llvm.sh 17
sudo apt install -y libclang-17-dev clang-tools-17 libomp-17-dev clang-tidy-17

git clone https://github.com/AdaptiveCpp/AdaptiveCpp.git
cmake -G Ninja -S AdaptiveCpp -B AdaptiveCpp/build -DCMAKE_INSTALL_PREFIX=acpp
cd AdaptiveCpp/build
ninja && ninja install
cd ../..

printf "#include<sycl/sycl.hpp>\nint main(){}" > main.cpp
clang-tidy-17 main.cpp  -- -Iacpp/include/AdaptiveCpp
```

Additionally the error can be traced to that specific commit : 
`c07b5023c8c9b7937a29b3b2b2ac754297ed4f18`

tdavidcl:
This PR add the `--acpp-dryrun-noplugin` to acpp which does the same as `--acpp-dryrun` but removes flags starting by `-fplugin` or `-fpass-plugin` from the output.

This is done in order to get the compile flags that should be passed to external tools such as clang-tidy. However, the plugin flags creates additional warning hence their removal from the output for clang-tidy or clangd. 

### Implementation

This is done by adding a new boolean to the `run_or_print` function which will remove those flags before printing. This new boolean is False by default to avoid modifying other cases.


illuhad:
Hm... I wonder if it might not be better to do the opposite and have a flag to print includes and a flag to print defines. Even if the `-fplugin` is gone, other compiler arguments like `-mllvm` stuff will still be in there which could trip up tooling. What do you think?

bin/acpp
```diff
@@ -925,7 +935,15 @@ class acpp_config:
...
-def run_or_print(command, print_only):
+def run_or_print(command, print_only, noplugin=False):
+
+  if(noplugin):
+    new_cmd = []
+    for arg in command:
+      if not (arg.startswith("-fplugin=") or arg.startswith("-fpass-plugin")):
+        new_cmd.append(arg)
+    command = new_cmd
+
```
illuhad, line 977: If we go with my suggestion, it might be easier to handle this in the `else` branch below (line 949), where we could filter out all unnecessary/uninteresting types of flags from the full command line.

tdavidcl:
For clang-tidy and clangd removing only those two flags was sufficient to suppress any issue. But it is true that other flags should be removed for other (non clang based) tools.
Would you rather go with a whitelist or blacklist approach then ?

also it was convient to implement it as I did since from the point of view of the cli arguments plugin flags are a single string instead of isystem with span two strings in the list 😅

illuhad:
I think it should be "whitelist" in the sense of "give me all flags relating to includes/defines/etc"

tdavidcl:
Alright, I do agree that it is probably the best approach here. I'll make the changes soon.
Also about the name of the flag is `--acpp-dryrun-noplugin` still fine or should a better name be chosen ?

tdavidcl:
I've changed the implementation of the flag. Now it pass the arg list through a `filter_cmd_args` function that uses a whitelist to select the relevant arguments.

bin/acpp
```diff
@@ -925,10 +935,51 @@ class acpp_config:
...
-def run_or_print(command, print_only):
+def filter_cmd_args(command, verbose = False):
+  new_cmd = []
+
+  whitelist = [
+    "-I", "-D", "-W", "-std=","-pedantic-errors"
+  ]
+
+  whitelist_enable_next = [
+    "-isystem", "-o", "-c"
```
illuhad: `-I` can also be used in conjunction with the next argument, e.g. `-I /path/to/includes`

bin/acpp
```diff
@@ -346,6 +346,9 @@ class acpp_config:
       'is-dryrun': option("--acpp-dryrun", "ACPP_DRYRUN", "default-is-dryrun",
 """  If set, only shows compilation commands that would be executed,
   but does not actually execute it. """),
+      'is-dryrun-noplugin': option("--acpp-dryrun-noplugin", "ACPP_DRYRUN_NOPLUGIN", "default-is-dryrun-noplugin",
```
illuhad: Is this still an appropriate name? It looks like what we are doing is more like a give-me-all-standard-flags type of thing.

bin/acpp
tdavidcl: done ! I documented it also in the implementation

bin/acpp
tdavidcl: agreed, i renamed it to `--acpp-dryrun-only-std-flags`

tdavidcl:
I've made the required changes : 

- The flag has been renamed to `--acpp-dryrun-only-std-flags` has it better reflect the actual behavior
- `-I <include>` is also supported in the list

illuhad (approved):
Thanks!
I think we might have to revisit that in the future with some more granular control (e.g. arent `-L` and `-l` also standard flags?) but I guess as a new not-widely advertised feature it's fine.

Issue #1469, opened 2024-06-01
Compatibility issue with latest Microsoft C++ Library on Windows


blinkfrog:
Hello. Thank you again for amazing SYCL implementation!

Recently I was using VS 17.8.3 with MSVC 14.38.33130. AdaptiveCpp was compiled using LLVM v17.0.4, and all worked just fine.
However, I decided to update VS to the latest 17.10.1 with MSVC 14.40.33807. After this AdaptiveCpp stopped to compile programs, targeted OpenMP.

Specifically, I got a lot of error messages like this:

```
 In file included from C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.40.33807\include\chrono:25:
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.40.33807\include\format(3204,54): error GBC023BAB: capturing a structured binding is not yet supported in OpenMP
   3204 |         return _STD _Widen_and_copy<_CharT>(_Buffer, _End, _STD move(_Out));
```

I suppose the problem is in updated C++ library?

The thing I tried to do the first was recompiling LLVM, boost and AdaptiveCpp with this new MS C++ library, but this didn't help.

After some search I found this bugreport: https://www.mail-archive.com/llvm-bugs@lists.llvm.org/msg69040.html
Then I tried the godbolt code by the link on that page, and discovered that there is no such error when using LLVM 19 trunk.

Then I tried to build LLVM 19.0.0git, boost, and AdaptiveCpp, but got a lot of errors. I suppose AdaptiveCpp is not guaranteed to be built with pre-release versions of LLVM?

So, the only solution which worked was reverting VS to 17.8.3. On this version, I also tried to build LLVM 18.1.6 and AdaptiveCpp, and can confirm that this combination works on Windows.

However, I am wondering if there is a way to make AdaptiveCpp to work with the latest MS C++ library?

Thanks.

nilsfriess:
I'm afraid there's not a lot we can do here (as you've already observed). It seems that MSVC started to use structured bindings somewhere which are not yet supported in Clang OpenMP. The reason that the small godbolt example works with clang 19 seems to be [this PR](
https://github.com/llvm/llvm-project/pull/80216), so it's actually still not implemented, but clang now only shows the diagnostic if you're actually using the structured binding inside a `#pragma omp parallel` or similar.

[...]

PR #1609, opened 2024-11-22
Allow linking AdaptiveCpp into LLVM tools

fodinabor:
With this PR, it is possible to integrate AdaptiveCpp into an LLVM toolchain instead of building it separately.
This means, you can use `clang` or `opt` without having to load the AdaptiveCpp plugin (you still need to use `acpp` in most cases for the required includes, ..)

There is a bit of motivation for this:
- you can distribute one monolithic LLVM that is also AdaptiveCpp, no need to figure out which LLVM version goes with your AdaptiveCpp, ...
- this should (in theory) also work nicely on non-Linux platforms, such as Windows, where using LLVM plugins is largely unsupported. This being said, I only tested it on Linux so far :)

On my system it works great, however, I haven't been able to get it married with my [out-of-tree Github Action](https://github.com/fodinabor/hipSYCL-LLVM-upstream-ci/blob/main/.github/workflows/weekly-acpp-llvm.yml).. (might be solvable when building OpenMP as part of LLVM .. or so?) 

I'd be glad if people test this - especially on other platforms, but also on your normal setup, that this doesn't break anything!
This should do it on Linux:
```sh
git clone https://github.com/llvm/llvm-project --single-branch -b release/18.x llvm
cd llvm
git clone https://github.com/AdaptiveCpp/AdaptiveCpp -b feature/link-into-llvm AdaptiveCpp
mkdir -p build && cd build
cmake ../llvm -GNinja -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=`pwd`/../../acppllvm-install -DLLVM_TARGETS_TO_BUILD="X86;NVPTX;AMDGPU" -DLLVM_ENABLE_PROJECTS="clang" -DLLVM_PARALLEL_LINK_JOBS=8 -DLLVM_BUILD_LLVM_DYLIB=ON -DLLVM_LINK_LLVM_DYLIB=ON -DLLVM_CCACHE_BUILD=ON -DLLVM_EXTERNAL_PROJECTS=AdaptiveCpp -DLLVM_EXTERNAL_ADAPTIVECPP_SOURCE_DIR=`pwd`/../AdaptiveCpp -DLLVM_ADAPTIVECPP_LINK_INTO_TOOLS=ON -DWITH_OPENCL_BACKEND=OFF
ninja install
cd ../..
```
Assuming you have `ninja` installed, this will install an LLVM instance to `acppllvm-install` in your initial working directory.
You can then use it to e.g. build the AdaptiveCpp tests:
```sh
export PATH="`pwd`/acppllvm-install/bin:$PATH"
export LD_LIBRARY_PATH="`pwd`/acppllvm-install/lib:$LD_LIBRARY_PATH"
CC=clang CXX=clang++ cmake .. -GNinja -DCMAKE_BUILD_TYPE=Release -DAdaptiveCpp_DIR=acppllvm-install/lib/cmake/AdaptiveCpp
ninja
```

CC @tom91136 

Closes: #1608


illuhad:
Full build takes around 10min for me with `make -j32` :P But yeah, seems to work mostly.

I noticed a llvm-to-host JIT failure for `sycl_tests` (CUDA and HIP are fine):

```
[AdaptiveCpp Error] from /home/aksel/store/src/cs/cpp/llvm/llvm-18/AdaptiveCpp/src/runtime/omp/omp_queue.cpp:482 @ submit_sscp_kernel_from_code_object(): omp_queue: Code object construction failed
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace, preprocessed source, and associated run script.
Stack dump:
0.      Program arguments: /home/aksel/store/src/cs/cpp/llvm/acppllvm-install/bin/clang-18 -cc1 -triple x86_64-unknown-linux-gnu -emit-obj -dumpdir acpp-sscp-host-edbf50.so- -disable-free -clear-ast-before-backend -disable-llvm-verifier -discard-value-names -main-file-name acpp-sscp-host-5a6550.bc -mrelocation-model pic -pic-level 2 -fhalf-no-semantic-interposition -mframe-pointer=none -fmath-errno -ffp-contract=on -fno-rounding-math -mconstructor-aliases -funwind-tables=2 -target-cpu x86-64 -target-feature +prfchw -target-feature -cldemote -target-feature +avx -target-feature +aes -target-feature +sahf -target-feature +pclmul -target-feature -xop -target-feature +crc32 -target-feature +xsaves -target-feature -avx512fp16 -target-feature -usermsr -target-feature -sm4 -target-feature +sse4.1 -target-feature +avx512ifma -target-feature +xsave -target-feature +sse4.2 -target-feature -avx512pf -target-feature -tsxldtrk -target-feature -ptwrite -target-feature -widekl -target-feature -sm3 -target-feature +invpcid -target-feature +64bit -target-feature +xsavec -target-feature -avx10.1-512 -target-feature +avx512vpopcntdq -target-feature +cmov -target-feature +avx512vp2intersect -target-feature +avx512cd -target-feature +movbe -target-feature -avxvnniint8 -target-feature -avx512er -target-feature -amx-int8 -target-feature -kl -target-feature -avx10.1-256 -target-feature +evex512 -target-feature +avxvnni -target-feature -rtm -target-feature +adx -target-feature +avx2 -target-feature -hreset -target-feature +movdiri -target-feature -serialize -target-feature -sha512 -target-feature +vpclmulqdq -target-feature +avx512vl -target-feature -uintr -target-feature +clflushopt -target-feature -raoint -target-feature -cmpccxadd -target-feature +bmi -target-feature -amx-tile -target-feature +sse -target-feature +gfni -target-feature -avxvnniint16 -target-feature -amx-fp16 -target-feature +xsaveopt -target-feature +rdrnd -target-feature +avx512f -target-feature -amx-bf16 -target-feature +avx512bf16 -target-feature +avx512vnni -target-feature +cx8 -target-feature +avx512bw -target-feature +sse3 -target-feature +pku -target-feature +fsgsbase -target-feature +clzero -target-feature +mwaitx -target-feature -lwp -target-feature +lzcnt -target-feature +sha -target-feature +movdir64b -target-feature +wbnoinvd -target-feature -enqcmd -target-feature -prefetchwt1 -target-feature -avxneconvert -target-feature -tbm -target-feature -pconfig -target-feature -amx-complex -target-feature +ssse3 -target-feature +cx16 -target-feature +bmi2 -target-feature +fma -target-feature +popcnt -target-feature -avxifma -target-feature +f16c -target-feature +avx512bitalg -target-feature +rdpru -target-feature +clwb -target-feature +mmx -target-feature +sse2 -target-feature +rdseed -target-feature +avx512vbmi2 -target-feature -prefetchi -target-feature +rdpid -target-feature -fma4 -target-feature +avx512vbmi -target-feature +shstk -target-feature +vaes -target-feature -waitpkg -target-feature -sgx -target-feature +fxsr -target-feature +avx512dq -target-feature +sse4a -debugger-tuning=gdb -fdebug-compilation-dir=/home/aksel/store/src/cs/cpp/llvm/llvm-18/build/acpp-tests -fcoverage-compilation-dir=/home/aksel/store/src/cs/cpp/llvm/llvm-18/build/acpp-tests -resource-dir /home/aksel/store/src/cs/cpp/llvm/acppllvm-install/lib/clang/18 -O3 -Wno-pass-failed -ferror-limit 19 -fgnuc-version=4.2.1 -fskip-odr-check-in-gmf -fcolor-diagnostics -vectorize-loops -vectorize-slp -faddrsig -D__GCC_HAVE_DWARF2_CFI_ASM=1 -o /tmp/acpp-sscp-host-5a6550-82c30d.o -x ir acpp-sscp-host-5a6550.bc
1.      Optimizer
 #0 0x000075e8ec0780fc llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (/home/aksel/store/src/cs/cpp/llvm/acppllvm-install/bin/../lib/libLLVM.so.18.1+0x8780fc)
 #1 0x000075e8ec07559b SignalHandler(int) Signals.cpp:0:0
 #2 0x000075e8eb24c1d0 (/usr/lib/libc.so.6+0x3d1d0)
 #3 0x000075e8ef83746e (anonymous namespace)::getLocalSizeValues(llvm::Function&, int, bool) SubCfgFormation.cpp:0:0
 #4 0x000075e8ef83f219 (anonymous namespace)::createLoopsAroundKernel(llvm::Function&, llvm::DominatorTree&, llvm::LoopInfo&, llvm::PostDominatorTree&, bool) (.isra.0) SubCfgFormation.cpp:0:0
 #5 0x000075e8ef847f9a hipsycl::compiler::SubCfgFormationPass::run(llvm::Function&, llvm::AnalysisManager<llvm::Function>&) (/home/aksel/store/src/cs/cpp/llvm/acppllvm-install/bin/../lib/libLLVM.so.18.1+0x4047f9a)
 #6 0x000075e8ef8316cf llvm::detail::PassModel<llvm::Function, hipsycl::compiler::SubCfgFormationPass, llvm::PreservedAnalyses, llvm::AnalysisManager<llvm::Function>>::run(llvm::Function&, llvm::AnalysisManager<llvm::Function>&) (/home/aksel/store/src/cs/cpp/llvm/acppllvm-install/bin/../lib/libLLVM.so.18.1+0x40316cf)
 #7 0x000075e8eead2858 llvm::detail::PassModel<llvm::Function, llvm::PassManager<llvm::Function, llvm::AnalysisManager<llvm::Function>>, llvm::PreservedAnalyses, llvm::AnalysisManager<llvm::Function>>::run(llvm::Function&, llvm::AnalysisManager<llvm::Function>&) (/home/aksel/store/src/cs/cpp/llvm/acppllvm-install/bin/../lib/libLLVM.so.18.1+0x32d2858)
 #8 0x000075e8ec2824ba llvm::ModuleToFunctionPassAdaptor::run(llvm::Module&, llvm::AnalysisManager<llvm::Module>&) (/home/aksel/store/src/cs/cpp/llvm/acppllvm-install/bin/../lib/libLLVM.so.18.1+0xa824ba)
 #9 0x000075e8eeac992f llvm::detail::PassModel<llvm::Module, llvm::ModuleToFunctionPassAdaptor, llvm::PreservedAnalyses, llvm::AnalysisManager<llvm::Module>>::run(llvm::Module&, llvm::AnalysisManager<llvm::Module>&) (/home/aksel/store/src/cs/cpp/llvm/acppllvm-install/bin/../lib/libLLVM.so.18.1+0x32c992f)
#10 0x000075e8ec279817 llvm::PassManager<llvm::Module, llvm::AnalysisManager<llvm::Module>>::run(llvm::Module&, llvm::AnalysisManager<llvm::Module>&) (/home/aksel/store/src/cs/cpp/llvm/acppllvm-install/bin/../lib/libLLVM.so.18.1+0xa79817)
#11 0x000075e8f37c9492 (anonymous namespace)::EmitAssemblyHelper::RunOptimizationPipeline(clang::BackendAction, std::unique_ptr<llvm::raw_pwrite_stream, std::default_delete<llvm::raw_pwrite_stream>>&, std::unique_ptr<llvm::ToolOutputFile, std::default_delete<llvm::ToolOutputFile>>&, clang::BackendConsumer*) BackendUtil.cpp:0:0
#12 0x000075e8f37cce37 clang::EmitBackendOutput(clang::DiagnosticsEngine&, clang::HeaderSearchOptions const&, clang::CodeGenOptions const&, clang::TargetOptions const&, clang::LangOptions const&, llvm::StringRef, llvm::Module*, clang::BackendAction, llvm::IntrusiveRefCntPtr<llvm::vfs::FileSystem>, std::unique_ptr<llvm::raw_pwrite_stream, std::default_delete<llvm::raw_pwrite_stream>>, clang::BackendConsumer*) (/home/aksel/store/src/cs/cpp/llvm/acppllvm-install/bin/../lib/libclang-cpp.so.18.1+0x1fcce37)
#13 0x000075e8f3c8c176 clang::CodeGenAction::ExecuteAction() (/home/aksel/store/src/cs/cpp/llvm/acppllvm-install/bin/../lib/libclang-cpp.so.18.1+0x248c176)
#14 0x000075e8f46f4f59 clang::FrontendAction::Execute() (/home/aksel/store/src/cs/cpp/llvm/acppllvm-install/bin/../lib/libclang-cpp.so.18.1+0x2ef4f59)
#15 0x000075e8f466a0ab clang::CompilerInstance::ExecuteAction(clang::FrontendAction&) (/home/aksel/store/src/cs/cpp/llvm/acppllvm-install/bin/../lib/libclang-cpp.so.18.1+0x2e6a0ab)
#16 0x000075e8f47895f6 clang::ExecuteCompilerInvocation(clang::CompilerInstance*) (/home/aksel/store/src/cs/cpp/llvm/acppllvm-install/bin/../lib/libclang-cpp.so.18.1+0x2f895f6)
#17 0x000062e087de957b cc1_main(llvm::ArrayRef<char const*>, char const*, void*) (/home/aksel/store/src/cs/cpp/llvm/acppllvm-install/bin/clang-18+0x1357b)
#18 0x000062e087de19c1 ExecuteCC1Tool(llvm::SmallVectorImpl<char const*>&, llvm::ToolContext const&) driver.cpp:0:0
#19 0x000062e087de4ea2 clang_main(int, char**, llvm::ToolContext const&) (/home/aksel/store/src/cs/cpp/llvm/acppllvm-install/bin/clang-18+0xeea2)
#20 0x000062e087de0df4 main (/home/aksel/store/src/cs/cpp/llvm/acppllvm-install/bin/clang-18+0xadf4)
#21 0x000075e8eb234e08 (/usr/lib/libc.so.6+0x25e08)
#22 0x000075e8eb234ecc __libc_start_main (/usr/lib/libc.so.6+0x25ecc)
#23 0x000062e087de0e55 _start (/home/aksel/store/src/cs/cpp/llvm/acppllvm-install/bin/clang-18+0xae55)
```
I'm not sure if this is related to this change, or whether this PR just exposes some problem here.

Any particular reason why you've disabled the OpenCL backend? It built fine for me.

Shouldn't we also enable `openmp;lld` when building LLVM since `--acpp-targets=hip` needs lld, and host backend needs OpenMP?

As I said, it doesn't have to be perfect since this will certainly remain an experimental  feature for now, but I'd like to see at least *some* CI and documentation for merging :-)

tom91136:
Would be really cool to bundle a binary release of this built on CentOS 8 (or just anything with older glibc) and published to GH releases, although not sure if Acpp can handle relocation for the ROCm and CUDA backend...

illuhad:
@fodinabor Do you want this to be a part of 24.10?

fodinabor:
This really is pretty experimental still... So not sure how useful it will be..
I fixed a few issues, including, it works for host on Windows now (didn't get Windows to find CUDA, yet).
But I'll have to cleanup a bit as well..

fodinabor:
We should be ready now... got a lot bigger than planned :D

CI coverage of this, is currently only out of tree:
https://github.com/fodinabor/hipSYCL-LLVM-upstream-ci/actions/runs/12390232178

I would additionally update the Windows wiki page to:
# Using AdaptiveCpp on Windows
There are two ways that make AdaptiveCpp work on Windows, experimentally.
Both require some manual steps.

The two options are "standard" installation and building AdaptiveCpp as part of LLVM.
The latter is the currently recommended variant on Windows (even though still very experimental).

## Building AdaptiveCpp as part of LLVM
This is currently the recommended workflow on Windows as it enables using the `generic` target (for NVIDIA GPUs and CPUs) and the accelerated CPU compilation flow.
This is thanks to it removing the need to use Clang/LLVM plugins on Windows, which is still not fully supported.
However, it is tested only in a limited manner as well, so consider this experimental as well.

### Dependencies
- CMake
- Ninja (https://github.com/ninja-build/ninja/releases)
- Python 3
- (optional) sccache (https://github.com/mozilla/sccache/releases) for faster rebuilds
- Pre-built vanilla LLVM (https://github.com/llvm/llvm-project/releases or part of Visual Studio)

#### Building Boost
First, let's build boost. Download and extract e.g. https://boostorg.jfrog.io/artifactory/main/release/1.81.0/source/boost_1_81_0.7z.
Then build Boost using the pre-built `clang-cl`:
```batch
@rem set me! This should point to a bin directory, that includes ninja.exe and, if you want, sccache.exe
set PATH_TO_TOOLS=<YOUR_PATH_TO_NINJA_AND_SCCACHE>
set PATH=%PATH_TO_TOOLS%;C:\Program Files\LLVM\bin;%PATH%
set CC=clang-cl
set CXX=clang-cl

cd <boost_dir>

bootstrap.bat
@rem Minimal Boost, if you don't need boost yourself
.\b2 toolset=clang-win address-model=64 variant=release,debug --build-type=complete stage --with-fiber --with-context --with-test --with-filesystem --with-atomic --with-headers
@rem Full Boost, if you need libraries form Boost yourself
.\b2 toolset=clang-win address-model=64 variant=release,debug --build-type=complete stage
```

### Building LLVM + AdaptiveCpp
Now, with the dependencies sorted, let's build LLVM with AdaptiveCpp included:
```batch
@rem set me!
set LLVM_VERSION=18
@rem set me (allow around 3GB of RAM per link job)
set LLVM_PARALLEL_LINK_JOBS=8
@rem set me!
set ACPP_INSTALL_PREFIX=%cd%/../../install
@rem set me! if built as shown above, path to "boost_1_xx_y\stage"
set BOOST_INSTALL_DIR=<PATH_CONTAINING_BOOSTS_lib>
@rem set me! follow the 1.xx.y format
set BOOST_VERSION=1.83.0
@rem set me! This should point to a bin directory, that includes ninja.exe and, if you want, sccache.exe
set PATH_TO_TOOLS=<YOUR_PATH_TO_NINJA_AND_SCCACHE>
@rem remove value, if you don't want to use `sccache` if you have ccache/sccache installed for faster rebuild times
set CCACHE_CMAKE_ARGS=-DCMAKE_C_COMPILER_LAUNCHER=sccache -DCMAKE_CXX_COMPILER_LAUNCHER=sccache

set PATH=%PATH_TO_TOOLS%;C:\Program Files\LLVM\bin;%PATH%
set CC=clang-cl
set CXX=clang-cl

git clone https://github.com/llvm/llvm-project --single-branch -b release/${LLVM_VERSION}.x llvm
cd llvm
git clone https://github.com/AdaptiveCpp/AdaptiveCpp AdaptiveCpp
md build
cd build

cmake ../llvm -GNinja -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=%ACPP_INSTALL_PREFIX% -DLLVM_TARGETS_TO_BUILD="X86;NVPTX" -DLLVM_ENABLE_PROJECTS="clang;openmp;lld" -DLLVM_PARALLEL_LINK_JOBS=%LLVM_PARALLEL_LINK_JOBS% -DLLVM_EXTERNAL_PROJECTS=AdaptiveCpp -DLLVM_EXTERNAL_ADAPTIVECPP_SOURCE_DIR=%cd%/../../AdaptiveCpp -DLLVM_ADAPTIVECPP_LINK_INTO_TOOLS=ON -DWITH_OPENCL_BACKEND=OFF -DBoost_DIR=%BOOST_INSTALL_DIR%\lib\cmake\Boost-%BOOST_VERSION% -DLLVM_TOOL_BUGPOINT_BUILD=OFF -DLLVM_INCLUDE_TESTS=OFF %CCACHE_CMAKE_ARGS%

ninja install
```

### Using AdaptiveCpp
Now, you're ready to use this LLVM built with AdaptiveCpp!

To use it, add `%ACPP_INSTALL_PREFIX%/bin` to your `PATH` and set the to-be-used copmiler:
```batch
set PATH=%ACPP_INSTALL_PREFIX%/bin;%PATH%
set CC=clang
set CXX=clang
```

Then you can use `acpp` directly or use the CMake integration as you're used to.

Note, `ACPP_TARGETS` should work now with `generic`, `omp` and `cuda` on Windows.

## Standard Flow
It is possible to make the OpenMP and CUDA backend work on Windows, using separately-built LLVM.
Also NOTE: this requires a Clang Plugin and Clang CUDA, which both are not officially supported on Windows. Thus this is considered highly experimental.
We are providing compatible LLVM 11.1 and Boost 1.75 binaries to ease the process. The binaries require the [VC++ 2019 redistributable](https://aka.ms/vs/16/release/vc_redist.x64.exe) to be installed (which you most likely already have).

### Dependencies (non-exhaustive)
- CMake
- Ninja
- LLVM (patched with [D69322](https://reviews.llvm.org/D69322), included in LLVM 12+, [prebuilt 11.1.0 binary](http://repo.urz.uni-heidelberg.de/sycl/windows/llvm.7z))
- Boost (built for Clang toolchain, tested with 1.70 and 1.75, [prebuilt 1.75 binary](http://repo.urz.uni-heidelberg.de/sycl/windows/boost_1_75_0.7z))
- Python 3

AdaptiveCpp depends on Boost and mostly uses Clang for compilation. Thus these are our main dependencies.

#### Getting the LLVM with plugin support
LLVM currently does not officially support Clang plugins on Windows. As AdaptiveCpp requires the use of a Clang plugin for compiling for CUDA, LLVM has to be built manually:

- Get the LLVM source (e.g. `git clone https://github.com/llvm/llvm-project`)
- Apply the required [patch](https://reviews.llvm.org/D69322)
- Open Visual Studio Command (usually want to use x64 Native Tools)
  - if using Ninja, execute `set CC=cl` and `set CXX=cl` to use VS compiler
- Configure LLVM with CMake with enabled RTTI and plugin support `-DLLVM_EXPORT_SYMBOLS_FOR_PLUGINS=ON -DLLVM_ENABLE_RTTI=ON`
  - See [LLVM docs](https://llvm.org/docs/CMake.html#id12) for available options.
- Build and install LLVM using `ninja` and `ninja install`

A full install might look like this:
```batch
git clone https://github.com/llvm/llvm-project llvm
cd llvm
@rem if not using LLVM 12+ execute the next line:
@rem git am --3way --ignore-space-change --keep-cr "D69322.diff"
@rem check whether there were merge conflicts.. and if so, solve them
mkdir build && cd build
set CC=cl
set CXX=cl
cmake ../llvm -G Ninja -DCMAKE_INSTALL_PREFIX=<INSTALL_PREFIX> -DLLVM_ENABLE_PROJECTS="clang;lld" -DLLVM_EXPORT_SYMBOLS_FOR_PLUGINS=ON -DLLVM_ENABLE_RTTI=ON -DLLVM_TARGETS_TO_BUILD=X86 -DCMAKE_BUILD_TYPE=RelWithDebInfo
ninja
ninja install
@rem the clang lib required for the AdaptiveCpp plugin is not installed, do it manually
copy RelWithDebInfo\lib\clang.lib <INSTALL_PREFIX>/lib
```

You might want to add your `<INSTALL PREFIX>` to your PATH variable. Otherwise, all steps below require you to actively specify the full path to clang.

Building the OpenMP runtime fails for LLVM 11 and probably 12, if using MSVC, but seems to be fixed on `main` in [this commit](https://github.com/llvm/llvm-project/commit/498c4b6fc45ae4704a7f93bd03e018392bb87945). Building the runtime separately, using Clang might work, otherwise, it also is possible to just copy `libomp.dll` and `libomp.lib` from a binary distribution of LLVM to `<INSTALL_PREFIX>/bin` and `<INSTALL_PREFIX>/lib` folders respectively.

#### Getting Boost built with Clang
Get a recent boost source archive, extract and open a console in the extracted folder.
Then execute:
```shell
bootstrap.bat
.\b2 toolset=clang-win address-model=64 variant=release --build-type=complete stage
```
For faster building, you might want to add `--with-fiber --with-context --with-test`.

### Building AdaptiveCpp
Now that we have all dependencies installed, we can go about building AdaptiveCpp, just as one would for any other platform.

Change over to your AdaptiveCpp clone.
Note, AdaptiveCpp is currently not ready to work with multi-config build tools like the Visual Studio generators, thus ninja is the recommended build tool.
```shell
mkdir build && cd build
# Configure with CMake, standard mechanisms for issues apply
cmake .. -G Ninja -DCMAKE_C_COMPILER=clang.exe -DCMAKE_CXX_COMPILER=clang++.exe -DBOOST_ROOT=<BOOST_DIR> -DLLVM_DIR=<INSTALL_PREFIX>/lib/cmake/llvm -DCMAKE_BUILD_TYPE=RelWithDebInfo
ninja
ninja install # might want to add a -DCMAKE_INSTALL_PREFIX=<INSTALL_PREFIX> to the CMake command above to install locally..
```

As Windows has no idea about `-rpath`, to use AdaptiveCpp in a project, you either have to copy `hipSYCL-rt.dll` next to your executable or ensure that `<INSTALL_PREFIX>/bin` is in your `PATH`.

## Pitfalls
### `python.exe` not found or Microsoft Store always opens when trying to build with AdaptiveCpp
Ensure your python installation is activated (in case of virtual environments or Anaconda), your python.exe is in your system path.

If the MS Store always opens, when executing `python`, remove the dummy `python.exe`s as explained [here](https://stackoverflow.com/a/58773979/3452807).

### Linker error with something like iterator level
Make sure all your libraries are built with the same build type (debug vs. release).

### Something doesn't work
This is super likely to happen and something that worked today might stop working tomorrow, as there's little commitment to supporting LLVM Plugins and Clang CUDA on Windows. So everything is experimental!

### Something is not found or there is a parsing error
Windows is special when it comes to paths, so check, whether you have any whitespaces in a path and if so, try adding quotes around the path. Also, avoid using backslashes `\` in paths and instead just replace them with forward-slashes `/`.

## Known issues
### Itanium and MS-ABI mismatch
CUDA uses the Itanium ABI on Windows (regardless of Clang / NVCC). As host code still uses MS-ABI there, mismatches regarding struct/class layout are possible. One such example is the empty bases issue, which was fixed for `sycl::accessor` in [#623](https://github.com/illuhad/hipSYCL/pull/623).
There isn't much we can do about this but hope that [MSVC changes it's ABI](https://devblogs.microsoft.com/cppblog/optimizing-the-layout-of-empty-base-classes-in-vs2015-update-2-3/) sooner than later or Clang adds a [warning](https://llvm.org/PR51689) for these cases.

### MS-STL versions incompatible with Clang CUDA
Due to a version check introduced in the [MS-STL](https://github.com/microsoft/STL/commit/2ce62b58489bcf93f81a39d883597e666bc34bfe), Clang CUDA cannot be used with a current version of the MS-STL. Multiple workarounds are possible: use a version without that change (probably broken since 16.10), specify `-D_ALLOW_COMPILER_AND_STL_VERSION_MISMATCH` (unverified) or apply the [proposed patch](https://github.com/microsoft/STL/pull/2075).

### Building the tests with `cuda` fails in Boost's debug.ipp or execution_monitor.ipp
A good solution has to be found... for now a workaround is to add a `#ifndef __CUDACC__` around the areas in the headers that use `__try`...
The `generic` flow is not suffering from this.

doc/installing.md
```diff
@@ -135,12 +135,48 @@ The default installation prefix is `/usr/local`. Change this to your liking.
...
+```bash
+export LLVM_VERSION=18 # set me!
+export LLVM_PARALLEL_LINK_JOBS=8 # set me (when using the default GNU ld, allow around 4GB of RAM per link job)
+export ACPP_INSTALL_PREFIX=`pwd`/../../install # set me
+export USE_CCACHE=ON # leave set to on, if you have ccache/sccache installed for faster rebuild times
+git clone https://github.com/llvm/llvm-project --single-branch -b release/${LLVM_VERSION}.x llvm
+cd llvm
+git clone https://github.com/AdaptiveCpp/AdaptiveCpp AdaptiveCpp
+mkdir -p build && cd build
+cmake ../llvm -GNinja -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=${ACPP_INSTALL_PREFIX} -DLLVM_TARGETS_TO_BUILD="X86;NVPTX;AMDGPU" -DLLVM_ENABLE_PROJECTS="clang;openmp;lld" -DLLVM_PARALLEL_LINK_JOBS=${LLVM_PARALLEL_LINK_JOBS} -DLLVM_BUILD_LLVM_DYLIB=ON -DLLVM_LINK_LLVM_DYLIB=ON -DLLVM_CCACHE_BUILD=${USE_CCACHE} -DLLVM_EXTERNAL_PROJECTS=AdaptiveCpp -DLLVM_EXTERNAL_ADAPTIVECPP_SOURCE_DIR=`pwd`/../AdaptiveCpp -DLLVM_ADAPTIVECPP_LINK_INTO_TOOLS=ON
```
illuhad, line 162: Is disabling assertions still a thing? Might be possible that we don't need it anymore purely for SSCP.

include/hipSYCL/common/filesystem.hpp
```diff
@@ -36,22 +36,24 @@ std::string
...
+                                            std::error_code &EC);
 
 /// Writes data atomically to filename
 bool atomic_write(const std::string& filename, const std::string& data);
 
 /// Removes a file, returns true if successful.
 bool remove(const std::string &filename);
 
+// Replaces runtime specialized variables, such as $HIPSYCL_PATH in the path.
+std::string replace_known_variables(const std::string path);
```
illuhad: Why is this necessary? the replacement that is done there is likely to support a different set of substitutions than what `acpp` supports, and I'd like to avoid having to maintain such an engine twice.
This seems to be used mainly for the clang path - maybe we just add a function that specifically returns the clang path?

include/hipSYCL/compiler/llvm-to-backend/LLVMToBackend.hpp
```diff
@@ -11,6 +11,11 @@
 #ifndef HIPSYCL_LLVM_TO_BACKEND_HPP
 #define HIPSYCL_LLVM_TO_BACKEND_HPP
 
+#ifndef _WIN32
+#define HIPSYCL_BACKEND_API_EXPORT
+#else
+#define HIPSYCL_BACKEND_API_EXPORT __declspec(dllexport)
```
illuhad: Can we have new defines with `ACPP_*` naming scheme?

include/hipSYCL/compiler/llvm-to-backend/NameHandling.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+#include <string>
+
+namespace hipsycl::compiler {
+
+inline bool isValidCharInSymbolName(char C) {
+  return std::isalnum(C) || C == '_' || C == '$' || C == '.';
+}
+
+inline void replaceInvalidCharsInSymbolName(std::string &Name) {
```
illuhad: Can we name this more specifically? Like `replaceInvalidMSABICharsInSymbolName()`

include/hipSYCL/runtime/dylib_loader.hpp
```diff
@@ -20,6 +20,12 @@
 #define HIPSYCL_PLUGIN_API_EXPORT extern "C" __declspec(dllexport)
 #endif
 
+#ifdef _WIN32
+#define HIPSYCL_SHARED_LIBRARY_EXTENSION "dll"
+#else
+#define HIPSYCL_SHARED_LIBRARY_EXTENSION "so"
```
illuhad: `ACPP_*` ?

src/compiler/llvm-to-backend/host/LLVMToHost.cpp
```diff
@@ -13,16 +13,13 @@
...
 #include "hipSYCL/compiler/cbs/PipelineBuilder.hpp"
-#include "hipSYCL/compiler/cbs/SimplifyKernel.hpp"
 #include "hipSYCL/compiler/cbs/SplitterAnnotationAnalysis.hpp"
-#include "hipSYCL/compiler/llvm-to-backend/AddressSpaceInferencePass.hpp"
 #include "hipSYCL/compiler/llvm-to-backend/AddressSpaceMap.hpp"
 #include "hipSYCL/compiler/llvm-to-backend/Utils.hpp"
 #include "hipSYCL/compiler/llvm-to-backend/host/HostKernelWrapperPass.hpp"
-#include "hipSYCL/compiler/sscp/IRConstantReplacer.hpp"
 #include "hipSYCL/glue/llvm-sscp/jit-reflection/queries.hpp"
+#include "hipSYCL/runtime/dylib_loader.hpp"
```
illuhad: Would it be a lot of work to move `dylib_loader.hpp` and its implementation file to common? Don't really want LLVM to load stuff from the runtime - if we mess something up, and some of the globals of the runtime gets constructed, then the runtime starts up during compilation....

src/compiler/llvm-to-backend/host/LLVMToHost.cpp
```diff
@@ -91,57 +88,71 @@ bool LLVMToHostTranslator::toBackendFlavor(llvm::Module &M, PassHandler &PH) {
...
     MAM.registerPass([] { return SplitterAnnotationAnalysis{}; });
   });
-  PH.PassBuilder->registerModuleAnalyses(*PH.ModuleAnalysisManager);
+  // PH.PassBuilder->registerModuleAnalyses(*PH.ModuleAnalysisManager);
+
+  PH.PassBuilder->registerModuleAnalyses(MAM);
+  PH.PassBuilder->registerCGSCCAnalyses(CGAM);
+  PH.PassBuilder->registerFunctionAnalyses(FAM);
+  PH.PassBuilder->registerLoopAnalyses(LAM);
+  PH.PassBuilder->crossRegisterProxies(LAM, FAM, CGAM, MAM);
```
illuhad: The `PassBuilder` from `PH` should already have these things registered from here:
https://github.com/AdaptiveCpp/AdaptiveCpp/blob/ba3fb96fbeb4a00929f4f93b57f0831e44ed903a/include/hipSYCL/compiler/llvm-to-backend/Utils.hpp#L89

So doing it again seems wrong. Why is it necessary? If something is missing and not set up, we should probably do it there.

src/compiler/llvm-to-backend/host/LLVMToHost.cpp
```diff
@@ -91,57 +88,71 @@ bool LLVMToHostTranslator::toBackendFlavor(llvm::Module &M, PassHandler &PH) {
     return false;
 
   llvm::ModulePassManager MPM;
-  PH.ModuleAnalysisManager->clear(); // for some reason we need to reset the analyses... otherwise
-                                     // we get a crash at IPSCCP
+  // TODO: Clean me up.
+  // PH.ModuleAnalysisManager->clear(); // for some reason we need to reset the analyses... otherwise
+  //                                    // we get a crash at IPSCCP
```
illuhad: Clean this up as the todo says? :P

src/compiler/llvm-to-backend/host/LLVMToHost.cpp
```diff
@@ -91,57 +88,71 @@ bool LLVMToHostTranslator::toBackendFlavor(llvm::Module &M, PassHandler &PH) {
...
                                                     std::string &out) {
-  auto InputFile = llvm::sys::fs::TempFile::create("acpp-sscp-host-%%%%%%.bc");
-  auto OutputFile = llvm::sys::fs::TempFile::create("acpp-sscp-host-%%%%%%.so");
 
-  if (auto E = InputFile.takeError()) {
-    this->registerError("LLVMToHost: Could not create temp file: " + InputFile->TmpName);
+  llvm::SmallVector<char> InputFile;
+  int InputFD;
+  // don't use fs::TempFile, as we can't unlock the file for the clang invocation later... (Windows)
+  if(auto E = llvm::sys::fs::createTemporaryFile("acpp-sscp-host", "bc", InputFD, InputFile, llvm::sys::fs::OF_None)){
```
illuhad, line 146: If there is a problem with locking on Windows, can we do this such that we still use the properly file-locked code path on Linux?

src/compiler/sscp/TargetSeparationPass.cpp
```diff
@@ -284,6 +302,8 @@ std::unique_ptr<llvm::Module> generateDeviceIR(llvm::Module &M,
     }
   }
 
+  replaceInvalidCharsInSymbolNames(*DeviceModule);
```
illuhad: Please only do this on Windows. Otherwise all our symbols may get non-demangleable.

illuhad:
> I would additionally update the Windows wiki page to:

Yes, we should update the wiki page.

But it seems to me the "standard installation" is outdated in many ways (prebuilt LLVM 11, issues that are non-issues in SSCP like MSABI-Itanium ABI mismatch for structs)  and might be more confusing than helpful. Should we maybe just remove this? Especially when we say that the `omp.accelerated` and `generic` compilation flows are supported, which require the other route.

I wonder... if we already point people to prebuilt LLVM binaries from the release page, how involved would it be to provide prebuilt AdaptiveCpp binaries for Windows?

src/compiler/llvm-to-backend/host/LLVMToHost.cpp
fodinabor: Hmm.. it's more that on Windows, a file that's opened for writing is immediately completely locked for any other accesses.. this isn't anything special that LLVM does here .. it's just that this API allows us to close the file handle at our will, while keeping the file around.

doc/installing.md
fodinabor: Default for `!=Debug` is `OFF` 🤔

src/compiler/llvm-to-backend/host/LLVMToHost.cpp
fodinabor: Will have to test later.. this was one of the things where reusing the MAM lead to crashes (seemingly).
Maybe one of the other, later, fixes actually solved this.

src/compiler/sscp/TargetSeparationPass.cpp
fodinabor: Well, the implementation already only did it on windows, felt like it was less annoying to not have macros here, but macrod out all of it.

fodinabor (commented):
Will have to come back to it, later, once I'm back on my Windows machine.

I agree, the "old" documentation is probably too outdated and limited to be kept around. Maybe @blinkfrog would be able to write a better version for that part, their experience with it.
On that note, given the recent work on enabling LLVM shared libs on Windows, these things might be supportable (in the future).

> I wonder... if we already point people to prebuilt LLVM binaries from the release page, how involved would it be to provide prebuilt AdaptiveCpp binaries for Windows?

Should be doable, I think 🤔 
One would have to decided what to package though - including `ninja` for example?

blinkfrog:
I was mentioned in this discussion and invited to share my experience with building AdaptiveCpp on Windows. Actually, current documentation is already quite good, I’d like to add a few additional notes based on my experience.

1. When building LLVM, it is needed also to add `NVPTX` or/and `AMDGPU` targets, otherwise AdaptiveCpp wont' be able to use CUDA or/and HIP backends respectively:

`-DLLVM_TARGETS_TO_BUILD="X86;NVPTX;AMDGPU"`

Also, it is worth enabling `compiler-rt` and `openmp` runtimes, this allows to use `omp.accelerated` AdaptiveCpp backend:

`-DLLVM_ENABLE_RUNTIMES="compiler-rt;openmp"`

2. When building AdaptiveCpp, it is better explicitly disable `SSCP`, `OpenCL` and `Level Zero` backends, otherwise configuration will fail. Also, if corresponding runtimes were built for LLVM, it is possible to enable `cpu.accelerated` backend:

`-DWITH_SSCP_COMPILER=OFF -DWITH_OPENCL_BACKEND=OFF -DWITH_LEVEL_ZERO_BACKEND=OFF -DWITH_ACCELERATED_CPU=ON`


However, as I mentioned in https://github.com/AdaptiveCpp/AdaptiveCpp/issues/1469, all this doesn't work correctly with recent versions of MSVC, and this is quite a problem, as many users usually keep their Visual Studio updated, or install it anew and updated just before building LLVM and AdaptiveCpp.

And while I mentioned there that the problem was in my code and not in AdaptiveCpp, this actually not exactly true. It is more complex. I probably will open a separate discussion (although this isn't AdaptiveCpp issue but rather LLVM/MSVC), but, in short, CUDA backend doesn't work anymore with compilation stopping with errors in CUDA backend of LLVM. OpenMP works fine. This is if to use LLVM 19. And older versions of LLVM don't work with the recent MSVC because of incompatibility of MSVC standard C++ library with OpenMP, which was fixed in LLVM 19.

blinkfrog:
I've tried to build LLVM + AdaptiveCpp from this branch, it worked, but only after I disabled OpenCL backend, which I tried to compile with initially. With enabled OpenCL building stopped with this error:
```
  Building Custom Rule D:/source/llvm-project/build/_deps/llvmspirvtranslator-src/tools/llvm-spirv/CMakeLists.txt
  llvm-spirv.cpp
  llvm-spirv.vcxproj -> D:\source\llvm-project\build\_deps\llvmspirvtranslator-build\tools\llvm-spirv\Release\llvm-spirv.exe
  Building Custom Rule D:/source/llvm-project/build/_deps/llvmspirvtranslator-src/CMakeLists.txt
FAILED: runtimes/runtimes-stamps/runtimes-configure D:/source/llvm-project/build/runtimes/runtimes-stamps/runtimes-configure
runtimes\CMakeFiles\runtimes-configure-3ebf4cf.bat f2602edd9640b321
```

So, after disabling OpenCL it compiled, but when I try to compile a simple program, it stops with these errors:

```
[1/2] Building CXX object CMakeFiles/Blur.dir/blur.cpp.obj
[2/2] Linking CXX executable Blur.exe
FAILED: Blur.exe
cmd.exe /C "cd . &&  python C:/llvm.acpp/lib/cmake/AdaptiveCpp/syclcc-launcher --launcher-cxx-compiler=C:/llvm.acpp/bin/clang++.exe --launcher-syclcc="python*C:/llvm.acpp/bin/acpp"  --acpp-targets="generic" C:\llvm.acpp\bin\clang++.exe -fuse-ld=lld-link -nostartfiles -nostdlib -O3 -DNDEBUG -D_DLL -D_MT -Xclang --dependent-lib=msvcrt -Xlinker /subsystem:console CMakeFiles/Blur.dir/blur.cpp.obj -o Blur.exe -Xlinker /MANIFEST:EMBED -Xlinker /implib:Blur.lib -Xlinker /pdb:Blur.pdb -Xlinker /version:0.0   C:/llvm.acpp/lib/acpp-rt.lib  -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32 -loldnames  && cd ."
lld-link: error: undefined symbol: public: __cdecl hipsycl::common::db::appdb::~appdb(void)
>>> referenced by CMakeFiles/Blur.dir/blur.cpp.obj:(public: class hipsycl::common::filesystem::persistent_storage & __cdecl hipsycl::common::filesystem::persistent_storage::operator=(class hipsycl::common::filesystem::persistent_storage &&))
>>> referenced by CMakeFiles/Blur.dir/blur.cpp.obj:(public: __cdecl hipsycl::common::filesystem::persistent_storage::~persistent_storage(void))
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
ninja: build stopped: subcommand failed.
```

blinkfrog:
Thank you @fodinabor for this amazing PR! After the recent commits, SSCP finally works on my computer!

First I tested with NVidia GPU, and it works, although I see warnings like this:
```
'+ptx86' is not a recognized feature for this target (ignoring feature)
```

Second, I tried to test it with CPU, and had a runtime problems: I got errors like this:
```
←[1;31m[AdaptiveCpp Error] ←[from D:\source\AdaptiveCpp.X\src\runtime\omp\omp_queue.cpp:503 @ submit_sscp_kernel_from_code_object(): omp_queue: Code object construction failed
acpp-sscp-host-88af68-87619b.o : error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MT_StaticRelease' doesn't match value 'MD_DynamicRelease' in acpp-sscp-host-88af68-87619b.o
acpp-sscp-host-88af68-87619b.o : error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MT_StaticRelease' doesn't match value 'MD_DynamicRelease' in acpp-sscp-host-88af68-87619b.o
   Creating library C:\Users\Daniil\AppData\Local\Temp\acpp-sscp-host-9c9eba.lib and object C:\Users\Daniil\AppData\Local\Temp\acpp-sscp-host-9c9eba.exp
LINK : warning LNK4098: defaultlib 'msvcrt.lib' conflicts with use of other libs; use /NODEFAULTLIB:library
C:\Users\Daniil\AppData\Local\Temp\acpp-sscp-host-9c9eba.dll : fatal error LNK1319: 2 mismatches detected
clang++: error: linker command failed with exit code 1319 (use -v to see invocation)
←[1;31m[AdaptiveCpp Error] ←[from D:\source\AdaptiveCpp.X\include\hipSYCL/glue/llvm-sscp/jit.hpp:320 @ compile(): jit::compile: Encountered errors:
0: LLVMToHost: clang invocation failed with exit code 1319
```

I haven't tested with AMD GPU yet, but I hope to test soon.

illuhad:
> First I tested with NVidia GPU, and it works, although I see warnings like this:

This can happen for newer NVIDIA GPUs (also happens on Linux) and can be ignored. It impacts neither functionality nor performance.
The background is that there's no API in CUDA to find out what the latest PTX version is that the driver supports, so we need to guess something. When the guess is wrong (as in your case), you get a warning and LLVM picks some default PTX version. It's been proven rather cumbersome to solve this more robustly so far.

> I haven't tested with AMD GPU yet, but I hope to test soon.

I'd be surprised if it worked given that so far SSCP on Windows for AMD is entirely untested, and uses very different code paths. But maybe you get lucky :)

blinkfrog:
Thank you @illuhad for explanations!

>I'd be surprised if it worked given that so far SSCP on Windows for AMD is entirely untested, and uses very different code paths. But maybe you get lucky :)

Running compiled host binaries doesn't seem as straightforward as it is with the SMCP-compiler. What additional runtime dependencies are required for app, compiled with SSCP, beyond the usual ACPP runtime and backend dynamic libraries? It appears to need the `llvm/lib/hipSYCL/bitcode` directory and its contents. Are there any other dependencies? Is there a way to bundle these dependencies with the compiled binaries?

The setup works on my computer, but I intend to deploy an SSCP-compiled SYCL application on end-user computers.

illuhad:
>  Running compiled host binaries doesn't seem as straightforward as it is with the SMCP-compiler. What additional runtime dependencies are required for app, compiled with SSCP, beyond the usual ACPP runtime and backend dynamic libraries? It appears to need the llvm/lib/hipSYCL/bitcode directory and its contents. Are there any other dependencies? Is there a way to bundle these dependencies with the compiled binaries?

The SSCP flow is a JIT compiler, so you need to ship our JIT compiler stack with your binaries. The JIT compilation is what gives the resulting binaries target independence, but at the expense of more runtime dependencies. I think it's not yet quite fully optimized for deployment simplicity, so some things are currently a little more complex than they need to be.

From the AdaptiveCpp side, you need
* runtime libraries and backend plugins as usual
* `llvm-to-backend` infrastructure, including the libraries (`llvm-to-*.so/dll` depending on which backends you need) as well as the bitcode library directory that you mention.

Additionally, you need the compiler dependencies for each of the backends. What is needed here varies a bit depending on the backend:
* For all: Main LLVM library (not sure what it's called on windows, probably `libLLVM.dll`). This library should be fairly monolithic, but I don't know about LLVM builds on Windows. If there are any transitive dependencies, you would need those as well.
* For host:
   * Currently needs `clang`, which is a bit overkill. In the future I think we could reduce this to `llc` and `lld`.
* For CUDA:
   * Currently needs `clang`, which is a bit overkill. In the future I think we could reduce this to `llc`.
* For AMD:
   * Only needs the hipRTC library which already comes as part of ROCm.
* For OpenCL:
   * Only needs the `llvm-spirv` translator binary, which is built as part of the OpenCL backend. In the future, we might be able to also use `llc` here. I don't think though that the OpenCL backend would already work on Windows, since `llvm-spirv` is built as part of AdaptiveCpp, and is supposed to be automatically patched during build. This currently needs the `patch` command which I don't think exists on Windows.

blinkfrog:
Thank you very much for the information! Ok, I think, to use ACPP to compile consumer-ready apps, it still is better to stick with SMCP for now. :)

illuhad:
> Thank you very much for the information! Ok, I think, to use ACPP to compile consumer-ready apps, it still is better to stick with SMCP for now. :)

I think this is only true if you limit yourself to just a few targeted architectures. As soon as you want your binary to run on as many targets as possible, SSCP will probably win, even if collecting the dependencies at the moment might be a bit cumbersome :)

In any case, SSCP implements a ton of optimizations that are unavailable in SMCP, so you might still want to try it for the performance. Also, if you tried deploying an SSCP application, your experience would be very valuable :)

fodinabor:
> For all: Main LLVM library (not sure what it's called on windows, probably libLLVM.dll). This library should be fairly monolithic, but I don't know about LLVM builds on Windows. If there are any transitive dependencies, you would need those as well.

On Windows, this PR links LLVM statically into AdaptiveCpp's `rt-backend-omp.dll` and friends. LLVM on Windows is only slowly making progress towards being dynamically linkable... but DLLs are a bit better in encapsulating local symbols, so it should be okay to do it this way for now..

Since that also means that clang.exe and friends are statically linked, one might get away with `clang.exe` (or `clang++.exe`.. depending on the configuration of the AdaptiveCpp build) as an additional dependency from the LLVM side.

illuhad:
> Since that also means that clang.exe and friends are statically linked, one might get away with clang.exe (or clang++.exe.. depending on the configuration of the AdaptiveCpp build) as an additional dependency from the LLVM side.

That would be cool if that worked! But I assume you would probably still need `lld`?

EDIT:
> On Windows, this PR links LLVM statically into AdaptiveCpp's rt-backend-omp.dll and friends.

Are there any issues when using multiple backends simultaneously?

fodinabor:
> That would be cool if that worked! But I assume you would probably still need lld?

Maybe. Afaict not necessarily for CPU. Afaict, clang might just use link.exe (VS Linker).
Also might want to include `libomp.dll` and friends, if it's what's linked against.

> Are there any issues when using multiple backends simultaneously?

I don't think our tests try this...? I did not add an additional testcase..

blinkfrog:
Thank you @fodinabor, now OpenMP backend works.

And I want to say, that its performance is spectacular! In the test (the only I conducted yet), with `AL=2` it is ~3 times faster than `omp.accelerated` SMCP compile flow, and is just 15% slower than OpenCL for CPU (via DPC++)!

(in the same test, CUDA backend is just 3% faster than SMCP, but this is quite simple code without much room for optimizations)

Amazing work!

illuhad:
> And I want to say, that its performance is spectacular! In the test (the only I conducted yet), with AL=2 it is ~3 times faster than omp.accelerated SMCP compile flow, and is just 15% slower than OpenCL for CPU (via DPC++)!

This is good to hear :) AL2 automatically detects invariant runtime kernel arguments across multiple kernel invocations, and at some point will hardwire them as constants in a new JIT-compiled kernel. This can significantly boost performance. However, it can also be surprising for users if suddenly a new kernel is JIT-compiled.

If you already know which runtime parameters will not typically change, then it might be better to do that selectively and use AL1 together with `sycl::specialized`, which can result in fewer JIT-compiler kernels: https://github.com/AdaptiveCpp/AdaptiveCpp/blob/develop/doc/extensions.md#acpp_ext_specialized

blinkfrog:
I've tried to assemble a self-containing test to try on another computer, but couldn't. It seems that binaries are hardwired with some files from llvm directory, for example, files from `C:\llvm.acpp\lib\hipSYCL\bitcode`, and, in runtime, binaries expect these files to be there. This greatly limits the deployment capabilities of binaries, which were compiled with `generic` target.
May be it is possible to make binaries to search these files in the current working directory, so it would be enough just to put `bitcode` subdirectory to the directory of binaries (or to `hipsycl` subdirectory which already is in the working directory)?

illuhad:
> I've tried to assemble a self-containing test to try on another computer, but couldn't. It seems that binaries are hardwired with some files from llvm directory, for example, files from C:\llvm.acpp\lib\hipSYCL\bitcode, and, in runtime, binaries expect these files to be there. This greatly limits the deployment capabilities of binaries, which were compiled with generic target.
May be it is possible to make binaries to search these files in the current working directory, so it would be enough just to put bitcode subdirectory to the directory of binaries (or to hipsycl subdirectory which already is in the working directory)?


I don't think it's hardcoded. The path is obtained here: 
https://github.com/AdaptiveCpp/AdaptiveCpp/blob/7677cf6eefd8ab46d66168cd07ab042109448124/src/compiler/llvm-to-backend/ptx/LLVMToPtx.cpp#L204
and `get_install_directory()` should attempt to find the directory of the executing binary, not the original AdaptiveCpp install directory. But there's a different code path for Windows, maybe something there is not working and it enters the fallback code at the end:
https://github.com/AdaptiveCpp/AdaptiveCpp/blob/7677cf6eefd8ab46d66168cd07ab042109448124/src/common/filesystem.cpp#L56

I think the next issue that you will hit then - at least for CUDA - is that it will need `libdevice.bc` from the CUDA toolkit. This path might currently be too hardcoded and should improved. But one problem at a time :)

fodinabor:
> and get_install_directory() should attempt to find the directory of the executing binary, not the original AdaptiveCpp install directory. But there's a different code path for Windows, maybe something there is not working and it enters the fallback code at the end:

`get_install_directory` is implemented in `libacpp-common.so/dll` and returns the parent directory of that so/dll - this assumes that these libs are in the typical `bin`, `lib`, ... structure (`bin` for Windows). This is the same for the fallback code.
So yes, either you model this structure as sub-folders or we would have to add logic to also search directly in the executable's/dll's directory.
When using SSCP with this PR's build model, it will also assume, that the relevant clang.exe is in `get_install_directory()/bin`. (see `getClangPath`)

blinkfrog:
If this helps, I've used ProcMon to track where the compiled binary searches for bitcodes. It appears to look exclusively in C:\llvm.acpp\lib\hipSYCL\bitcode\libkernel-sscp-ptx-full.bc—this is the directory where I installed LLVM with AdaptiveCpp. It does not attempt to access this file from any other location. Meanwhile, the acpp-common.dll file, successfully accessed by the compiled .exe, is located in the working directory containing this .exe. The same is true for the hipSYCL directory with backend DLLs; it is also in the working directory.

illuhad:
@fodinabor, what is your plan with this? Is this now ready for final review/merging?

fodinabor:
Resolved merge conflicts with develop.

I'm currently testing MacOS. Hope to push the few required changes before the weekend.
Otherwise, I have no current plans to do any more changes as part of this PR.

fodinabor:
Okay, this might feel a bit stupid, but one thing that really works well on MacOS (as far as I can tell), is using a bootstrapping process -> 2 stage build of LLVM.

If we just build LLVM directly with AdaptiveCpp, we get to use AppleClang to build the OpenMP stuff for us, which we likely don't want (I didn't manage to make it happy....).
So, I would recommend to first build LLVM proper (with clang + compiler-rt) and then use that for the actual build in stage2, including clang + openmp + AdaptiveCpp.

I assume we should also be able to skip stage1 when using `brew install llvm`... but I lately was pretty annoyed with the brew llvm shipping its own libc++, leading to some pretty annoying ABI issues in the tests...

Similar setup as in the original post: clone llvm and then as a subfolder AdaptiveCpp. in `llvm/build`, run:
```
cmake ../llvm -GNinja -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=`pwd`/../../acppllvm-install -DLLVM_TARGETS_TO_BUILD="AArch64" -DLLVM_ENABLE_PROJECTS="clang;lld" -DLLVM_ENABLE_RUNTIMES="compiler-rt" -DLLVM_PARALLEL_LINK_JOBS=8 -DLLVM_BUILD_LLVM_DYLIB=ON -DLLVM_LINK_LLVM_DYLIB=ON -DLLVM_CCACHE_BUILD=ON  -DDEFAULT_SYSROOT= -DCLANG_ENABLE_BOOTSTRAP=ON -DCLANG_BOOTSTRAP_PASSTHROUGH="CMAKE_INTSALL_PREFIX;LLVM_TARGETS_TO_BUILD;LLVM_PARALLEL_LINK_JOBS;LLVM_BUILD_LLVM_DYLIB;LLVM_LINK_LLVM_DYLIB;LLVM_CCACHE_BUILD" -DBOOTSTRAP_LLVM_ENABLE_PROJECTS="clang;lld;openmp" -DBOOTSTRAP_LLVM_EXTERNAL_PROJECTS=AdaptiveCpp -DBOOTSTRAP_LLVM_EXTERNAL_ADAPTIVECPP_SOURCE_DIR=`pwd`/../AdaptiveCpp -DBOOTSTRAP_LLVM_ADAPTIVECPP_LINK_INTO_TOOLS=ON -DBOOTSTRAP_WITH_OPENCL_BACKEND=OFF 
ninja stage2-install
```

Might wanna do some CI for this new MacOS setup.. but not sure I'll get to that any time soon..

fodinabor:
MacOS in out-of-tree CI is now also happy: https://github.com/fodinabor/hipSYCL-LLVM-upstream-ci/actions/runs/13656913245/job/38178319901

Do you want to integrate these workflows in upstream?
If so, in what way? For push/pull requests or on a cronjob for `develop`?
https://github.com/fodinabor/hipSYCL-LLVM-upstream-ci/tree/main/.github/workflows

fodinabor:
Now also with CI for LLVM builds.
Thanks to caching, once the caches are hot, the workflows should each take < 30min, which is pretty much on-par with the non-LLVM builds.
Note, however, that initial cache filling runs may take >2hrs.. 🙈
Note, note: there are certain cache/branch rules that GitHub imposes that make it use caches only from parent commits iirc, which is why the initial builds on this branch (and likely once merged to develop) will take the full 2hrs.

The Windows LLVM CI pushes artifacts as well, which should be fully self-contained (modulo visual studio redist & STL) functional distributions of LLVM + ACpp. In theory, we could publish/link these to releases or something..
We could also use something like https://nightly.link to get a link to the latest `develop` artifact: e.g.: https://nightly.link/AdaptiveCpp/AdaptiveCpp/workflows/windows-acppllvm/ci%2Flink-into-llvm/AdaptiveCpp-LLVM20-Win.zip (replace `ci%2link-to-llvm` with the appropriate branch name)

illuhad:
What is the status here? Looks like there are some CI issues?

fodinabor:
CI problems were mostly waiting for #1752 / a self-hosted glitch ... and subsequent issues with CMake 4.0 and LLVM 🙈 

Hopefully should be sorted now.

fodinabor:
Should be good (for review/merge :))

include/hipSYCL/common/appdb.hpp
```diff
@@ -29,7 +30,7 @@ struct kernel_arg_value_statistics {
   uint64_t last_used = 0; // The number of the kernel invocation where this value
                       // was last used
 
-  void dump(std::ostream& ostr, int indentation_level=0) const;
+  ACPP_COMMON_EXPORT void dump(std::ostream& ostr, int indentation_level=0) const;
```
illuhad, line 33: Note to future us: We may have to check whether the PCUDA runtime API that the AdaptiveCpp now exports also needs some Windows declspec export.

src/compiler/llvm-to-backend/host/LLVMToHost.cpp
illuhad: Okay. So there should be no change to semantics on Linux here, right? In this case do we also want to make this change for SPIRV and amdgcn JIT so that we don't stumble across the same issue when we/someone else tries to get AMD/Intel support to work on Windows?

src/compiler/llvm-to-backend/ptx/LLVMToPtx.cpp
```diff
@@ -70,15 +70,16 @@ class LibdevicePath {
...
+    auto Files = common::filesystem::list_regular_files(BitcodeDir, EC);
+    if(EC) return false;
+
+    for(const auto& F : Files) {
+      if (F.find("libdevice.") != std::string::npos && F.find(".bc") != std::string::npos) {
+        Out = F;
+        return true;
       }
-    }catch(...) { /* false will be returned anyway at this point */ }
+    }
```
illuhad, line 82: Why do we no longer need the try/catch?

src/compiler/sscp/TargetSeparationPass.cpp
```diff
@@ -196,6 +198,22 @@ struct KernelInfo {
...
+      S.setName(Name);
+    };
+
+    for(auto &F : M) {
+      UpdateName(F);
+    }
+    for(auto &G : M.globals()){
+      UpdateName(G);
+    }
+}
```
illuhad, line 215: 2 space indentation?

src/compiler/llvm-to-backend/ptx/LLVMToPtx.cpp
fodinabor: Using error code instead of exception handling, as this is compiler world and not runtime world..
See: https://github.com/AdaptiveCpp/AdaptiveCpp/commit/dc965823782910e45837d3cba52764d238a1e839

include/hipSYCL/common/appdb.hpp
fodinabor: This is already merged in - and it didn't complain so far in the CI... so maybe not? :)
(but CI isn't doing anything specific to test PCUDA.. should it?)

illuhad: Looks like we currently only test PCUDA in self-hosted CI on Linux: https://github.com/AdaptiveCpp/AdaptiveCpp/blob/2a2795b73afb77530de2d01907bef22267adfd02/.github/workflows/linux-self-hosted.yml#L102

illuhad (approved):
Okay, let's merge and see what happens :D

PR #1613, opened 2024-11-25
Replace FetchContent_Populate with FetchContent_MakeAvailable

### LINKED ISSUE #1611: Allow building AdaptiveCpp offline with OpenCL support
I packaged [AdaptiveCpp for openSUSE Tumbleweed](https://build.opensuse.org/package/show/devel:tools:compiler/adaptivecpp). The build environment (Open Build Service) requires packages to be built offline, so sources are downloaded before build and uploaded to the service.

When trying to build with OpenCL support, CMake tries to download some headers, like [SPIRV-Headers](https://github.com/KhronosGroup/SPIRV-Headers). Even if I add them as a build dependency (so that they are located in `/usr/include`) CMake tries (and fails) to download them.

CMake should check if the required headers are already present on the system before trying to download them. If not automatically, I should be able to provide a flag to make CMake search for them in a specific location.

Vendoring headers requested by CMake and then extracting them to the checked dir, but CMake does not provide an easy way to do this so it should be part of the RPM build process.

**Additional context**

VaiTon:
Should fix #1611

### DIFF
--- a/src/compiler/llvm-to-backend/CMakeLists.txt
+++ b/src/compiler/llvm-to-backend/CMakeLists.txt
@@ -144,17 +144,11 @@ if(WITH_SSCP_COMPILER)
       GIT_REPOSITORY https://github.com/KhronosGroup/SPIRV-LLVM-Translator
       GIT_TAG origin/${LLVMSPIRV_BRANCH}
       GIT_SHALLOW ON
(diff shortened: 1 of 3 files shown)


illuhad:
Seems like this broke llvm-spirv translator. Can you check whether the patch is still applied correctly?

VaiTon:
@illuhad could you try re-running the workflows?

illuhad:
I did, seems like there is still the same issue.

PR #1615, opened 2024-11-25
Add missing llvm/IR/Module.h header in KnownPtrParamAlignmentOptPass.cpp

VaiTon:
Didn't build on my machine without it.

LLVM/Clang 19


fodinabor:
Thanks!
We should probably add the already released LLVM 19 to the CI..

illuhad (approved):
Thanks!

PR #1618, opened 2024-11-27
Warn if buffers are used

illuhad:
Since users seem to stumble over this, and since it is not obvious in a lot of SYCL teaching material, this PR adds a warning when a buffer is used for the first time in an application recommending users to switch to USM.

### DIFF
--- a/include/hipSYCL/sycl/buffer.hpp
+++ b/include/hipSYCL/sycl/buffer.hpp
@@ -20,6 +20,7 @@
 #include <type_traits>
 #include <algorithm>
 #include <utility>
+#include <atomic>
 
 #include "hipSYCL/common/debug.hpp"
 #include "hipSYCL/runtime/allocator.hpp"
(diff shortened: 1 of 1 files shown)


tdavidcl:
Would it be possible to have an env flag to disable the warning ?

illuhad:
We have that - `ACPP_DEBUG_LEVEL` will do it for a sufficiently low setting. Will that work?
Also, it only prints the warning the very first time a buffer is used in an application, so you will not be bombarded with warnings either way. You either don't see the warning, or you see it exactly once.

tdavidcl:
it is enabled for `ACPP_DEBUG_LEVEL >= 1` i guess ?
That should be fine.

illuhad:
Yep, same as any other warning. And as I said, you would never see this warning more than once per application run, similarly to e.g. our JIT warning. So I don't think the warning would be too distracting.

PR #1622, opened 2024-11-30
Increment `cmake_minimum_required` to 3.10 to avoid deprecation warnings

marcosolanki:
CMake 3.31 [deprecates compatibility with versions of CMake older than 3.10](https://cmake.org/cmake/help/latest/release/3.31.html#deprecated-and-removed-features), which causes warnings of the form:

```
CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.
```

As even Debian Buster (which is old enough at this point to be unmaintained) [ships with CMake 3.13.4](https://packages.debian.org/buster/cmake), I would argue that there should be no harm in bumping the minimum required CMake version to 3.10.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 3.9)
+cmake_minimum_required(VERSION 3.10)
 if(NOT CMAKE_VERSION VERSION_LESS 3.12)
   cmake_policy(SET CMP0074 NEW) # Don't complain about using BOOST_ROOT...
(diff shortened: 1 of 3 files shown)


illuhad (approved):
Thanks!

PR #1623, opened 2024-12-01
Fix double negation for `has_system_svm`

carbotaniuman:
This was detecting a system that had SVM as one that did not.


illuhad (approved):
Thanks!

PR #1624, opened 2024-12-02
Set CMake policy CMP0167 to suppress FindBoost removal warning

marcosolanki:
CMake 3.30 [officially removes the FindBoost module](https://cmake.org/cmake/help/latest/release/3.30.html#deprecated-and-removed-features), which causes warnings of the form:

```
CMake Warning (dev) at CMakeLists.txt:96 (find_package):
  Policy CMP0167 is not set: The FindBoost module is removed.  Run "cmake
  --help-policy CMP0167" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.
```

However, setting `CMP0167` to `NEW`, which apparently makes CMake search directly for the upstream BoostConfig.cmake file, appears sufficient to make CMake find Boost without warning (while preserving AdaptiveCpp's ability to be compiled). Hence, I have added this setting to the CMakeLists.txt.

I have currently only enabled this new behaviour for CMake versions new enough such that they are affected by the above-described warning (using the logic that people who run CMake `>= 3.30`  will also use a Boost version new enough (`>= 1.70`) such that it provides a valid BoostConfig.cmake).

I have also replaced the `NOT CMAKE_VERSION VERSION_LESS` conditions with `CMAKE_VERSION VERSION_GREATER_EQUAL`, as `VERSION_GREATER_EQUAL` [was added in CMake 3.7](https://cmake.org/cmake/help/v3.7/release/3.7.html#commands), which is below the currently required minimum CMake version (3.10).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,8 @@ if(POLICY CMP0074) # Since CMake 3.12
 endif()
 if(POLICY CMP0167) # Since CMake 3.30
     # Suppress warning: "Policy CMP0167 is not set: The FindBoost module is removed."
-    cmake_policy(SET CMP0167 NEW)
+    # Set CMP0167 to OLD to avoid breakage in the Windows & macOS CI tests.
+    cmake_policy(SET CMP0167 OLD)
 endif()
 
 
@@ -57,16 +58,16 @@ if(NOT ACPP_VERSION_SUFFIX)
       RESULT_VARIABLE GIT_LOCAL_CHANGES_RETURN_CODE
       OUTPUT_STRIP_TRAILING_WHITESPACE
     )
-  
-    if (GIT_HASH_RETURN_CODE EQUAL 0 AND GIT_BRANCH_RETURN_CODE EQUAL 0 AND 
+
+    if (GIT_HASH_RETURN_CODE EQUAL 0 AND GIT_BRANCH_RETURN_CODE EQUAL 0 AND
         GIT_DATE_RETURN_CODE EQUAL 0 AND GIT_LOCAL_CHANGES_RETURN_CODE EQUAL 0)
-    
+
       if(NOT "${ACPP_LOCAL_CHANGES}" STREQUAL "")
         set(DIRTY_STR ".dirty")
       else()
         set(DIRTY_STR "")
       endif()
-  
+
       set(ACPP_VERSION_SUFFIX "+git.${ACPP_GIT_COMMIT_HASH}.${ACPP_GIT_DATE}.branch.${ACPP_GIT_BRANCH}${DIRTY_STR}")
     endif()
   endif()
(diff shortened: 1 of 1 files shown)


marcosolanki:
OK, from what I can see:

- On Windows, for whatever reason, CMake fails to locate the BoostConfig.cmake file. I have virtually no experience using CMake on Windows, so I could not really give an educated guess on what the underlying problem might be.
- On macOS, a Boost-related linking error pops up not while building AdaptiveCpp itself but while building the CPU tests (whose CMake file I notably have not even touched).

I suppose the easiest solution would be to just stick to the old behaviour by setting `CMP0167` to `OLD` and hope that by the time CMake decides to remove FindBoost properly, these issues will have sorted themselves out.

Your opinion, @illuhad?

CMakeLists.txt
```diff
@@ -1,6 +1,11 @@
 cmake_minimum_required(VERSION 3.10)
-if(NOT CMAKE_VERSION VERSION_LESS 3.12)
+
+if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.12)
   cmake_policy(SET CMP0074 NEW) # Don't complain about using BOOST_ROOT...
+  if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.30)
+    # Suppress warning: "Policy CMP0167 is not set: The FindBoost module is removed."
+    cmake_policy(SET CMP0167 NEW)
+  endif()
 endif()
```
al42and: FYI, CMake supports `if(POLICY policy-id)` syntax to directly check whether a policy is available. I think it's a bit more straightforward and less error-prone than just checking the CMake version, but up to you"

```suggestion
if(POLICY CMP0074) # Since CMake 3.12
  cmake_policy(SET CMP0074 NEW) # Don't complain about using BOOST_ROOT...
endif()
if(POLICY CMP0167) # Since CMake 3.30
    # Suppress warning: "Policy CMP0167 is not set: The FindBoost module is removed."
    cmake_policy(SET CMP0167 NEW)
endif()
```

marcosolanki: Ah, that is quite a bit more elegant. I will definitely be using this. Thanks! 👍

illuhad:
> I suppose the easiest solution would be to just stick to the old behaviour by setting CMP0167 to OLD and hope that by the time CMake decides to remove FindBoost properly, these issues will have sorted themselves out.

Yep. IIRC, there was a lot complexity involved in getting boost to work properly.

fodinabor:
> On Windows, for whatever reason, CMake fails to locate the BoostConfig.cmake file. I have virtually no experience using CMake on Windows, so I could not really give an educated guess on what the underlying problem might be. 

On windows I was quite successful setting `-DBoost_DIR=<path/to/boost/lib/cmake/Boost-1.83.0` instead, lately.

marcosolanki:
OK, I have changed the CMakeLists.txt so that `CMP0167` is set to `OLD` by default. The CI tests should now all pass.

_([Kate](https://kate-editor.org) has also done its part by eliminating all semantically superfluous trailing whitespace from the file...)_

marcosolanki:
> On windows I was quite successful setting `-DBoost_DIR=<path/to/boost/lib/cmake/Boost-1.83.0` instead, lately.

Hmmm, in that case, I suppose we should consider explicitly setting `-DBoost_DIR` in our Windows CI test and see whether that helps (although, either way, we probably should not set `CMP0167` to `NEW` in our CMakeLists.txt, as this policy value seems to also cause issues on macOS).

illuhad:
Thanks, I've rerun workflows - fingers crossed that it works this time :)

I've also sent you an invitation to join the AdaptiveCpp github organization. This is not tied to any expectations, but will allow workflows from your PRs to run automatically.

The background here is that we also have self-hosted CI runners on our own infrastructure, so we  have to be careful about running workflows from PRs of a public project like AdaptiveCpp, since this would allow someone to run arbitrary code on our machines. Because of this, it only runs workflows by default from PRs of members of the AdaptiveCpp organization.
But I don't want make organization membership exclusive, so I happily invite frequent contributors that I know :)

If you join, you also have privileges to review other PRs and approve running CI workflows. Please feel free to do that, but before you approve CI workflows, please have a look at the code to check that there's nothing obviously bogus and potentially malicious inside :)

illuhad (approved):
Thanks!

illuhad:
> Hmmm, in that case, I suppose we should consider explicitly setting -DBoost_DIR in our Windows CI test and see whether that helps (although, either way, we probably should not set CMP0167 to NEW in our CMakeLists.txt, as this policy value seems to also cause issues on macOS).

Feel free to play around here, and potentially create another PR :) I think this one is ready to merge as it seems to solve the most pressing issue with the warnings.

PR #1625, opened 2024-12-02
Fix pointer to a temporary returned from kernel_configuration::data_ptr

al42and:
`kernel_configuration::data_ptr` was creating a temporary string and returning a pointer to it.

The change to `data_size` is unrelated, just avoiding extra copies there.


illuhad (approved):
Good catch, thanks!

PR #1633, opened 2024-12-07
replace some c headers by std c++ headers

tdavidcl:
I was compiling acpp on some tight nixos configuration, which resulted in some c headers not being accessible even though c++ headers were working. This highlighted several c std includes that could be replaced by their c++ counterpart. 
This PR does exactly that.


illuhad:
I think for the bitcode libraries, we deliberately decided that we didn't want to use the C++ headers. C++ standard library headers may pull in some system configuration files which may fail to compile since effectively in the bitcode headers we are kind of in a cross-compilation scenarion.
For the `host` bitcode libraries this is somewhat different, but it could still be argued that we should align it with the others. @fodinabor what is your opinion?

But good catch with the group functions header! :)

fodinabor:
For host, I don't see any strong reason for one or the other at the moment... using these C++ headers should be okay, as in most scenarios, they're mostly wrappers/pulling the math/io functions into the `std` namespace.

illuhad (approved):
Thanks!

PR #1636, opened 2024-12-07
Build Clang plugin as module

VaiTon:
Otherwise `-Wl,--no-undefined` will make the build fail

## To test

From `develop` do
```shell
cmake -B build -DCMAKE_CXX_FLAGS=-Wl,--no-undefined
cmake --build build -j8
```

### DIFF
--- a/bin/acpp
+++ b/bin/acpp
@@ -745,8 +745,6 @@ class acpp_config:
   def acpp_plugin_path(self):
     if sys.platform.startswith('win32'):
       return os.path.join(self.acpp_installation_path, "bin", "acpp-clang.dll")
-    elif sys.platform == "darwin":
-      return os.path.join(self.acpp_installation_path, "lib", "libacpp-clang.dylib")
     else:
       return os.path.join(self.acpp_installation_path, "lib", "libacpp-clang.so")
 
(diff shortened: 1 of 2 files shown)


fodinabor:
This will break macos (changes the plugin from dylib to so)

(Rn we don't use the plugin in the CI, but it is possible to use the accelerated flow, would need to adjust the path in bin/acpp to remedy)

VaiTon:
@fodinabor what about now?

fodinabor:
I was more thinking of fixing `bin/acpp` to use so instead of dylib...

VaiTon:
@fodinabor unfortunately I don't know how to do so. Would you like to do it and for me to close this PR?

fodinabor:
https://github.com/AdaptiveCpp/AdaptiveCpp/blob/2d47fb0d533db54dbba3bf803f3997ac95929241/bin/acpp#L749
Just fix this line :-)

VaiTon:
@fodinabor should be done!

fodinabor (approved):
LGTM as far as I can tell without testing :)
Thanks.

@illuhad: in general, I believe it'd also be correct to do this for all the other loadables (backends & llvm-to-X).

illuhad:
> @illuhad: in general, I believe it'd also be correct to do this for all the other loadables (backends & llvm-to-X).

llvm-to-backend is not loaded with `dlopen()`, we link them at build time to rt backend plugins if SSCP is available:
https://github.com/AdaptiveCpp/AdaptiveCpp/blob/bf218d9b835efe2ac3c609a10abb1b997b00b116/src/runtime/CMakeLists.txt#L128

but I can see it for the runtime backend plugins.

illuhad (approved):

PR #1641, opened 2024-12-10
Add acpp namespace alias

illuhad:
This PR adds aliases for the current `hipsycl` namespace. This is in preparation for moving away from the `hipsycl` namespace in the longer term.
These aliases are available in both SYCL and stdpar mode.

The discord poll has resulted in a preference for the `acpp` namespace. However, the point was raised that fixing namespace collisions is fairly difficult, and more likely for a shorter namespace name like `acpp`.

Therefore, I propose the following strategy to address both arguments:

* Once we migrate away from the `hipsycl` namespace, the code will be declared in the `adaptivecpp` namespace.
* An alias will be available and active by default to also expose our functionality in the `acpp` namespace.
* In documentation and examples, we will primarily use the shorter `acpp` alias.
* In case of namespace collision, users can set a macro to disable the shorthand `acpp` namespace alias, and use the `adaptivecpp` namespace instead.

In light of this strategy, this PR introduces both `acpp` and `adaptivecpp` namespaces as aliases, with the macro to disable the `acpp` namespace.
The `adaptivecpp` alias will be replaced in the future by renaming the `hipsycl` namespace to `adaptivecpp`.

@al42and Are you happy with this approach?


al42and:
> @al42and Are you happy with this approach?

Yes. The current solution combines the best of both worlds :+1: 

All in all, I would not be too sad if we went with `acpp` namespace only. I was the one who raised the point of namespace clash and I _do_ think it's useful to consider this possibility, that's purely hypothetical at this point. 

Unrelated: why limit the new namespace to `<algorithm/X>` headers? There are some internal details that can be helpful for applications (such as querying `hipsycl::rt::settings`) which are currently only available in `hipsycl` namespace.

illuhad:
> All in all, I would not be too sad if we went with acpp namespace only. I was the one who raised the point of namespace clash and I do think it's useful to consider this possibility, that's purely hypothetical at this point.

Okay. I guess while we are still just working with aliases, we can always revisit the question.

> Unrelated: why limit the new namespace to <algorithm/X> headers? There are some internal details that can be helpful for applications (such as querying hipsycl::rt::settings) which are currently only available in hipsycl namespace.

It's not :-) The alias header is also pulled in when including `sycl.hpp`. The idea is to provide an alternative to the old `hipsycl` namespace everywhere.

A related issue that is not yet solved is that currently the algorithm and other internal headers are only in the `hipSYCL` directory. So users will have to e.g. `#include <hipSYCL/algorithms/numeric.hpp>`. A directory symlink would also be needed to allow users to use functionality like the algorithms without using something with `hipSYCL` somewhere.

al42and:
> It's not :-) The alias header is also pulled in when including sycl.hpp

Right, missed that it's included there too. :+1: 

> A related issue that is not yet solved is that currently the algorithm and other internal headers are only in the hipSYCL directory. So users will have to e.g. #include <hipSYCL/algorithms/numeric.hpp>. A directory symlink would also be needed to allow users to use functionality like the algorithms without using something with hipSYCL somewhere.

Why not make the new directory and install the new headers only there? The longer the old name lives the harder it will be to change eventually. And not adding the new functionality to the old headers will not break any backwards-comptibility.

This will not work with directory symlinks and will require installing two copies of the same file, but I don't think that the doubling the disk space used by headers would be a big problem.

illuhad:
> Why not make the new directory and install the new headers only there? The longer the old name lives the harder it will be to change eventually. And not adding the new functionality to the old headers will not break any backwards-comptibility.
> This will not work with directory symlinks and will require installing two copies of the same file, but I don't think that the doubling the disk space used by headers would be a big problem.

I'm not primarily concerned about backwards compatibility. Many of our headers do have the `hipSYCL/*` directory in their `#include` statements. Removing the old directory would therefore require changes to a lot of files, which I'd like to avoid this shortly before release.
It might be possible to copy the headers to two locations for now. I did some quick cmake experiments around this which didn't quite work out, but I did not spend more than a couple of minutes trying this.

PR #1642, opened 2024-12-10
[SSCP] Provide `sycl::buffer` device pointers to custom operations

feltech:
If a kernel passed to `AdaptiveCpp_enqueue_custom_operation` makes use of (i.e. captures) a `sycl::buffer`, then the captured buffer's internal pointer must point to its associated device memory, in order for the custom kernel to extract the pointer and provide to e.g. `cudaMemcpyAsync`.

The transformation from host pointer to device pointer is done at kernel launch time by rewriting the bytes of the kernel's captured variables using an `initialize_embedded_pointers` function. However, this function was not being called for custom operations enqueued using `AdaptiveCpp_enqueue_custom_operation`.

So decorate the custom kernel with a call to `initialize_embedded_pointers`, before calling the kernel itself.

This is required for e.g. oneMKL, which makes use of `AdaptiveCpp_enqueue_custom_operation` (well, `hipSYCL_enqueue_custom_operation`) to call cuBLAS functions, with the device memory pointer coming from a `sycl::buffer` (via `sycl::interop_handle::get_native_mem<...>(...)`

There is a unit test of `AdaptiveCpp_enqueue_custom_operation` (`custom_enqueue`) checking this case. However, I found that it was being skipped when running the suite locally. It seems there is a logic flaw in the macros that conditionally compile the test. After tweaking the macros and running the test, it does indeed fail without the above fix.

Unfortunately I'm currently limited in local testing to NVIDIA and to SSCP. Setting the target to either `cuda` or `cuda.explicit-multipass` causes compiler crashes (hopefully just a local config issue). 

This means I don't actually know whether the `custom_enqueue` test runs using other CUDA targets; and whether my changes to the CMake/macro config makes sense for other CUDA targets; and whether similar changes are needed for ROCM/HIP targets...

### DIFF
--- a/include/hipSYCL/glue/llvm-sscp/sscp_kernel_launcher.hpp
+++ b/include/hipSYCL/glue/llvm-sscp/sscp_kernel_launcher.hpp
@@ -295,7 +295,12 @@ public:
       
     } else if constexpr (type == rt::kernel_type::custom) {
       // handled at invoke time
-      data.custom_op = k;
+      data.custom_op = [k](rt::kernel_operation * kernel_op, sycl::interop_handle& ih) mutable {
+          kernel_op->initialize_embedded_pointers(
+              static_cast<void*>(&k),
+              sizeof(Kernel));
+          k(ih);
+       };
     }
     else {
       assert(false && "Unsupported kernel type");
@@ -313,8 +318,9 @@ public:
       assert(backend_params);
       sycl::interop_handle handle{node->get_assigned_device(),
                                   backend_params};
-
-      launch_config.custom_op(handle);
+      auto *kernel_op =
+          static_cast<rt::kernel_operation *>(node->get_operation());
+      launch_config.custom_op(kernel_op, handle);
 
       return rt::make_success();
     } else {
--- a/include/hipSYCL/glue/kernel_launcher_data.hpp
+++ b/include/hipSYCL/glue/kernel_launcher_data.hpp
@@ -31,6 +31,7 @@ class dag_node;
 class kernel_configuration;
 class backend_kernel_launch_capabilities;
 class hcf_kernel_info;
+class kernel_operation;
 }
 
 namespace glue {
@@ -53,7 +54,7 @@ struct kernel_launcher_data {
   rt::range<3> group_size; // <- indices must be flipped
   unsigned local_mem_size;
   // In case the launch is a custom operation
-  std::function<void(sycl::interop_handle&)> custom_op;
+  std::function<void(rt::kernel_operation*, sycl::interop_handle&)> custom_op;
 
   using invoker_function_t = rt::result (*)(
       const kernel_launcher_data &launch_config, rt::dag_node *node,


include/hipSYCL/glue/llvm-sscp/sscp_kernel_launcher.hpp
```diff
@@ -295,7 +295,12 @@ class sscp_kernel_launcher
       
     } else if constexpr (type == rt::kernel_type::custom) {
       // handled at invoke time
-      data.custom_op = k;
+      data.custom_op = [k](rt::kernel_operation * kernel_op, sycl::interop_handle& ih) mutable {
+          kernel_op->initialize_embedded_pointers(
+              static_cast<void*>(&k),
+              sizeof(Kernel));
+          k(ih);
```
illuhad, line 302: Why do this lambda wrapping exercise? Why not initialize the embedded pointers in `invoke()` a couple lines further down (before line 317), where this is also done for non-custom-op kernel invocations (see lines 327-331)?

tests/CMakeLists.txt
```diff
@@ -110,6 +110,19 @@ add_executable(sycl_tests
 # Also test instant submission mode
 target_compile_definitions(sycl_tests PRIVATE -DACPP_ALLOW_INSTANT_SUBMISSION=1)
 target_include_directories(sycl_tests PRIVATE ${Boost_INCLUDE_DIRS} ${CMAKE_CURRENT_SOURCE_DIR} ${OpenMP_CXX_INCLUDE_DIRS})
+
+if (ACPP_CUDA_BACKEND_AVAILABLE AND (NOT ACPP_TARGETS OR ACPP_TARGETS MATCHES "cuda" OR ACPP_TARGETS MATCHES "generic"))
+  message(STATUS "CUDA enabled - searching for CUDA toolkit")
+  find_package(CUDAToolkit QUIET)
+  if (CUDAToolkit_FOUND)
+    target_link_libraries(sycl_tests PRIVATE CUDA::cudart)
```
illuhad: I'm not really convinced this is the right approach. I understand that you want to test whether interop with custom operation works with CUDA, and I appreciate the effort :)

But the whole point of `generic` is to decouple the generated binary from backends in order to have backend-independent binaries. The code should be able to run (minus interop, of course) independently of whether the binary is directly linked against CUDA or not.

This PR would change our CI to no longer test this if a CUDA toolkit is found. In this case, CUDA would be linked directly to the application, and we would be unable to detect symbol creep from the CUDA backend into the application.

So, for SSCP I think we should at most make this an optional test, if we want to have it at all. Testing both cases would require building the tests twice - once with CUDA linked and once without, and I'm not sure this double stress on CI is worth it.
Also, I'm not sure why we should treat CUDA differently here than say, HIP (where we have even less CI resources).

Maybe we should have some dedicated interop test suite which could then link against backend libraries as much as it wants. This test suite could be fairly small, and hence only require little resources and be run on all backends that support interop.

This change might be out of scope for this PR, so I would be happy to maintain the status quo of interop testing with SSCP for now until we do it properly in a future PR. Would this be okay for you?

illuhad (commented):
Good catch :) This looks like it might have taken quite some debugging time, thank  you :)

include/hipSYCL/glue/llvm-sscp/sscp_kernel_launcher.hpp
feltech: Ah yes. I struggled due to the opaque way `std::function` allocates space for its stored callable.  I.e. given a `std::function`, how can we retrieve a pointer (and size) to pass to `initialize_embedded_pointers`?  I did try a few approaches (it was a few weeks ago now so the details are fuzzy), but kept hitting segfaults and other nasty errors.

tests/CMakeLists.txt
feltech: Totally on board. Honestly, the large number of macros and defines and targets and their interaction still confuses me. 

> I understand that you want to test whether interop with custom operation works with CUDA

Indeed, well, I was surprised to learn there was already a test for this, but it wasn't running. I suspect there's something I'm just not understanding here, but..

As best I can tell, before this PR the tests would never run, since the macro `ACPP_LIBKERNEL_COMPILER_SUPPORTS_CUDA` is basically the same as `__CUDACC__`. 

The [CUDA docs say](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/)

> Defined when compiling CUDA source files

The [Clang docs say](https://llvm.org/docs/CompileCudaWithLLVM.html)

> Both clang and nvcc define __CUDACC__ during CUDA compilation

Both of which to my mind are ambiguous, but I assume mean compiling `.cu` files regardless of host or device pass.  So the `custom_enqueue` tests would have only run if treating `.cpp` files as `.cu` files?  Perhaps this is the case for the `cuda` targets of `acpp`, but not for the `generic` target, hence my confusion?


> Also, I'm not sure why we should treat CUDA differently here than say, HIP (where we have even less CI resources).

Yeah I was reluctant to touch HIP since I have no way to develop/test this locally.


> I would be happy to maintain the status quo of interop testing with SSCP for now until we do it properly in a future PR. Would this be okay for you?

This is fine.  So to be clear, I can drop all commits apart from the last one?  This will mean my change is not tested (yet), at least with the `generic` target.

feltech:
> Good catch :) This looks like it might have taken quite some debugging time, thank you :)

:smile: Indeed it was an epic adventure into the internals of AdaptiveCpp, in a quest to make oneMKL work with it (via cuBLAS).

tests/CMakeLists.txt
illuhad: > Totally on board. Honestly, the large number of macros and defines and targets and their interaction still confuses me.

Yeah, flexibility inevitably pulls in complexity. There's no other SYCL compiler that supports as many compilation flows as AdaptiveCpp.

> Indeed, well, I was surprised to learn there was already a test for this, but it wasn't running. I suspect there's something I'm just not understanding here, but..
> As best I can tell, before this PR the tests would never run, since the macro ACPP_LIBKERNEL_COMPILER_SUPPORTS_CUDA is basically the same as __CUDACC__.

Yes. `ACPP_LIBKERNEL_COMPILER_SUPPORTS_CUDA` basically tests whether we are in `--acpp-targets=cuda`, *not* `--acpp-targets=generic`.
This might have been deliberate to avoid the the questions around linking of `libcudart` that you discovered :)

> Perhaps this is the case for the cuda targets of acpp, but not for the generic target, hence my confusion?

Yes, exactly. `--acpp-targets=cuda` runs through the clang CUDA toolchain, so in this case you are effectively compiling your SYCL code with a CUDA compiler in disguise that was augmented by AdaptiveCpp to also understand SYCL. The same is true for `--acpp-targets=hip` and the clang HIP toolchain respectively.

On the other hand, with `--acpp-targets=generic` everything is controlled by AdaptiveCpp and there are no other programming models involved.

The `cuda`/`hip` targets integrate tightly with backend models CUDA/HIP and in particular also support mix-and-match of CUDA/HIP code with SYCL. As such, they are primarily useful when migrating code.

The advantage of `generic` is that, since it's completely under the control of AdaptiveCpp, it offers much more features and AdaptiveCpp-specific performance optimizations. Since this compiler is a JIT compiler with a backend-unified code representation, binary portability is much better. After all, every binary can dispatch to all backends depending on what hardware it finds in the system at runtime.

> Yeah I was reluctant to touch HIP since I have no way to develop/test this locally.

Understandable. We do have AMD hardware in CI, but of course iterating on code while relying solely on CI for testing is cumbersome.

> This is fine. So to be clear, I can drop all commits apart from the last one? This will mean my change is not tested (yet), at least with the generic target.

Yes, this would be my suggestion for now.

include/hipSYCL/glue/llvm-sscp/sscp_kernel_launcher.hpp
illuhad: Ah, I missed that, right. Hm... for the other kernel invocation types this is solved by storing the lambda in the kernel argument fields. However, in these cases, the lambda never needs to be invoked since we only care about its members to pass them in as argument for the kernel.

Doing something similar for custom ops might be equally, if not more complicated than what you did. So I guess your conclusion is probably the right one :)

tests/CMakeLists.txt
feltech: That all makes total sense, thanks for the detailed explanation.

I've dropped the other commits, rebased on `develop`, and re-worded the remaining commit message to elaborate on the questions raised here.

illuhad (approved):
Thanks!

PR #1643, opened 2024-12-11
[SSCP][llvm-to-host] Support internal local memory for group algorithms

illuhad:
This PR adds support for an additional section of dedicated local memory to be passed into the kernel for internal use, particularly for the implementation of group algorithms.

It turned out to not be possible to allocate local memory as `thread_local` memory inside the bitcode libraries due to TLS limitations.

While it might have been possible to use the existing local memory and set aside some part of that allocation for group algorithms, this would have introduced some additional, potentially undesired, offset calculations when accessing local memory.

Therefore, this PR solves this at the compiler level with an additional local memory builtin for internal use.

### DIFF
--- a/src/runtime/omp/omp_queue.cpp
+++ b/src/runtime/omp/omp_queue.cpp
@@ -186,14 +186,36 @@ std::size_t get_page_size() {
 #endif
 }
 
+void *resize_and_align(std::vector<char> &data, std::size_t size,
+                       std::size_t alignment) {
+  data.resize(size + alignment);
+  return reinterpret_cast<void*>(
+        next_multiple_of(reinterpret_cast<std::uint64_t>(data.data()),
+                         alignment));
+}
+
+void *resize_and_strongly_align(std::vector<char> &data, std::size_t size) {
+  // compiler/libkernel builtins assume alignment of at least
+  // 512 byte boundaries
+  std::size_t alignment = std::max(std::size_t{512}, get_page_size());
+  return resize_and_align(data, size, alignment);
+}
+
 result
 launch_kernel_from_so(omp_sscp_executable_object::omp_sscp_kernel *kernel,
                       const rt::range<3> &num_groups,
                       const rt::range<3> &local_size, unsigned shared_memory,
                       void **kernel_args) {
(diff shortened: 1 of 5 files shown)


src/runtime/omp/omp_queue.cpp
```diff
@@ -210,23 +232,20 @@ launch_kernel_from_so(omp_sscp_executable_object::omp_sscp_kernel *kernel,
...
-    local_memory.resize(shared_memory + local_mem_alignment);
-    auto aligned_local_memory = reinterpret_cast<void *>(
-        next_multiple_of(reinterpret_cast<std::uint64_t>(local_memory.data()),
-                         local_mem_alignment));
-
+    static thread_local std::vector<char> internal_local_memory;
+    auto aligned_local_memory =
+        resize_and_strongly_align(local_memory, shared_memory);
+    auto aligned_internal_local_memory = resize_and_strongly_align(
+        internal_local_memory, local_size.size() * sizeof(uint64_t));
```
fodinabor, line 239: Just to confirm if the size is sufficient?
In OpenMP we had: 128*1024 bytes (IIRC, this was due to `vec<double,16>`). We won't need this here?

fodinabor (commented):
LGTM so far.
I'm wondering, if there's ways to inform LLVM that these pointers are noalias and dereferencable. The latter, I gues would only work, once @moritz-2001 sends in the WG specialization patch :)

src/runtime/omp/omp_queue.cpp
illuhad: Yes, this will be fine, at least for now. SSCP group algorithms operate component-wise, so we will never need more than 64-bit per element (well, unless `int128_t` becomes a proper thing in SYCL). The motivation here is that we also need to support `marray` which has arbitrary number of components - so it makes sense to just handle non-scalar types component-wise.

illuhad:
> I'm wondering, if there's ways to inform LLVM that these pointers are noalias and dereferencable

I guess we could add some compiler logic to add `noalias` attributes. Perhaps it's already sufficient to add `__restrict__` to the return type of the builtins?

PR #1645, opened 2024-12-11
Add `is_device_copyable`

carbotaniuman:
This is added for compatibility with libraries that expect this type trait. I decided to not make it always true in case that breaks less code.

### DIFF
--- a/include/hipSYCL/sycl/is_device_copyable.hpp
+++ b/include/hipSYCL/sycl/is_device_copyable.hpp
@@ -1,0 +1,61 @@
+/*
+ * This file is part of AdaptiveCpp, an implementation of SYCL and C++ standard
+ * parallelism for CPUs and GPUs.
+ *
+ * Copyright The AdaptiveCpp Contributors
+ *
+ * AdaptiveCpp is released under the BSD 2-Clause "Simplified" License.
+ * See file LICENSE in the project root for full license details.
+ */
+// SPDX-License-Identifier: BSD-2-Clause
+#ifndef HIPSYCL_IS_DEVICE_COPYABLE_HPP
+#define HIPSYCL_IS_DEVICE_COPYABLE_HPP
+
+#include <array>
+#include <optional>
+#include <utility>
+#include <tuple>
+#include <type_traits>
+
+// AdaptiveCPP does not use this type trait to restrict allowed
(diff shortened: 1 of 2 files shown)


illuhad:
Thank you!

Have you checked whether it actually works with code that expects to specialize it?

SYCL is very vague w.r.t how its `sycl` namespace works. In AdaptiveCpp, `sycl` lives in the namespace `hipsycl` (`adaptivecpp` in future versions). The SYCL header then sets an alias.
With this implementation, I'm not sure if it would work if users attempted to specialize like so:

```c++
namespace sycl {

template<>
struct is_device_copyable<MyType> : public std::true_type {};
} //sycl
```
because the declaration of `is_device_copyable` does not live in `::sycl`.
I suspect you may have to declare it in `::sycl` instead of `::hipsycl::sycl`, which would be a bit unfortunate since it would be the only thing that we pollute the global `::sycl` namespace with.

carbotaniuman:
Ah, this was tested on specializations of the form:

```
template<>
struct sycl::is_device_copyable<MyType> : std::true_type {};
```
which is what oneDPL uses, which does work.

The form you provided does not work and would indeed require us to pollute the `::sycl` namespace.

illuhad:
Interesting. DPC++ seems to declare it in `sycl::_V1`: https://github.com/intel/llvm/blob/a1355e8717c4c10113b9f3b0e2e214e02c70c421/sycl/include/sycl/detail/is_device_copyable.hpp#L46

Not sure if it makes a difference that `_V1` is an inline namespace, but it's quite possible then that DPC++ has the same issue.

The form that you mentioned is what is used as example by the spec: https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_is_device_copyable_type_trait

However, I'm not seeing any mention in the spec that the form I used is illegal, and the synopsis for `is_device_copyable` suggests that it should be declared in `::sycl`.

As a first stab at this, we could merge your implementation, and simultaneously ask for clarification of the spec.
There's no SYCL WG meeting scheduled before january, but we could open an issue in `KhronosGroup/SYCL-Docs` to get some opinions from other people from the SYCL working group.

carbotaniuman:
I'm pretty sure the inline namespace allows the DPC++ implementation to work because that's the primary use-case for them, and it does work with that form in testing.

PR #1648, opened 2024-12-11
Print JIT compile time, cleanup kernel_cache.hpp

VaiTon:
- Put `#include` after header guard
- Collapse namespaces
- Remove unneeded namespace specification (rt::)
- Add debug log with JIT timing
- [Use SFINAE for get_or_construct_jit_code_object](https://github.com/AdaptiveCpp/AdaptiveCpp/pull/1648/commits/62772c94e6a2aaaddf5b280db7944b6ae7a8b057)

### DIFF
--- a/include/hipSYCL/runtime/kernel_cache.hpp
+++ b/include/hipSYCL/runtime/kernel_cache.hpp
@@ -318,9 +318,19 @@ public:
     // TODO: We might want to allow JIT compilation in parallel at some point
     std::lock_guard<std::mutex> lock{_mutex};
 
-    if(!persistent_cache_lookup(id_of_binary, compiled_binary)){
+    if(!persistent_cache_lookup(id_of_binary, compiled_binary)) {
+      HIPSYCL_DEBUG_INFO << "kernel_cache: JIT-compiling binary for id "
+                         << kernel_configuration::to_string(id_of_binary)
+                         << "\n";
+
+      const auto start_time = std::chrono::high_resolution_clock::now();
       if(!jit_compile(compiled_binary))
         return nullptr;
+      const auto elapsed_time = std::chrono::high_resolution_clock::now() - start_time;
+
+      HIPSYCL_DEBUG_INFO << "kernel_cache: JIT compilation took "
+                         << std::chrono::duration_cast<std::chrono::milliseconds>(elapsed_time).count()
+                         << "ms\n";
 
       if(_is_first_jit_compilation) {
         _is_first_jit_compilation = false;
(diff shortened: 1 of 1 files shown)


illuhad:
Can you please separate all style changes from functional changes? It gets difficult to figure out where any functional changes are.

Overall, I'm a bit hesitant as far as PRs are concerned that apply some style in non-localized manner to entire files. Tools like `clang-format` and similar are great, but they are not consistent across different LLVM versions. And since AdaptiveCpp supports many LLVM versions, chances are everybody uses something else.

Note that there's also a PR #912 which also enables timing information for S2 (jit) compilation at the LLVM level with a bit finer granularity.

VaiTon:
> Note that there's also a PR https://github.com/AdaptiveCpp/AdaptiveCpp/pull/912 which also enables timing information for S2 (jit) compilation at the LLVM level with a bit finer granularity.

Oh that's great! But AFAIS it's printed only for CUDA, HIP and ZE. I'd like to see the total time it took on each queue type (OMP, OpenCL, ...). 

I'll factor this PR and remove the generalized style changes.

VaiTon:
@illuhad done, review by commit should be more effective now

illuhad:
>  Oh that's great! But AFAIS it's printed only for CUDA, HIP and ZE. I'd like to see the total time it took on each queue type (OMP, OpenCL, ...).

Ah, I remembered the PR incorrectly, apologies. We do something similar for stage 1 compilation inside the compiler, see e.g. here https://github.com/AdaptiveCpp/AdaptiveCpp/blob/36281ee925b297cc2729abf08d3db363f6e1b969/src/compiler/sscp/TargetSeparationPass.cpp#L439
so if you run `ACPP_DEBUG_LEVEL=3 acpp -o test test.cpp` the compiler will show you the overall time of the AdaptiveCpp specific part of compilation, as well as more fine-grained timings.

I thought the PR I mentioned did something similar, but apparently that's not the case. Nevertheless it might be worthwhile to consider whether we want to do JIT timing inside the `llvm-to-backend` infrastructure (particularly `LLVMToBackend.cpp`, which is the main driver) for more fine-grained information, similarly to what happens at compile time already.

EDIT: Another, smaller component of JIT-time that is not contained in your measurements and also would not be contained in the measurements if we did it at llvm-to-backend level is the time it actually takes to construct the code object. (this would correspond to measuring the code object constructor).

On backends that rely on ingesting an IR, loading a code object will trigger another step of JIT inside drivers and can also be noticable. This currently only affects PTX and SPIR-V.
In the future we might also want to look into persistently caching this where possible (AFAIK e.g not every OpenCL implementation supports this), but we don't do this yet.

VaiTon:
I'm going to decompose this PR

PR #1651, opened 2024-12-12
[SSCP] Add support for group algorithms

### LINKED ISSUE #1432: Fail to run kernels with the JIT compiler
**Bug summary**
A simple application compiles but fails to run with the JIT compiler. AOT compilation with CUDA results in a working executable.
I'm afraid I've missed something when building acpp, but I can't figure out what's wrong.
The CPU through OpenCL works, while the CUDA backend, the Intel GPU (with both OpenCL and Level Zero), and the host device fail.

CUDA output:
```
Running on Tesla T4
[AdaptiveCpp Error] from /data/user/aperego/AdaptiveCpp/src/runtime/cuda/cuda_code_object.cpp:96 @ build_cuda_module_from_ptx(): cuda_executable_object: could not load module (error code = CU:218)
[AdaptiveCpp Error] from /data/user/aperego/AdaptiveCpp/src/runtime/cuda/cuda_queue.cpp:692 @ submit_sscp_kernel_from_code_object(): cuda_queue: Code object construction failed
============== hipSYCL error report ==============
hipSYCL has caught the following undhandled asynchronous errors:

   0. from /data/user/aperego/AdaptiveCpp/src/runtime/cuda/cuda_code_object.cpp:96 @ build_cuda_module_from_ptx(): cuda_executable_object: could not load module (error code = CU:218)
   1. from /data/user/aperego/AdaptiveCpp/src/runtime/cuda/cuda_queue.cpp:692 @ submit_sscp_kernel_from_code_object(): cuda_queue: Code object construction failed
The application will now be terminated.
terminate called without an active exception
Aborted (core dumped)
```

Intel GPU (OpenCL):
```
Running on Intel(R) Data Center GPU Flex 170
[AdaptiveCpp Error] from /data/user/aperego/AdaptiveCpp/src/runtime/ocl/ocl_code_object.cpp:90 @ ocl_executable_object(): ocl_code_object: Building CL program failed. Build log: error : unresolved external symbol puts at offset 244 in instructions segment #0 (aka kernel : _Z21__hipsycl_sscp_kernelIN7hipsycl4glue15__sscp_dispatch20ndrange_parallel_forIZZ4mainENK3$_0clERNS0_4sycl7handlerEEUlNS5_7nd_itemILi1EEEE_Li1EEEEvRKT_)
error : unresolved external symbol puts at offset 356 in instructions segment #0 (aka kernel : _Z21__hipsycl_sscp_kernelIN7hipsycl4glue15__sscp_dispatch20ndrange_parallel_forIZZ4mainENK3$_0clERNS0_4sycl7handlerEEUlNS5_7nd_itemILi1EEEE_Li1EEEEvRKT_)
 (error code = CL:-42)
[AdaptiveCpp Error] from /data/user/aperego/AdaptiveCpp/src/runtime/ocl/ocl_code_object.cpp:90 @ ocl_executable_object(): ocl_code_object: Building CL program failed. Build log: error : unresolved external symbol puts at offset 244 in instructions segment #0 (aka kernel : _Z21__hipsycl_sscp_kernelIN7hipsycl4glue15__sscp_dispatch20ndrange_parallel_forIZZ4mainENK3$_0clERNS0_4sycl7handlerEEUlNS5_7nd_itemILi1EEEE_Li1EEEEvRKT_)
error : unresolved external symbol puts at offset 356 in instructions segment #0 (aka kernel : _Z21__hipsycl_sscp_kernelIN7hipsycl4glue15__sscp_dispatch20ndrange_parallel_forIZZ4mainENK3$_0clERNS0_4sycl7handlerEEUlNS5_7nd_itemILi1EEEE_Li1EEEEvRKT_)
 (error code = CL:-42)
[AdaptiveCpp Error] from /data/user/aperego/AdaptiveCpp/src/runtime/ocl/ocl_queue.cpp:513 @ submit_sscp_kernel_from_code_object(): ocl_queue: Code object construction failed
============== hipSYCL error report ==============
hipSYCL has caught the following undhandled asynchronous errors:

   0. from /data/user/aperego/AdaptiveCpp/src/runtime/ocl/ocl_code_object.cpp:90 @ ocl_executable_object(): ocl_code_object: Building CL program failed. Build log: error : unresolved external symbol puts at offset 244 in instructions segment #0 (aka kernel : _Z21__hipsycl_sscp_kernelIN7hipsycl4glue15__sscp_dispatch20ndrange_parallel_forIZZ4mainENK3$_0clERNS0_4sycl7handlerEEUlNS5_7nd_itemILi1EEEE_Li1EEEEvRKT_)
error : unresolved external symbol puts at offset 356 in instructions segment #0 (aka kernel : _Z21__hipsycl_sscp_kernelIN7hipsycl4glue15__sscp_dispatch20ndrange_parallel_forIZZ4mainENK3$_0clERNS0_4sycl7handlerEEUlNS5_7nd_itemILi1EEEE_Li1EEEEvRKT_)
 (error code = CL:-42)
   1. from /data/user/aperego/AdaptiveCpp/src/runtime/ocl/ocl_code_object.cpp:90 @ ocl_executable_object(): ocl_code_object: Building CL program failed. Build log: error : unresolved external symbol puts at offset 244 in instructions segment #0 (aka kernel : _Z21__hipsycl_sscp_kernelIN7hipsycl4glue15__sscp_dispatch20ndrange_parallel_forIZZ4mainENK3$_0clERNS0_4sycl7handlerEEUlNS5_7nd_itemILi1EEEE_Li1EEEEvRKT_)
error : unresolved external symbol puts at offset 356 in instructions segment #0 (aka kernel : _Z21__hipsycl_sscp_kernelIN7hipsycl4glue15__sscp_dispatch20ndrange_parallel_forIZZ4mainENK3$_0clERNS0_4sycl7handlerEEUlNS5_7nd_itemILi1EEEE_Li1EEEEvRKT_)
 (error code = CL:-42)
   2. from /data/user/aperego/AdaptiveCpp/src/runtime/ocl/ocl_queue.cpp:513 @ submit_sscp_kernel_from_code_object(): ocl_queue: Code object construction failed
The application will now be terminated.
terminate called without an active exception
Aborted (core dumped)
```

Host:
```
Running on hipSYCL OpenMP host device
[AdaptiveCpp Warning] [LoopSplitterInlining] puts is not defined!
/cvmfs/cms.cern.ch/el9_amd64_gcc11/external/gcc/11.4.1-30ebdc301ebd200f2ae0e3d880258e65/bin/ld: cannot find crtbeginS.o: No such file or directory
/cvmfs/cms.cern.ch/el9_amd64_gcc11/external/gcc/11.4.1-30ebdc301ebd200f2ae0e3d880258e65/bin/ld: cannot find -lstdc++: No such file or directory
clang-16: error: linker command failed with exit code 1 (use -v to see invocation)
[AdaptiveCpp Error] from /data/user/aperego/AdaptiveCpp/include/hipSYCL/glue/llvm-sscp/jit.hpp:265 @ compile(): jit::compile: Encountered errors:
0: LLVMToHost: clang invocation failed with exit code 1

[AdaptiveCpp Error] from /data/user/aperego/AdaptiveCpp/src/runtime/omp/omp_queue.cpp:494 @ submit_sscp_kernel_from_code_object(): omp_queue: Code object construction failed
============== hipSYCL error report ==============
hipSYCL has caught the following undhandled asynchronous errors:

   0. from /data/user/aperego/AdaptiveCpp/include/hipSYCL/glue/llvm-sscp/jit.hpp:265 @ compile(): jit::compile: Encountered errors:
0: LLVMToHost: clang invocation failed with exit code 1

   1. from /data/user/aperego/AdaptiveCpp/src/runtime/omp/omp_queue.cpp:494 @ submit_sscp_kernel_from_code_object(): omp_queue: Code object construction failed
The application will now be terminated.
terminate called without an active exception
Aborted (core dumped)
```
In this case, it cannot find `crtbeginS.o` and `-lstdc++` (at least the first is in the path set with the `--gcc-install-dir` flag).
**To Reproduce**
`test.cpp`:
```cpp
#include <sycl/sycl.hpp>

int main(int argc, char** argv) {
  static const std::vector<sycl::device> devices =
      sycl::device::get_devices(sycl::info::device_type::all);

  std::cerr << "Found " << devices.size() << " SYCL devices:" << std::endl;
  for (auto const &device : devices)
    std::cerr << "  - " << device.get_info<sycl::info::device::name>() << " ["
              << device.get_info<sycl::info::device::driver_version>() << "]"
              << std::endl;
  std::cerr << std::endl;
  if (argc < 2) {
    std::cout << "Provide the device number N: ./test N" << std::endl;
    return 1;
  }
  auto q = sycl::queue{devices[atoi(argv[1])]};
  std::cout << "Running on "
            << q.get_device().get_info<sycl::info::device::name>() << std::endl;

     cgh.parallel_for(sycl::nd_range<1>(1, 1),
                      [=](sycl::nd_item<1> item) { printf("Running the kernel\n"); });
   }).wait();

  return 0;
}
```
compile and run:
```
acpp --gcc-install-dir=/cvmfs/cms.cern.ch/el9_amd64_gcc11/external/gcc/11.4.1-30ebdc301ebd200f2ae0e3d880258e65/lib/gcc/x86_64-redhat-linux-gnu/11.4.1 -std=c++17 --acpp-targets="generic" -O2  -o test test.cpp
./test N # where N is the device's position in the list of available devices
```

**Expected behavior**
It should run without errors

**Describe your setup**
* AdaptiveCpp cloned from github: `develop@57773099 + PR 1031` and built with:
   ```
  cmake \
    -DCMAKE_INSTALL_PREFIX=$PWD \
    -DCUDA_TOOLKIT_ROOT_DIR=/usr/local/cuda \
    -DWITH_CUDA_BACKEND=ON \
    -DROCM_PATH=/opt/rocm \
    -DWITH_ROCM_BACKEND=ON \
    -DWITH_LEVEL_ZERO_BACKEND=ON \
    -DOpenCL_LIBRARY=/opt/intel/oneapi/compiler/2024.0/lib/libOpenCL.so \
    -DWITH_OPENCL_BACKEND=ON \
    -DLLVM_DIR=/cvmfs/cms.cern.ch/el9_amd64_gcc11/external/llvm/16.0.3-d5387186335b0dd85e1c294a1fd64dd0/lib64/cmake/llvm \
    ..
    ```
* ```
  clang version 16.0.3
  Target: x86_64-redhat-linux-gnu
  Thread model: posix
  InstalledDir: /cvmfs/cms.cern.ch/el9_amd64_gcc11/external/llvm/16.0.3-d5387186335b0dd85e1c294a1fd64dd0/bin
  Found CUDA installation: /usr/local/cuda-12.2, version
  Found HIP installation: /opt/rocm, version 6.0.32831
  ```
* `gcc (GCC) 11.4.1 20230601`
 
* from `syclcc --hipsycl-version` I see:
  ```bash
  AdaptiveCpp version: 24.02.0+git.dbaf585a.20240330.branch.develop.dirty
  Installation root: /data/user/aperego/AdaptiveCpp/build
  Plugin LLVM version: 16, can accelerate CPU: True
  Available runtime backends:
     librt-backend-omp.so
     librt-backend-cuda.so
     librt-backend-ze.so
     librt-backend-hip.so
     librt-backend-ocl.so
  ```

sbalint98:
Add support for the SYCL group algorithms. To enable the use of `__acpp_sscp_typed_get_global_linear_id<DIM>()` for builtin functions minor changes were required to the CBS infrastructure. @fodinabor  Could you have a look at that?


include/hipSYCL/sycl/libkernel/sscp/builtins/broadcast.hpp
```diff
@@ -29,6 +33,40 @@ HIPSYCL_SSCP_CONVERGENT_BUILTIN
 __acpp_int64 __acpp_sscp_work_group_broadcast_i64(__acpp_int32 sender,
                                                         __acpp_int64 x);
 
+template<typename T>
+T __acpp_sscp_sub_group_broadcast(T, __acpp_int32) = delete;
```
illuhad: This file is supposed to define the public builtin interface. Can you please separate internal content that you need for the implementation out from the public interface? E.g. create a `detail` subdirectory or something like that.

include/hipSYCL/sycl/libkernel/sscp/builtins/broadcast.hpp
```diff
@@ -29,6 +33,40 @@ HIPSYCL_SSCP_CONVERGENT_BUILTIN
...
+
+
+template<typename T>
+T __acpp_sscp_work_group_broadcast(__acpp_int32, T) = delete;
+
+#define TEMPLATE_DECLARATION_WG_BROADCAST(size) \
+template<> \
+__acpp_int##size __acpp_sscp_work_group_broadcast<__acpp_int##size>(__acpp_int32 id, __acpp_int##size value); \
+
+#define TEMPLATE_DEFINITION_WG_BROADCAST(size) \
```
illuhad: Please namespace macros with `ACPP_*` prefix - especially in header files.

include/hipSYCL/sycl/libkernel/sscp/builtins/broadcast.hpp
```diff
@@ -45,4 +83,38 @@ HIPSYCL_SSCP_CONVERGENT_BUILTIN
 __acpp_int64 __acpp_sscp_sub_group_broadcast_i64(__acpp_int32 sender,
                                                        __acpp_int64 x);
 
+
+#define SUBGROUP_BCAST(fn_suffix,input_type) \
+HIPSYCL_SSCP_CONVERGENT_BUILTIN \
+__acpp_##input_type __acpp_sscp_sub_group_broadcast_##fn_suffix(__acpp_int32 sender, \
+                                                     __acpp_##input_type x){ \
+    return __acpp_sscp_sub_group_select_##fn_suffix(x, sender); \
+                                                     } \
```
illuhad: Formatting looks off

include/hipSYCL/sycl/libkernel/sscp/builtins/broadcast.hpp
```diff
@@ -45,4 +83,38 @@ HIPSYCL_SSCP_CONVERGENT_BUILTIN
...
+T __acpp_sscp_work_group_broadcas_impl(__acpp_int32 sender, 
+                                                     T x){
+      #ifndef ACPP_SSCP_OMP_LIBKERNEL
+      ACPP_CUDALIKE_SHMEM_ATTRIBUTE T shrd_x[1];
+      #else
+      T* shrd_x = static_cast<T*>(__acpp_sscp_host_get_internal_local_memory());
+      #endif
+     
+
+     if(sender == __acpp_sscp_typed_get_local_linear_id<3, int>()){
```
illuhad: Formatting looks off. AdaptiveCpp code style is not the most consistent, but one thing we are consistent on is indentation of 2 spaces ;)

include/hipSYCL/sycl/libkernel/sscp/builtins/broadcast.hpp
```diff
@@ -45,4 +83,38 @@ HIPSYCL_SSCP_CONVERGENT_BUILTIN
...
+                                                     } \
+
+template<typename T> 
+T __acpp_sscp_work_group_broadcas_impl(__acpp_int32 sender, 
+                                                     T x){
+      #ifndef ACPP_SSCP_OMP_LIBKERNEL
+      ACPP_CUDALIKE_SHMEM_ATTRIBUTE T shrd_x[1];
+      #else
+      T* shrd_x = static_cast<T*>(__acpp_sscp_host_get_internal_local_memory());
+      #endif
```
illuhad: Instead of introducing new macros, can you just add the local memory as an additional parameter? Then it's the caller's responsibility to obtain local memory in some backend-specific way. Doesn't look too good to have backend-specific stuff appear in generic headers ;)

And again, please move internal impl stuff away from the public interface.

include/hipSYCL/sycl/libkernel/sscp/builtins/broadcast.hpp
```diff
@@ -45,4 +83,38 @@ HIPSYCL_SSCP_CONVERGENT_BUILTIN
...
+     if(sender == __acpp_sscp_typed_get_local_linear_id<3, int>()){ 
+        shrd_x[0] = x; 
+     }; 
+     __acpp_sscp_work_group_barrier(__acpp_sscp_memory_scope::work_group, __acpp_sscp_memory_order::relaxed); 
+     x = shrd_x[0]; 
+    __acpp_sscp_work_group_barrier(__acpp_sscp_memory_scope::work_group, __acpp_sscp_memory_order::relaxed); 
+    return x; 
+  } 
+
+#define GROUP_BCAST(fn_suffix,input_type) \
```
illuhad: Same thing w.rt. macros (`ACPP_` prefix, and move to internal detail header)

include/hipSYCL/sycl/libkernel/sscp/builtins/core_typed.hpp
```diff
@@ -0,0 +1,198 @@
...
+HIPSYCL_SSCP_BUILTIN __acpp_uint64 __acpp_sscp_get_group_id_y();
+HIPSYCL_SSCP_BUILTIN __acpp_uint64 __acpp_sscp_get_group_id_z();
+
+HIPSYCL_SSCP_BUILTIN __acpp_uint64 __acpp_sscp_get_local_size_x();
+HIPSYCL_SSCP_BUILTIN __acpp_uint64 __acpp_sscp_get_local_size_y();
+HIPSYCL_SSCP_BUILTIN __acpp_uint64 __acpp_sscp_get_local_size_z();
+
+HIPSYCL_SSCP_BUILTIN __acpp_uint64 __acpp_sscp_get_num_groups_x();
+HIPSYCL_SSCP_BUILTIN __acpp_uint64 __acpp_sscp_get_num_groups_y();
+HIPSYCL_SSCP_BUILTIN __acpp_uint64 __acpp_sscp_get_num_groups_z();
```
illuhad, line 32: These functions are already declared in `core.hpp`. We redeclare them?

include/hipSYCL/sycl/libkernel/sscp/builtins/reduction.hpp
```diff
@@ -9,12 +9,16 @@
  */
 // SPDX-License-Identifier: BSD-2-Clause
 #include "builtin_config.hpp"
+#include "core_typed.hpp"
+#include "broadcast.hpp"
+#include "utils.hpp"
 #include "hipSYCL/sycl/libkernel/detail/half_representation.hpp"
+#include "hipSYCL/sycl/libkernel/sscp/builtins/shuffle.hpp"
+#include "hipSYCL/sycl/libkernel/sscp/builtins/subgroup.hpp"
```
illuhad: Ahhhh...what's the point of having separate headers when everything is pulled in anyway :P

You wouldn't have this issue with internal `detail` headers as I suggested.

include/hipSYCL/sycl/libkernel/sscp/builtins/reduction.hpp
```diff
@@ -84,4 +88,99 @@ HIPSYCL_SSCP_CONVERGENT_BUILTIN
 __acpp_f64 __acpp_sscp_sub_group_reduce_f64(__acpp_sscp_algorithm_op op, __acpp_f64 x);
 
 
+
+
+// reduce
+template <typename OutType, typename BinaryOperation> 
+OutType __acpp_reduce_over_subgroup_impl(OutType x, BinaryOperation binary_op, __acpp_int32 active_threads) {
```
illuhad: Namingwise: In internal detail header, these could just be normal C++ functions without `__acpp` prefix, and live in some `hipsycl::libkernel` namespace or so. Then you wouldn't need to haul around that `_impl` suffix.

include/hipSYCL/sycl/libkernel/sscp/builtins/reduction.hpp
```diff
@@ -84,4 +88,99 @@ HIPSYCL_SSCP_CONVERGENT_BUILTIN
...
+  return __acpp_reduce_over_subgroup_impl(x, op{}, lrange);
+}
+
+template <typename OutType, typename BinaryOperation> 
+OutType __acpp_reduce_over_work_group_impl(OutType x,BinaryOperation op){
+  const constexpr __acpp_uint32 shmem_array_length = 32;
+  #ifndef ACPP_SSCP_OMP_LIBKERNEL
+  ACPP_CUDALIKE_SHMEM_ATTRIBUTE OutType  shrd_mem[shmem_array_length];
+  #else
+  OutType* shrd_mem = static_cast<OutType*>(__acpp_sscp_host_get_internal_local_memory());
```
illuhad: Same here...just pass in local memory as argument?

include/hipSYCL/sycl/libkernel/sscp/builtins/reduction.hpp
```diff
@@ -84,4 +88,99 @@ HIPSYCL_SSCP_CONVERGENT_BUILTIN
...
+  using op = typename get_op<binary_op>::type;
+  const __acpp_uint32       lrange     = __acpp_sscp_get_subgroup_size(); 
+  return __acpp_reduce_over_subgroup_impl(x, op{}, lrange);
+}
+
+template <typename OutType, typename BinaryOperation> 
+OutType __acpp_reduce_over_work_group_impl(OutType x,BinaryOperation op){
+  const constexpr __acpp_uint32 shmem_array_length = 32;
+  #ifndef ACPP_SSCP_OMP_LIBKERNEL
+  ACPP_CUDALIKE_SHMEM_ATTRIBUTE OutType  shrd_mem[shmem_array_length];
```
illuhad: This is called `CUDALIKE` (we use `hiplike` in other places), but is active for all backends but host. What about OpenCL?

include/hipSYCL/sycl/libkernel/sscp/builtins/scan_exclusive.hpp
```diff
@@ -0,0 +1,202 @@
...
+SUBGROUP_DECL(i64, int64);
+
+SUBGROUP_DECL(u8,  uint8);
+SUBGROUP_DECL(u16, uint16);
+SUBGROUP_DECL(u32, uint32);
+SUBGROUP_DECL(u64, uint64);
+
+SUBGROUP_DECL(f16, f16);
+SUBGROUP_DECL(f32, f32);
+SUBGROUP_DECL(f64, f64);
```
illuhad: Move all that stuff to internal `detail` header. And `ACPP_*` prefix :P

include/hipSYCL/sycl/libkernel/sscp/builtins/scan_inclusive.hpp
```diff
@@ -0,0 +1,233 @@
...
+    __acpp_uint32 next_id = lid -i; 
+    auto other_x=bit_cast<T>(__acpp_sscp_sub_group_select(bit_cast<typename integer_type<T>::type>(local_x), next_id)); 
+    if (next_id >= 0 && i <= lid) 
+        local_x = binary_op(local_x, other_x); 
+    } 
+  return local_x; 
+}
+
+template <typename OutType, typename BinaryOperation> 
+OutType __acpp_group_inclusive_scan_cudalike_impl(OutType x,BinaryOperation op){
```
illuhad: What exactly makes this `cudalike`? Looks pretty generic to me.

include/hipSYCL/sycl/libkernel/sscp/builtins/scan_inclusive.hpp
```diff
@@ -0,0 +1,233 @@
...
+  __acpp_sscp_work_group_barrier(__acpp_sscp_memory_scope::work_group, __acpp_sscp_memory_order::relaxed);
+  OutType result = subgroup_id > 0 ? op(shrd_mem[subgroup_id-1], sg_scan_result) : sg_scan_result;
+  //WAR dependency in case scan is called in a loop
+  __acpp_sscp_work_group_barrier(__acpp_sscp_memory_scope::work_group, __acpp_sscp_memory_order::relaxed);
+
+  return result;
+}
+
+template <typename OutType, typename BinaryOperation> 
+OutType __acpp_group_exclusive_scan_host_impl(OutType x,BinaryOperation op){
```
illuhad: Can we have different headers or subdirectories for implementations of different backends? I'd like to keep them separate.

Why do we even need to put these into a header? Presumably the only user will be the host implementation of `exclusive_scan`, so just put it in the `.cpp` file?

include/hipSYCL/sycl/libkernel/sscp/builtins/scan_inclusive.hpp
```diff
@@ -0,0 +1,233 @@
...
+      shrd_mem[wg_lid] = local_x;
+    }
+    __acpp_sscp_work_group_barrier(__acpp_sscp_memory_scope::work_group, __acpp_sscp_memory_order::relaxed);
+  }
+  return local_x; 
+}
+
+template <typename OutType, typename BinaryOperation> 
+OutType __acpp_group_inclusive_scan_impl(OutType x,BinaryOperation op){
+  const constexpr __acpp_uint32 shmem_array_length = 32;
```
illuhad: Maybe make this a template argument? Then the caller can configure the function appropriately for the backend.

include/hipSYCL/sycl/libkernel/sscp/builtins/utils.hpp
```diff
@@ -0,0 +1,185 @@
...
+ * Copyright The AdaptiveCpp Contributors
+ *
+ * AdaptiveCpp is released under the BSD 2-Clause "Simplified" License.
+ * See file LICENSE in the project root for full license details.
+ */
+// SPDX-License-Identifier: BSD-2-Clause
+#include "builtin_config.hpp"
+
+
+#define  ACPP_CUDALIKE_SHMEM_ATTRIBUTE static __attribute__((loader_uninitialized))  __attribute__((address_space(3)))
```
illuhad: Isn't it exactly the same on SPIR-V?

include/hipSYCL/sycl/libkernel/sscp/builtins/utils.hpp
```diff
@@ -0,0 +1,185 @@
...
+
+__acpp_uint64 get_active_mask();
+
+struct plus
+{
+    template<typename T>
+    T operator()(T lhs, T rhs){
+        return lhs+rhs;
+    }
+};
```
illuhad: Put all of that stuff in a namespace..

include/hipSYCL/sycl/libkernel/sscp/builtins/utils.hpp
```diff
@@ -0,0 +1,185 @@
...
+ * AdaptiveCpp is released under the BSD 2-Clause "Simplified" License.
+ * See file LICENSE in the project root for full license details.
+ */
+// SPDX-License-Identifier: BSD-2-Clause
+#include "builtin_config.hpp"
+
+
+#define  ACPP_CUDALIKE_SHMEM_ATTRIBUTE static __attribute__((loader_uninitialized))  __attribute__((address_space(3)))
+
+HIPSYCL_SSCP_BUILTIN void* __acpp_sscp_host_get_internal_local_memory();
```
illuhad: This is only valid for host. Can we make the declaration not spill into all the other backends?

include/hipSYCL/sycl/libkernel/sscp/builtins/utils.hpp
```diff
@@ -0,0 +1,185 @@
...
+    out = u.union_out;                                                         \
+  }
+#endif
+
+template <class Tout, class Tin>
+Tout bit_cast(Tin x) {
+  Tout result;
+  HIPSYCL_INPLACE_BIT_CAST(Tin, Tout, x, result);
+  return result;
+}
```
illuhad: You don't need all of this complexity. We're in the bitcode libraries, so you can safely assume clang and just use `__builtin_bit_cast`.

include/hipSYCL/sycl/libkernel/sscp/builtins/utils.hpp
```diff
@@ -0,0 +1,185 @@
...
+
+__acpp_uint64 get_active_mask();
```
illuhad: What is this for? On what backends is this valid etc? Put it somewhere appropriate or name it appropriately ;)

include/hipSYCL/sycl/libkernel/sscp/builtins/utils.hpp
```diff
@@ -0,0 +1,185 @@
...
+MAP_SSCP_ALGORITHM_OP(__acpp_sscp_algorithm_op::plus,plus)
+MAP_SSCP_ALGORITHM_OP(__acpp_sscp_algorithm_op::multiply,multiply)
+MAP_SSCP_ALGORITHM_OP(__acpp_sscp_algorithm_op::min,min)
+MAP_SSCP_ALGORITHM_OP(__acpp_sscp_algorithm_op::max,max)
+MAP_SSCP_ALGORITHM_OP(__acpp_sscp_algorithm_op::bit_and,bit_and)
+MAP_SSCP_ALGORITHM_OP(__acpp_sscp_algorithm_op::bit_or,bit_or)
+MAP_SSCP_ALGORITHM_OP(__acpp_sscp_algorithm_op::bit_xor,bit_xor)
+MAP_SSCP_ALGORITHM_OP(__acpp_sscp_algorithm_op::logical_and,logical_and)
+MAP_SSCP_ALGORITHM_OP(__acpp_sscp_algorithm_op::logical_or,logical_or)
+
```
illuhad: Might be a good practice to `#undef` this macro here again. Especially if you're in a header. Especially if you're not using `ACPP_` prefix :P

include/hipSYCL/sycl/libkernel/sscp/builtins/utils.hpp
```diff
@@ -0,0 +1,185 @@
...
+struct integer_type{
+    using type = T;
+};
+
+// __acpp_f16 is unsigned short currently so
+// This will resolve to the same type
+// template<>
+// struct integer_type<__acpp_f16>{
+//   using type = __acpp_int16;
+// };
```
illuhad: Does this mean we don't currently support fp16 in the group algorithms?

include/hipSYCL/sycl/libkernel/sscp/group_functions.hpp
```diff
@@ -67,15 +69,14 @@ __acpp_group_barrier(sub_group g,
 
 // broadcast
 
-template <int Dim, typename T>
+template <int Dim, typename T, std::enable_if_t<sizeof(T) <= 8, int> = 0 >
 HIPSYCL_BUILTIN 
-T __acpp_group_broadcast(
+ std::enable_if_t<(sizeof(T) <= 8), T> __acpp_group_broadcast(
```
illuhad: I think you want to optimize the case where we have a `vec` or `marray` of small size. Do we actually gain something performance-wise from this complexity?

include/hipSYCL/sycl/libkernel/sscp/group_functions.hpp
```diff
@@ -151,11 +144,11 @@ __acpp_group_broadcast(
   return result;
 }
 
-template <class Group, typename T, int N>
+template<typename T, int N, class Group>
```
illuhad: Why put `Group` at the end? It goes in the beginning everywhere else.

include/hipSYCL/sycl/libkernel/sscp/group_functions.hpp
```diff
@@ -556,74 +557,516 @@ T __acpp_joint_reduce(Group g, Ptr first, Ptr last, T init,
...
+  //Ptr start_ptr = first + lid;
+  using type = decltype(*first);
+  auto identity = sscp_binary_operation_identity<std::decay_t<type>, sscp_binary_operation_v<BinaryOperation>>::get();
+  size_t segment = 0;
+  size_t num_segments = (num_elements+lrange-1)/lrange;
+
+  // for (Ptr p = start_ptr + lrange; p < last; p += lrange){
+  for(size_t segment = 0; segment < num_segments; segment++){
+    size_t element_idx = segment*lrange+lid;
+      auto local_element = element_idx < num_elements ? first[element_idx] : identity;
```
illuhad: Strange indentation?

src/libkernel/sscp/host/broadcast.cpp
```diff
@@ -0,0 +1,92 @@
...
+  return __acpp_sscp_sub_group_broadcast_i64(value, id);
+}
+
+
+
+template<typename T> 
+T __acpp_sscp_work_group_broadcast_host_impl(__acpp_int32 sender, 
+                                                     T x){        
+     static int shrd_mem;
+     #pragma omp threadprivate(shrd_mem)
```
illuhad: Why risk any TLS issues here - just use the new builtin to obtain local memory?

src/libkernel/sscp/host/scan_exclusive.cpp
```diff
@@ -0,0 +1,151 @@
...
+        case __acpp_sscp_algorithm_op::bit_xor: \
+            return __acpp_subgroup_exclusive_scan_impl(x, bit_xor{}, init); \
+        case __acpp_sscp_algorithm_op::logical_and: \
+            return __acpp_subgroup_exclusive_scan_impl(x, logical_and{}, init); \
+        case __acpp_sscp_algorithm_op::logical_or: \
+            return __acpp_subgroup_exclusive_scan_impl(x, logical_or{}, init); \
+        default: \
+            return __acpp_##type{}; \
+    } \
+} \
```
illuhad: This is a case where clang-format might be helpful. For me it moves all `\` at the end of the line :)

src/libkernel/sscp/ptx/broadcast.cpp
```diff
@@ -0,0 +1,65 @@
...
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
```
illuhad, line 9: We don't use this preamble anymore. See other headers for the current one.

src/libkernel/sscp/ptx/shuffle.cpp
```diff
@@ -0,0 +1,205 @@
...
+template<>
+__acpp_int64 __acpp_sscp_sub_group_select<__acpp_int64>(__acpp_int64 value, __acpp_int32 id){
+  return __acpp_sscp_sub_group_select_i64(value, id);
+}   
+
+// __acpp_uint32 get_active_mask(){
+//     //__acpp_int64 mask = __nvvm_activemask();
+//     __acpp_uint32 subgroup_size = __acpp_sscp_get_subgroup_size();
+//     return (1 << subgroup_size)-1;
+// }
```
illuhad: What is this code? Do we still need it?

src/libkernel/sscp/ptx/shuffle.cpp
```diff
@@ -0,0 +1,205 @@
...
+//                                  int __width = warpSize) {                     \
+//     return __IntIntrinsic(__val, __offset,                                     \
+//                           ((warpSize - __width) << 8) | (__Mask));             \
+//   }   
+// __MAKE_SHUFFLES(__shfl_up, __nvvm_shfl_up_i32, __nvvm_shfl_up_f32, 0,
+//                 unsigned int);
+// __MAKE_SHUFFLES(__shfl_down, __nvvm_shfl_down_i32, __nvvm_shfl_down_f32, 0x1f,
+//                 unsigned int);
+// __MAKE_SHUFFLES(__shfl_xor, __nvvm_shfl_bfly_i32, __nvvm_shfl_bfly_f32, 0x1f,
+//                 int);
```
illuhad: What is the relevance of this?

src/libkernel/sscp/ptx/shuffle.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+HIPSYCL_SSCP_CONVERGENT_BUILTIN
+__acpp_int32 __acpp_sscp_sub_group_select_i32(__acpp_int32 value,
+                                                     __acpp_int32 id){
+    //int max_subgroup_size = __acpp_sscp_get_subgroup_max_size();
+    // int index = id%max_subgroup_size;
+    // return __builtin_amdgcn_ds_bpermute(index<<2, value);
+    // __acpp_uint32 mask = get_active_mask();
+    // This doesn't work
+    return __nvvm_shfl_sync_idx_i32(FULL_MASk, value, id, 31);
```
illuhad: What doesn't work? The code or the thing that is commented out?

Also: Typo `FULL_MASk` with lower case `k` at the end?

src/libkernel/sscp/spirv/shuffle.cpp
```diff
@@ -0,0 +1,231 @@
...
+    out = u.union_out;                                                         \
+  }
+#endif
+
+template <class Tout, class Tin>
+Tout bit_cast(Tin x) {
+  Tout result;
+  HIPSYCL_INPLACE_BIT_CAST(Tin, Tout, x, result);
+  return result;
+}
```
illuhad: Just use `__builtin_bit_cast`.

illuhad (commented):
Awesome to see!! :D

src/libkernel/sscp/CMakeLists.txt
```diff
@@ -18,7 +18,7 @@ function(libkernel_generate_bitcode_library)
   set(target_output_file ${output})
   add_custom_command(
       OUTPUT ${target_output_file}
-      COMMAND ${CLANG_EXECUTABLE_PATH} -target ${triple} -fno-exceptions -fno-rtti -O3 -emit-llvm -std=c++17 -DHIPSYCL_SSCP_LIBKERNEL_LIBRARY
+      COMMAND ${CLANG_EXECUTABLE_PATH} -target ${triple} -fPIC -fno-exceptions -fno-rtti -O3 -emit-llvm -std=c++17 -DHIPSYCL_SSCP_LIBKERNEL_LIBRARY
```
illuhad: Why do we need `-fPIC`?

src/compiler/llvm-to-backend/host/HostKernelWrapperPass.cpp
```diff
@@ -172,6 +182,7 @@ llvm::Function *makeWrapperFunction(llvm::Function &F, std::int64_t DynamicLocal
     replaceUsesOfGVWith(*Wrapper, cbs::NumGroupsGlobalNames[I], NumGroups[I]);
     replaceUsesOfGVWith(*Wrapper, cbs::GroupIdGlobalNames[I], GroupIds[I]);
     replaceUsesOfGVWith(*Wrapper, cbs::LocalSizeGlobalNames[I], LocalSize[I]);
+    replaceUsesOfGVWith(*Wrapper, cbs::LocalIdGlobalNames[I], LocalIds[I]);
```
fodinabor: I'm not sure if I understand what you are trying to achieve here / why this is necessary.
Is this an attempt to 0 initialize the `__acpp_cbs_local_id_[x-z]`?
Wouldn't this override all uses to `0` instead?

On the other hand, I'm not even sure, why there should any uses of these variables be around at this point.. 🤔 
Maybe the correct fix would be to fix the uses to be deleted correctly, beforehand.

Maybe this should be `mergeGVLoadsInEntry` instead:
https://github.com/AdaptiveCpp/AdaptiveCpp/blob/bf218d9b835efe2ac3c609a10abb1b997b00b116/src/compiler/cbs/SubCfgFormation.cpp#L1339
So that the, then unique loads, can be effectively deleted here: https://github.com/AdaptiveCpp/AdaptiveCpp/blob/bf218d9b835efe2ac3c609a10abb1b997b00b116/src/compiler/cbs/SubCfgFormation.cpp#L1355

And we should also delete those loads here as well:
https://github.com/AdaptiveCpp/AdaptiveCpp/blob/bf218d9b835efe2ac3c609a10abb1b997b00b116/src/compiler/cbs/SubCfgFormation.cpp#L1294

Or, potentially, I'm misunderstanding what is happening here/why this should be necessary :)

illuhad: I think the motivation is that the SSCP group algorithms interface is unaware of the dimensionality of the kernel. So often times we need to calculate the general linearized local id `lid_z * size_x * size*y + lid_y * size_x + lid_x`.

The problem is that the CBS builtins for the local id remain undefined if the dimension is greater than the dimensionality of the kernel, which creates linker errors.

So we need to have all builtin local ids defined always, independently from the dimensionality of the kernels. The ones that are outside the dimensionality of the kernel can just be set to constant 0. 

The idea behind @sbalint98's change here is that references to local ids that are still in the code after the CBS pipeline  has run must be cases where the dimension is outside the dimensionality of the kernel, and can thus be set to 0.
But it's not clear whether this is the best way to handle this :)

sbalint98: `__acpp_cbs_local_id_[x-z]` is around because in the builtins need to know the local_id, however the dimensionality  is only available in runtime. Therefore we call  `__acpp_sscp_typed_get_local_linear_id<3, int>()` which uses `__acpp_cbs_local_id_[x-z]`. In case the kernel is not 3D the unused `__acpp_cbs_local_id_[x-z]` references remain in the IR and cause a JIT time error.

Maybe we could just set these to zero in the SubCfgFormation createLoopsAroundKernel function? Or would deleting them have the same effect?

illuhad: @sbalint98 By the way, have you tried `ACPP_ADAPTIVITY_LEVEL=0`? I wonder if we get the same problem with the local sizes, and you're just not seeing this because the the default `ACPP_ADAPTIVITY_LEVEL=1` replaces them before the bitcode libraries are processed, so CBS never sees these function calls.

sbalint98: Same happens with ` ACPP_ADAPTIVITY_LEVEL=0 ` unfortunately.

include/hipSYCL/sycl/libkernel/sscp/builtins/scan_inclusive.hpp
sbalint98: The fact that we know that on HIP and CUDA we will have num_subgroups <= 32. Contrary to the OpenCL backend where subgroup size of 1 and workgroup size of 8192 are possible at the same time. In that case we would need to allocate much more shared memory statically.

include/hipSYCL/sycl/libkernel/sscp/builtins/core_typed.hpp
sbalint98: As far as I remember it is not possible for some reason to include the rest of the types from `core.hpp` in the bulitins. Therefore there only the new `core_typed.hpp` is included, therefore I think it is needed to have these redeclared.

include/hipSYCL/sycl/libkernel/sscp/builtins/broadcast.hpp
sbalint98: Fixed moved to headers in the detail subfolder.

include/hipSYCL/sycl/libkernel/sscp/builtins/broadcast.hpp
sbalint98: everything supposed to be formatted with two spaces now

include/hipSYCL/sycl/libkernel/sscp/builtins/broadcast.hpp
sbalint98: Good point, added extra parameter for local memory

include/hipSYCL/sycl/libkernel/sscp/builtins/broadcast.hpp
sbalint98: reformatted with two spaces

include/hipSYCL/sycl/libkernel/sscp/builtins/reduction.hpp
sbalint98: Removed all unnecessary includes from the interface headers

include/hipSYCL/sycl/libkernel/sscp/builtins/reduction.hpp
sbalint98: switched to `hipsycl::libkernel::sscp` namespace

include/hipSYCL/sycl/libkernel/sscp/builtins/reduction.hpp
sbalint98: renamed to `ACPP_GPU_SHMEM_ATTRIBUTE`

include/hipSYCL/sycl/libkernel/sscp/builtins/reduction.hpp
sbalint98: yes, fixed

include/hipSYCL/sycl/libkernel/sscp/builtins/scan_exclusive.hpp
sbalint98: These are just for stamping out the interface. However this approach is only present in this file, therefore replaced with the extension of the macros.

include/hipSYCL/sycl/libkernel/sscp/builtins/scan_inclusive.hpp
sbalint98: Currently only scan has different implementations for different backends. Therefore I think it is best to just have `scan_<backend_name>.hpp` files. (changed)

Currently the same templated function is used for both the exclusive and the inclusive scan functions.

include/hipSYCL/sycl/libkernel/sscp/builtins/scan_inclusive.hpp
sbalint98: done

include/hipSYCL/sycl/libkernel/sscp/builtins/utils.hpp
sbalint98: true, renamed to `ACPP_SHMEM_ATTRIBUTE`

include/hipSYCL/sycl/libkernel/sscp/builtins/utils.hpp
sbalint98: Added to the appropriate cpp files

include/hipSYCL/sycl/libkernel/sscp/builtins/utils.hpp
sbalint98: true, removed the macros

include/hipSYCL/sycl/libkernel/sscp/builtins/utils.hpp
sbalint98: removed, (left over from earlier version)

include/hipSYCL/sycl/libkernel/sscp/builtins/utils.hpp
sbalint98: I am not sure. We don't test with half, and it seems that adding `sycl::half` support to the tests is not straightforward. My very limited local tests show that half works.

src/libkernel/sscp/CMakeLists.txt
sbalint98: It was needed for TLS, removed

src/libkernel/sscp/host/broadcast.cpp
sbalint98: removed

src/libkernel/sscp/host/scan_exclusive.cpp
sbalint98: Thanks, started using clang-format

src/libkernel/sscp/ptx/broadcast.cpp
sbalint98: replaced with the current one

src/libkernel/sscp/ptx/shuffle.cpp
sbalint98: no, removed it

src/libkernel/sscp/ptx/shuffle.cpp
sbalint98: removed

src/libkernel/sscp/ptx/shuffle.cpp
sbalint98: This was from some previous debuging, removed.

src/libkernel/sscp/spirv/shuffle.cpp
sbalint98: replaced

include/hipSYCL/sycl/libkernel/sscp/builtins/broadcast.hpp
sbalint98: I hope I replaced all of them now

include/hipSYCL/sycl/libkernel/sscp/builtins/broadcast.hpp
sbalint98: replaced

include/hipSYCL/sycl/libkernel/sscp/builtins/utils.hpp
sbalint98: done

include/hipSYCL/sycl/libkernel/sscp/builtins/utils.hpp
sbalint98: done

src/compiler/llvm-to-backend/host/HostKernelWrapperPass.cpp
fodinabor: Hm okay, I think, I understand the issue then.

Can you test if applying the following patch fixes this, without your compiler change?

```diff
diff --git a/src/compiler/cbs/SubCfgFormation.cpp b/src/compiler/cbs/SubCfgFormation.cpp
index 9b73472c..c62efdcf 100644
--- a/src/compiler/cbs/SubCfgFormation.cpp
+++ b/src/compiler/cbs/SubCfgFormation.cpp
@@ -351,6 +351,12 @@ void createLoopsAround(llvm::Function &F, llvm::BasicBlock *AfterBB,
 
   VMap[mergeGVLoadsInEntry(F, LocalIdGlobalNamesRotated[0])] = IndVars[0];
 
+  // in case code references all dimensions, we need to set the remaining dimensions to 0
+  for (size_t D = Dim; D < 3; ++D) {
+    VMap[mergeGVLoadsInEntry(F, LocalIdGlobalNamesRotated[D])] =
+        Builder.getIntN(Idx->getType()->getIntegerBitWidth(), 0);
+  }
+
   VMap[ContiguousIdx] = Idx;
   ContiguousIdx = Idx;
 }
@@ -1311,7 +1317,6 @@ void createLoopsAroundKernel(llvm::Function &F, llvm::DominatorTree &DT, llvm::L
 
   Body = Body->getSingleSuccessor();
 
-
   llvm::SmallVector<llvm::BasicBlock *, 4> ExitBBs;
   llvm::BasicBlock *ExitBB = llvm::BasicBlock::Create(F.getContext(), "exit", &F);
   llvm::IRBuilder<> Bld{ExitBB};
@@ -1352,10 +1357,13 @@ void createLoopsAroundKernel(llvm::Function &F, llvm::DominatorTree &DT, llvm::L
   llvm::remapInstructionsInBlocks(Blocks, VMap);
 
   // remove uses of the undefined global id variables
-  for (int D = 0; D < Dim; ++D)
-    if (auto *Load =
-            llvm::cast_or_null<llvm::LoadInst>(getLoadForGlobalVariable(F, LocalIdGlobalNames[D])))
+  for (int D = 0; D < 3; ++D)
+    if (auto *Load = llvm::cast_or_null<llvm::LoadInst>(
+                  mergeGVLoadsInEntry(F, LocalIdGlobalNames[D]))) {
+      if (D >= Dim)
+        Load->replaceAllUsesWith(llvm::ConstantInt::get(Load->getType(), 0));
       Load->eraseFromParent();
+    }
   HIPSYCL_DEBUG_EXECUTE_VERBOSE(F.viewCFG())
 }
 ```

Comment (sbalint98) [reply src/compiler/llvm-to-backend/host/HostKernelWrapperPass.cpp]: Hi, It seems to work as intended, do you want to push a commit somewhere that I could cherry pick for proper attribution?

Comment (fodinabor) [reply src/compiler/llvm-to-backend/host/HostKernelWrapperPass.cpp]: https://github.com/AdaptiveCpp/AdaptiveCpp/tree/cleanup-higher-dim-ids
Can you test this?
I'm not sure, I tested this a bit, and I'm not entirely sure, if it is sufficient..

File: include/hipSYCL/sycl/libkernel/sscp/group_functions.hpp (hunk 23 above)
Comment (sbalint98) [reply include/hipSYCL/sycl/libkernel/sscp/group_functions.hpp]: This was intended as a functional change. (At least one of `enable_if_t`s :P) Without this, in case T is `sycl::vec<int, 3>` for instance the template will be ambiguous with the specialization meant for vectors. 

Removing this change yields:
```
/data/sbalint/acpp_gmx_practical/AdaptiveCpp-upstreaming/build/install/bin/../include/AdaptiveCpp/hipSYCL/sycl/libkernel/group_functions.hpp:62:3: error: call to '__acpp_group_broadcast' is ambiguous
  HIPSYCL_RETURN_DISPATCH_GROUP_ALGORITHM(__acpp_group_broadcast, g, x,
.....

/data/sbalint/acpp_gmx_practical/AdaptiveCpp-upstreaming/build/install/bin/../include/AdaptiveCpp/CL/../hipSYCL/sycl/libkernel/sscp/group_functions.hpp:74:4: note: candidate function [with Dim = 1, T = hipsycl::sycl::vec<int, 3>]
 T __acpp_group_broadcast(
   ^
/data/sbalint/acpp_gmx_practical/AdaptiveCpp-upstreaming/build/install/bin/../include/AdaptiveCpp/CL/../hipSYCL/sycl/libkernel/sscp/group_functions.hpp:137:1: note: candidate function [with T = int, N = 3, Group = hipsycl::sycl::group<1>]
__acpp_group_broadcast(
```

File: include/hipSYCL/sycl/libkernel/sscp/group_functions.hpp (hunk 25 above)
Comment (sbalint98) [reply include/hipSYCL/sycl/libkernel/sscp/group_functions.hpp]: reformatted with clangformat

File: include/hipSYCL/sycl/libkernel/sscp/group_functions.hpp (hunk 24 above)
Comment (sbalint98) [reply include/hipSYCL/sycl/libkernel/sscp/group_functions.hpp]: changed

File: src/compiler/llvm-to-backend/host/HostKernelWrapperPass.cpp (hunk 34 above)
Comment (sbalint98) [reply src/compiler/llvm-to-backend/host/HostKernelWrapperPass.cpp]: Interestingly this fails with:
```
[AdaptiveCpp Warning] /home/sbalint/.acpp/apps/sycl_tests-5273160718515756707/jit-cache/14570224624806148754.11565643320243800617.jit.so: undefined symbol: __acpp_cbs_local_id_y
```

Even though the symbols seem to be defined in the IR
```llvm
@__acpp_cbs_local_id_y = external local_unnamed_addr global i64, align 8
@__acpp_cbs_local_id_z = external local_unnamed_addr global i64, align 8

; Function Attrs: inaccessiblememonly mustprogress nocallback nofree nosync nounwind willreturn
declare void @llvm.assume(i1 noundef) #0

; Function Attrs: nofree nosync nounwind
define void @_Z18__acpp_sscp_kernelIN7hipsycl4glue15__sscp_dispatch20ndrange_parallel_forIZZ25test_nd_group_function_1dILi102EiZN21group_functions_tests15group_broadcastIiE11test_methodEvEUlRSt6vectorIiSaIiEEmmE_ZNS7_11test_methodEvEUlT_mNS0_4sycl9sub_groupET0_iE_ZNS7_11test_methodEvEUlRKSA_SJ_mmmE_EvmT1_T2_T3_ENKUlRNSE_7handlerEE_clESP_EUlNSE_7nd_itemILi1EEEE_Li1EEEEvRKSD_(ptr nocapture readonly %0, ptr nocapture readonly %1) local_unnamed_addr #1 {
entry:
  %num_groups_x = load i64, ptr %0, align 8
  %2 = getelementptr inbounds { [3 x i64], [3 x i64], [3 x i64], ptr, ptr }, ptr %0, i64 0, i32 1, i64 0
  %group_id_x = load i64, ptr %2, align 8
  %3 = getelementptr inbounds { [3 x i64], [3 x i64], [3 x i64], ptr, ptr }, ptr %0, i64 0, i32 2, i64 0
  %local_size_x = load i64, ptr %3, align 8
  %4 = getelementptr inbounds { [3 x i64], [3 x i64], [3 x i64], ptr, ptr }, ptr %0, i64 0, i32 2, i64 1
  %local_size_y = load i64, ptr %4, align 8
  %5 = getelementptr inbounds { [3 x i64], [3 x i64], [3 x i64], ptr, ptr }, ptr %0, i64 0, i32 4
  %internal_local_mem_ptr = load ptr, ptr %5, align 8
  %6 = load ptr, ptr %1, align 8
  %7 = load ptr, ptr %6, align 8
  %8 = getelementptr inbounds ptr, ptr %1, i64 3
  %9 = load ptr, ptr %8, align 8
  %10 = load ptr, ptr %9, align 8
  %_alloca2.i = alloca i64, i64 %local_size_x, align 64, !hipSYCL.arrayified !372
  %11 = icmp ult i64 %num_groups_x, 8388608
  %12 = shl nuw nsw i64 %group_id_x, 8
  %13 = icmp ult i64 %group_id_x, 8388607
  %14 = load i64, ptr @__acpp_cbs_local_id_y, align 8
  %15 = trunc i64 %14 to i32
  %16 = load i64, ptr @__acpp_cbs_local_id_z, align 8
 %17 = trunc i64 %16 to i32
  %18 = trunc i64 %local_size_x to i32
  %19 = trunc i64 %local_size_y to i32
  %20 = mul i32 %17, %19
  %21 = add i32 %20, %15
  %22 = mul i32 %21, %18
  %umax = tail call i64 @llvm.umax.i64(i64 %local_size_x, i64 1)

```

Here is the full dump of the LLVM IR for all the stages:  [out.txt](https://github.com/user-attachments/files/18155053/out.txt)

fodinabor: `external` means undefined :)
OK.. So doesn't really fix it.. I'll have a look tomorrow.

src/libkernel/sscp/ptx/shuffle.cpp
```diff
@@ -0,0 +1,141 @@
...
+__acpp_int32 hipsycl::libkernel::sscp::sg_select<__acpp_int32>(__acpp_int32 value, __acpp_int32 id) {
+  return __acpp_sscp_sub_group_select_i32(value, id);
+}
+
+template <>
+__acpp_int64 hipsycl::libkernel::sscp::sg_select<__acpp_int64>(__acpp_int64 value, __acpp_int32 id) {
+  return __acpp_sscp_sub_group_select_i64(value, id);
+}
+
+unsigned int FULL_MASK = 0xffffffff;
```
illuhad: This must (!) be `constexpr`. Otherwise you get load from global memory for every shuffle.

include/hipSYCL/sycl/libkernel/sscp/builtins/detail/utils.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+#define ACPP_SHMEM_ATTRIBUTE                                                                       \
+  static __attribute__((loader_uninitialized)) __attribute__((address_space(3)))
+
+namespace hipsycl::libkernel::sscp {
+
+template <class Tout, class Tin> Tout bit_cast(Tin x) {
+  Tout result;
+  result = __builtin_bit_cast(Tout, x);
+  return result;
```
illuhad: What is the point of this function? Doesn't seem to add anything of value over just calling `__builtin_bit_cast`

src/libkernel/sscp/amdgpu/shuffle.cpp
```diff
@@ -0,0 +1,156 @@
...
+  int index = (self + delta);
+  index = (int)((self & (sg_size - 1)) + delta) > sg_size ? self : index;
+
+  return __builtin_amdgcn_ds_bpermute(index << 2, value);
+}
+
+HIPSYCL_SSCP_CONVERGENT_BUILTIN
+__acpp_int64 __acpp_sscp_sub_group_shl_i64(__acpp_int64 value, __acpp_uint32 delta) {
+  int tmp[2];
+  __builtin_memcpy(tmp, &value, sizeof(tmp));
```
illuhad, line 45: Have you checked whether this affects registerization? `memcpy` semantics forces the compiler to think in terms of memory addresses, which could prevent registerization. Maybe it's smart enough to figure out that it does not need that here.

Might be better to use bit shifts and bitmasks instead.

src/libkernel/sscp/amdgpu/shuffle.cpp
```diff
@@ -0,0 +1,156 @@
...
+ */
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include "hipSYCL/sycl/libkernel/sscp/builtins/shuffle.hpp"
+#include "hipSYCL/sycl/libkernel/sscp/builtins/detail/shuffle.hpp"
+#include "hipSYCL/sycl/libkernel/sscp/builtins/subgroup.hpp"
+
+template <>
+__acpp_int8 hipsycl::libkernel::sscp::sg_select<__acpp_int8>(__acpp_int8 value, __acpp_int32 id) {
+  return __acpp_sscp_sub_group_select_i8(value, id);
```
illuhad: What is the purpose of these function definitions? I see the same trivial definitions copy-pasted to all backends.

If the implementation is trivial, just directly call `__acpp_sscp_sub_group_select_*`.
If you want a templated convenience wrapper, just define it one in the header for all backends.

src/compiler/llvm-to-backend/host/HostKernelWrapperPass.cpp
fodinabor: Used the wrong name array :)
Should work now (did at least testing with your changes - minus the HostKernelWrapperPass)

illuhad:
As this is still marked as a draft PR, what do we need to make progress?

src/libkernel/sscp/ptx/shuffle.cpp
sbalint98: added

include/hipSYCL/sycl/libkernel/sscp/builtins/detail/utils.hpp
sbalint98: I would prefer to keep this in a separate function for now, so that I can explicitly set the cast type.

src/libkernel/sscp/amdgpu/shuffle.cpp
sbalint98: Good point, implemented with inline templates.

include/hipSYCL/sycl/libkernel/sscp/builtins/detail/utils.hpp
illuhad: You also explicitly specify the type with `__builtin_bit_cast` using the first argument, no?

Can we at least remove all of the unnecessary lines and just implement it as `return __builtin_bit_cast(Tout, x)` which is exactly the same thing?

src/libkernel/sscp/amdgpu/shuffle.cpp
illuhad: Thanks :)

include/hipSYCL/sycl/libkernel/sscp/builtins/core_typed.hpp
illuhad: I don't understand what you mean. But I think the problem is that you have added an include to `core_typed.hpp` to `core.hpp`, so therefore, including `core.hpp` here causes an include cycle.

IMO, this indicates that there is a design flaw w.r.t what is defined in which includes. There should probably be 3 headers:
* `core.hpp`, only defining the core builtins. Can be included in bitcode libraries.
* `core_typed_linear_id.hpp`, which includes `core.hpp` and defines the type-template function to calculate linear id. Can be included in bitcode libraries.
* `core_linear_id.hpp`, which includes `core_typed_linear_id.hpp` and defines the `__acpp_sscp_get_*_linear_id<Dim>()` functions. Cannot be included in bitcode libraries.

Headers currently using `core.hpp` will then potentially also have to include the other headers. Outside of the bitcode libraries, we only include `core.hpp` in `sscp_backend.hpp` so that might not be too difficult - just add the additional includes there.

include/hipSYCL/sycl/libkernel/sscp/builtins/detail/utils.hpp
sbalint98: Ahh true, sorry. Removed the `bit_cast` function from `utils.hpp` and switched to using `__builtin_bit_cast`

illuhad (approved):
Okay, let's get it in now. We can work out any remaining issues with smaller subsequent PRs. Thanks for your work :)

include/hipSYCL/sycl/libkernel/sscp/builtins/core_typed.hpp
sbalint98: Okay, I think I understand. Implemented your suggestion

include/hipSYCL/sycl/libkernel/sscp/group_functions.hpp
```diff
@@ -556,74 +554,442 @@ T __acpp_joint_reduce(Group g, Ptr first, Ptr last, T init,
...
+          std::enable_if_t<is_group_v<std::decay_t<Group>>, bool> = true>
+HIPSYCL_BUILTIN OutPtr __acpp_joint_exclusive_scan(Group g, InPtr first, InPtr last, OutPtr result,
+                                                   T init, BinaryOperation binary_op) {
+
+  const size_t lrange = g.get_local_range().size();
+  const size_t num_elements = last - first;
+  const size_t lid = g.get_local_linear_id();
+  __acpp_group_barrier(g);
+  if (lid == 0 && num_elements > 0) {
+    first[0] = binary_op(first[0], init);
```
al42and, line 980: The SYCL standard does not say this explicitly, but it would be more consistent with `std::exclusive_scan` if the `[first; last)` range was not modified.

sbalint98: Oops, Thanks, good catch!

sbalint98: Should be fixed in https://github.com/AdaptiveCpp/AdaptiveCpp/pull/1709

PR #1653, opened 2024-12-12
[SSCP][llvm-to-host] Apply known group size optimization for generation of WI loops

moritz-2001:
The SubCfgFormationPass uses the work-group size global variables (e.g., for the trip counts of the loops) internally.
This PR adds a pass that replaces the GVs with their respective constant value after running the CBS pipeline.


tests/compiler/cbs/accumulator_for.cpp
```diff
@@ -2,6 +2,10 @@
 // RUN: %t | FileCheck %s
 // RUN: %acpp %s -o %t --acpp-targets=omp --acpp-use-accelerated-cpu -O
 // RUN: %t | FileCheck %s
+// RUN: %acpp %s -o %t --acpp-targets=generic --acpp-use-accelerated-cpu
+// RUN: %t | FileCheck %s
+// RUN: %acpp %s -o %t --acpp-targets=generic --acpp-use-accelerated-cpu -O
```
illuhad: Are you sure this actually runs on host backend? When using `generic` you need to be careful, since it can pick whatever device it finds. It also prioritizes OpenCL CPU backend over OpenMP backend typically.
I think in CI it's probably fine since in the CBS lit workflow we don't install other backends. But it might not do the expected thing when run locally.
So better use `ACPP_VISIBILITY_MASK=omp` or explicitly select the OpenMP device programmatically.

illuhad (commented):
Thank you for this contribution :)

src/compiler/llvm-to-backend/host/LLVMToHost.cpp
```diff
@@ -101,6 +102,7 @@ bool LLVMToHostTranslator::toBackendFlavor(llvm::Module &M, PassHandler &PH) {
   registerCBSPipeline(MPM, hipsycl::compiler::OptLevel::O3, true);
 
   llvm::FunctionPassManager FPM;
+  FPM.addPass(HostKnownWgSizePass{KnownGroupSizeX, KnownGroupSizeY, KnownGroupSizeZ});
```
fodinabor: While in general, I'm in favor of splitting things in different passes, I'm not entirely sure, what the benefit is of doing this in a separate pass compared to doing it here:
https://github.com/AdaptiveCpp/AdaptiveCpp/blob/bf218d9b835efe2ac3c609a10abb1b997b00b116/src/compiler/llvm-to-backend/host/HostKernelWrapperPass.cpp#L177

One benefit of keeping it as-is, would be the following optimization: only add the pass if any of the `KnownGroupSize[X-Z]` is not `0`.
Did you observe behavior that would make it important to do it the way it currently is?

fodinabor (commented):
Generally looks good.
I'm just wondering a bit, if the extra pass isn't a bit overkill... :)

@illuhad W.r.t. tests.. maybe it would make sense to have some lit tests as well, that just run the `llvm-to-host-tool` & friends to check if certain expectations are actually fulfilled (such as: is the group size really specialized, if given?)
But that's nothing I'd expect in this PR

illuhad:
> W.r.t. tests.. maybe it would make sense to have some lit tests as well, that just run the llvm-to-host-tool & friends to check if certain expectations are actually fulfilled (such as: is the group size really specialized, if given?)
But that's nothing I'd expect in this PR

@fodinabor Yep, that would be good, but I don't think it's something we can do at the moment. The tools don't expose the interface for most of the JIT-time configuration that we do, including known group size.

src/compiler/llvm-to-backend/host/HostKernelWrapperPass.cpp
```diff
@@ -167,6 +157,13 @@ llvm::Function *makeWrapperFunction(llvm::Function &F, std::int64_t DynamicLocal
     replaceUsesOfGVWith(*Wrapper, cbs::GroupIdGlobalNames[I], GroupIds[I]);
     replaceUsesOfGVWith(*Wrapper, cbs::LocalSizeGlobalNames[I], LocalSize[I]);
   }
+
+  for (auto i = 0ul; i < 3ul; ++i) {
+    if (KnownWgSize.at(i) != 0)
+      utils::replaceUsesOfGVWith(F, cbs::LocalSizeGlobalNames.at(i),
+                                 llvm::ConstantInt::get(SizeT, KnownWgSize.at(i)), PassPrefix);
```
fodinabor: Any reason you didn't include this in the loop directly above?
Please use the normal array access operator `[]` as the rest of the code does - this part is usually built without exceptions, so bounds checking isn't a thing here really and the arrays are all fixed size...

Also, why use `F` not `*Wrapper`? `F` is already "dead" here (see inline call above).

fodinabor (changes requested):
Thanks! I think this is a cleaner change.

illuhad:
Can we finalize this PR? I would like to release AdaptiveCpp 24.10 this week, and I'd like this to be included.

fodinabor (approved):
LGTM :)
Thanks!

PR #1666, opened 2024-12-18
Fix atomic min/max on non-native integral types in HIP-like backends

psalz:
The previous implementation had a cast from `long long` to `unsigned long long`, which works for something like `atomicAdd`, where the bitwise result is identical regardless of the sign of the operands, but breaks for `atomicMin` and `atomicMax`.

### DIFF
--- a/include/hipSYCL/sycl/libkernel/generic/hiplike/atomic_builtins.hpp
+++ b/include/hipSYCL/sycl/libkernel/generic/hiplike/atomic_builtins.hpp
@@ -329,8 +329,9 @@ HIPSYCL_HIPLIKE_BUILTIN bool __acpp_atomic_compare_exchange_weak(
 // *********************** Integral values only *****************************
 
 // Defines overloads for integral atomics for types that are not natively
-// supported using bitcasts: long long, long and unsigned long
-#define HIPSYCL_DEFINE_HIPLIKE_ATOMIC_INTEGRAL_NONNATIVE_OVERLOADS(            \
+// supported using bitcasts (long long, long and unsigned long), for operations
+// that are agnostic to the sign of the operands (e.g. atomicAdd).
+#define HIPSYCL_DEFINE_HIPLIKE_ATOMIC_INTEGRAL_NONNATIVE_OVERLOADS_AGNOSTIC(   \
     name, bitcast_name)                                                        \
   template <access::address_space S, class InvokedT, class T>                  \
   HIPSYCL_HIPLIKE_BUILTIN T bitcast_name(T *addr, T x, memory_order order,     \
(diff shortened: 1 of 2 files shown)


psalz:
Great, [looks like](https://rocm.docs.amd.com/projects/HIP/en/docs-5.2.3/reference/kernel_language.html#atomic-functions) AMD forgot to copy-paste the `long long` overload of `atomicMin|Max`.

Also why am I getting an error from some ROCm header in the [CUDA build](https://github.com/AdaptiveCpp/AdaptiveCpp/actions/runs/12397123472/job/34606755561?pr=1666#step:7:350)?!

fodinabor:
> Also why am I getting an error from some ROCm header in the [CUDA build](https://github.com/AdaptiveCpp/AdaptiveCpp/actions/runs/12397123472/job/34606755561?pr=1666#step:7:350)?!

Thats an explicit multipass build - also builds hip.

illuhad:
looks like this is not really supported on AMD...also, are you on SSCP yet? It's free of this issue and should work ;)

psalz:
How should I proceed? Disable the overloads and test cases for AMD?

Edit: [Looks like the overloads were added in ROCm 5.7](https://github.com/ROCm/clr/issues/2)

illuhad:
> How should I proceed? Disable the overloads and test cases for AMD?

Another option would be to use the `__hip_atomic_fetch_min/max` builtins on AMD like SSCP is doing: https://github.com/AdaptiveCpp/AdaptiveCpp/blob/ba3fb96fbeb4a00929f4f93b57f0831e44ed903a/src/libkernel/sscp/amdgpu/atomic.cpp#L496

psalz:
> Another option would be to use the `__hip_atomic_fetch_min/max` builtins on AMD like SSCP is doing:
> 
> https://github.com/AdaptiveCpp/AdaptiveCpp/blob/ba3fb96fbeb4a00929f4f93b57f0831e44ed903a/src/libkernel/sscp/amdgpu/atomic.cpp#L496

I implemented the workaround, but now it looks like the AMD SSCP CI is failing the adjusted unit tests? Could it be that SSCP also has this bug? Unfortunately I don't have an AMD setup locally to test this.

illuhad:
> Could it be that SSCP also has this bug? Unfortunately I don't have an AMD setup locally to test this.

Unlikely. It uses the same builtin that you are using now...I've looked at the SSCP code path and it looks like it correctly distinguishes signed/unsigned, so either there is a bug in this builtin, or there is some other issue, or bug in the test or so.

include/hipSYCL/sycl/libkernel/generic/hiplike/atomic_builtins.hpp
```diff
@@ -517,11 +556,18 @@ __acpp_atomic_fetch_min(double *addr, double x, memory_order order,
 template <access::address_space S, class T>
 HIPSYCL_HIPLIKE_BUILTIN T __acpp_atomic_fetch_max(
     T *addr, T x, memory_order order, memory_scope scope) noexcept {
+#if defined(__HIP__)
+  // HIP does not support some of the overloads of atomicMax until ROCm 5.7
+  // https://github.com/ROCm/clr/issues/2
+  return __hip_atomic_fetch_max(addr, x, builtin_memory_order(order),
+      __HIP_MEMORY_SCOPE_AGENT);
```
illuhad, line 563: `__HIP_MEMORY_SCOPE_AGENT` is a pessimization for all scopes that are weaker than `device`, and incorrect in case a user wants to use `system` memory scope.
You will probably need a switch statement, like SSCP does: https://github.com/AdaptiveCpp/AdaptiveCpp/blob/7677cf6eefd8ab46d66168cd07ab042109448124/src/libkernel/sscp/amdgpu/atomic.cpp#L51

psalz: It should be the same as before, as `atomicMax` has device scope, and I didn't want to widen the scope (pun intended) of the PR. In any case I'm blocked by the SSCP issue and have no way of investigating this further 🤷‍♂️

illuhad: Ah, well... SMCP atomics are not great. I forgot that they don't handle that :( Alright then.

It might be useful to add some diagnostics to figure out which test exactly fails (which data type and operation). The atomic tests currently seem to be implemented as many checks in one test, so we can't see clearly where exactly things go wrong. Maybe it's not min/max with long/unsigned long that fails.

One difference I see in the atomic implementation between SMCP and SSCP is that SMCP uses some explicit CAS loops, whereas SSCP always calls the builtin and lets LLVM handle putting in CAS loops where needed.

Once I am back at the office I can take a look too.

PS: Happy holidays :)

psalz: > Maybe it's not min/max with long/unsigned long that fails.

You were right! I've added a context message to print the data type and operation, and it looks like ROCm SSCP is [failing for `atomicMin` and `atomicMax` on `float` and `double`](https://github.com/AdaptiveCpp/AdaptiveCpp/actions/runs/12545162134/job/34978935338?pr=1666#step:10:343). Since [I changed](https://github.com/AdaptiveCpp/AdaptiveCpp/pull/1666/files#diff-e1eca73484f076818fe8dd3928b360ceadd5bc3c1d9e66a2fbd38986d2315962R84) the way the input data is initialized from `i` to `-i` (with `i` in `[0, 2047]`), the results are clearly wrong. In fact, it looks like the two operations are doing exactly the opposite..?!

> PS: Happy holidays :)

Thanks! And to you!

psalz: Any thoughts on this?

illuhad: I can't reproduce the issue on LLVM 18 with ROCm 6.2. Our CI image is older and still on LLVM 15, ROCm 5.4. Most likely there is a bug in either old LLVM or old ROCm.

I think this might probably be a good moment to update our CI image, unless you're super motivated and want to add a workaround ;)

psalz: Sounds good (the updating the CI part, not the workaround ;-)). I'll rebase onto #1687 once that is green!

illuhad:
Oopsie, #1687 closed this because I mentioned in the description that the motivation for updating CI was to fix this issue - sorry!

Reopened and merged `develop` in now that the updated CI seems to work. Fingers crossed now :)

psalz:
Looking good! 👍

illuhad (approved):
Thanks!

PR #1670, opened 2024-12-23
Allow to override the default install locations

nullr0ute:
Usually CMake allows specifying, or detecting, where things are installed to, like lib vs lib64, so allow the build process to override locations where CMake has default variables for things like bin/lib/include so that users like Linux distros which might have different defaults can easily specify their defaults using the usual CMake variables.


nullr0ute:
This fixes building, and worked as expected, but I also need to add a patch to convert /usr/bin/acpp to find the right location as it's also hard coded there.

PR #1674, opened 2025-01-01
[CBS] Fix hanging on kernels with unreachable exits.

fodinabor:
This applies the patch from https://github.com/pocl/pocl/pull/1742


illuhad (approved):
Thanks!

PR #1675, opened 2025-01-06
Add Unified Runtime backend

VaiTon:
This is an attempt to add an Unified Runtime backend to AdaptiveCpp.

The code in this PR has been written taking inspiration from the OpenCL backend.

The new backend can be activated with the CMake `-DWITH_UNIFIED_RUNTIME_BACKEND=ON` option.

- [x] Listing devices via `acpp-info`
- [ ] Submitting kernels
- [ ] JIT via SPIR-V (to be tested, should work OOTB)

### DIFF
--- a/include/hipSYCL/runtime/device_id.hpp
+++ b/include/hipSYCL/runtime/device_id.hpp
@@ -24,6 +24,7 @@ enum class hardware_platform
   rocm,
   cuda,
   level_zero,
+  unified_runtime,
   ocl,
   cpu
 };
@@ -32,6 +33,7 @@ enum class api_platform {
   cuda,
   hip,
   level_zero,
+  unified_runtime,
   ocl,
   omp
 };
@@ -41,7 +43,8 @@ enum class backend_id {
   hip,
   level_zero,
   ocl,
-  omp
+  omp,
+  unified_runtime
 };
 
 struct backend_descriptor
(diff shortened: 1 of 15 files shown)


illuhad:
Wow, interesting. Can you elaborate a little on the motivation behind this work?

I don't think UR would allow us to target any hardware that we can not already target directly, so what is the benefit of pulling in an additional layer, which then also needs maintenance, testing etc?

VaiTon:
AFAIK It wouldn't add any hardware to the one we can target right now, as UR has adapters for OpenCL, Level Zero, CUDA, HIP and native CPU.

It is more of an experiment to see if ACPP could be stacked upon UR, which is what DPC++ uses to run on the different devices [^1].

[^1]: https://github.com/intel/llvm/blob/69da08871a2799dfd9941025e663f2a814282941/sycl/doc/FAQ.md#q-what-devices-are-supported-by-dpc-compiler

illuhad:
> AFAIK It wouldn't add any hardware to the one we can target right now, as UR has adapters for OpenCL, Level Zero, CUDA, HIP and native CPU.

Yes, this is also my understanding.

> It is more of an experiment to see if ACPP could be stacked upon UR, which is what DPC++ uses to run on the different devices

I'm aware of DPC++'s design. If this is just for experimentation purposes, does the experiment need to be upstreamed? Or are you just looking for some feedback without merging the PR?

VaiTon:
> If this is just for experimentation purposes, does the experiment need to be upstreamed?

Nope, it doesn't!

> Or are you just looking for some feedback without merging the PR?

It was just to signal that I'm working on this, if anyone is interested in experimenting on it (maybe, still don't know what could be gained by this but there is a lot of work going on on UR as far as I can see from git activity). I can close this one if you'd like to keep the list clean.

illuhad:
>  I can close this one if you'd like to keep the list clean.

No, that's alright, just wanted to know whether and what kind of reviews you were expecting :)

PR #1676, opened 2025-01-07
[SSCP][llvm-to-host] Internalize global variables

illuhad:
Not internalizing GVs that are generated e.g. for indexing into constant arrays can cause relocation errors at JIT time. Internalizing is in principle safe because all symbol resolution should be complete before doing the backend flavoring.

`__acpp_cbs_*` variables *must not* be internalized as this will again cause failures. It's not yet clear to me whether this is the only case where we must not internalize.
Otherwise, we could also try to internalize `constant` GVs, but I'm not sure that restriction is needed.

### DIFF
--- a/src/compiler/llvm-to-backend/host/LLVMToHost.cpp
+++ b/src/compiler/llvm-to-backend/host/LLVMToHost.cpp
@@ -91,10 +91,13 @@ bool LLVMToHostTranslator::toBackendFlavor(llvm::Module &M, PassHandler &PH) {
   if (!this->linkBitcodeFile(M, BuiltinBitcodeFile))
     return false;
 
-  // Internalize all global variables - this is fine because llvm-to-backend
-  // lowering always happens *after* linking in all dependencies
+  // Internalize all constant global variables that don't their definition
+  // to be imported from external sources - this is fine because llvm-to-backend
+  // lowering always happens *after* linking in all dependencies, and therefore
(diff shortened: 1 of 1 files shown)


fodinabor:
I believe this to be okay... I'm just wondering if there's anything we're potentially destroying here, that's implicit.. I'm thinking `errno` or things like that..

I'm still not sure if we maybe still should drop the PIC/PIE metadata, as it isn't helpful to keep the application's PIC mode for the JIT - we always want fPIC for the shared object.. Therefore keeping it, is likely going to come up again somewhere?

illuhad:
> I believe this to be okay... I'm just wondering if there's anything we're potentially destroying here, that's implicit.. I'm thinking errno or things like that..

I'm considering restricting this change to only apply to `constant` globals. This should be much more conservative. We could also restrict it to `constant` globals that have an initializer. I think this should still fix the problematic case, while avoiding anything that actually needs to be linked.

> I'm still not sure if we maybe still should drop the PIC/PIE metadata, as it isn't helpful to keep the application's PIC mode for the JIT - we always want fPIC for the shared object.. Therefore keeping it, is likely going to come up again somewhere?

It's been some time now, but IIRC I did try what you suggested but it didn't work. I'm not opposed to perhaps also drop PIC/PIE independently from this change.

fodinabor (approved):
LGTM

PR #1678, opened 2025-01-14
Handle Expected<T> from TempFile::create() to prevent potential crashes

devajithvs:
Add proper error handling for `TempFile::create()` by checking `takeError()` and `discard()`

I get a crash with the following error without this change:

```
Expected<T> must be checked before access or destruction.
Expected<T> value was in success state. (Note: Expected<T> values in
success mode must still be checked prior to being destroyed).
```

### DIFF
--- a/src/compiler/llvm-to-backend/amdgpu/LLVMToAmdgpu.cpp
+++ b/src/compiler/llvm-to-backend/amdgpu/LLVMToAmdgpu.cpp
@@ -92,7 +92,7 @@ bool getCommandOutput(const std::string &Program, const llvm::SmallVector<std::s
     return false;
   }
 
-  AtScopeExit DestroyOutputFile([&]() { auto Err = OutputFile->discard(); });
+  AtScopeExit DestroyOutputFile([&]() { consumeError(std::move(OutputFile->discard())); });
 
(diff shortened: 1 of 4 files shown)


illuhad (commented):
Sorry, this PR kind of went under my radar, apologies.

Do we really need the if-statements? My guess is that LLVM just checks whether the error has been cast to bool. In this case we could solve the problem without creating useless branching everywhere:

```c++
AtScopeExit DestroyInputFile([&]() {
  auto Err = InputFile->discard(); static_cast<bool>(Err);
});
```

We could make it a bit more readable using some thing like


```c++
#define ACPP_DISCARD_LLVM_ERROR(Expression) \
  {auto Err = (Expression); static_cast<bool>(Err); }

AtScopeExit DestroyInputFile([&]() {
  ACPP_DISCARD_LLVM_ERROR(InputFile->discard());
});
```

devajithvs:
Thanks @illuhad,

I realized that LLVM already provides consumeError() for this exact purpose. I've now replaced all branching occurrences with `consumeError()`

illuhad (approved):
Lovely, thanks! Let's just wait for CI to come back, then we can merge :)

PR #1679, opened 2025-01-17
Make inclusive and exclusive scans more generic

carbotaniuman:
This makes both inclusive and exclusive group scans more generic by supporting smart pointer types.


illuhad:
Hi, can you comment more on the use case? I think these functions can only be allowed inside kernels, and I don't think smart pointers can be supported inside kernels since there is no dynamic memory allocation/deallocation possible inside kernels.

carbotaniuman:
The main use case for this is `local_ptr` alongside probably the other `get_pointer` return types.

illuhad (approved):
Oh, I see. These are not tyically referred to as smart pointers, which had me confused. Got it now, this looks good, thank you! Let's wait for CI to come back, then we can merge if all is well.

PR #1681, opened 2025-01-17
Fix undefined reference in case HIPSYCL_ONDEMAND_ITERATION_SPACE_INFO is defined

sbalint98:
In case `HIPSYCL_ONDEMAND_ITERATION_SPACE_INFO` is defined the private members of the group class are not defined. Let's call the public interface to carry out the comparison. 

I encountered this issue during the SMCP build of GROMACS


illuhad (approved):
Thanks! Let's wait for CI to come back and then we can merge.

PR #1682, opened 2025-01-18
Fix implementation of group algorithms

carbotaniuman:
- Make sure that `sycl::group_broadcast` cannot accidentally become infinite via self recursion by using a separate namespace.
- Add group SFINAE to all functions without it for uniformity

### DIFF
--- a/include/hipSYCL/sycl/libkernel/sscp/group_functions.hpp
+++ b/include/hipSYCL/sycl/libkernel/sscp/group_functions.hpp
@@ -17,6 +17,7 @@
 #include "../detail/mem_fence.hpp"
 #include "../functional.hpp"
 #include "../group.hpp"
+#include "../group_traits.hpp"
 #include "../id.hpp"
 #include "../sub_group.hpp"
 #include "../vec.hpp"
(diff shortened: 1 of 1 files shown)


sbalint98:
Nice catch, Thanks! Lets wait for @illuhad to have a look at it but it looks good to me

include/hipSYCL/sycl/libkernel/sscp/group_functions.hpp
```diff
@@ -120,16 +124,8 @@ __acpp_group_broadcast(sub_group g, T x, typename sub_group::linear_id_type loca
...
-T __acpp_group_broadcast(sub_group g, T x,
-                  typename sub_group::id_type local_id) {
-
-  return __acpp_group_broadcast(g, x, static_cast<int>(local_id[0]));
-}
-
-
-template<typename T, int N, class Group>
+template<typename T, int N, class Group,
+          std::enable_if_t<is_group_v<std::decay_t<Group>>, bool> = true>
```
illuhad, line 129: We might want to include `group_traits.hpp` for `is_group_v`. Otherwise this will break if the file is ever included in an order that differs from what `sycl.hpp` currently does (e.g., in the stdpar case, we might want to directly include headers without going through the full SYCL header).

illuhad (commented):
Thanks, one small comment, looks great otherwise :)

illuhad (approved):

PR #1684, opened 2025-01-20
Add property_list parameters in USM allocation functions to match SYCL 2020 specification

thomasRoglin:
The SYCL 2020 specification  includes an optional property_list parameter to modify behavior of queues, [allocators](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_device_allocation_functions), reductions etc

As-of-today, AdaptiveCpp does not implement this parameter for USM memory allocators. 

This PR  proposes a first step toward this support.


illuhad (approved):
Lovely, thank you!

PR #1687, opened 2025-01-22
Update self-hosted CI image to ROCm 6.2, LLVM 18

illuhad:
Update self-hosted CI to ROCm 6.2 and LLVM 18.

This hopefully fixes #1666. The image is already live, so when CI passes on this PR, we should be good.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -41,7 +41,7 @@ jobs:
           destination: ./_site
 
       - name: Upload artifact
-        uses: actions/upload-pages-artifact@v1
+        uses: actions/upload-pages-artifact@v3
 
 
 
--- a/.github/workflows/linux-self-hosted.yml
+++ b/.github/workflows/linux-self-hosted.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: [self-hosted, gpu-nvidia]
     strategy:
       matrix:
-        clang_version: ['15']
+        clang_version: ['18']
         cuda: ['11.0']
         nvhpc_version: ['22.11']
     steps:
@@ -98,7 +98,7 @@ jobs:
     runs-on: [self-hosted, gpu-amd]
     strategy:
       matrix:
-        clang_version: ['15']
+        clang_version: ['18']
         cuda: ['11.0'] # Just to be able to build the backend for explicit multipass
     steps:
     - uses: actions/checkout@v4
@@ -154,7 +154,7 @@ jobs:
     runs-on: [self-hosted, gpu-intel]
     strategy:
       matrix:
-        clang_version: ['15']
+        clang_version: ['18']
     steps:
     - uses: actions/checkout@v4
     - name: clear JIT cache
--- a/.github/workflows/runner.def
+++ b/.github/workflows/runner.def
@@ -3,13 +3,16 @@ From: ubuntu:22.04
 
 %post
 export DEBIAN_FRONTEND=noninteractive
-export ROCM_VERSION="5.4.3"
+export ROCM_VERSION="6.2.4"
 
 export NVHPC_MAJOR_VERSION="22"
 export NVHPC_MINOR_VERSION="11"
 
 apt-get update
 apt-get install -y libboost-all-dev wget git libnuma-dev cmake curl unzip apt-transport-https ca-certificates software-properties-common sudo build-essential gettext libcurl4-openssl-dev openssh-client libnuma-dev jq libtbb-dev
+
+wget https://github.com/oneapi-src/level-zero/releases/download/v1.13.5/level-zero-devel_1.13.5+u22.04_amd64.deb
+wget https://github.com/oneapi-src/level-zero/releases/download/v1.13.5/level-zero_1.13.5+u22.04_amd64.deb
 
 wget https://github.com/intel/intel-graphics-compiler/releases/download/igc-1.0.16695.4/intel-igc-core_1.0.16695.4_amd64.deb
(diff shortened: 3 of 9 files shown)


illuhad:
@al42and This is a heads up for you due to a potential change in behavior when building against ROCm LLVM 6.2 or newer and using `--acpp-targets=hip`.

Turns out that ROCm LLVM 6.2 implements (and uses in its bitcode libraries) the `amdgcn.readlane.i32` intrinsic which is not available in mainline LLVM 18. `amdgcn.readlane.i32` is the generalized and renamed successor of `amdgcn.readlane` which *does* exist in vanilla LLVM 18.

Therefore, building AdaptiveCpp against mainline LLVM 18 fails in SMCP mode with ROCm 6.2, when certain group algorithms are invoked since this causes calls to this intrinsic to be pulled in from ROCm 6.2 bitcode libraries.

To get mainline LLVM to work, this PR adds a workaround, where `amdgcn.readlane.i32` is replaced in IR with its earlier  form `amdgcn.readlane`, when building with LLVM 18.
This workaround is disabled when `ROCM_CLANG_VERSION` is defined, i.e. we are building against ROCm clang/LLVM - in this case the replacement is not needed and would create calls to an intrinsic that no longer exists in this LLVM distribution.
`ROCM_CLANG_VERSION` is defined when we have the `ROCM_VERSION_MAJOR` variable in cmake, which in turn depends on parsing the output of `clang++ --version`: https://github.com/AdaptiveCpp/AdaptiveCpp/blob/f336ab843ffa52e54018417491cc892c00f8696a/CMakeLists.txt#L337

**Long story short**: If you're building against ROCm LLVM with a ROCm version based on LLVM 18, please double check that it can correctly parse `clang++ --version`. If this fails, then any code using this intrinsic will likely error during linking.

SSCP is not affected by any of this.

al42and:
Thanks for the heads-up and the workaround, @illuhad. Appreciate handling ROCm LLVM :)

I noticed the issue recently when updating the CI, but was a bit busy (*) and just switched to a build with no asserts: for us, this was only triggered by `__assert_fail`, which, in turn, used printf-related functionality; no asserts -- no problem. But we perhaps should add it to the installation docs.

(*) testing the performance impact of replacing `arr[i]` with `*reinterpret_cast<T*>(reinterpret_cast<char*>(arr) + i * sizeof(T))`, which led to nearly 10% improvement on MI250X for some kernels. That's not related to this issue, I'm just ranting.

illuhad:
> I noticed the issue recently when updating the CI, but was a bit busy (*) and just switched to a build with no asserts: for us, this was only triggered by __assert_fail, which, in turn, used printf-related functionality; no asserts -- no problem. But we perhaps should add it to the installation docs.

Oh, you had `amdgcn.readlane` use in `printf`? I didn't test that, but it's of course possible that ROCm `printf` uses this intrinsic.
I only noticed the issue when running the full set of unit tests, and it was only triggered by some specific group algorithm tests. So it's quite possible that you didn't encounter the problem otherwise.

> (*) testing the performance impact of replacing arr[i] with *reinterpret_cast<T*>(reinterpret_cast<char*>(arr) + i * sizeof(T)), which led to nearly 10% improvement on MI250X for some kernels. That's not related to this issue, I'm just ranting.

WTF. How does that make a difference, and how does it impact code generation?? My impression is that the amdgcn backend in LLVM is really erratic sometimes, but this is a whole new level...

al42and:
> Oh, you had amdgcn.readlane use in printf?

We had `assert`, which in turn had `printf`-related things in it.

We also use `__builtin_amdgcn_readfirstlane` directly, but only on 64-wide devices and I haven't tested mainline LLVM on those.

> WTF. How does that make a difference, and how does it impact code generation?? My impression is that the amdgcn backend in LLVM is really erratic sometimes, but this is a whole new level...

[AMD added it in their HIP fork a few years ago](https://github.com/ROCm/Gromacs/blob/d2cacc4257a125035128b6721ab36b06d2a3c9a3/src/gromacs/nbnxm/hip/nbnxm_hip_kernel_utils.hpp#L64-L77), but we low-key hoped that it got fixed during this time; and also the speedup is minor in some benchmarks, so not easy to notice when trying to quickly gauge the impact of different optimizations. But no, it is still necessary for register-heavy kernels in ROCm 6.3 (with ROCm LLVM; haven't tested upstream):

```
// These functions help hipcc to generate faster code for loads and atomic operations where
// 64-bit scalar + 32-bit vector registers are used instead of 64-bit vector saving a few
// instructions for computing 64-bit vector addresses.
```

illuhad:
> We also use __builtin_amdgcn_readfirstlane directly, but only on 64-wide devices and I haven't tested mainline LLVM on those.

I haven't looked into `readfirstlane`, only `readlane`, as the CI errors in our tests were triggered by `readlane`. The LLVM change suggests that `readfirstlane` might be affected as well:
https://github.com/llvm/llvm-project/pull/89217

The workaround currently only handles `readlane`. So if you try with mainline LLVM, it could be that you are still hitting this issue - if that is the case, let me know and we can also do the same workaround for `readfirstlane`.

> [AMD added it in their HIP fork a few years ago](https://github.com/ROCm/Gromacs/blob/d2cacc4257a125035128b6721ab36b06d2a3c9a3/src/gromacs/nbnxm/hip/nbnxm_hip_kernel_utils.hpp#L64-L77), but we low-key hoped that it got fixed during this time; and also the speedup is minor in some benchmarks, so not easy to notice when trying to quickly gauge the impact of different optimizations. But no, it is still necessary for register-heavy kernels in ROCm 6.3 (with ROCm LLVM; haven't tested upstream):

Yikes :/

PR #1691, opened 2025-01-23
Allow building with external OpenCL headers

VaiTon:
First step to make OCL backend packageable in offline environments (like Open Build Service). 

The `FIND_PACKAGE_ARGS` tells CMake to try `find_package` before pulling the repo.

Also, replaces `FetchContent_Populate` with `FetchContent_MakeAvailable`, as the former is deprecated for this purpose.

This is another try at #1613.

### DIFF
--- a/src/runtime/CMakeLists.txt
+++ b/src/runtime/CMakeLists.txt
@@ -215,33 +215,19 @@ endif()
 
 if(WITH_OPENCL_BACKEND)
   include(FetchContent)
-  FetchContent_Declare(ocl-headers
-    GIT_REPOSITORY https://github.com/KhronosGroup/OpenCL-Headers
-    GIT_TAG c860bb551eeef9a47d56286a70cea903db3d6ed2
-    CONFIGURE_COMMAND ""
-    BUILD_COMMAND ""
+  FetchContent_Declare(OpenCLHeaders
+    GIT_REPOSITORY "https://github.com/KhronosGroup/OpenCL-Headers"
+    GIT_TAG "c860bb551eeef9a47d56286a70cea903db3d6ed2"
+    FIND_PACKAGE_ARGS
   )
-  FetchContent_GetProperties(ocl-headers)
-  if(NOT ocl-headers_POPULATED)
-    FetchContent_Populate(ocl-headers)
-    add_subdirectory(${ocl-headers_SOURCE_DIR} ${ocl-headers_BINARY_DIR} EXCLUDE_FROM_ALL)
-  endif()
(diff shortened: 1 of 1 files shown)


tdavidcl:
We tried also to enable OpenCL for acpp on nixos, but couldn't do it in the end for that reason
See: https://github.com/NixOS/nixpkgs/pull/360893

VaiTon:
> We tried also to enable OpenCL for acpp on nixos, but couldn't do it in the end for that reason See: [NixOS/nixpkgs#360893](https://github.com/NixOS/nixpkgs/pull/360893)

Be aware that even after merging this we still download the LLVM-SPIR-V translator and patch it in the CMake, and that's why I specified this as a "first step".

illuhad:
IMO it will only be reasonably possible to build the OpenCL backend offline once we switch to `llc` instead of `llvm-spirv`, i.e. the SPIR-V backend in LLVM. This requires generic address space to be the default AS 0, instead of the ridiculous traditional SPIR convention in LLVM, which is a change that was only done recently and therefore will at least require LLVM 19, perhaps even something newer.

https://discourse.llvm.org/t/rfc-the-spir-v-backend-should-change-its-address-space-mappings/82640/17

As far as I can see this change seems to only be partially implemented in LLVM 19.

VaiTon:
Ouch. @illuhad does the runner use the same directory every time? Could we try putting an `rm -rf build` before building?

```
[ 11%] Creating directories for 'openclheaders-populate'
[ 22%] Performing download step (git clone) for 'openclheaders-populate'
-- Avoiding repeated git clone, stamp file is up to date: '/tmp/unihd-sycl-ci-fatboy-0/actions-runner/_work/AdaptiveCpp/AdaptiveCpp/build/_deps/openclheaders-subbuild/openclheaders-populate-prefix/src/openclheaders-populate-stamp/openclheaders-populate-gitinfo.txt'
[ 33%] Performing update step for 'openclheaders-populate'
```

illuhad:
> Ouch. @illuhad does the runner use the same directory every time? Could we try putting an rm -rf build before building?

I would have expected that the `actions/checkout` action that we use handles this. Also, while it does not create a new directory for every PR, I don't think it just operates on an existing directory. Otherwise this line here would fail:
https://github.com/AdaptiveCpp/AdaptiveCpp/blob/b528cd432c47e4c58657e4522704176205798f61/.github/workflows/linux-self-hosted.yml#L21

(note: `mkdir`, not `mkdir -p`, so that job would fail if the directory already existed)

PR #1692, opened 2025-01-24
Use execution_hints to propagate properties informations to backends allocator

thomasRoglin:
This PR builds upon  #1684 , to support properties on USM memory allocator.   
  
Goal is to support runtime allocators properties such as :  

```
int *p = sycl::malloc_device<int>(size, queue, sycl::property_list{my_property} );
```
  
As of today, SYCL has no predefined allocator-specific properties. Nonetheless adding this support could help to give, for instance, hints about NUMA allocation memory nodes.  
  
This PR propagates properties expressed  by the developer to each backend allocators.
This has been done by reusing the `hints` mechanism (already used for `queues`, `handlers`, `buffers` and `dags`) to be consistent with the rest of AdaptiveCpp.  
  
A closer look to AdpativeCpp reveal that  `hints` are stored in a class member of `dag_node`.
This solution is not well suited for the `backend_allocator` class since two consecutives allocations may have differents properties.

Therefore I opted to pass `hints` as function argument of every function of each backend allocators.

### DIFF
--- a/include/hipSYCL/runtime/allocator.hpp
+++ b/include/hipSYCL/runtime/allocator.hpp
@@ -13,6 +13,7 @@
 
 #include "device_id.hpp"
 #include "error.hpp"
+#include "hints.hpp"
 
 namespace hipsycl {
 namespace rt {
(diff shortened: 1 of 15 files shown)


illuhad:
Hi Thomas, this is just about customizing the way that allocations work, right?

In that case it might be better not to use `execution_hints` but a dedicated property struct for allocations. `execution_hints` are intended for things that affect task submission. Putting new fields in there that only affect allocation would mean unnecessarily bloating up the data that needs to be hauled around for every kernel launch.

thomasRoglin:
Thank for you @illuhad  for your feedback.

Yes this PR only focuses on modifying the allocation behavior. 

I just updated the branch to account your comment. I opted for a fresh `allocation_hint` class to be consistent with `execution_hint`. 
Nonetheless part of the code could be factorized between these two classes. 
Do you prefer to keep it like this or do you want me to refactorize?

illuhad:
> Nonetheless part of the code could be factorized between these two classes.
Do you prefer to keep it like this or do you want me to refactorize?

No, I don't think we need to refactor things. Actually, for the purpose of this PR, I think it would even be sufficient to just have `allocation_hint` be an empty struct. 

Without yet having actual use cases for it, trying to make something more elaborate or share code with `execution_hints`  might lead to overengineering since we don't know yet what we actually want to do with it. Perhaps it might be sufficient to just store these hints in a very simple manner like so:

```
struct allocation_hints {
  std::optional<int> numa_domain;
  std::optional<std::size_t> some_other_property;
};
```

Would you prefer to mirror the `execution_hints` design, or would you prefer something simpler?

thomasRoglin:
Thanks for you feedback. 
I think you are right about not overengineering hints for this PR.
I just update code to have an empty structure `allocation_hints`

illuhad (approved):
Thanks Thomas!

PR #1693, opened 2025-01-24
Remove unused includes and move includes from header to source in GlobalsPruningPass

VaiTon:
Other than being a best practice, it reduces the amount of code included every time the .hpp file gets included.


VaiTon:
Tests also failing on develop

illuhad (approved):
Thanks!

PR #1694, opened 2025-01-24
Add `local_accessor` replacing local `accessor`s

carbotaniuman:
This also adds a `get_multi_ptr` method, allowing users to avoid deprecation warnings.

I accidentally deleted the branch for the last PR, which caused GitHub to close the PR.

### DIFF
--- a/include/hipSYCL/sycl/libkernel/accessor.hpp
+++ b/include/hipSYCL/sycl/libkernel/accessor.hpp
@@ -398,6 +398,61 @@ private:
   mutable sycl::id<dimensions> _access_id;
 };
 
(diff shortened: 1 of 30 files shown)


illuhad (commented):
Thanks! Can you scan the code base for uses of the old local accessor (e.g. search for `target::local` uses) and replace with `local_accessor`? I think there are still quite a lot of tests that use the old one, and I don't want to create a lot of deprecation warnings there.

carbotaniuman:
I've changed out all the uses of `target::local` I've found. There's one that uses the atomic access mode, which I can't swap over, but that mode is already deprecated in SYCL 2020.

illuhad (approved):
Thanks!

PR #1695, opened 2025-01-25
Add tuple support for group functions

carbotaniuman:
This adds support to broadcast, both shifts, select, and permute for tuple-like classses that can be destructured, such as `std::tuple` or oneDPL's internal tuple.

### DIFF
--- a/include/hipSYCL/sycl/libkernel/detail/tuple_helper.hpp
+++ b/include/hipSYCL/sycl/libkernel/detail/tuple_helper.hpp
@@ -1,0 +1,76 @@
+/*
+ * This file is part of AdaptiveCpp, an implementation of SYCL and C++ standard
+ * parallelism for CPUs and GPUs.
+ *
+ * Copyright The AdaptiveCpp Contributors
+ *
+ * AdaptiveCpp is released under the BSD 2-Clause "Simplified" License.
+ * See file LICENSE in the project root for full license details.
+ */
+// SPDX-License-Identifier: BSD-2-Clause
+#ifndef HIPSYCL_TUPLE_HELPER_HPP
+#define HIPSYCL_TUPLE_HELPER_HPP
+
+#include <cstddef>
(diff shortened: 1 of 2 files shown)


illuhad:
I think this change is not in line with the SYCL specification, which does not define `tuple` support for these functions:

> All algorithms are supported for the fundamental scalar types supported by SYCL (see [Table 172](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#table.types.fundamental)) and instances of the SYCL vec and marray classes.

https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:algorithms

carbotaniuman:
Hmm yeah, this seems to be an extension that DPCPP++ implements, so feel free to close, but it doesn't feel like that spec restricts extending the allowed types.

illuhad:
That's true, it would be possible to add an extension here. That would require defining and documenting it as such, and having a clear value proposition why it's useful to add an extension and bear the maintenance burden.
For an extension it would probably also be desirable to support it uniformly across all compilation flows, not just SSCP.

carbotaniuman:
Ah yeah, the other flows... This was added to support oneDPL and then extended to allow something resembling composite types to be used - but not sure whether that's valuable enough to upstream.

illuhad:
oneDPL support is probably not really part of the strategy of AdaptiveCpp, since there are many DPC++ assumptions in there - even if we add all the DPC++ extensions to make it work now, chances are that it will very likely be broken again in the future.
AdaptiveCpp ships its own parallel algorithm library (https://github.com/AdaptiveCpp/AdaptiveCpp/blob/develop/doc/algorithms.md) that we can make much stronger support and optimization guarantees for with AdaptiveCpp, and I'd rather push people towards that.

carbotaniuman:
Yeah, makes sense.

PR #1696, opened 2025-01-25
[NFC] Update the year listed by the acpp compilation driver

marcosolanki:
While the recent refactoring of our copyright headers has eliminated most hard-coded references to the current year, the one remaining has once again gone out of date.


illuhad (approved):
Thanks!

PR #1700, opened 2025-01-30
fix clz on hip SMCP

tdavidcl:
On hip SMCP the return value of clz(0) is inconsistent. As a result a fix was introduced in #1687 but was inoperant as llvm was optimizing the if away. This pr avoid the optimization by using a noinline to force the check and hopefully fix clz on hip SMCP.

This PR require many run of the tests as the issue is somewhat random in the first place...


illuhad (commented):
Can we limit this to the actually problematic case? having a non-inlined function call will have a performance impact, even on e.g. NVIDIA where clz was working fine.

Even on AMD not all LLVM/ROCm combinations are not affected - ideally we could figure out what exactly causes the problem.

tdavidcl:
> Can we limit this to the actually problematic case? having a non-inlined function call will have a performance impact, even on e.g. NVIDIA where clz was working fine.
> 
> Even on AMD not all LLVM/ROCm combinations are not affected - ideally we could figure out what exactly causes the problem.

Indeed, my bad, i totally forgot that hiplike was also used for CUDA while coding it. I'll change that this weekend

illuhad:
This is still waiting for the weekend's change ;)

tdavidcl:
Yeah I know 😅 I have to finish it

include/hipSYCL/sycl/libkernel/generic/hiplike/builtins.hpp
```diff
@@ -450,7 +456,19 @@ HIPSYCL_HIPLIKE_BUILTIN T __acpp_clz(T x) noexcept {
   constexpr T size = CHAR_BIT*sizeof(T);
 
   auto v = static_cast<__acpp_int32>(static_cast<Usigned>(x));
-  return v ? __clz(v)-diff : size;
+
+  #if ACPP_LIBKERNEL_IS_DEVICE_PASS_CUDA 
+    // here we force noinline on clz to avoid the if(v) to be optimized away as
+    // llvm rightfully assume that clz(0) == bitsize
+    // see : https://llvm.org/docs/LangRef.html#llvm-ctlz-intrinsic
```
illuhad: I don't think the comment applies to this branch of the ifdef, right?

include/hipSYCL/sycl/libkernel/generic/hiplike/builtins.hpp
```diff
@@ -450,7 +456,19 @@ HIPSYCL_HIPLIKE_BUILTIN T __acpp_clz(T x) noexcept {
...
+    return v ? __builtin_clz(v)-diff : size;
+  #else
+    // on hip we have to circumvent the clz(0) bug
+    // here we force noinline on clz to avoid the if(v) to be optimized away as
+    // llvm rightfully assume that clz(0) == bitsize
```
illuhad:
```suggestion
    // llvm rightfully assume that clz(0) == bitsize. However, in some LLVM versions, the amdgpu
    // backend does not seem to honor this guarantee from the LLVM LangRef.
```
I think this was the issue, IIRC?

illuhad (approved):
Thanks, LGTM - perhaps some small improvements could be made to the comments if you're motivated :)

PR #1703, opened 2025-02-04
Extract getPointerType and ocl info_query to their own headers

VaiTon:
This allow building AdaptiveCpp using [unity builds](https://cmake.org/cmake/help/latest/prop_tgt/UNITY_BUILD.html). 

This mode merges .cpp files via `#includes` into a single file and then compile it. That clashes with .cpp having functions with the same name (even in anonymous namespaces).

This can speed up compilation by a lot.

### DIFF
--- a/include/hipSYCL/compiler/LLVMUtils.hpp
+++ b/include/hipSYCL/compiler/LLVMUtils.hpp
@@ -1,0 +1,24 @@
+/*
+ * This file is part of AdaptiveCpp, an implementation of SYCL and C++ standard
+ * parallelism for CPUs and GPUs.
+ *
+ * Copyright The AdaptiveCpp Contributors
+ *
+ * AdaptiveCpp is released under the BSD 2-Clause "Simplified" License.
+ * See file LICENSE in the project root for full license details.
+ */
+// SPDX-License-Identifier: BSD-2-Clause
+#ifndef HIPSYCL_SSCP_LLVMUTILS_H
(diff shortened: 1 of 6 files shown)


illuhad:
> That clashes with .cpp having functions with the same name (even in anonymous namespaces).

I'm not willing to make such guarantees that go beyond what C++ allows. This is the same as mandating that the project now adhere only to some dialect or subset of C++. Unless there is some very, very good reason backed up by data for this, it's not clear to me why we should commit to something like this.

VaiTon:
> This is the same as mandating that the project now adhere only to some dialect or subset of C++. 

You're right, but with this small change (along with some code deduplication), we could enable unity builds for the project. Of course, you could always merge something tomorrow that breaks it again. :)  

> Unless there is some very, very good reason backed up by data for this, it's not clear to me why we should commit to something like this.

On my machine (AMD Ryzen 7 PRO 4750U, 8 cores), the build time drops from ~90 seconds to ~60. It’s not a huge difference, but we still have a lot of unnecessary headers in source files, especially those related to LLVM.

> There are also efforts by people who build chrome often (and have gotten very good results - parts of it get built in 30% of the original time) to bring native support for unity builds into clang to minimize the code changes needed for the technique (gives unique names to objects with internal linkage (static or in anonymous namespaces), undefines macros between source files, etc.):

(https://onqtam.com/programming/2018-07-07-unity-builds/)

PR #1704, opened 2025-02-05
runtime: precompile most used headers

VaiTon:
[Precompiling headers](https://cmake.org/cmake/help/latest/command/target_precompile_headers.html) can accelerate build times by preventing headers from being recompiled with each inclusion.  

This patch updates the `PRECOMPILE_HEADERS` property of relevant targets to include the most frequently used headers.

### DIFF
--- a/src/runtime/CMakeLists.txt
+++ b/src/runtime/CMakeLists.txt
@@ -199,6 +199,8 @@ if(WITH_LEVEL_ZERO_BACKEND)
   target_compile_options(rt-backend-ze PRIVATE ${HIPSYCL_RT_EXTRA_CXX_FLAGS})
   target_link_libraries(rt-backend-ze PRIVATE ${HIPSYCL_RT_EXTRA_LINKER_FLAGS})
 
+  target_precompile_headers(rt-backend-ze PRIVATE <level_zero/ze_api.h>)
+
   if(WITH_SSCP_COMPILER)
     target_compile_definitions(rt-backend-ze PRIVATE -DHIPSYCL_WITH_SSCP_COMPILER)
     target_link_libraries(rt-backend-ze PRIVATE llvm-to-spirv)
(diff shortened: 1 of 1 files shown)


illuhad:
"most frequently used headers" - these headers are only needed when building AdaptiveCpp (at least when using `--acpp-targets=generic`).
I am a) skeptical that the workflow of most users is limited by the time it takes to build the AdaptiveCpp runtime and b) that these particular headers are responsible for the build times.

If there's value in using PCH for AdaptiveCpp, then I see it primarily in using it with `sycl.hpp`, although PCH like would only work with the generic JIT compiler.

VaiTon:
> "most frequently used headers" - these headers are only needed when building AdaptiveCpp

Maybe I wasn’t clear. Yes, I’m specifically referring to project build time (not the files compiled with acpp).

>  I am a) skeptical that the workflow of most users is limited by the time it takes to build the AdaptiveCpp runtime and b) that these particular headers are responsible for the build times.

That’s true—users typically only need to build once. However, I believe reducing development compile time is also an important goal. Additionally, this seems like an easy optimization since nearly every file in those directories includes omp.h, CL/cl.h, ecc.

illuhad:
> That’s true—users typically only need to build once. However, I believe reducing development compile time is also an important goal. Additionally, this seems like an easy optimization since nearly every file in those directories includes omp.h, CL/cl.h, ecc.

Yes, but those are C headers, mostly containing only declarations.... Have you looked at `clang -ftime-trace` or similar? I'd bet that compile times for those pale in comparison to C++ standard library stuff that we use...
There's also value in not over-complicating the build process.

AdaptiveCpp for me builds in <5 seconds with all backends enabled.

VaiTon:
Yeah... this was just a starting point to test this feature on the codebase, as my first idea was to add llvm headers to PCH, as the longest compile times are the one of files in the `compiler` directory (~1 min only AdaptiveCppClangPlugin on my machine :/)

illuhad:
Oh..What kind of machine do you have?

If this is just an experiment, can we please not use upstream AdaptiveCpp for that?

EDIT: My guess is that LLVM code is always going to take its time compiling. Even if there is a PCH, the frontend might still have to do a lot of template instantiations etc...

VaiTon:
> If this is just an experiment, can we please not use upstream AdaptiveCpp for that?

I mean, it isn't an experiment. Instead of one big PR I thought it was better to break it into smaller chunks.

Still, if you don't think this single change is worth upstreaming, I can close this np

illuhad:
I like small PRs in principle :)

If you'd like to do bigger changes, it might be a good idea to open an RFC issue or similar first to discuss what you want to achieve first, and then start implementation work once there is consensus how to achieve that in the issue. In that way, there's less risk for you that things won't get merged, and it's clear to reviewers how your smaller PRs fit into the larger picture :)

PR #1705, opened 2025-02-05
Remove the use of raw pointers in backend.hpp get(backend_id)

VaiTon:
Rename backends() to backend_mgr() in device.hpp and introduce a backend() function in backend.hpp, to remove the unneeded for_each lambda.

Adapt user code.

### DIFF
--- a/include/hipSYCL/runtime/backend.hpp
+++ b/include/hipSYCL/runtime/backend.hpp
@@ -64,22 +64,16 @@ class backend_manager
 {
 public:
   using backend_list_type =
-      std::vector<std::unique_ptr<backend>>;
+      std::vector<std::shared_ptr<backend>>;
 
   backend_manager();
   ~backend_manager();
-  
-  backend* get(backend_id) const;
+
+  const backend_list_type &backends() const { return _backends; }
+  std::shared_ptr<backend> get(backend_id) const;
+
   hw_model& hardware_model();
   const hw_model& hardware_model() const;
-
-  template<class F>
-  void for_each_backend(F f)
-  {
-    for(auto& b : _backends){
-      f(b.get());
-    }
-  }
 
 private:
   backend_loader _loader;
--- a/include/hipSYCL/algorithms/util/allocation_cache.hpp
+++ b/include/hipSYCL/algorithms/util/allocation_cache.hpp
@@ -53,7 +53,7 @@ public:
     std::lock_guard<std::mutex> lock{_mutex};
     
     for(auto& allocation : _allocations) {
-      auto* allocator = _rt.get()->backends()
+      auto* allocator = _rt.get()->backend_mgr()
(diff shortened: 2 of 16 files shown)


include/hipSYCL/runtime/backend.hpp
```diff
@@ -64,23 +64,17 @@ class backend_manager
 {
 public:
   using backend_list_type =
-      std::vector<std::unique_ptr<backend>>;
+      std::vector<std::shared_ptr<backend>>;
```
illuhad, line 67: I stopped reading here. `backend` IMO should be owned by the `backend_manager`. We should not allow a situation where stale backends are still alive somehow because *something* still holds a shared ptr to them.
(If anything, I'd like to *reduce* the amount of `shared_ptr` we have, not increase it)

This is a fundamental change to the ownership model of runtime objects, and I don't understand what you even hope to achieve with this PR.

VaiTon: We can keep a vector of `unique_ptr` there, no problem. The problem is that we pass raw pointers around using `.get`, and that's something I want to reduce as much as possible.

illuhad: Why? If the ownership and lifetime model is clear, there is no danger to that.

VaiTon: >  because something still holds a shared ptr to them.
(If anything, I'd like to reduce the amount of shared_ptr we have, not increase it)

Because if something still holds a ptr to a deallocated backend that's a sigsegv. What about using weak_ptrs?

illuhad: > Because if something still holds a ptr to a deallocated backend that's a sigsegv

Not possible because of the runtime lifetime management model. The runtime and its managed backends cannot die while there is a user-facing SYCL object (like `sycl::device`, `sycl::queue` etc) alive.
Effectively, these classes hold something similar like a `shared_ptr` to the entire runtime. This design allows us to have only a single `shared_ptr` instead of having to manage lifetime of runtime objects at lower granularity for no reason.

`shared_ptr` has a cost and so does `weak_ptr` - perhaps even more so because in order to access a `weak_ptr`, you first need to construct `shared_ptr` and there is also some synchronization involved.

VaiTon: I understand, thank you for your explanation and time :)

illuhad: No problem :) If you are interested in `shared_ptr` and lifetime management stuff, as I said I'd be interested in reducing the amount of `shared_ptr` for performance reasons. In particular, I wonder whether `dag_node_ptr` needs to be a `shared_ptr` or whether a raw pointer would be sufficient due to the same reasoning.

PR #1709, opened 2025-02-05
Do not modify the input data for joint group scans

sbalint98:
As spotted by @al42and in https://github.com/AdaptiveCpp/AdaptiveCpp/pull/1651#discussion_r1940269492 The joint group scan algorithms modify the input data, which although allowed by the SYCL standard, not in line with user expectations. 

This MR changes this behaviour

### DIFF
--- a/include/hipSYCL/sycl/libkernel/sscp/group_functions.hpp
+++ b/include/hipSYCL/sycl/libkernel/sscp/group_functions.hpp
@@ -541,7 +541,7 @@ __acpp_joint_reduce(Group g, Ptr first, Ptr last, BinaryOperation binary_op) {
   const size_t num_elements = last - first;
   const size_t lid          = g.get_local_linear_id();
 
-  using value_type = std::remove_reference_t<decltype(*first)>;
+  using value_type = std::remove_cv_t<std::remove_reference_t<decltype(*first)>>;
 
   if(num_elements == 0)
     return value_type{};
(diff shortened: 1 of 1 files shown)


include/hipSYCL/sycl/libkernel/sscp/group_functions.hpp
```diff
@@ -743,48 +747,36 @@ HIPSYCL_BUILTIN OutPtr __acpp_joint_inclusive_scan(Group g, InPtr first, InPtr l
...
+
   size_t segment = 0;
   size_t num_segments = (num_elements + lrange - 1) / lrange;
 
-  // for (Ptr p = start_ptr + lrange; p < last; p += lrange){
   for (size_t segment = 0; segment < num_segments; segment++) {
     size_t element_idx = segment * lrange + lid;
     auto local_element = element_idx < num_elements ? first[element_idx] : identity;
-    auto segment_result = __acpp_inclusive_scan_over_group(g, local_element, binary_op);
+    value_type segment_result;
```
al42and, line 757: Needs `std::remove_const`, in case `value_type` is `const T`. Or use a tertiary

sbalint98: Good point, thanks. Added `remove_const`. I was wondering if `std::decay_t` would be better here, but remove_ref, remove_const is probably more expressive...

al42and: > I was wondering if std::decay_t would be better here, but remove_ref, remove_const is probably more expressive...

I only suggested `remove_const` because that's the first idea that came to my mind :)

Thanks for fixing, now all GROMACS 2025.0 kernels build fine with SSCP :) And on RTX3060, there is a net win in performance compared to the multipass mode :+1: (on a per-kernel basis, it's a mixed bag).

al42and (approved):

include/hipSYCL/sycl/libkernel/sscp/group_functions.hpp
illuhad: @al42and you should get in touch with @sbalint98 , he has quite the experience around performance analysis of SSCP with Gromacs :)

al42and: That's why I was a bit surprised that our reduction kernel from GROMACS 2025 failed to build :smiley_cat:

sbalint98: Yeah I was mainly benchmarking the 2024.5 release with some extra changes to enable the generic target. That one does not have the sci sorting yet I think.

al42and: Besides the sci sorting (which should have the same effect on SSCP), we have introduces some [optimizations](https://gitlab.com/gromacs/gromacs/-/merge_requests/4858/diffs) for multipass HIP, so at least on MI250 the performance noticeably improved. If SSCP avoid the need for such cursed workarounds, that would be pretty great.

PR #1710, opened 2025-02-07
Minor CTS fixes

carbotaniuman:
This addresses a few changes I had to make while testing some other changes with the CTS.

This PR mainly moves half into its own namespace to avoid some ADL issues, as well as overloads the relational operators for some integer types in order to remove ambiguity in the CTS and in some implementation. I also deprecated some things marked deprecated while I was making these changes.

### DIFF
--- a/include/hipSYCL/sycl/libkernel/bit_cast.hpp
+++ b/include/hipSYCL/sycl/libkernel/bit_cast.hpp
@@ -19,7 +19,7 @@ namespace sycl {
 
 template <class Tout, class Tin>
 ACPP_UNIVERSAL_TARGET
-Tout bit_cast(Tin x) {
+Tout bit_cast(const Tin& x) {
   Tout result;
   HIPSYCL_INPLACE_BIT_CAST(Tin, Tout, x, result);
   return result;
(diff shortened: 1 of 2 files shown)


include/hipSYCL/sycl/libkernel/half.hpp
```diff
@@ -149,72 +152,128 @@ class half {
...
-      return __acpp_sscp_half_lte(a._data, b._data),
-      return __hle(fp16::as_cuda_half(a._data), fp16::as_cuda_half(b._data)),
-      return fp16::builtin_less_than_equal(a._data, b._data))
-  }
+  OP(<)
+  OP(<=)
+  OP(>)
+  OP(>=)
+  OP(==)
+  OP(!=)
```
illuhad: Is it actually correct to implement these operators by casting everything to `half`, even when this causes loss of information? For example, when comparing with `double`, I would expect the `half` to be cast to `double` and the comparison being carried out with double precision.

illuhad (commented):
Thanks, this looks mostly good, just one comment :)

include/hipSYCL/sycl/libkernel/half.hpp
carbotaniuman: I decided to rip out all the half changes from this PR and put them in #1710.

illuhad (approved):
Thanks!

PR #1711, opened 2025-02-07
Refactor float builtins that take an int argument

carbotaniuman:
This changes the SSCP form associated 32-bit ints with float and 64-bit ints with double - change them both to 32-bit ints for consistency and code simplification. We were really just converting back and forth for no reason.

### DIFF
--- a/src/libkernel/sscp/amdgpu/math.cpp
+++ b/src/libkernel/sscp/amdgpu/math.cpp
@@ -111,11 +111,16 @@ HIPSYCL_SSCP_BUILTIN float __acpp_sscp_frexp_f32(float x, __acpp_int32 *y) {
   }
 }
 
-HIPSYCL_SSCP_BUILTIN double __acpp_sscp_frexp_f64(double x, __acpp_int64 *y) {
+HIPSYCL_SSCP_BUILTIN double __acpp_sscp_frexp_f64(double x, __acpp_int32 *y) {
   __acpp_int32 w;
-  double res = __ocml_frexp_f64(x, to_private(&w));
-  *y = static_cast<__acpp_int64>(w);
-  return res;
+  if(__ockl_is_private_addr(y))
+    return __ocml_frexp_f64(x, to_private(y));
+  else {
+    __acpp_int32 w;
+    float res = __ocml_frexp_f64(x, to_private(&w));
+    *y = w;
+    return res;
+  }
 }
 
 HIPSYCL_SSCP_MAP_OCML_FLOAT_BUILTIN2(hypot, __ocml_hypot)
@@ -125,7 +130,7 @@ HIPSYCL_SSCP_BUILTIN float __acpp_sscp_ldexp_f32(float x, __acpp_int32 k) {
   return __ocml_ldexp_f32(x, k);
 }
 
-HIPSYCL_SSCP_BUILTIN double __acpp_sscp_ldexp_f64(double x, __acpp_int64 k) {
+HIPSYCL_SSCP_BUILTIN double __acpp_sscp_ldexp_f64(double x, __acpp_int32 k) {
   return __ocml_ldexp_f64(x, k);
 }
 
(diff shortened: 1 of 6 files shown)


include/hipSYCL/sycl/libkernel/builtin_interface.hpp
```diff
@@ -181,8 +181,8 @@ HIPSYCL_BUILTIN T __acpp_fmod(T x, T y) noexcept {
   HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_fmod, x, y);
 }
 
-template<class T>
-T __acpp_fract(T x, T* ptr) noexcept {
+template<class T, class FloatPtr>
+T __acpp_fract(T x, FloatPtr ptr) noexcept {
```
illuhad: I think this will fail for `multi_ptr` or other pointer types, as the pointer here is directly forwarded to the dispatch-builtin macro and may therefore end up e.g. in SSCP builtin calls which only exist for specific raw pointer types.

If we allow any type that provides a pointer interface, we should ensure at a high-level that the type that is actually passed into backend builtins remains a supported pointer type.

src/libkernel/sscp/amdgpu/math.cpp
```diff
@@ -111,11 +111,16 @@ HIPSYCL_SSCP_BUILTIN float __acpp_sscp_frexp_f32(float x, __acpp_int32 *y) {
...
-  *y = static_cast<__acpp_int64>(w);
-  return res;
+  if(__ockl_is_private_addr(y))
+    return __ocml_frexp_f64(x, to_private(y));
+  else {
+    __acpp_int32 w;
+    float res = __ocml_frexp_f64(x, to_private(&w));
+    *y = w;
+    return res;
+  }
```
illuhad, line 123: Is there actually a performance benefit for this additional complexity? It looks to me like we're optimizing away one copy to `w` at the expense of more complex control flow, which might eat up the benefit.

src/libkernel/sscp/amdgpu/math.cpp
```diff
@@ -144,11 +149,15 @@ HIPSYCL_SSCP_BUILTIN float __acpp_sscp_lgamma_r_f32(float x, __acpp_int32* y ) {
...
-  return res;
+HIPSYCL_SSCP_BUILTIN double __acpp_sscp_lgamma_r_f64(double x, __acpp_int32* y) {
+  if(__ockl_is_private_addr(y)) {
+    return __ocml_lgamma_r_f64(x, to_private(y));
+  } else {
+    __acpp_int32 w;
+    auto res = __ocml_lgamma_r_f64(x, to_private(&w));
+    *y = w;
+    return res;
+  }
```
illuhad, line 160: Same question here?

src/libkernel/sscp/ptx/math.cpp
```diff
@@ -101,11 +101,8 @@ HIPSYCL_SSCP_BUILTIN float __acpp_sscp_frexp_f32(float x,
   return __nv_frexpf(x, y);
 }
 HIPSYCL_SSCP_BUILTIN double __acpp_sscp_frexp_f64(double x,
-                                                     __acpp_int64 *y) {
-  __acpp_int32 w;
-  double res = __nv_frexp(x, &w);
-  *y = static_cast<__acpp_int64>(w);
-  return res;
+                                                     __acpp_int32 *y) {
+  return __nv_frexpf(x, y);
```
illuhad: This seems wrong, shouldn't it be `__nv_frexp` for `double` type?

src/libkernel/sscp/amdgpu/math.cpp
carbotaniuman: I copied this from the preexisting code above when I changed the second parameter type to match (and same for the other comment reference this).

```c++
HIPSYCL_SSCP_BUILTIN float __acpp_sscp_frexp_f32(float x, __acpp_int32 *y) {
  if(__ockl_is_private_addr(y))
    return __ocml_frexp_f32(x, to_private(y));
  else {
    __acpp_int32 w;
    float res = __ocml_frexp_f32(x, to_private(&w));
    *y = w;
    return res;
  }
}
```

include/hipSYCL/sycl/libkernel/builtin_interface.hpp
carbotaniuman: Yeah, my ideal solution would be to change the `HIPSYCL_RETURN_DISPATCH_BUILTIN` call to handle passing it in a raw pointer, as that would be a decent decrease in complexity, but the functions here directly delegate and I wasn't sure if that was intentional. Something like the following would be ideal imo (and could provide future hooks for private addr spaces and all:

```
template<class T, class FloatPtr, std::enable_if_t<std::is_same_v<FloatPtr, T*>, bool> = true>
T __acpp_fract(T x, FloatPtr ptr) noexcept {
  HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_fract, x, ptr);
}

template<class T, class FloatPtr, std::enable_if_t<!std::is_same_v<FloatPtr, T*>, bool> = true>
T __acpp_fract(T x, FloatPtr ptr) noexcept {
  HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_fract, x, ptr.get());
}
```

carbotaniuman: It seems that I was mistaken by how some of the builtins implementation worked, and it appears that this wouldn't really solve the problem. I've ripped out the changes related to this until I can come back and take a better look at the whole dispatch chain.

src/libkernel/sscp/amdgpu/math.cpp
illuhad: Oh, I see. I missed that this pattern already existed.

illuhad (approved):
Thanks!

PR #1713, opened 2025-02-09
Allow building ACpp without Boost

al42and:
The vestigial functionality was there, but not easily usable.
Fixed and refreshed:

- Rearranged the CMake handling of Boost to minimize duplication and group
  Boost-related things together.
- New `ACPP_NO_FIBERS` CMake option to disable the need for Boost even
  when the accelerated CPU is disabled. In this case, we can't run
  nd_range kernels on the CPU, but the rest still works.
  - It replaces the old, undocumented `HIPSYCL_NO_FIBERS` option passed
    when compiling to user code. It's ACpp's job to figure out if Boost is
    needed or not.
- Boost is not searched and not linked anymore when accelerated CPU is enabled,
  which is the most common use-case. No reason to require library that's never used.
  - This can break things if someone builds ACpp with accelerated CPU enabled,
    but then passes the compiler flag to disable it: even if Boost is
    present, we will fall back to "semi-functional" path. Do we care about such cases?
- Boost is still needed for tests and for the rare case of broken `std::filesystem`.

Draft because I'm not 100% sure if I did not miss some potential use case where
we still want to let the user decide whether we want boost or not when invoking `acpp`.

Another alternative would be to move it back to user-selection, adding, e.g., `--acpp-no-boost` option.

Also, do we want to keep these two recommendations? 

```
          " * users targeting SYCL 1.2.1 may use hierarchical parallel for, which\n"
          "   can express the same algorithms, but may have functionality caveats in AdaptiveCpp\n"
          "   and/or other SYCL implementations.\n"
          " * if you use AdaptiveCpp exclusively, you are encouraged to use scoped parallelism:\n"
          "   https://github.com/AdaptiveCpp/AdaptiveCpp/blob/develop/doc/scoped-parallelism.md\n"
```

The first two commits (perhaps incorporating improvements to the recommendations mentioned above) could be made into separate PRs if that's helpful.


illuhad:
I agree that we should work towards removing that dependency wherever possible. One issue I see is that this approach kind of backtracks on the direction that we started with `ACPP_COMPILER_FEATURE_PROFILE`, i.e. there should only be a small set of different feature profiles that build upon each other for simplicity.

Making fibers optional means optionally cutting out core functionality from existing feature profiles.

> Boost is not searched and not linked anymore when accelerated CPU is enabled,
which is the most common use-case. No reason to require library that's never used

Even if AdaptiveCpp is built with accelerated CPU, I think there's still value in `omp.library-only` for debugging use cases. Debugging with `accelerated` may not be as convenient. I still occasionally turn to `omp.library-only`, e.g. when using sanitizers or similar tooling.


> Another alternative would be to move it back to user-selection, adding, e.g., --acpp-no-boost option.

I think this would be less controversial.

I think the *best* solution would be to just use another `fiber` implementation that is not boost. There are some header-only fiber implementations that we could try (and ultimately just ship them with AdaptiveCpp), or perhaps C++ coroutines could also be interesting.
In that way, we could retain all of the functionality without having to resort to complex dependencies.

The functionality that we need is actually fairly limited: Spawn a function as a new fiber, and when we hit a barrier, cycle through all running fibers.

> Also, do we want to keep these two recommendations?

I think we can drop those. `nd_range` is well-established and optimized with AdaptiveCpp on CPU.


EDIT: This is what HIP-CPU uses: https://github.com/ROCm/HIP-CPU/tree/master/external/libco

MarkusBuettner:
> Even if AdaptiveCpp is built with accelerated CPU, I think there's still value in `omp.library-only` for debugging use cases. Debugging with `accelerated` may not be as convenient. I still occasionally turn to `omp.library-only`, e.g. when using sanitizers or similar tooling.
> 
> > Another alternative would be to move it back to user-selection, adding, e.g., --acpp-no-boost option.
> 
> I think this would be less controversial.

Could this be implied from `--acpp-targets`? If `omp.accelerated` is in the target list, Boost is not linked, but if `omp.library-only` is in the target list Boost has to be linked.

al42and:
> I think the best solution would be to just use another fiber implementation that is not boost.

Good point. Messing with userland threads runtime is more fun than messing with build systems :)

illuhad:
> Could this be implied from --acpp-targets? If omp.accelerated is in the target list, Boost is not linked, but if omp.library-only is in the target list Boost has to be linked.

Good idea, I think you are completely right! In that case it would be almost a kind of optimization as the feature would just transparently simplify the build process when it can.

al42and:
> The functionality that we need is actually fairly limited: Spawn a function as a new fiber, and when we hit a barrier, cycle through all running fibers.

A quick test using [Tina](https://github.com/slembcke/Tina) looks promising. Faster than Boost.Fiber on sycl-bench/reduction test and a couple others. Work-items can now do an early exit without messing up barriers, but also we distinguish barriers from synchronization when switching to the next work-group, so better correctness checks. Not opening a PR since still need to polish the code _and_ fix builds here and there _and_ do more tests _and_ gracefully handle unsupported architectures (notably, PPC64), but the code is at https://github.com/AdaptiveCpp/AdaptiveCpp/tree/issue1713/tina.

Other alternatives considered: 
- [`libco`](https://github.com/higan-emu/libco) used in HIP does not support RISC-V, making it less aligned with my current projects, but looks ok.
- [`minicoro`](https://github.com/edubart/minicoro/) looks nice, but a quick prototype (getting DeepSeek R1 to rewrite Tina version) showed _much_ worse performance (but could be easy to fix; I so far took the LLM output with zero modifications, pretty amazing that it worked at all).
- C++ coroutines are very appealing, but switching to C++20 can break existing C++17 applications.

illuhad:
Great work!

Tina looks interesting. One thing I'm a bit concerned about is that the header is basically a blob of inline assembly for all of the supported architectures... So *if* there's an issue on a particular platform, it might require expert knowledge to fix. But I guess if it works and we are confident that we don't have to touch it, that's fine.

The performance boost is nice too, although I would say that probably performance is secondary. If you have barriers in `omp.library-only`, most likely performance is going to be terrible anyway compared to `omp.accelerated`, `generic` or other competing compilers. Not sure if it makes a real-world difference whether it's 100x slower or 1000x slower.

al42and:
> One thing I'm a bit concerned about is that the header is basically a blob of inline assembly for all of the supported architectures... So if there's an issue on a particular platform, it might require expert knowledge to fix. But I guess if it works and we are confident that we don't have to touch it, that's fine.

I understand your concern. But ASM is unavoidable; [libco](https://github.com/higan-emu/libco/blob/master/amd64.c), [minicoro](https://github.com/edubart/minicoro/blob/main/minicoro.h), and [boost.context](https://github.com/boostorg/context/tree/develop/src/asm) are the same. Don't think there are other ways to manipulate the stack without help from the compiler. At least the developer of Tina seems active and the assembly is well documented (but some quality time might be needed with the architecture spec to meaningfully debug things).

A more specific risk is that Tina is less mature than Boost, so bugs are more likely.  From what I gathered, there are tons of other fiber libraries, but none are clearly better in that regard. `minicoro` has more GitHub stars but its code is, IMO, harder to understand, so I'd say it's a wash?

Since there's some clear interest in avoiding Boost (the Miosix PR, Markus's comments here, and this PR itself), I see these 3⅔ options:

- Keep Boost, and (a) allow building ACpp without it, (b) when built with Boost, only link it when needed. More preprocessor/CMake, more options for the config file. But least change in the actual implementation.
  - Bundling Boost.Context seems unpleasant.
- Bundle Tina or minicoro (or other small coroutine library). Still can happen that it's not supported by the architecture, so cannot avoid preprocessor, but at least ACpp's CMake machinery and the linking of user applications are simpler. The library can be buggy.
- Use C++20 coroutines, and say that you need C++20 if you want to use `nd_range` kernels with `omp.library-only` build, deal with it :sunglasses:.
  - One can isolate C++20 code into one file, and expose C++17-compatible API in a header to the user code. Sounds too messy to me.

illuhad:
> I understand your concern. But ASM is unavoidable; [libco](https://github.com/higan-emu/libco/blob/master/amd64.c), [minicoro](https://github.com/edubart/minicoro/blob/main/minicoro.h), and [boost.context](https://github.com/boostorg/context/tree/develop/src/asm) are the same. Don't think there are other ways to manipulate the stack without help from the compiler. At least the developer of Tina seems active and the assembly is well documented (but some quality time might be needed with the architecture spec to meaningfully debug things).

Hm... I thought that at least on Linux fibers could also be implemented using `setcontext`/`makecontext` and `ucontext` friends (https://www.man7.org/linux/man-pages/man2/setcontext.2.html). From the `minicoro.h` header, it seems that indeed they can use `ucontext` as a fallback method:

```
Most platforms are supported through different methods:

| Platform     | Assembly Method  | Fallback Method   |
|--------------|------------------|-------------------|
| Android      | ARM/ARM64        | N/A               |
| iOS          | ARM/ARM64        | N/A               |
| Windows      | x86_64           | Windows fibers    |
| Linux        | x86_64/i686      | ucontext          |
| Mac OS X     | x86_64/ARM/ARM64 | ucontext          |
| WebAssembly  | N/A              | Emscripten fibers / Binaryen asyncify |
| Raspberry Pi | ARM              | ucontext          |
| RISC-V       | rv64/rv32        | ucontext          |
```

So, while ASM may be required for best performance, there may be other mechanisms that are potentially more maintainable/portable, at least on some platforms.

I'm by no means opposed though to give Tina (or minicoro or something else of your choice) a try. If it causes issues, we could still switch to something else.

> Bundling Boost.Context seems unpleasant.

Yes.

> Bundle Tina or minicoro (or other small coroutine library). Still can happen that it's not supported by the architecture, so cannot avoid preprocessor, but at least ACpp's CMake machinery and the linking of user applications are simpler. The library can be buggy.

I think overall I'm still in favor of this approach.
*If* someone runs on a super-exotic system, then at least the other compilation flows that don't need fibers should still work. We'd need to double check though that things build if you have an invalid target for, say, tina but the target is *not* `omp.library-only`.

EDIT: I think I have a slight preference for `minicoro` over `tina` since it seems that thanks to `ucontext` they can also support ppc64le, mips, sparc, and all the other exotic Linux targets. (but we should double check whether this is actually the case)

This might be a good moment to also ping @NiccoloN, since the Miosix PR is specifically about `omp.library-only` on an embedded platform - so whatever fiber library we pick, we should make sure that it doesn't break their use case.

> Use C++20 coroutines, and say that you need C++20 if you want to use nd_range kernels with omp.library-only build, deal with it 😎.
> One can isolate C++20 code into one file, and expose C++17-compatible API in a header to the user code. Sounds too messy to me.

Yeah, let's not do that.

fedetftpolimi:
Thanks for the ping on this issue. If there are multiple coroutine libraries different users prefer, maybe you could try creating an AdaptiveCpp-specific coroutine.h header providing a common interface and using `#ifdef` to switch between implementations.

That should make it easy to use C++20 coroutines where possible, and for us to plug some Miosix-specific code until we support C++20.

al42and:
> If there are multiple coroutine libraries different users prefer, maybe you could try creating an AdaptiveCpp-specific coroutine.h header providing a common interface and using #ifdef to switch between implementations.

Not that it's hard (all couroutine-related code is already isolated in a single file), but it's a pain to test and maintain, especially with the header included in user applications.

And the fiber APIs are quite varied, so a trivial, function-level abstraction is not an option.

Do you have in mind a specific fibers implementation that can work for Miosix, so I can see how best to structure the code to ease switching?

> I thought that at least on Linux fibers could also be implemented using setcontext/makecontext and ucontext friend

Yeah, except `ucontext` got removed from POSIX 2008, with the recommendation to use pthreads instead. It is widely supported, but that still smells.

 > I think I have a slight preference for minicoro over tina

Okay!

illuhad:
> Yeah, except ucontext got removed from POSIX 2008, with the recommendation to use pthreads instead. It is widely supported, but that still smells.

I don't think that most Linux distributions have a goal of strict POSIX-compliance (most are not AFAIK), so I'd be surprised if this functionality went away. Is there evidence to suggest otherwise?

al42and:
Superseded by #1844

PR #1721, opened 2025-02-16
ci: fix and bump Boost download on Windows

al42and:
https://github.com/boostorg/boost/issues/997

And while we're at it, switch to a newer version


illuhad (approved):
Thanks!

PR #1726, opened 2025-02-20
Add a dark theme to the documentation pages

marcosolanki:
This pull request adds a button next to the search bar in the [Material for MkDocs-based documentation pages](https://adaptivecpp.github.io/AdaptiveCpp/), which, when clicked, changes the theme of said pages.

In particular, in addition to the existing light theme, this PR adds a dark theme, which, as is typical, comes with a darker background colour while making the foreground text light.

The default theme (dark or light) is automatically chosen based on the preference communicated by the user's browser through the [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) media feature.

### DIFF
--- a/doc/stylesheets/extra.css
+++ b/doc/stylesheets/extra.css
@@ -1,3 +1,9 @@
 :root {
-    --md-primary-fg-color:        #c50d29;
-  }
+  --md-primary-fg-color: #c50d29;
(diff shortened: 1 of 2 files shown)


tdavidcl (approved):
LGTM

PR #1727, opened 2025-02-21
[stdpar][algorithms] Add support for `move`

normallytangent:
This PR adds automatic offloading support for [move](https://en.cppreference.com/w/cpp/algorithm/move).

### DIFF
--- a/include/hipSYCL/std/stdpar/detail/algorithm_fwd.hpp
+++ b/include/hipSYCL/std/stdpar/detail/algorithm_fwd.hpp
@@ -17,6 +17,8 @@
 
 namespace std {
 
+///////////////////////////// par_unseq policy /////////////////////////////
+
 template <class ForwardIt, class UnaryFunction2>
 HIPSYCL_STDPAR_ENTRYPOINT void for_each(hipsycl::stdpar::par_unseq, ForwardIt first,
                                         ForwardIt last, UnaryFunction2 f);
@@ -54,6 +56,11 @@ template <class ForwardIt1, class Size, class ForwardIt2>
 HIPSYCL_STDPAR_ENTRYPOINT ForwardIt2 copy_n(hipsycl::stdpar::par_unseq,
                                             ForwardIt1 first, Size count,
                                             ForwardIt2 result);
+
+template <class ForwardIt1, class ForwardIt2>
+HIPSYCL_STDPAR_ENTRYPOINT ForwardIt2 move(hipsycl::stdpar::par_unseq,
+                                          ForwardIt1 first, ForwardIt1 last,
+                                          ForwardIt2 d_first);
 
 template <class ForwardIt, class T>
 HIPSYCL_STDPAR_ENTRYPOINT void fill(hipsycl::stdpar::par_unseq, ForwardIt first,
@@ -123,6 +130,12 @@ HIPSYCL_STDPAR_ENTRYPOINT
 bool none_of(hipsycl::stdpar::par_unseq, ForwardIt first, ForwardIt last,
             UnaryPredicate p );
 
+///////////////////////////// par policy /////////////////////////////
+
+template <class ForwardIt1, class ForwardIt2>
+HIPSYCL_STDPAR_ENTRYPOINT ForwardIt2 move(hipsycl::stdpar::par,
+                                          ForwardIt1 first, ForwardIt1 last,
+                                          ForwardIt2 d_first);
 }
 
 #endif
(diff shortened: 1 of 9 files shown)


include/hipSYCL/std/stdpar/detail/algorithm_fwd.hpp
```diff
@@ -55,6 +55,11 @@ HIPSYCL_STDPAR_ENTRYPOINT ForwardIt2 copy_n(hipsycl::stdpar::par_unseq,
                                             ForwardIt1 first, Size count,
                                             ForwardIt2 result);
 
+template <class ForwardIt1, class ForwardIt2>
```
normallytangent, line 60: Why are only the `par_unseq` policy variants forward declared?

tests/pstl/pstl_test_suite.hpp
```diff
@@ -58,4 +58,38 @@ struct non_trivial_copy {
...
+
+  non_trivial_move(non_trivial_move&& other) {
+    x = std::move(other.x);
+    __acpp_if_target_host(++move_counter;)
+  }
+
+  non_trivial_move& operator=(non_trivial_move&& other) {
+    if (this != &other) {
+      x = std::move(other.x);
+      // x = std::exchange(other.x, {});
```
normallytangent: What should the state of the moved object be? The standard mentions, "..the elements in the moved-from range will still contain valid values of the appropriate type, but not necessarily the same values as before the move". Using `std::exchange` would give the user more guarantees, but this implementation would then deviate from the standard by being more specific.

include/hipSYCL/std/stdpar/detail/algorithm_fwd.hpp
illuhad: Ohhh... good catch! Even for `par_unseq`, it's incomplete, as some newer algorithms like `scan` or `merge` are missing...I think at some point probably I just forgot to add the forward declarations :/

They all should be forward-declared (for both `par` and `par_unseq`). The actual implementations are pulled in by `execution`, so those forward declares might only matter in specific cases (perhaps depending on the order in which `execution` and `algorithm`/`numeric` are included?).

tests/pstl/pstl_test_suite.hpp
illuhad: I think for the purpose of this test, the code is fine (not sure if we even need a test case for non-trivial move, but it can't hurt)...The class doesn't do anything useful and is purely artifically made for the test, so I don't think it matters a lot how exactly it implements move assignment.

tests/pstl/move.cpp
```diff
@@ -0,0 +1,69 @@
...
+    std::vector<T> dest_host(problem_size);
+
+    auto ret = std::move(pol, data.begin(), data.end(),
+                         dest_device.begin());
+    std::move(data.begin(), data.end(), dest_host.begin());
+
+    BOOST_CHECK(ret == dest_device.begin() + problem_size);
+    BOOST_CHECK(dest_device == dest_host);
+    BOOST_CHECK(data == dest_device);
+    BOOST_CHECK(data == dest_host);
```
normallytangent, line 45: Is this a valid check? It is not guaranteed that the elements of the test vector remain intact after a move operation. I could not come up with a rigorous test

illuhad: That's right, this might be problematic, depending on what type `T` is. You could just copy the original `data` though before you run the test:

```
std::vector<T> data(problem_size)
// init data

auto device_input = data;
auto host_input = data;

auto ret = std::move(pol, device_input.begin(), device_input.end(),
                         dest_device.begin());
std::move(host_input.begin(), host_output.end(), dest_host.begin());
```

include/hipSYCL/std/stdpar/detail/algorithm_fwd.hpp
normallytangent: Thanks for the explanation! I'll forward declare both.

normallytangent:
It's now ready :)

tests/pstl/move.cpp
```diff
@@ -0,0 +1,73 @@
...
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include <algorithm>
+#include <execution>
+#include <utility>
+#include <vector>
+
+#include <boost/test/unit_test.hpp>
+#include <boost/mp11/list.hpp>
+#include <boost/mp11/mpl_list.hpp>
```
illuhad: Can you change this header in accordance with this PR that I've just merged? https://github.com/AdaptiveCpp/AdaptiveCpp/pull/1734

normallytangent: Done :)

illuhad (approved):
Thanks!

PR #1729, opened 2025-02-24
Fix the group algorithms operating on marray

carbotaniuman:
`sycl::marray` takes in a `size_t` instead of an `int`, so these template overloads just had substitution failures and were never actually callable.


illuhad (approved):
Thanks!

(at some point, we should consider also testing this in our unit tests. I don't think they currently test group algorithms with marray at all)

PR #1730, opened 2025-02-24
Fix behaviour of sycl::free

Stastez:
As far as I can tell, calling `free` on a `nullptr` should have no effect (https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_memory_deallocation_functions).
This works for the OMP backend; the HIP backend, however, fails with `[AdaptiveCpp Error] from /AdaptiveCpp/src/runtime/hip/hip_allocator.cpp:118 @ query_pointer(): hip_allocator: query_pointer(): pointer is unknown by backend (error code = HIP:1)` upon calling `free` on a `nullptr`.

Minimal example:
```cpp
#include <sycl/sycl.hpp>

int main() {
    // Backend must be HIP
    sycl::free(nullptr, queue);

    return 0;
}
```

### DIFF
--- a/include/hipSYCL/sycl/usm.hpp
+++ b/include/hipSYCL/sycl/usm.hpp
@@ -283,7 +283,8 @@ T *aligned_alloc(std::size_t alignment, std::size_t count, const sycl::queue &q,
 }
 
 inline void free(void *ptr, const sycl::context &ctx) {
-  return rt::deallocate(detail::select_usm_allocator(ctx), ptr);
+  if (ptr != nullptr)
+    return rt::deallocate(detail::select_usm_allocator(ctx), ptr);
 }
 
 inline void free(void *ptr, const sycl::queue &q) {


illuhad (approved):
Thanks! This is potentially a bug in HIP, but it's good if we can work around!

Stastez:
According to the latest docs, `hipFree` and `hipFreeHost` both return `hipSuccess` when called with `nullptr` (https://rocm.docs.amd.com/projects/HIP/en/latest/doxygen/html/group___memory.html#ga740d08da65cae1441ba32f8fedb863d1).
This is equivalent to the behaviour of `cudaFree` (https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1ga042655cbbf3408f01061652a075e094).

Similarly, it seems as though both `cudaPointerGetAttributes` and `hipPointerGetAttributes` are **supposed** to return an error when passed `nullptr`.
This behaviour seems unintuitive to me - how else should you decide between calling `hipFree` or `hipFreeHost`? - but I believe it's just a case of AMD mirroring the behaviour of the CUDA API (as it states in the documentation for `hipPointerGetAttributes`).

https://rocm.docs.amd.com/projects/HIP/en/latest/doxygen/html/group___memory.html#ga7c3e8663feebb7be9fd3a1e5139bcefc

https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__UNIFIED.html#group__CUDART__UNIFIED_1gd89830e17d399c064a2f3c3fa8bb4390

---

Edited to add:
The minimal example provided in the original comment also errors when using CUDA 12.6.3, so this is almost certainly not a bug with HIP.

PR #1732, opened 2025-02-26
Half changes

carbotaniuman:
I had to make slightly more invasive changes which made the diff really painful to review, so I extracted the half changes out of #1710. Half seems ridiculously underspecified in the SYCL standard, so the following changes were done to make the CTS compile properly, as well as fix some issues before our builtins could be converted to allow half types.

This is best reviewed commit-by-commit.


illuhad:
Hi, the `half_impl` namespace seems to already be introduced in #1710. Are there any dependencies or conflicts between those two PRs; should one be merged before the other?

carbotaniuman:
> Is it actually correct to implement these operators by casting everything to half, even when this causes loss of information? For example, when comparing with double, I would expect the half to be cast to double and the comparison being carried out with double precision.

That makes sense and casting everything to half is likely incorrect. I think the way to handle this properly is to promote the half to a float, and then let C++ floating point comparison conversions handle the rest.

illuhad:
> That makes sense and casting everything to half is likely incorrect. I think the way to handle this properly is to promote the half to a float, and then let C++ floating point comparison conversions handle the rest.

Agreed, this sounds reasonable. Do you want to add this change to this PR, or would you like to leave it for a future PR? Of course, this PR clearly already improves the status quo by itself.

I took the liberty to merge in the latest changes from `develop`, as we've made some improvements to CI.

carbotaniuman:
I've gone and fixuped the change so that comparison happens via float, so this should be ready to merge once CI passes.

illuhad:
I see, great! (I suspect there might be some more issues around this... for example, the arithmetic operators should probably also promote to float whenever `half` and float/double is involved, instead of returning `half`)

illuhad (approved):

PR #1733, opened 2025-02-26
[stdpar][algorithms] Add support for `reverse` and `reverse_copy`

normallytangent:
This PR adds automatic offloading support for:
- [x] [reverse](https://en.cppreference.com/w/cpp/algorithm/reverse)
- [x] [reverse_copy](https://en.cppreference.com/w/cpp/algorithm/reverse_copy)

### DIFF
--- a/include/hipSYCL/algorithms/algorithm.hpp
+++ b/include/hipSYCL/algorithms/algorithm.hpp
@@ -411,6 +411,36 @@ sycl::event replace_copy(sycl::queue &q, ForwardIt1 first, ForwardIt1 last,
       new_value, deps);
 }
 
+template <class BidirIt, class ForwardIt>
+sycl::event reverse_copy(sycl::queue &q, BidirIt first,
+                         BidirIt last, ForwardIt d_first,
+                         const std::vector<sycl::event> &deps = {}) {
+  if (first == last)
+    return sycl::event{};
+
+  auto size = std::distance(first, last);
+
+  using value_type1 = typename std::iterator_traits<BidirIt>::value_type;
+  using value_type2 = typename std::iterator_traits<ForwardIt>::value_type;
+
+  // ??
+  if (std::is_trivially_copyable_v<BidirIt> &&
+      std::is_same_v<value_type1, value_type2> &&
+      util::is_contiguous<BidirIt>() && util::is_contiguous<ForwardIt>() &&
+      detail::should_use_memcpy(q.get_device())) {
+    return q.memcpy(&(*d_first), &(*first), size * sizeof(value_type1), deps);
+  } else {
+    return q.parallel_for(sycl::range{size}, deps,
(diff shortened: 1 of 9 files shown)


include/hipSYCL/algorithms/algorithm.hpp
```diff
@@ -383,6 +383,36 @@ sycl::event replace_copy(sycl::queue &q, ForwardIt1 first, ForwardIt1 last,
...
+  using value_type1 = typename std::iterator_traits<BidirIt>::value_type;
+  using value_type2 = typename std::iterator_traits<ForwardIt>::value_type;
+
+  // ??
+  if (std::is_trivially_copyable_v<BidirIt> &&
+      std::is_same_v<value_type1, value_type2> &&
+      util::is_contiguous<BidirIt>() && util::is_contiguous<ForwardIt>() &&
+      detail::should_use_memcpy(q.get_device())) {
+    return q.memcpy(&(*d_first), &(*first), size * sizeof(value_type1), deps);
+  } else {
```
normallytangent: I would like to modify this if-condition but I don't exactly understand its purpose.  For `reverse_copy` this condition will always be false since the two types are different (Bidirectional and Forward iterators).  It's also in `std::copy`.

illuhad: This condition is there for algorithms like `copy`, because on some backends, using `memcpy` is faster than a custom kernel.

So, if we can use a bitwise-memcpy (e.g. if the type is trivially copyable) and it makes sense for the backend, use `memcpy`.

For `reverse_copy`, you can remove the `if` branch entirely because there's no way a `memcpy` can do a reverse as well. So this optimization won't apply.

> this condition will always be false since the two types are different (Bidirectional and Forward iterators)

I don't think this is the case, because we're comparing the value types, not the iterator types. So if the iterators point to values of the same type, the condition could be true.

include/hipSYCL/algorithms/algorithm.hpp
```diff
@@ -383,6 +383,36 @@ sycl::event replace_copy(sycl::queue &q, ForwardIt1 first, ForwardIt1 last,
...
+    return q.parallel_for(sycl::range{size}, deps,
+                          [=](sycl::id<1> id) {
+                            auto input = last;
+                            auto output = d_first;
+                            std::advance(input, -id[0]);
```
illuhad: Does this actually work?

`id[0]` will likely return a `size_t`, so an unsigned type... So this might cause an underflow.

It would be better to implement this by calculating an offset that is then applied to `first`.

normallytangent: That might explain why the tests fail in CI. It did work on Jarvis, for integers I  was able to verify the test results with some print statements 😅 I'll fix that

illuhad (approved):
Thanks :)

PR #1734, opened 2025-02-28
Fix AdaptiveCpp build on Ubuntu 20.04

carbotaniuman:
The specific header included does not exist on Boost 1.71, replace it with the more general header.


illuhad (approved):
Thanks!

PR #1736, opened 2025-02-28
[KHR] Implement sycl_khr_max_work_group_queries extension

Aympab:
Implementation proposal for the `max_num_work_groups<3>` [KHR extension](https://github.com/KhronosGroup/SYCL-Docs/pull/712)


include/hipSYCL/sycl/device.hpp
```diff
@@ -314,6 +314,25 @@ HIPSYCL_SPECIALIZE_GET_INFO(device, max_compute_units)
...
+      rt::device_uint_property::max_num_work_groups0);
+}
+
+HIPSYCL_SPECIALIZE_GET_INFO(device, max_num_work_groups1)
+{
+  return get_rt_device()->get_property(
+      rt::device_uint_property::max_num_work_groups1);
+}
+
+HIPSYCL_SPECIALIZE_GET_INFO(device, max_num_work_groups2)
```
illuhad: Should we not return some `range<3>` instead of having a separate query for each dimension?
Or, even better, template it with the dimension and return `range<Dim>`.
This also would solve another issue: In typical backends like CUDA/HIP/OpenCL etc the first dimension is the vectorizing dimension (or fast index or whatever you want to call it) in the linerization of a multi-dimensional id.

In SYCL, like in C++, it is always the last one (e.g `id[2]` for 3d, `id[1]` for 2d). So for many backends, SYCL implementations perform a flip of dimensions. In other words: The values that you return here might not correspond to the ones that you expect in SYCL.

illuhad (commented):
Thanks! What about the OpenCL backend though? It seems there is an implementation for it missing?

include/hipSYCL/sycl/device.hpp
Aympab: > What about the OpenCL backend though

I updated the file, but I am not sure about `CL_DEVICE_MAX_WORK_GROUP_SIZE`, I actually think in OpenCL there are no limitation? Since it's not in the SYCL spec that would be logical? If so, should we return `std::numeric_limits<int>::max`? I am not familiar with OpenCL backend. Will update the PR accrodingly.

> Should we not return some range<3> instead of having a separate query for each dimension? Or, even better, template it with the dimension

In `acpp-info` and `hardware.hpp` there are 3 lines/enum for `max_global_sizeN`, `max_group_sizeN`  one for each dimension. I wrote `max_num_work_groups0`, `max_num_work_groups1`, `max_num_work_groups2` for consistency. I am not sure what you mean by returning a range?

> The values that you return here might not correspond to the ones that you expect in SYCL.

I think the implementation should return the maximum number of work-group that can be submitted in any dimension, regardless of the backend mapping. It is then user's responsibility to check before submitting a kernel if the maximum sizes are not exceeded. And it's the implementation responsibility to map the right max_size onto the right dimension so the user gets the actual information. Then maybe when users change SYCL implementation the kernel won't execute because of the flipping, but at least they have the tools to check.  

Dpc++ flips the dimensions of the loops and ACpp doesn't, right? I guess depending on that answer I should update so that `max_num_work_groups<3> = max_X` or `max_num_work_groups<3> = max_Z`.

illuhad: > I updated the file, but I am not sure about CL_DEVICE_MAX_WORK_GROUP_SIZE, I actually think in OpenCL there are no limitation? Since it's not in the SYCL spec that would be logical? If so, should we return std::numeric_limits<int>::max? I am not familiar with OpenCL backend. Will update the PR accrodingly.

Could be. I'm also not seeing anything that would give us the max global size (in that case, we could just divide global size by the minimum group size or so). There's `CL_DEVICE_MAX_WORK_ITEM_SIZES` which however seems to refer to work group sizes, not the global size -- although we seem to incorrectly use it for max global size in AdaptiveCpp: https://github.com/AdaptiveCpp/AdaptiveCpp/blob/206760b95ab2a81f70de468cb87db0999a959c4d/src/runtime/ocl/ocl_hardware_manager.cpp#L295

If it's unlimited, we should probably use `std::numeric_limits<size_t>::max()`, not the int max. Could you experimentally check on, say, an iGPU whether it can indeed handle arbitrary global sizes?

 > In acpp-info and hardware.hpp there are 3 lines/enum for max_global_sizeN, max_group_sizeN one for each dimension. I wrote max_num_work_groups0, max_num_work_groups1, max_num_work_groups2 for consistency. I am not sure what you mean by returning a range?

I'm talking about the user-facing SYCL interface.
The `hardware.hpp` stuff is for internal runtime usage. You also need to define a `get_info` query for `sycl::device` (which seems to only be partially implemented in this PR). I'm saying that this `get_info` query should be templated with the dimensions and then return `range<Dim>`.

I.e., the user should be able to do something like
```c++
sycl::range<2> max_groups = mydevice.get_info<sycl::info::device::max_num_work_groups<2>>()
```
I don't think this PR currently exposes a working interface on the SYCL level for users, which is also why CI fails.

>  Then maybe when users change SYCL implementation the kernel won't execute because of the flipping, but at least they have the tools to check.

If the return type of the user-facing query is a `range<Dim>`, then the entries in the range can be mapped appropriately for a kernel of dimension `Dim`. So the user just needs to check whether `num_submitted_groups[i] <= max_groups[i]`. Otherwise they will know whether they will need to do the flip, and if so, do it themselves in this comparison which is a non-portable, backend-specific exercise.

> Dpc++ flips the dimensions of the loops and ACpp doesn't, right?

No! The flip depends less on implementation and more on backend.
The SYCL specification mandates that in SYCL, the fast-moving index is always the last one.
If you implement SYCL with a backend where the fast-moving index is the first one (e.g. CUDA, HIP, OpenCL), then any SYCL implementation for this backend will have to do the flip.

The only case where AdaptiveCpp does currently *not* flip indices is the OpenMP backend when not going through `--acpp-targets=generic`. With `generic` target, the OpenMP backend flips too. (The OpenMP backend is unlimited in terms of max global size, so this does not really matter however).

There is a field `device_uint_property::needs_dimension_flip` that you can query for whether a flip is needed.

Aympab: > If it's unlimited, we should probably use std::numeric_limits<size_t>::max(), not the int max. Could you experimentally check on, say, an iGPU whether it can indeed handle arbitrary global sizes?

From the OpenCL spec:
> - [CL_INVALID_GLOBAL_WORK_SIZE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_INVALID_GLOBAL_WORK_SIZE) if any of the values specified in global_work_size exceed the maximum value representable by size_t on the device associated with command_queue.

Which doesn't correspond exactly to the number of work-groups, but since we can have work-groups of size 1, we can expect OCL to run any NDRange kernel with a global size smaller than `numeric_limits<size_t>::max()`. I've tested it on Intel PVC and I confirm I can run kernels up to `2**31 - 1`. Updated the PR with `numeric_limits<size_t>::max()` in every dimension for the OCL backend

> I don't think this PR currently exposes a working interface on the SYCL level for users, which is also why CI fails.

I've updated the API for users and used the same methodology as `max_work_item_sizes<Dim>` with the check `device_uint_property::needs_dimension_flip`

illuhad:
Thanks, the implementation looks reasonable now. As there are no tests in this PR (and I assume it might be a bit difficult to test) - Have you tried it? Does it do what you expect?

> and I confirm I can run kernels up to 2**31 - 1.

Does that mean that it fails beyond that (in which case there might still be a 32-bit limit), or have you just not tried anything larger?

include/hipSYCL/sycl/info/device.hpp
```diff
@@ -82,6 +82,9 @@ namespace device {
   HIPSYCL_DEFINE_INFO_DESCRIPTOR(max_compute_units, detail::u_int);
   HIPSYCL_DEFINE_INFO_DESCRIPTOR(max_work_item_dimensions, detail::u_int);
 
+  template<int Dimensions = 3>
+  HIPSYCL_DEFINE_INFO_DESCRIPTOR(max_num_work_groups, range<Dimensions>);
```
illuhad: One thing that's still not entirely aligned from what I can see is that this would not declare the query in `sycl::khr::info::device` as the spec for the extension says, but in `info::device`. So you might have to declare it outside the `info::device` namespace in a new `khr::info::device` namespace.

If you're lucky this will just work, however, the macro so far hasn't really been used outside of `sycl::info::device`, so I don't know if it makes any assumptions here.

Aympab:
> Does that mean that it fails beyond that (in which case there might still be a 32-bit limit), or have you just not tried anything larger?

I was testing with an OpenCL code on PVC, and when I exceed the 2**31-1 limit I get the `CL_INVALID_GLOBAL_WORK_SIZE` error and the kernel doesn't execute.

>  - Have you tried it? Does it do what you expect?

It does. I can query the `max_num_work_groups` before running my kernel and am guaranteed it runs. Although I have not experimentally tried all backends, I didn't encounter a bug to this date (meaning I did not exceed the number of work-groups and my kernel didn't run).

I think the idea is to get a 0 overhead implementation, without any assertions before submitting a kernel, but now as a user I have the information.

include/hipSYCL/sycl/info/device.hpp
Aympab: Indeed I should put it in the right namespace. And we have to remove the `khr` when the extensions is merged into the spec.

Aympab: Ok doesn't work as is, I should not use the macro. Will look into that and commit changes. For now I just revert the last commit so the CI runs

illuhad:
> I was testing with an OpenCL code on PVC, and when I exceed the 2**31-1 limit I get the CL_INVALID_GLOBAL_WORK_SIZE error and the kernel doesn't execute.

Doesn't that imply  though that the limit on OpenCL is not `std::numeric_limits<size_t>::max()` as we currently use?

> It does. I can query the max_num_work_groups before running my kernel and am guaranteed it runs. Although I have not experimentally tried all backends, I didn't encounter a bug to this date (meaning I did not exceed the number of work-groups and my kernel didn't run).

Okay, that's fine :)

include/hipSYCL/sycl/info/device.hpp
illuhad: Ok!

illuhad:
Waiting for namespace change.

Aympab:
> Waiting for namespace change.

Done! Had some issues with my local env so this explains the large number of commits, you might want to squash that when merging, I think it's maintainers role right?  

> Doesn't that imply though that the limit on OpenCL is not `std::numeric_limits<size_t>::max()` as we currently use?

Ah yes indeed, so it seems like we should use  `std::numeric_limits<int>::max()`  for the OCL backend. I'm updating that right now

src/runtime/ocl/ocl_hardware_manager.cpp
```diff
@@ -293,15 +293,15 @@ std::size_t ocl_hardware_context::get_property(device_uint_property prop) const
     break;
   case device_uint_property::max_num_work_groups0:
     return static_cast<std::size_t>(
```
Aympab, line 295: I'm still casting to a `size_t` for the return type I think that's better.

illuhad: Yes, that's alright - the interface is designed to return `size_t` after all.

include/hipSYCL/sycl/info/info.hpp
```diff
@@ -27,4 +27,9 @@
   inline typename info::class_name::specialization::return_type                \
   sycl::class_name::get_info<info::class_name::specialization>() const
 
+#define HIPSYCL_SPECIALIZE_GET_INFO_EXTENSION(class_name, specialization)      \
```
illuhad:
```suggestion
#define HIPSYCL_SPECIALIZE_GET_INFO_KHR_EXTENSION(class_name, specialization)      \
```
?

include/hipSYCL/sycl/device.hpp
```diff
@@ -314,6 +315,41 @@ HIPSYCL_SPECIALIZE_GET_INFO(device, max_compute_units)
...
+  std::size_t size1 = static_cast<std::size_t>(get_rt_device()->get_property(
+      rt::device_uint_property::max_num_work_groups1));
+  if (get_rt_device()->get_property(
+      rt::device_uint_property::needs_dimension_flip))
+    return range<2>{size1, size0};
+  else
+    return range<2>{size0, size1};
+}
+
+HIPSYCL_SPECIALIZE_GET_INFO_EXTENSION(device, max_num_work_groups<3>)
```
illuhad:
```suggestion
HIPSYCL_SPECIALIZE_GET_INFO_KHR_EXTENSION(device, max_num_work_groups<3>)
```
?

include/hipSYCL/sycl/device.hpp
```diff
@@ -314,6 +315,41 @@ HIPSYCL_SPECIALIZE_GET_INFO(device, max_compute_units)
...
 }
 
+HIPSYCL_SPECIALIZE_GET_INFO_EXTENSION(device, max_num_work_groups<1>)
+{
+  std::size_t size0 = static_cast<std::size_t>(get_rt_device()->get_property(
+      rt::device_uint_property::max_num_work_groups0));
+  return range<1>{size0};
+}
+
+HIPSYCL_SPECIALIZE_GET_INFO_EXTENSION(device, max_num_work_groups<2>)
```
illuhad:
```suggestion
HIPSYCL_SPECIALIZE_GET_INFO_KHR_EXTENSION(device, max_num_work_groups<2>)
```
?

include/hipSYCL/sycl/device.hpp
```diff
@@ -314,6 +315,41 @@ HIPSYCL_SPECIALIZE_GET_INFO(device, max_compute_units)
       rt::device_uint_property::max_compute_units);
 }
 
+HIPSYCL_SPECIALIZE_GET_INFO_EXTENSION(device, max_num_work_groups<1>)
```
illuhad:
```suggestion
HIPSYCL_SPECIALIZE_GET_INFO_KHR_EXTENSION(device, max_num_work_groups<1>)
```
?

illuhad (commented):
Thanks, this looks really good :)
Some minor naming suggestions, but apart from that I think we're done :)

Thank you!

(I'm only not adding the green "approval" check so that we remember to delay the merge until the spec is agreed by the SYCL WG)

Aympab:
@illuhad I've updated the PR following the discussion in the [SYCL-Docs PR](https://github.com/KhronosGroup/SYCL-Docs/pull/712).

I've also added `max_work_group_range_size` which basically corresponds to the maximum linear id inside an ndrange, I'll comment that directly on the code (`include/hipSYCL/sycl/device.hpp`) so you can tell me if you prefer to implement it another way

illuhad:
Hi, there seem to be a lot of unrelated commits  in this PR that have already been merged - I assume something went wrong with rebasing? Can you clean up the PR so that it only contains your changes? It's a bit difficult to review as is :)

Aympab:
Huh yeah sorry for that :sweat_smile: , it took all the commits from the rebasing.
Files changed should be fine now

src/runtime/hip/hip_hardware_manager.cpp
```diff
@@ -258,6 +258,18 @@ hip_hardware_context::get_property(device_uint_property prop) const {
...
+    return _properties->maxGridSize[0];
+    break;
+  case device_uint_property::max_work_group_range1:
+    return _properties->maxGridSize[1];
+    break;
+  case device_uint_property::max_work_group_range2:
+    return _properties->maxGridSize[2];
+    break;
+  case device_uint_property::max_work_group_range_size:
+    return std::numeric_limits<std::size_t>::max();
```
illuhad, line 271: I'm not sure if it might not be better to effectively do something like `min(maxGridSize[0]*maxGridSize[1]*maxGridSize[2], numeric_limits<size_t>::max())` here (with some handling of potential overflow in the multiplication). Similarly for other backends that don't have direct support for the query.

It seems a little counterintuitive to say that the number of groups is limited in all dimensions (presumably much stricter limits than `numeric_limits<size_t>::max()`), but the overall number of work groups is unlimited.

The wording in the current KHR proposal implies to me that it should always hold that `maxGridSize[0]*maxGridSize[1]*maxGridSize[2] >= max_work_group_range_size`, but maybe I'm reading it wrong (perhaps this should also be clarified in the KHR).

illuhad: Okay, Greg thinks this is fine as is :) https://github.com/KhronosGroup/SYCL-Docs/pull/712#issuecomment-2980288008

illuhad (approved):
Since the KHR spec PR is now merged, we should merge the implementation once CI comes back good.
Thanks @Aympab for your work :)

PR #1738, opened 2025-03-06
Create a USM property to specify the nodes on which to allocate memory

thomasRoglin:
This PR builds upon #1692 that allow the support of properties on USM memory allocator. 
Its goal is to implement a new USM properties allowing the user to specify which nodes to allocate memory to.

This PR introduces a new USM property : `target_numa_node` which contains a set of user-specified numa nodes id. 
The `rt::allocation_hints` now contains an `optional<vector>` which is populated with the information from  `target_numa_node`property. 

Now, the property is only effective when using the OpenMP backend. 
To allocate memory on specific numa nodes, `numa_alloc_interleaved_subset` allocator from the numa library is used.
the allocation size is stored with the `allocation_map`, so it can be retrieved before calling `numafree`. 

This extension does not change the default behavior of AdaptiveCpp. To activate the support of this extension, the `-DWITH_TARGET_NUMA_NODE_PROPERTY=1` must be passed at the cmake call.


src/runtime/omp/omp_allocator.cpp
```diff
@@ -31,6 +51,62 @@ void *omp_allocator::raw_allocate(size_t min_alignment, size_t size_bytes,
...
+    numa_free_nodemask(bm);
+    numa_free_nodemask(available_bm);
+
+    // return if allocation fails
+    if (!mem)
+      return mem;
+
+    // the allocation size is stored in an allocation map so that it can be
+    // retrieved when calling numafree.
+    using value_type = amap_t::value_type;
```
illuhad: I suspect, allocation map is actually more powerful than what is needed here - e.g. it can efficiently map an address at arbitrary offset within an allocation to the root address.

I think for your purpose, a simple key-value mapping (e.g. `std::unordered_map`) might be sufficient since we only need the map in `free()`, where we would expect the root address to already be given as argument.

However, if in the future your goal is e.g. to be able to retrieve for arbitrary addresses some information - e.g. on which NUMA domain they were allocated - then allocation map will indeed be the better fit.

illuhad (commented):
What an awesome contribution, thank you!

Isn't libnuma fairly common? Do we actually need to make this support optional?

If it's optional, then we should probably also have some test where this code path is used.
And we'd probably also want to advertise a feature test macro so that user code can know whether they can use this extension.

In general, for extensions,
* there should be a feature test macro in `extensions.hpp`
* there should be some documentation of the API in `extensions.md`
* If you want to have long-term API stability, you may want to use `AdapitveCpp_*` prefix (e.g. `AdaptiveCpp_target_numa_node`) to avoid problems in case the SYCL WG standardizes something different with `target_numa_node` name.

thomasRoglin:
Hi, thanks for your feedback.

I've updated the code following your comments. 
 
- Now the new property (and therefore the depedancy to numa ) is no longer optional and does not need to be  explicitly enabled when building AdaptiveCpp.
 - I've added the `AdapitveCpp_` prefix to the property name and hints for API stability. 
 - I changed the allocation_map to `std::unordered_map`, and protected it with an atomic lock to prevent data races.

I'll add tests and documentation in a later commit.

illuhad:
Hm... We may have to limit the NUMA requirement to Linux.

@NiccoloN @fedetftpolimi Is libnuma available on Miosix? I guess probably not?

fedetftpolimi:
No, we don't have support for libnuma

thomasRoglin:
So if the machine does not have libNuma, should we: 
- disable the property completely (using AdaptiveCpp_target_numa_property would raise an error saying that the property does not exist) 
- leave the property available but using it would have no effect 
- or raise a warning to the user saying that the property is not available ?

Also, should the property be enabled "if we are on Linux" or "if libNuma is available"?

fedetftpolimi:
Well, I guess it's a matter of how many platforms you support other than Linux and how many of them support an API like libnuma

illuhad:
Hm... I would like that in the common case of Linux, this extension "just works". 

As for handling the case when libnuma is unavailable:
* I think it would be better to have the property be ignored. I see it as a performance hint (do you agree @thomasRoglin ?), so ignoring it should not lead to incorrect behavior. This would simplify user code compared to not having the API be available at all, as users wishing to hint to the runtime a specific NUMA behavior would not need to `#ifdef` the code to handle all sorts of cases for portability.
* Since ignoring it silently may lead to a performance bug *if the system supports libnuma, but we missed it/did not find it*, there should be a runtime warning if the property is used, but ignored.
    * It might be a good idea to limit this warning to systems where we know that libnuma is in principle available. I suppose there's no point warning, say, a Windows user that AdaptiveCpp was built without libnuma support.
* There might still be value to have a preprocessor definition exposed (e.g. configured from cmake via `config.hpp.in`) to user code that allows users to query whether AdaptiveCpp has been built with NUMA support. `ACPP_EXT_TARGET_NUMA_PROPERTY` would then specify whether the API is available, and an additional macro would specify whether the property has an effect.

> Also, should the property be enabled "if we are on Linux" or "if libNuma is available"?

Probably, for the greatest generality, it might be best to just `find_library(numa)` in cmake, and enable it if we find it. There might be other non-Linux systems where libnuma is available (*BSD maybe?).
On Linux as our main target, we can also issue a cmake warning if we don't find libnuma.

It seems we need to take a more cautious approach instead of just mandating libnuma as I originally had in mind. Would the strategy above be a reasonable starting point? We can then (perhaps also once we have user feedback) in the future revisit the question of whether we want cmake to emit an error (instead of warning) on Linux specifically if libnuma is unavailable.

I realize that this implies that you have to reintroduce the conditional support for numa in this PR. Sorry for backtracking :(

thomasRoglin:
Hi. I agree that the property should be ignored without raising an error if the system does not support libNUMA. 

So now, when the libNUMA is available, CMAKE defined the `LIB_NUMA_AVAILABLE` MACRO that will be used in `omp_allocator` to test whether the property backend should be built. 

I added a warning to warn a user using the property with the OpenMP backend on a machine not supporting libNUMA. (The warning will only appear once to avoid flooding the user). Maybe I should limit the warning to non-windows and non-apple machines ?

You mention adding a Macro in  `sycl/extension.hpp`, would you want something like this :
```
#ifdef LIB_NUMA_AVAILABLE
#define ACPP_EXT_TARGET_NUMA_FULLY_AVAILABLE
#endif 
```
This would requiert to move the query of the NUMA library from `src/runtime/CMakeLists.txt` to `CMakeLists.txt` since the `LIB_NUMA_AVAILABLE` is currently only used in the omp backend. 


I also added some tests and documentation. 
When the libNUMA is not available, it will use the property with random nodes id expecting no error since the backend of the property is not built. 
When libNUMA is available, it will use it to query the available nodes. Success is expected when targeting the first, the last and all available nodes. And Failure is expected when specifying an empty set of nodes, or when specifying a node out of range.

thomasRoglin:
I forgot an argument to the `sycl::free()` in the tests. CI should not fail now

normallytangent:
> I forgot an argument to the `sycl::free()` in the tests. CI should not fail now

no problem, started it 👍

thomasRoglin:
I rebased the PR to include the CI of LLVM 19&20, this should fix the last problems

thomasRoglin:
Apparently, the NUMA nodes are not accessible in the ci docker, which causes undefined behavior. 
To avoid this, I check that the library can be used on the machine with `numa_available()`. 
If not, the property is ignored and the user is warned that property is not available.

thomasRoglin:
It seems that even when `numa_available()` does not return 0, the NUMA nodes may not be accessible, and that's apparently the case in the docker that run the CI. Now if `numa_get_mems_allowed()` to returns an empty bitmask,  we expect all allocations to fail.

normallytangent:
> It seems that even when `numa_available()` does not return 0, the NUMA nodes may not be accessible, and that's apparently the case in the docker that run the CI. Now if `numa_get_mems_allowed()` to returns an empty bitmask, we expect all allocations to fail.

AFAICT the larger Github provided runners are perhaps _NUMA capable_ single node systems which might explain the previous CI failures in those workflows. Whereas the self-hosted workflow runs on our internal dual-socket systems.

illuhad:
> You mention adding a Macro in sycl/extension.hpp, would you want something like this :

No, I meant it in the way you are doing it now :)

- The macro in `extensions.hpp` is always defined and set to 1. It tells client code whether the *API* for the new property is available, and could help write portable code when it's not clear whether the compiler might be an older version of AdaptiveCpp, or perhaps another SYCL compiler that does not implement the API.
- We could debate whether it might be useful for users to have a separate macro which could indicate whether the property actually has an effect and AdaptiveCpp was built with NUMA support. But I think we could also postpone this until we have more experience with what actual user requirements are.

illuhad (approved):
Thank you :)

PR #1690, opened 2025-01-22
Add LLVM 19 to Linux CI

VaiTon:

PR #1742, opened 2025-03-13
[CI] Restructure CI to use docker containers (and enable LLVM 19&20)

fodinabor:
This adds a CI dockerfile and infrastructure to use the created docker images for the Linux CI flows.
Ideally, this should give us overall faster turn-around times for the CI builds.
The docker images are always built on-demand if something about the dockerfile or create-ci-docker action has changed.
This is also true for PRs and non-develop branches.
For these, the image tags are suffixed with `-<branch-name>` or `-pr-<id>`.
We need to use some hackery-pokery to propagate the image names properly from the docker job to the actual CI jobs
We use the beacon-biosignals/matrix-output@v1 for this.

As an additional change, given that job starting time got quite a bit cheaper, I also split up the linux.yml tests into multiple groups: `cpu`, `gpu` and `sscp` with the goal to reduce overall CI latency from ~50min to ~35min at the cost of adding more parallel jobs. (moving explicit multi pass from GPU to CPU variant, might improve this even more... but is unintuitive 🤔 :) )

Oh.. and also, we're using ccache for building AdaptiveCpp now -> for smaller changes (e.g. test only), building AdaptiveCpp can take as little as 30s instead of 5min+.

This PR, furthermore, includes compatibility fixes with LLVM 19 and 20 and enables them in CI.

Note: it's also kinda nice to have these docker images around for a "stable" container environment one can use on clusters and so on :)

Fixes #1690 
Prepares having CI for #1609

### DIFF
--- a/.github/workflows/create-ci-docker/AdaptiveCpp-CI-Linux.dockerfile
+++ b/.github/workflows/create-ci-docker/AdaptiveCpp-CI-Linux.dockerfile
@@ -1,0 +1,102 @@
+ARG BASE="ubuntu:22.04"
+
+FROM ${BASE}
+
+ARG LLVM_VERSION=18
+# make empty to skip installation of ROCM and Intel Stack
+ARG ROCM_VERSION="6.2.4"
+# make empty to skip installation of CUDA
+ARG CUDA_MAJOR_VERSION="11"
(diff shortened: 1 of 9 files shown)


fodinabor:
Super seeded by #1743

PR #1743, opened 2025-03-13
[CI] Restructure CI to use docker containers (and enable LLVM 19&20)

fodinabor:
This adds a CI dockerfile and infrastructure to use the created docker images for the Linux CI flows.
Ideally, this should give us overall faster turn-around times for the CI builds.
The docker images are always built on-demand if something about the dockerfile or create-ci-docker action has changed.
This is also true for PRs and non-develop branches.
For these, the image tags are suffixed with `-<branch-name>` .
We need to use some hackery-pokery to propagate the image names properly from the docker job to the actual CI jobs
We use the beacon-biosignals/matrix-output@v1 for this.

As an additional change, given that job starting time got quite a bit cheaper, I also split up the linux.yml tests into multiple groups: `cpu`, `gpu` and `sscp` with the goal to reduce overall CI latency from ~50min to ~35min at the cost of adding more parallel jobs. (moving explicit multi pass from GPU to CPU variant, might improve this even more... but is unintuitive 🤔 :) )

Oh.. and also, we're using ccache for building AdaptiveCpp now -> for smaller changes (e.g. test only), building AdaptiveCpp can take as little as 30s instead of 5min+.

This PR, furthermore, includes compatibility fixes with LLVM 19 and 20 and enables them in CI.

Since we can't push the docker images from PRs, ignore the CI failures of the PR for now and just look at the "push" CI.
Once this PR is merged, fallback images are created as necessary to fuel future PRs.

Note: it's also kinda nice to have these docker images around for a "stable" container environment one can use on clusters and so on :)

Fixes #1690 
Superseeds #1742 
Prepares having CI for #1609


.github/workflows/create-ci-docker/AdaptiveCpp-CI-Linux.dockerfile
```diff
@@ -0,0 +1,102 @@
...
+FROM ${BASE}
+
+ARG LLVM_VERSION=18
+# make empty to skip installation of ROCM and Intel Stack
+ARG ROCM_VERSION="6.2.4"
+# make empty to skip installation of CUDA
+ARG CUDA_MAJOR_VERSION="11"
+ARG CUDA_MINOR_VERSION="0"
+ARG CUDA_PATCH_VERSION="2"
+ARG CUDA_DRIVER_VERSION="450.51.05"
```
illuhad, line 12: Why do we now need this? I don't think we've had to fix a specific driver version before? (which wouldn't matter anyway, because we don't have CUDA hardware in the github runners, so no driver needed)

.github/workflows/linux-lit.yml
```diff
@@ -46,95 +74,87 @@ jobs:
           mkdir build && cd build
           echo ${CXXFLAGS}
           
-          cmake -DCMAKE_CXX_COMPILER=/usr/bin/clang++-${{matrix.clang}} -DOMP_CXX_FLAGS="$OMP_CXX_FLAGS -fopenmp" -DCLANG_EXECUTABLE_PATH=/usr/bin/clang++-${{matrix.clang}} -DLLVM_DIR=/usr/lib/llvm-${{matrix.clang}}/cmake .. -DCMAKE_INSTALL_PREFIX=`pwd`/install
-          make -j2 install
+          cmake -GNinja -DCMAKE_CXX_COMPILER=/usr/bin/clang++-${{matrix.docker.clang}} -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DOMP_CXX_FLAGS="$OMP_CXX_FLAGS -fopenmp" -DCLANG_EXECUTABLE_PATH=/usr/bin/clang++-${{matrix.docker.clang}} -DLLVM_DIR=/usr/lib/llvm-${{matrix.docker.clang}}/cmake .. -DCMAKE_INSTALL_PREFIX=`pwd`/install
+          ninja install
```
illuhad, line 78: Why was it necessary to switch to `ninja`?

illuhad (commented):
Thanks!
There may be some conflicts incoming.. #1737 adds additional stuff to CI. Not sure whether it's a better strategy to first merge #1737, or this PR?

.github/workflows/create-ci-docker/AdaptiveCpp-CI-Linux.dockerfile
fodinabor: It's just a part of the installer URL ... didn't want to hardcode it down there... https://github.com/AdaptiveCpp/AdaptiveCpp/blob/f9273f1075f5ab982164388dab90350fe3ce8ce0/.github/workflows/linux.yml#L39

.github/workflows/linux-lit.yml
fodinabor: No necessary... just makes things easier, as we don't have to hardcode the core count and dependency resolution works a bit better... (more optimal)

illuhad: ok! :)

.github/workflows/linux.yml
```diff
@@ -3,116 +3,136 @@ name: Linux build and test
...
+            image: ${{ steps.docker.outputs.docker_image }}
+
+  test-clang-based:
+    name: clang ${{ matrix.docker.clang }}, ${{ matrix.test_kind }}, CUDA ${{matrix.docker.cuda}}, ROCm ${{matrix.docker.rocm}}, ${{ matrix.docker.os }}
+    runs-on: ${{ matrix.docker.os }}
+    needs: prep-container
+    strategy:
+      matrix:
+        docker: ${{fromJson(needs.prep-container.outputs.docker_matrix)}}
+        test_kind: [cpu, sscp, gpu]
```
illuhad: Perhaps we should consider renaming this, as SSCP is also CPU/GPU in a way.... perhaps `omp, sscp, smcp`? Or `omp, sscp, smcp-interop`?
 But we can also leave it as is for the time being, not a strong issue for me :)

illuhad (approved):
Nice nice nice!

PR #1744, opened 2025-03-17
[stdpar][algorithms] Add support for `equal`

normallytangent:
This PR adds automatic offloading support for all of the four overloads of [equal](https://en.cppreference.com/w/cpp/algorithm/equal)


include/hipSYCL/algorithms/algorithm.hpp
```diff
@@ -543,6 +543,99 @@ sycl::event none_of(sycl::queue &q,
...
+    return sycl::event{};
+  auto evt = detail ::early_exit_for_each(q, problem_size, out,
+                                          [=](sycl::id<1> idx) -> bool {
+                                            auto it1 = first1;
+                                            auto it2 = first2;
+                                            std::advance(it1, idx[0]);
+                                            std::advance(it2, idx[0]);
+                                            return !(*it1 == *it2);
+                                          }, deps);
+  // use memcmp to short-circuit sequential implementation
```
illuhad: I don't understand this comment. We don't use `memcmp` here?

normallytangent: My comment here is wrong, the standard talks about the sequential implementation. Since it's not needed so I'll remove it. The comment was to note what was mentioned on cppreference:

> Sequential std::equal is not guaranteed to be short-circuit. E.g. if the first pair elements of both ranges do not compare equal, the rest of elements may also be compared. Non-short-circuit comparison may happen when the ranges are compared with [std::memcmp](https://en.cppreference.com/w/cpp/string/byte/memcmp) or implementation-specific vectorized algorithms.

illuhad (approved):
Thanks!

PR #1751, opened 2025-03-21
Refactor the builtins to use more overloads

carbotaniuman:
This is a wholesale refactor of the builtins, splitting them into 3 files for generic, math + integer, as well as relational builtins. As part of the refactoring, some functions have been refactored from solely using templates to using overloads and templates, while others have had their templates updated according to the latest spec.

Because of the support for floats and vectors/marrays, the actual `builtin_interface` has been changed, particularly with the functions taking pointer arguments.

Note: The new builtins are designed to support `half` in varying states, sometimes by just casting to float and doing the operations on float, but the half support should be treated as broken on most platforms. (The support for half on device code in multipass is very weird and spews lots of errors, and the accuracy tests are fail heavily if extended to half).


illuhad:
Great work! I'm a bit confused - you say in the title that you make the builtins use more overloads, but the code seems to use less overloads, and more templates instead? Or am I misunderstanding the change?

carbotaniuman:
Yeah, the wording there is a bit imprecise. If you look at the 3 headers that replace the old `builtin_utils`, `math_builtins` uses (some) overloads, while `generic_builtins` and `relational_builtins` are all templates. The code duplication and repetition is sort of unfortunate, but the SYCL spec for all the categories is different enough that it's not really possible to nicely share code there.

illuhad:
I see. I was just asking because there was a change in the SYCL 2020 spec to define the builtins as a mixture of templates and overloads (mostly overload for `float/double/half` and template for `marray/vec`), which matters for overload resolution. But it seems your PR already takes this into account :)

illuhad:
To be merged after 25.02 release

illuhad (approved):
Thanks! Very cool work. Our test cases only test builtin functions selectively, and such a big change is a little scary :) Have you by chance tried to run CTS with this? 

> Note: The new builtins are designed to support half in varying states, sometimes by just casting to float and doing the operations on float, but the half support should be treated as broken on most platforms. (The support for half on device code in multipass is very weird and spews lots of errors, and the accuracy tests are fail heavily if extended to half).

half is an optional feature; we could in principle say that it is e.g. only fully supported with `generic` target (SSCP compiler).

carbotaniuman:
I got the math builtin CTS working (for some definition of working, see [this branch](https://github.com/carbotaniuman/SYCL-CTS/tree/update-acpp-bypasses)), but I did get the math builtins test to run. I also hacked ACPP, see [this branch](https://github.com/carbotaniuman/AdaptiveCpp/tree/hack), but I think it works fineish.

These are the results: https://github.com/carbotaniuman/SYCL-CTS/blob/e51802a759ebda335c53b5952f41f487afe24bd2/foo

Those are sinpi and cospi like things which we fail because of precision, which I think is existing. Unfortunately this reveals that the way we handle half is just broken and blows up on a couple of edge cases. We also don't advertise fp16 support so the CTS doesn't build it, and if we enable it everything crashes terribly. It probably needs another whole refactor.

I am fairly confident that the non-half functions are correct, but I'm a bit unsure how to proceed with the half ones. I think the best way to do it would be to properly add overloads for all the builtins for half and probably to deprecate/remove the SCMP support.

I would like to get this merged, but given the absolutely shambolic half support I feel a bit uncomfortable doing so. Do you have any thoughts?

illuhad:
> I got the math builtin CTS working (for some definition of working, see [this branch](https://github.com/carbotaniuman/SYCL-CTS/tree/update-acpp-bypasses)), but I did get the math builtins test to run. I also hacked ACPP, see [this branch](https://github.com/carbotaniuman/AdaptiveCpp/tree/hack), but I think it works fineish.

That's great :)

> Those are sinpi and cospi like things which we fail because of precision, which I think is existing. 

IMO that might be a bug in the CTS and we should consider opening an issue there. The SYCL math builtin precision is implementation-defined, so CTS would be wrong to check for a specific accuracy here. This was recently-ish clarified: https://github.com/KhronosGroup/SYCL-Docs/pull/511

> Unfortunately this reveals that the way we handle half is just broken and blows up on a couple of edge cases. We also don't advertise fp16 support so the CTS doesn't build it, and if we enable it everything crashes terribly. It probably needs another whole refactor. I am fairly confident that the non-half functions are correct, but I'm a bit unsure how to proceed with the half ones. I think the best way to do it would be to properly add overloads for all the builtins for half and probably to deprecate/remove the SCMP support.

Can you summarize what the issue is?

I think in the long run we probably won't get around adding `half` overloads to the builtin interface which the backends can then implement in whatever way makes sense for them/the target hardware. In this case, the SYCL builtin definitions in the header would just call into the builtin interface in the same way as for float or double.

We could then either implement these overloads in the builtin interface by blindly casting to float for SMCP, or implement them properly if we are more motivated.

> I would like to get this merged, but given the absolutely shambolic half support I feel a bit uncomfortable doing so. Do you have any thoughts?

Even if the PR is perhaps not perfect, as far as I understand it, it doesn't make anything worse, right? As you point out we don't *really* advertise half. So I wouldn't be opposed to merging it. We can still improve half support in the future.

carbotaniuman:
Basically the way I implemented half by delegating to double is slightly broken. After thinking about it I think it would be fine to merge it now, half is currently not that very well supported anyways and it's probably better to address that in a follow up.

illuhad:
Alright then, in it goes :) Thanks!

fxzjshm:
Sorry to disturb, but this pull request breaks at least `sycl::modf` on host:
```
In file included from modf.cpp:1:
In file included from /opt/hipSYCL/bin/../include/AdaptiveCpp/sycl/sycl.hpp:14:
In file included from /opt/hipSYCL/bin/../include/AdaptiveCpp/sycl/../hipSYCL/sycl/sycl.hpp:52:
In file included from /opt/hipSYCL/bin/../include/AdaptiveCpp/sycl/../hipSYCL/sycl/libkernel/builtins.hpp:14:
In file included from /opt/hipSYCL/bin/../include/AdaptiveCpp/hipSYCL/sycl/libkernel/builtins/math_builtins.hpp:22:
/opt/hipSYCL/bin/../include/AdaptiveCpp/hipSYCL/sycl/libkernel/builtin_interface.hpp:261:3: error: no matching function for call to '__acpp_modf'
  261 |   HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_modf, x, y);
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/opt/hipSYCL/bin/../include/AdaptiveCpp/hipSYCL/sycl/libkernel/detail/builtin_dispatch.hpp:24:32: note: expanded from macro 'HIPSYCL_RETURN_DISPATCH_BUILTIN'
   24 |   __acpp_backend_switch(return host_builtins::name(__VA_ARGS__),            \
      |   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   25 |                                   return sscp_builtins::name(__VA_ARGS__),     \
      |                                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   26 |                                   return hiplike_builtins::name(__VA_ARGS__),  \
      |                                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   27 |                                   return hiplike_builtins::name(__VA_ARGS__))
      |                                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/opt/hipSYCL/bin/../include/AdaptiveCpp/sycl/../hipSYCL/sycl/libkernel/backend.hpp:136:5: note: expanded from macro '__acpp_backend_switch'
  136 |     host_code;                                                                 \
      |     ^~~~~~~~~
/opt/hipSYCL/bin/../include/AdaptiveCpp/hipSYCL/sycl/libkernel/builtins/math_builtins.hpp:363:1: note: in instantiation of function template specialization 'hipsycl::sycl::detail::__acpp_modf<float>' requested here
  363 | HIPSYCL_MATH_BUILTIN(BINARY_T_TPTR, modf)
      | ^
/opt/hipSYCL/bin/../include/AdaptiveCpp/hipSYCL/sycl/libkernel/builtins/math_builtins.hpp:299:5: note: expanded from macro 'HIPSYCL_MATH_BUILTIN'
  299 |     HIPSYCL_PP_CONCATENATE(detail::__acpp_, builtin_name)                    \
      |     ^
/opt/hipSYCL/bin/../include/AdaptiveCpp/hipSYCL/sycl/libkernel/builtins/builtin_utils.hpp:22:38: note: expanded from macro 'HIPSYCL_PP_CONCATENATE'
   22 | #define HIPSYCL_PP_CONCATENATE(x, y) HIPSYCL_PP_CONCATENATE_IMPL(x,y)
      |                                      ^
/opt/hipSYCL/bin/../include/AdaptiveCpp/hipSYCL/sycl/libkernel/builtins/builtin_utils.hpp:21:42: note: expanded from macro 'HIPSYCL_PP_CONCATENATE_IMPL'
   21 | #define HIPSYCL_PP_CONCATENATE_IMPL(x,y) x ## y
      |                                          ^
<scratch space>:67:1: note: expanded from here
   67 | __acpp_modf
      | ^
modf.cpp:23:34: note: in instantiation of function template specialization 'hipsycl::sycl::modf<hipsycl::sycl::multi_ptr<float, hipsycl::sycl::access::address_space::private_space, hipsycl::sycl::access::decorated::yes>>' requested here
   23 |                 d_c[idx] = sycl::modf(d_a[idx], sycl::decorated_private_ptr<float>{&d_b[idx]});
      |                                  ^
/opt/hipSYCL/bin/../include/AdaptiveCpp/hipSYCL/sycl/libkernel/host/builtins.hpp:267:19: note: candidate template ignored: couldn't infer template argument 'T'
  267 | HIPSYCL_BUILTIN T __acpp_modf(float x, float* y) noexcept {
      |                   ^
/opt/hipSYCL/bin/../include/AdaptiveCpp/hipSYCL/sycl/libkernel/host/builtins.hpp:272:19: note: candidate template ignored: couldn't infer template argument 'T'
  272 | HIPSYCL_BUILTIN T __acpp_modf(double x, double* y) noexcept {
      |                   ^
1 error generated.
```

Not seen on hiplike.

source:

<details><summary>Details</summary>
<p>

```c++
#include <sycl/sycl.hpp>
#include <vector>
#include <iostream>

int main() {
    const size_t N = 1024;
    std::vector<float> a(N, 1.2f);
    std::vector<float> b(N, 2.0f);
    std::vector<float> c(N, 0.0f);

    try {
        sycl::queue q(sycl::default_selector_v);
        std::cout << q.get_device().get_info<sycl::info::device::name>() << "\n";

        float *d_a = sycl::malloc_device<float>(N, q);
        float *d_b = sycl::malloc_device<float>(N, q);
        float *d_c = sycl::malloc_device<float>(N, q);
        q.memcpy(d_a, a.data(), N * sizeof(float));
        // q.memcpy(d_b, b.data(), N * sizeof(float));

        q.submit([&](sycl::handler& h) {
            h.parallel_for(sycl::range<1>(N), [=](sycl::id<1> idx) {
                d_c[idx] = sycl::modf(d_a[idx], sycl::decorated_private_ptr<float>{&d_b[idx]});
            });
        }).wait();

        q.memcpy(b.data(), d_b, N * sizeof(float)).wait();
        q.memcpy(c.data(), d_c, N * sizeof(float)).wait();

        bool correct = true;
        for (size_t i = 0; i < N; i++) {
            if (std::abs(b[i] - 1.0f) > 1e-5) {
                std::cout << "error: b[" << i << "] = " << b[i] << "\n";
                correct = false;
                break;
            }
            if (std::abs(c[i] - 0.2f) > 1e-5) {
                std::cout << "error: c[" << i << "] = " << c[i] << "\n";
                correct = false;
                break;
            }
        }
        if (correct) {
            std::cout << "correct!\n";
        }

        sycl::free(d_a, q);
        sycl::free(d_b, q);
        sycl::free(d_c, q);
    } catch (const sycl::exception& e) {
        std::cerr << e.what() << std::endl;
        return 1;
    }

    return 0;
}

```

</p>
</details>

carbotaniuman:
Sorry about that, put a PR up in #1893.

PR #1752, opened 2025-03-21
[CI] Add a dispatchable workflow to prime docker images.

fodinabor:
This is necessary (at least, I haven't been able to think of anything way smarter, yet) to prime the cache for pull_requests that need a new docker image variant.

### DIFF
--- a/.github/workflows/prime-docker.yml
+++ b/.github/workflows/prime-docker.yml
@@ -1,0 +1,52 @@
+name: Prime docker container
+
+on:
+  workflow_dispatch:
+    inputs:
+      clang:
+        description: 'LLVM Version to install'
+        required: true
+        default: 20
+      os:
+        description: 'Base Operating System'
+        required: true
+        default: "ubuntu-22.04"
+      cuda_major_version:
+        description: 'CUDA Major Version'
+        required: true
+        default: 11
+      cuda_minor_version:
+        description: 'CUDA Minor Version'
+        required: true
+        default: 0
+      cuda_patch_version:
+        description: 'CUDA Patch Version'
+        required: true
+        default: 2
+      rocm:
+        description: 'ROCm Version'
+        required: true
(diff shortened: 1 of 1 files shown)


illuhad (approved):
Thanks! What would the docker image workflow look like as a whole?

Whenever we change the dockerfile, we run this workflow? And we need to manually run the workflow for each combination of LLVM/CUDA/ROCm that we run tests with?

fodinabor:
It's only really needed/useful for Pull Requests that add new matrix entries, i.e. use a different version of LLVM, CUDA, ...
After a PR that changes the Dockerfile / docker ci action, is merged to `develop`, the "default" images are updated - these, in turn, are then used for all Pull Requests (and branches that don't touch the Dockerfile / docker-ci-action) thereafter.

So, one case where this is useful is: https://github.com/AdaptiveCpp/AdaptiveCpp/pull/1609/files#diff-6e0188d806ebf79b6ae348d43ffdb70ec4ec0b0d8ee83deb246d9bd4c02eb247R18-R23
Where we require a new image (LLVM 17 with CUDA **or** ROCM, vs the existing ones being LLVM 17 + no GPU stuff or LLVM 17 + all GPU stuff) that has not been created/used before - but doesn't necessarily change much about the image itself... I haven't figured out how to deal well with that scenario (esp with PRs from forks).
In that example, we would run this workflow for `[{clang: 17, cuda: 11, rocm: ""}, {clang: 17, cuda: "", rocm: 5.6.1}]`.

Note, this does not update the images with changes to the Dockerfile from the Pull Request.

Note 2, the push actions don't suffer from this, they create branch-specific new docker images that are immediately used for the rest of the actions. Therefore we could also argue that for PRs that change the CI in this way, we just look at the push action (which has the downside that it might be on a fork & we don't see what happens after the automated merge with `develop`..)

PR #1754, opened 2025-03-23
Add `get_backend` for the platform

Oblomov:
The first commit is a trivial change that implements `get_backend` for the SYCL platform class too, as required by the spec.

The second commit is a trivial cleanup patch that removes some spurious whitespace and fixes a typo (I felt there was no need for a separate PR for this, but do let me know if it would be preferred).


illuhad (approved):
Thanks!

PR #1756, opened 2025-03-24
[stdpar][algorithms] Add support for `find`, `find_if`, and related STL algorithms

normallytangent:
This PR adds automatic offloading support for the following:
- [x] `find`
- [x] `find_if`
- [x] `find_if_not`
- [x] `find_end`
- [x] `find_first_of`


include/hipSYCL/algorithms/numeric.hpp
```diff
@@ -191,6 +191,23 @@ transform_reduce(sycl::queue &q, util::allocation_group &scratch_allocations,
...
+                 BinaryReductionOp reduce, UnaryTransformOp transform,
+                 const std::vector<sycl::event>& deps = {}) {
+  if(first == last)
+    return sycl::event{};
+
+  std::size_t n = std::distance(first, last);
+
+  return detail::transform_reduce_impl(q, scratch_allocations, out, init, n,
+                                       kernel, reduce, deps);
+}
```
normallytangent: `find_helper` here is essentially `transform_reduce`. Since I wanted to adapt the `transform` lambda, I added a separate function instead of modifying the already implemented overloads.

I face the issue that a call to find either returns an uninitialised memory :( or just the first position of the vector. Not sure where I make a mistake, maybe in the call to reducer?

include/hipSYCL/algorithms/algorithm.hpp
```diff
@@ -411,6 +412,32 @@ sycl::event replace_copy(sycl::queue &q, ForwardIt1 first, ForwardIt1 last,
...
+    return (*input == value ? std::distance(first, input) : problem_size);
+  };
+
+  auto kernel = [=](sycl::id<1> idx, auto& reducer) {
+    auto input = first;
+    std::advance(input, idx[0]);
+    reducer.combine(transform(input));
+  };
+
+  auto reduce = sycl::minimum<std::size_t>{};
```
normallytangent: My suspicion is that I need to implement a reduction specific to find

include/hipSYCL/algorithms/algorithm.hpp
```diff
@@ -411,6 +412,36 @@ sycl::event replace_copy(sycl::queue &q, ForwardIt1 first, ForwardIt1 last,
...
+    std::advance(val, i);
+    return (input == value ? std::distance(first, val) : problem_size);
+  };
+
+  auto kernel = [=](sycl::id<1> idx, auto& reducer) {
+    auto input = first;
+    std::advance(input, idx[0]);
+    auto i = std::distance(first, input);
+    reducer.combine(transform(*input, i));
+  };
```
normallytangent, line 497: I think I've narrowed it down; I don't understand nested lambdas.

normallytangent:
@illuhad I reordered the Self-hosted Intel workflow. Looks like the same test as in #1744 fails (transform_exclusive_scan on Intel L0) but for different  execution policies `par_large` here whereas `par_unseq_large` in `std::equal`

illuhad (approved):
Thanks! :) Sorry for the wrong brace, looks like I messed up when updating the PR with current `develop` branch.

PR #1763, opened 2025-03-27
[CBS] Get way more conservative on when not to widen allocas.

### LINKED ISSUE #1759: [Omp] capturing a sycl::group in lambda messes with group reductions
Here is the issue @fodinabor @illuhad 

If we take the following code snippet:
```c++
#include <sycl/sycl.hpp>

using f64_3 = sycl::vec<double, 3>;

template<class T, class Func>
inline T map_vector(const T &in, Func && f) {
    return {f(in[0]), f(in[1]), f(in[2])};
}

int main() {
    sycl::queue q{};

    std::vector<f64_3> input_data
        = {{1.0, 2.0, 3.0}, {4.0, 5.0, 6.0}, {7.0, 8.0, 9.0}, {10.0, 11.0, 12.0}};
    std::vector<f64_3> output_data(input_data.size(), {0.0, 0.0, 0.0});

    sycl::buffer<f64_3> input_buf(input_data.data(), input_data.size());
    sycl::buffer<f64_3> output_buf(output_data.data(), output_data.size());

        sycl::accessor in(input_buf, cgh, sycl::read_only);
        sycl::accessor out(output_buf, cgh, sycl::write_only);

        cgh.parallel_for(
            sycl::nd_range<1>(input_data.size(), input_data.size()), [=](sycl::nd_item<1> item) {
                sycl::group<1> group = item.get_group();

                auto v = in[item.get_global_linear_id()];

                f64_3 sum
                    = map_vector(v, [&](auto component) {
                        auto ret = sycl::reduce_over_group(group, component, sycl::plus<decltype(component)>{});
                        return ret;
                    });

                std::cout << v[0] << " -> " << sum[0] << " | " << v[1] << " -> " << sum[1] << " | "
                          << v[2] << " -> " << sum[2] << std::endl;

                out[item.get_global_linear_id()] = sum;
            });
    });
}
```
and compile it as follows `acpp --acpp-targets=omp -O3 test_issue.cpp` and run it the result is
```
❯ ./a.out                                                              
1 -> 22 | 2 -> 26 | 3 -> 30
4 -> 22 | 5 -> 26 | 6 -> 30
7 -> 22 | 8 -> 26 | 9 -> 30
10 -> 22 | 11 -> 26 | 12 -> 30
```
which is expected behaviour.

Now if we instead compile without -O3 `acpp --acpp-targets=omp test_issue.cpp`, the result is
```
❯ ./a.out                                                          
1 -> 22 | 2 -> 22 | 3 -> 22
4 -> 22 | 5 -> 22 | 6 -> 22
7 -> 22 | 8 -> 22 | 9 -> 22
10 -> 22 | 11 -> 22 | 12 -> 22
```
As if only the first reduction is performed and the others are just copy paste of the first one result.

## Configurations we tried

- If we compile using `--acpp-targets=omp.library-only` the issue does not appear.
- If we capture the sycl::group by copy in the lambda or if we just forward it and have a lambda without any captures the issue does not appear either.

In general this seems to only occur whenever a sycl::group is passed by reference to a lambda.

fodinabor:
We have to widen allocas, so that each work item get's it's own alloca slot, whenever, the alloca is used across multiple SubCFGs.
This commit adds into consideration that this might as well happen, when the pointer to the alloca is stored somewhere else.
This can happen whenever the pointer is stored or if it is being used by a function.
When it's used by a function, we try to be nice and don't arrayify whenever the pointer operand is marked as "nocapture" or as "sret" (struct return, i.e. being initialized).

Fixes #1759

### DIFF
--- a/src/compiler/cbs/SubCfgFormation.cpp
+++ b/src/compiler/cbs/SubCfgFormation.cpp
@@ -158,10 +158,11 @@ getLocalSizeArgumentFromAnnotation(llvm::Function &F) {
 }
 
 // identify the local size values by the store to it
-void fillStores(llvm::Value *V, int Idx, bool IsIdxBytewise, llvm::SmallVector<llvm::Value *, 3> &LocalSize) {
+void fillStores(llvm::Value *V, int Idx, bool IsIdxBytewise,
+                llvm::SmallVector<llvm::Value *, 3> &LocalSize) {
   if (auto *Store = llvm::dyn_cast<llvm::StoreInst>(V)) {
 #if LLVM_VERSION_MAJOR >= 17
-    if(IsIdxBytewise)
+    if (IsIdxBytewise)
       Idx /= Store->getAccessType()->getScalarSizeInBits() / 8;
 #endif
     LocalSize[Idx] = Store->getOperand(0);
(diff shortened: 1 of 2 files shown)


illuhad (approved):
Thanks :)

PR #1770, opened 2025-04-10
Fix atomic_ref math operators

Stastez:
According to the [SYCL 2020 spec](https://github.khronos.org/SYCL_Reference/iface/atomic_ref.html), AdaptiveCpp currently does not adhere to the standard for the `operator*` functions in `atomic_ref`. This commit aligns the implementation with the standard.

(Thanks to @coledea for pointing out this inconsistency.)


illuhad (approved):
Thanks, good catch!

PR #1772, opened 2025-04-16
Make LLVM_VERSION_MAJOR not being set not fail CMake configure on Darwin

KipHamiltons:
This change checks if the variable is not set and issues a warning that advanced features of AdaptiveCpp won't be enabled, when it isn't set. The current behaviour when LLVM_VERSION_MAJOR is not set is to fail because `( LESS 14)` isn't a valid condition.


illuhad:
Why would `LLVM_VERSION_MAJOR` not be set? If I see it correctly, this part of the code is already in the branch where we want to build LLVM-based AdaptiveCpp components, so if we don't have `LLVM_VERSION_MAJOR`, doesn't this mean that something is broken?

KipHamiltons:
> Why would `LLVM_VERSION_MAJOR` not be set? If I see it correctly, this part of the code is already in the branch where we want to build LLVM-based AdaptiveCpp components, so if we don't have `LLVM_VERSION_MAJOR`, doesn't this mean that something is broken?

Good point! I ran into this before I set `LLVM_DIR`, which was required anyway. A better change is to make it clearer in the docs that `LLVM_DIR` may be required (on Apple).

Thanks for the review @illuhad !

PR #1774, opened 2025-04-17
[NFC][doc] Clarify that LLVM_DIR may be required when building on Mac.

KipHamiltons:
I can't build on Mac without `LLVM_DIR` set.


KipHamiltons:
Following up from #1772

illuhad (approved):
Thanks! There is also a homebrew made by @tdavidcl  package that we could mention: https://formulae.brew.sh/formula/adaptivecpp#default

PR #1777, opened 2025-04-18
Add note about homebrew distribution to installation docs

KipHamiltons:
We have a homebrew distribution, so we might as well mention it!

### DIFF
--- a/doc/installing.md
+++ b/doc/installing.md
@@ -6,6 +6,12 @@
 
 Operating system support currently strongly focuses on Linux. On Mac, only the CPU backend is expected to work. Windows support with CPU and CUDA backends is experimental, see [Using AdaptiveCpp on Windows](https://github.com/OpenSYCL/OpenSYCL/wiki/Using-AdaptiveCpp-on-Windows).
 
+## Installation as a package
+
+The following package managers have AdaptiveCpp available:
+* The Arch Linux AUR has a package under the name [adaptivecpp](https://aur.archlinux.org/packages/adaptivecpp).
+* A Homebrew distribution for MacOS is available under the name [adaptivecpp](https://formulae.brew.sh/formula/adaptivecpp#default).
+
 ## Installation from source (Linux)
 
 ### Software dependencies


doc/installing.md
```diff
@@ -4,7 +4,7 @@
 
 ## Operating system support
 
-Operating system support currently strongly focuses on Linux. On Mac, only the CPU backend is expected to work. Windows support with CPU and CUDA backends is experimental, see [Using AdaptiveCpp on Windows](https://github.com/OpenSYCL/OpenSYCL/wiki/Using-AdaptiveCpp-on-Windows).
+Operating system support currently strongly focuses on Linux. On Mac, only the CPU backend is expected to work. A Homebrew distribution is available under the name [adaptivecpp](https://formulae.brew.sh/formula/adaptivecpp#default). Windows support with CPU and CUDA backends is experimental, see [Using AdaptiveCpp on Windows](https://github.com/OpenSYCL/OpenSYCL/wiki/Using-AdaptiveCpp-on-Windows).
```
tdavidcl: Maybe we could also add the command to clarify `brew install adaptivecpp`. (maybe even a subsection about packaging)

Also I think that it is important to mention that, as is the home-brew package currently only support/test OpenMP backend. SSCP, CUDA or others might work but are not tested at all. If I remember correctly I even set SSCP to off in the home-brew package when I upstreamed it.

tdavidcl (commented):
Nice, good to see that people are using the home-brew package (It was a pain to package properly on all platforms). 

I just had a few comments on this side as I'm currently maintaining it :)

doc/installing.md
illuhad: Oh... I thought we had SSCP in the homebrew package.

CUDA or other GPU backends is not a problem, since they are unlikely to work on Mac anyway (Apple closes off everything regarding GPU).

But SSCP for OpenMP backend would almost be a must for advertising the package in the main documentation... Without SSCP, there's no PCUDA support, and other functionality is lost as well. Can you check whether the homebrew package as it to confirm?

tdavidcl: https://github.com/Homebrew/homebrew-core/blob/a6c7d924f102692d20b88f1f5d9ea81e94d85d32/Formula/a/adaptivecpp.rb

Never mind it’s not disabled (explicitly). Though if i remember well SSCP is not enabled on MacOS.

KipHamiltons: I added an installation subsection. Good idea @tdavidcl ! I wish there were more distros packaging it though 😆

illuhad: @tdavidcl We do have some CI flows that test SSCP on Mac: https://github.com/AdaptiveCpp/AdaptiveCpp/blob/develop/.github/workflows/macos-acppllvm.yml
So in principle it should work.

Can you try whether it works with the homebrew package?

tdavidcl: I see that you get it using acpp linking in llvm. Does it work with a standard acpp over a in-house built llvm ? Wasn't SSCP unsupported with apple clang or something like that ?

illuhad: @tdavidcl To my knowledge, the linking-in-llvm is mainly needed for Windows where clang plugins are unsupported. The fact that SSCP is not tested on Mac with plugin is most likely simply due to nobody adding a workflow for it. I'm fairly sure that other flows with the plugin work as well on Mac (e.g. `omp.accelerated`), so SSCP should work too. Whether it requires normal LLVM or Apple clang is sufficient I don't know though.

tdavidcl: So the home-brew package does not seem to compile anything with SSCP, and cloning acpp on develop and running `cmake .` on MacOS yield:
```
CMake Warning at CMakeLists.txt:603 (message):
  The generic SSCP compiler is disabled; without it, offloading code to
  accelerators requires explicitly specifying --acpp-targets when invoking
  acpp.  A binary generated by invoking acpp without --acpp-targets will only
  be able to run on the host.  Set -DDEFAULT_TARGETS to your desired default
  targets during cmake configuration to avoid this warning.  In the future,
  configurations without the SSCP compiler enabled might potentially be only
  supported if -DDEFAULT_TARGETS is provided explicitly.
```

tdavidcl: in the main CmakeLists.txt
```cmake
if((APPLE OR WIN32) AND ACPP_LLVM_COMPONENT)
  set(DEFAULT_COMPILER_FEATURES_ENABLED true)
  set(ACPP_COMPILER_FEATURE_PROFILE "full" CACHE STRING "The level of AdaptiveCpp compiler features to enable. If not set to 'full', AdaptiveCpp will run in a degraded mode that may have implications for certain functionalities or performance in certain cases.")
  if(WIN32)
    add_definitions(-DWIN32_LEAN_AND_MEAN -DNOMINMAX -D_USE_MATH_DEFINES)
  endif()
elseif(APPLE OR WIN32)
  set(DEFAULT_COMPILER_FEATURES_ENABLED false)
  set(ACPP_COMPILER_FEATURE_PROFILE "none" CACHE STRING "The level of AdaptiveCpp compiler features to enable. If not set to 'full', AdaptiveCpp will run in a degraded mode that may have implications for certain functionalities or performance in certain cases.")
else()
  set(DEFAULT_COMPILER_FEATURES_ENABLED true)
  set(ACPP_COMPILER_FEATURE_PROFILE "full" CACHE STRING "The level of AdaptiveCpp compiler features to enable. If not set to 'full', AdaptiveCpp will run in a degraded mode that may have implications for certain functionalities or performance in certain cases.")
endif()
set_property(CACHE ACPP_COMPILER_FEATURE_PROFILE PROPERTY STRINGS full minimal none)
```
So unless it is linked into LLVM it seems disabled by default on macOS

fodinabor: In my attempts I was unable to find a compatible abi mix of libraries only using brew.. That's why I set sscp enabled for the linking into llvm only.

PR #1781, opened 2025-04-19
[stdpar][algorithms] Add support for `is_sorted` and `is_sorted_until`

normallytangent:
This PR adds automatic offloading support for [is_sorted](https://en.cppreference.com/w/cpp/algorithm/is_sorted) and [is_sorted_until](https://en.cppreference.com/w/cpp/algorithm/is_sorted_until)


illuhad (approved):
Thanks! :)

PR #1782, opened 2025-04-22
Add static cast to a sycl::range

Aympab:
I noticed a bug when compiling the CTS, these lines don't compile unless I explicitely `static_cast` them.
The ACpp compilation works just fine, the error happens when compiling the CTS.

`/home/user/install/acpp/bin/../include/AdaptiveCpp/sycl/../hipSYCL/sycl/libkernel/accessor.hpp:2056:26: error: no member named 'size' in 'hipsycl::sycl::specialized<hipsycl::sycl::range<3>>'
 2056 |     return _num_elements.size();`

### DIFF
--- a/include/hipSYCL/sycl/libkernel/accessor.hpp
+++ b/include/hipSYCL/sycl/libkernel/accessor.hpp
@@ -2287,7 +2287,7 @@ public:
   ACPP_KERNEL_TARGET
   size_t get_count() const
   {
-    return _num_elements.size();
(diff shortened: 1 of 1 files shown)


illuhad:
Thanks! Can you make the same change in the SYCL 2020 `local_accessor` class a bit further down below your changes? I'm seeing the same code patterns there, so likely it is also affected.

PS: There are some issues with CI on Intel/AMD at the moment. If it hangs or fails, restart the flows.

illuhad (approved):
Thanks!

PR #1783, opened 2025-04-22
[stdpar][algorithms] Add support for `min_element` and `max_element`

normallytangent:
This PR adds automatic offloading support for  [min_element](https://en.cppreference.com/w/cpp/algorithm/min_element) and [max_element](https://en.cppreference.com/w/cpp/algorithm/max_element).


normallytangent:
Here's the [libstdc++](https://github.com/gcc-mirror/gcc/blob/master/libstdc++-v3/include/bits/stl_algo.h#L5633) sequential implementation:
```
template<class ForwardIt, class Compare>
ForwardIt min_element(ForwardIt first, ForwardIt last, Compare comp)
{
    if (first == last)
        return last;
 
    ForwardIt smallest = first;
 
    while (++first != last)
        if (comp(*first, *smallest))
            smallest = first;
 
    return smallest;
}
```

include/hipSYCL/algorithms/algorithm.hpp
```diff
@@ -607,6 +608,190 @@ sycl::event count_if(sycl::queue &q, util::allocation_group &scratch_allocations
...
+        return a;
+      else
+        return b;
+    else
+      if (comp(a.second, b.second) == false)
+        return b;
+      else
+        return a;
+  };
+
```
normallytangent, line 958: I have now simplified the logic for the min- and max-element algorithms.  This way we can continue offloading also in the case of non-trivial classes without Equality. Our implementation now mimics the default behaviour observed on host.

illuhad (approved):

PR #1791, opened 2025-04-28
[Omp] set global_mem_size to physical mem size if available

### LINKED ISSUE #1573: global_mem_size == 2^64-1
**Bug summary**
get_device().get_info<sycl::info::device::global_mem_size>() returns 18446744073709551615 for cpu device.

**To Reproduce**
e.g.
```cpp
spdlog::info("Device: {}", Globals::Queue.get_device().get_info<sycl::info::device::name>());
auto global_mem = Globals::Queue.get_device().get_info<sycl::info::device::global_mem_size>();
spdlog::info("Global memory: {} MB", global_mem / 1024 / 1024);
```

**Expected behavior**
Intel dpc++ returns "Global memory: 128700 MB" as expected for the available host memory for device "Device: AMD Ryzen 9 5950X 16-Core Processor", while in case of SYCL the byte count from above ("Global memory: 17592186044415 MB") for "Device: hipSYCL OpenMP host device". This happens for both the OMP and generic workflow.

When checking with the GPU backend (via generic workflow), everything works as expected with AdaptiveCPP.
```
[2024-09-27 14:12:13.153] [logger_rank0] [info] Device: NVIDIA GeForce RTX 3090
[2024-09-27 14:12:17.371] [logger_rank0] [info] Global memory: 24145 MB
```

**Describe your setup**
Installed from git commit ~ 1 month ago.

tdavidcl:
This PR set the value of global_mem_size in the OpenMP device properties to be the physical memory size.

 Closes #1573

TODOs: 

- [x] Windows case
- [ ] testing ??? @illuhad What approach do you I think I should use to test this in CI if possible ?

### DIFF
--- a/include/hipSYCL/runtime/omp/omp_phys_mem.hpp
+++ b/include/hipSYCL/runtime/omp/omp_phys_mem.hpp
@@ -8,8 +8,8 @@
  * See file LICENSE in the project root for full license details.
  */
 // SPDX-License-Identifier: BSD-2-Clause
-#ifndef HIPSYCL_OMP_PHYS_MEM_HPP
-#define HIPSYCL_OMP_PHYS_MEM_HPP
+#ifndef ACPP_OMP_PHYS_MEM_HPP
+#define ACPP_OMP_PHYS_MEM_HPP
 
 #include <cstddef>
 #include <optional>
@@ -27,7 +27,7 @@ namespace rt {
  * @details This function is implemented for Mac OS X and Linux. Other platforms
  * will return std::nullopt.
  */
-std::optional<std::size_t> getPhysicalMemory();
+std::optional<std::size_t> get_physical_memory();
 } // namespace rt
 } // namespace hipsycl
 
--- a/include/hipSYCL/runtime/omp/omp_phys_mem.hpp
+++ b/include/hipSYCL/runtime/omp/omp_phys_mem.hpp
@@ -1,0 +1,34 @@
+/*
+ * This file is part of AdaptiveCpp, an implementation of SYCL and C++ standard
+ * parallelism for CPUs and GPUs.
+ *
+ * Copyright The AdaptiveCpp Contributors
+ *
+ * AdaptiveCpp is released under the BSD 2-Clause "Simplified" License.
+ * See file LICENSE in the project root for full license details.
+ */
+// SPDX-License-Identifier: BSD-2-Clause
+#ifndef HIPSYCL_OMP_PHYS_MEM_HPP
+#define HIPSYCL_OMP_PHYS_MEM_HPP
+
+#include <cstddef>
+#include <optional>
+
+namespace hipsycl {
+namespace rt {
+
+/**
+ * @brief Get the amount of physical memory (RAM) available on the system, in
+ * bytes.
+ *
+ * @return The amount of physical memory available, or std::nullopt if the
+ * information cannot be retrieved.
+ *
+ * @details This function is implemented for Mac OS X and Linux. Other platforms
+ * will return std::nullopt.
+ */
+std::optional<std::size_t> getPhysicalMemory();
+} // namespace rt
+} // namespace hipsycl
+
+#endif
--- a/src/runtime/omp/omp_hardware_manager.cpp
+++ b/src/runtime/omp/omp_hardware_manager.cpp
@@ -14,6 +14,7 @@
 #include "hipSYCL/runtime/omp/omp_hardware_manager.hpp"
 #include "hipSYCL/runtime/error.hpp"
 #include "hipSYCL/runtime/device_id.hpp"
+#include "hipSYCL/runtime/omp/omp_phys_mem.hpp"
 
 namespace hipsycl {
 namespace rt {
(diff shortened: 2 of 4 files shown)


src/runtime/omp/omp_phys_mem.cpp
```diff
@@ -0,0 +1,58 @@
...
+}
+return value;
+}
+
+// Linux/BSD implementation:
+#elif (defined(linux) || defined(__linux__) || defined(__linux))                                   \
+|| (defined(__DragonFly__) || defined(__FreeBSD__) || defined(__NetBSD__)                      \
+    || defined(__OpenBSD__))
+
+#include <sys/sysinfo.h>
```
illuhad: Maybe the following is a better heuristic?
```c++
#ifdef __has_include
  #if __has_include(<sys/sysinfo.h>)
   ...
  #endif
#endif
```
or are there cases where the header exists, but doesn't provide the functionality?

src/runtime/omp/omp_phys_mem.cpp
```diff
@@ -0,0 +1,58 @@
...
+ */
+// SPDX-License-Identifier: BSD-2-Clause
+#include "hipSYCL/runtime/omp/omp_phys_mem.hpp"
+
+// Mac OSX implementation:
+#if defined(__MACH__)
+#include <sys/sysctl.h>
+#include <sys/types.h>
+
+std::optional<std::size_t> hipsycl::rt::getPhysicalMemory() {
```
illuhad: We generally only use camelCase for LLVM compiler code to conform to LLVM conventions. So maybe better to use snake_case here?

src/runtime/omp/omp_hardware_manager.cpp
```diff
@@ -247,7 +257,7 @@ omp_hardware_context::get_property(device_uint_property prop) const {
     return 1; // TODO
     break;
   case device_uint_property::global_mem_size:
-    return std::numeric_limits<std::size_t>::max(); // TODO
+    return phys_mem_or_max(); // TODO
```
illuhad: I think the TODO is there because of the placeholder value that we used to return, so we can get rid of that comment with this PR :)

include/hipSYCL/runtime/omp/omp_phys_mem.hpp
```diff
@@ -0,0 +1,34 @@
...
+ * @brief Get the amount of physical memory (RAM) available on the system, in
+ * bytes.
+ *
+ * @return The amount of physical memory available, or std::nullopt if the
+ * information cannot be retrieved.
+ *
+ * @details This function is implemented for Mac OS X and Linux. Other platforms
+ * will return std::nullopt.
+ */
+std::optional<std::size_t> getPhysicalMemory();
```
illuhad: Is there a benefit to this being a free function? Would it perhaps make more sense to have this be a private member function of `omp_hardware_context`, or even just defined in an anonymous namespace in `omp_hardware_manager.cpp`?

I suppose the question boils down to whether there is some need to know the physical memory size outside of that class. Hm....

include/hipSYCL/runtime/omp/omp_phys_mem.hpp
```diff
@@ -0,0 +1,34 @@
...
+ * This file is part of AdaptiveCpp, an implementation of SYCL and C++ standard
+ * parallelism for CPUs and GPUs.
+ *
+ * Copyright The AdaptiveCpp Contributors
+ *
+ * AdaptiveCpp is released under the BSD 2-Clause "Simplified" License.
+ * See file LICENSE in the project root for full license details.
+ */
+// SPDX-License-Identifier: BSD-2-Clause
+#ifndef HIPSYCL_OMP_PHYS_MEM_HPP
```
illuhad:
```suggestion
#ifndef ACPP_OMP_PHYS_MEM_HPP
```

include/hipSYCL/runtime/omp/omp_phys_mem.hpp
```diff
@@ -0,0 +1,34 @@
...
+#define HIPSYCL_OMP_PHYS_MEM_HPP
```
illuhad:
```suggestion
#define ACPP_OMP_PHYS_MEM_HPP
```

src/runtime/omp/omp_phys_mem.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+int mib[]     = {CTL_HW, HW_MEMSIZE};
+int64_t value = 0;
+size_t length = sizeof(value);
+
+if (-1 == sysctl(mib, 2, &value, &length, NULL, 0)) {
+    return std::nullopt;
+}
+return value;
```
illuhad: Can you please fix indentation for this function?

illuhad (commented):
Thanks! I don't see how we could really test this in CI... Perhaps check that it returns a value greater than 0 and smaller than some unrealistic large number, like 100 Petabyte or so...

include/hipSYCL/runtime/omp/omp_phys_mem.hpp
tdavidcl: yeah since there is some #ifdef like switches I wanted to hide it in a separate file to have a clean interface.

src/runtime/omp/omp_phys_mem.cpp
tdavidcl: I move the linux case in the else and guard the return by `#if __has_include(<sys/sysinfo.h>)`

illuhad (approved):
Thanks!

PR #1794, opened 2025-04-29
Bump minimum supported LLVM version to 15

illuhad:
Currently, the minimum supported LLVM version is 14. We already only have partial CI testing for LLVM 14, and most tests are only run for newer LLVM.

This PR therefore bumps the minimum version to 15.


fodinabor (approved):
Alright, LGTM!

PR #1798, opened 2025-05-02
Don't offload task submission to worker thread

illuhad:
Currently we're using an asynchronous worker thread to handle submission of non-instant tasks.

This can make task submission be sensitive to NUMA effects and thread-pinning. I've observed a case where `OMP_PROC_BIND=true/false` has ~40% perf impact for a kernel running on CUDA (not CPU!).

This PR is an experiment to remove the asynchronous submission worker thread, bringing instant and non-instant task submission very close together in their semantics. This can allow further simplification in the future.

Obviously this kind of change requires careful testing which I haven't done yet, so draft PR.

@al42and @PeterTh - I imagine you might like this. If you have feedback let me know :)


illuhad:
* fixes 40% performance regression for HeCBench mandelbrot on NVIDIA GPU with `OMP_PROC_BIND=true` if AdaptiveCpp was built with clang.
* EasyWave: +8% performance
* CloverLeaf: No change in performance

This PR now also enables instant submission semantics for all tasks. By this I mean: All tasks are submitted greedily. However, note that for pure USM kernels as before with instant submission, additional scheduling layers that are only relevant for buffer/accessor are bypassed, so there's still a scheduling performance advantage for USM over buffers. As such, the distinction between cases where previously instant submission was possible, and cases where it used to be impossible still is measurable.

The following macros are now without effect
- `ACPP_ALLOW_INSTANT_SUBMISSION`
- `ACPP_FORCE_INSTANT_SUBMISSION`

The following environment variables have been removed, and setting them has no effect:
- `ACPP_RT_MAX_CACHED_NODES`

illuhad:
There is probably more potential for optimization of the submission process, but I believe this change is good enough now to experiment with by more users.

Due to interaction with OpenMP runtime `OMP_PROC_BIND` and the old submission thread, there can be an impact even for GPU backends running USM kernels as in the mandelbrot case that I've mentioned.

PR #1803, opened 2025-05-05
[SSCP] Add experimental device support for assert()

### LINKED ISSUE #1795: Strip libstdc++ asserts from device code?
**Bug summary**
When running SYCL kernels that have a std::array parameter (valid as per https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec::device.copyable , in my understanding) from a debug build, Libstdc++ assertions are active in kernel code (https://gcc.gnu.org/wiki/LibstdcxxDebugMode), but cannot be compiled. Because of this, the run fails (at least for the HIP backend).

I would therefor suggest always setting the preprocessor directive `_GLIBCXX_NO_ASSERTIONS` for all non-CPU offloads.

**To Reproduce**
Compile with acpp on a system with Libstdc++. A non-optimized build will fail to run, an optimized build (`-O3`) or a non-optimized build with `-D_GLIBCXX_NO_ASSERTIONS` will work as Libstdc++ disables assertions in this case.
```cpp
#include <sycl/sycl.hpp>

#include <array>

int main() {
    sycl::queue q{sycl::gpu_selector_v};

    std::array<uint32_t, 1> data;

    q.parallel_for(1, [data](const sycl::id<1>& id) {
        (void) data[id];
    });
}
```

**Expected behavior**
A successful run as with `-O3`.

**Actual behavior**
```
[trimmed for brevity]

amd_comgr_do_action:
          ActionKind: AMD_COMGR_ACTION_LINK_RELOCATABLE_TO_EXECUTABLE
             IsaName: amdgcn-amd-amdhsa--gfx1103
             Options:
                Path: 
            Language: AMD_COMGR_LANGUAGE_NONE
 Comgr Branch-Commit: not-available-not-available
         LLVM Commit: 
    Compilation Args:  "-target" "amdgcn-amd-amdhsa" "-mcpu=gfx1103" "/tmp/comgr-214abc/input/LLVMBitcode.bc.o" "-o" "/tmp/comgr-214abc/output/a.so"
     Driver Job Args: lld "--no-undefined" "-shared" "/tmp/comgr-214abc/input/LLVMBitcode.bc.o" "-plugin-opt=mcpu=gfx1103" "-o" "/tmp/comgr-214abc/output/a.so"
ld.lld: error: undefined symbol: std::__glibcxx_assert_fail(char const*, int, char const*, char const*)
>>> referenced by /tmp/comgr-214abc/input/LLVMBitcode.bc.o:(void __acpp_sscp_kernel<hipsycl::glue::__sscp_dispatch::basic_parallel_for<main::$_0, 1>>(hipsycl::glue::__sscp_dispatch::basic_parallel_for<main::$_0, 1> const&))
>>> referenced by /tmp/comgr-214abc/input/LLVMBitcode.bc.o:(void __acpp_sscp_kernel<hipsycl::glue::__sscp_dispatch::basic_parallel_for<main::$_0, 1>>(hipsycl::glue::__sscp_dispatch::basic_parallel_for<main::$_0, 1> const&))
        ReturnStatus: AMD_COMGR_STATUS_ERROR

```

**Describe your setup**
Acpp version: develop/593c2f42fd507da99454a3bc87096bece05bd3f3
ROCm 6.3.1
LLVM 18
Libstdc++ 15.1.1
Fedora 42

illuhad:
Add device `assert()` support for the generic SSCP compiler.

* Supports both normal `__assert_fail` from `assert()` as well as `std::__glibcxx_assert_fail` as used inside libstdc++
* Behavior of assert failure is backend specific:
   - On `omp`, the error is printed and the application aborted
   - On `cuda`, the error is printed and the kernel aborted
   - On `hip`, the error is printed, and then the kernel enters UB (is there a way to abort kernels on AMD?)
   - On `ocl/ze`, there seems to be an issue with the SPIR-V `printf` builtin on the Intel OpenCL CPU runtime which we use in CI. So we cannot even print :( The kernel just enters UB if the assertion failure is triggered.

Fixes #1795

### DIFF
--- a/include/hipSYCL/compiler/sscp/DeviceAssertPass.hpp
+++ b/include/hipSYCL/compiler/sscp/DeviceAssertPass.hpp
@@ -1,0 +1,31 @@
+/*
+ * This file is part of AdaptiveCpp, an implementation of SYCL and C++ standard
+ * parallelism for CPUs and GPUs.
+ *
+ * Copyright The AdaptiveCpp Contributors
+ *
+ * AdaptiveCpp is released under the BSD 2-Clause "Simplified" License.
+ * See file LICENSE in the project root for full license details.
+ */
+// SPDX-License-Identifier: BSD-2-Clause
+#ifndef ACPP_DEVICE_ASSERT_PASS_HPP
+#define ACPP_DEVICE_ASSERT_PASS_HPP
+
+#include <llvm/IR/PassManager.h>
+
+namespace hipsycl {
+namespace compiler {
+
+class DeviceAssertPass : public llvm::PassInfoMixin<DeviceAssertPass> {
+public:
+  llvm::PreservedAnalyses run(llvm::Module &M,
+                              llvm::ModuleAnalysisManager &MAM);
+
+};
+
+}
+}
+
+
+#endif
+
--- a/include/hipSYCL/sycl/libkernel/sscp/builtins/assert.hpp
+++ b/include/hipSYCL/sycl/libkernel/sscp/builtins/assert.hpp
@@ -1,0 +1,25 @@
+/*
+ * This file is part of AdaptiveCpp, an implementation of SYCL and C++ standard
+ * parallelism for CPUs and GPUs.
+ *
+ * Copyright The AdaptiveCpp Contributors
+ *
+ * AdaptiveCpp is released under the BSD 2-Clause "Simplified" License.
+ * See file LICENSE in the project root for full license details.
+ */
+// SPDX-License-Identifier: BSD-2-Clause
+#ifndef HIPSYCL_SSCP_BUILTINS_ASSERT_HPP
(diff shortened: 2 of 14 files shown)


Stastez:
Nice! According to AMD's docs, there should be both an `assert` and `abort` function in HIP, but they terminate the application completely. Is this the desired behaviour? 

https://rocm.docs.amd.com/projects/HIP/en/docs-develop/how-to/kernel_language_cpp_support.html#assertions

illuhad:
@Stastez This would be acceptable behavior. We should check whether `__builtin_trap()` can be used in our bitcode libraries - although in practice, I suspect that it wouldn't change much. At the moment we hit `__builtin_unreachable()` after an assertion failure on AMD, which is likely already lowered to a trap by the LLVM amdgcn backend.

PR #1816, opened 2025-05-12
[SubCfgFormation] Handle nullptrs explicitly to prevent assertion

devajithvs:
Fixes assertion when replacing values of pointer type with integer zero. Now uses `llvm::Constant::getNullValue()` to correctly handle pointer-typed IDs.


fodinabor (approved):
Lgtm, thanks!

PR #1824, opened 2025-05-14
[stdpar][algorithms] Add support for `mismatch`

normallytangent:
This PR adds automatic offloading support for all overloads of [mismatch](https://en.cppreference.com/w/cpp/algorithm/mismatch)


normallytangent:
@illuhad I answered too soon earlier.. This too fails on Level Zero

illuhad (approved):
Thanks :)

Issue #1819, opened 2025-05-12
RFC: Deployment model for applications built with generic JIT compiler


illuhad:
I think it's time that we start defining the deployment model for applications built with the generic JIT compiler.

This is about the use case where an AdaptiveCpp user would like to distribute binaries generated by AdaptiveCpp to end users, where the binary is then supposed to "just run" on whatever hardware the end user has available.

The generic JIT compiler already helps us in that direction by decoupling the device binary format from the actual target thanks to its unified JIT infrastructure.

However, there are still some issues around the *dependencies*, i.e. it's not very clear which parts of LLVM, ROCm or CUDA would need to be distributed along with the AdaptiveCpp runtime and the application binary. Also, some parts of the JIT infrastructure use hardcoded paths e.g. to LLVM tooling.

I have an idea how I would like to approach this, and would appreciate feedback whether people feel that this is the right direction.

My goal is to allow users to just grab AdaptiveCpp's `lib` directory with everything inside, and have everything they need to ship with their application. Required LLVM components, backend bitcode libraries etc
(For simplicity, we could additionally add an `acpp-deploy` tool which copies the needed content if desired).

The key to achieve this would be to make AdaptiveCpp work exclusively in terms of relative paths w.r.t the location of its runtime library. This means no hardcoded CUDA, ROCm, LLVM paths etc.

I propose that, when `make install` is executed, our cmake infrastructure also copies needed bits from LLVM, CUDA and ROCm to `$INSTALL_PREFIX/lib`. Then the AdaptiveCpp installation will be self-contained and have no direct dependencies anymore to those large external software stacks.
* One alternative here would be to instead let the runtime library discover those dependencies at runtime. This would lead to a smaller AdaptiveCpp redistributable package, however this is more brittle since the found ROCm might be the wrong version etc.

To make this work, we should do the following:
* All backends that currently use `clang` during JIT compilation should instead use `llc` and `lld` for reduced dependency surface. This should allow our LLVM dependencies to be limited to `llc`, `lld` and `libLLVM.so`.
* All backends should use relative paths for any JIT tooling that they invoke, and the needed JIT tooling should be part of the AdaptiveCpp directory structure.
* ROCm bitcode libraries and CUDA's `libdevice.bc` will have to be copied into the AdaptiveCpp bitcode directory. Backends can then discover them there alongside the AdaptiveCpp bitcode libraries.
[...]

PR #1826, opened 2025-05-15
Look for CUDA libdevice harder

Oblomov:
This is a two-commit patchset that helps AdaptiveCpp find CUDA‌ libdevice files even when they are installed in /usr. The first is a maintenance commit that introduces a variadic version of join_path, which is used in the second to add lib/cuda before nvvm/libdevice when nvvm/libdevice isn't found in HIPSYCL_CUDA_PATH.

### DIFF
--- a/src/compiler/llvm-to-backend/ptx/LLVMToPtx.cpp
+++ b/src/compiler/llvm-to-backend/ptx/LLVMToPtx.cpp
@@ -69,6 +69,17 @@ private:
     std::string CUDAPath = HIPSYCL_CUDA_PATH;
     std::vector<std::string> SubDir {"nvvm", "libdevice"};
     std::string BitcodeDir = common::filesystem::join_path(CUDAPath, SubDir);
+    // When CUDA is installed via system packages in /usr, the nvvm/libdevice structure
+    // will be under /usr/lib/cuda. Generalize this by checking first if the currently assumed
+    // BitcodeDir exists, and if not, trying under CUDAPath/lib/cuda/SubDir
+    if(!common::filesystem::exists(BitcodeDir)) {
+      std::vector<std::string> libCuda { "lib", "cuda" };
+      BitcodeDir = common::filesystem::join_path(CUDAPath, libCuda, SubDir);
+      // if this doesn't exist either, just give up
+      if(!common::filesystem::exists(BitcodeDir)) {
+        return false;
+      }
+    }
 
     std::error_code EC;
     auto Files = common::filesystem::list_regular_files(BitcodeDir, EC);
--- a/include/hipSYCL/common/filesystem.hpp
+++ b/include/hipSYCL/common/filesystem.hpp
@@ -27,6 +27,7 @@ namespace filesystem {
 ACPP_COMMON_EXPORT std::string get_install_directory();
 
 ACPP_COMMON_EXPORT std::string join_path(const std::string& base, const std::string& extra);
+ACPP_COMMON_EXPORT std::string join_path(const std::string& base, const char* extra);
 
 ACPP_COMMON_EXPORT bool exists(const std::string& path);
 
@@ -35,6 +36,14 @@ ACPP_COMMON_EXPORT std::string absolute(const std::string& path);
 ACPP_COMMON_EXPORT std::string
 join_path(const std::string &base,
           const std::vector<std::string> &additional_components);
+
+/// Variadic form of join_path, concatenates any number of strings or vectors of strings
+template<typename T, typename... Ts>
+std::string
+join_path(const std::string& base, const T& next_component,
+  const Ts& ...other_components) {
+  return join_path(join_path(base, next_component), other_components...);
+}
 
 ACPP_COMMON_EXPORT std::vector<std::string> list_regular_files(const std::string &directory,
                                             std::error_code &EC);
--- a/src/common/filesystem.cpp
+++ b/src/common/filesystem.cpp
@@ -86,6 +86,10 @@ std::string join_path(const std::string& base, const std::string& extra) {
   return (fs::path(base) / extra).string();
 }
 
+std::string join_path(const std::string& base, const char* extra) {
+  return join_path(base, std::string(extra));
+}
+
 std::string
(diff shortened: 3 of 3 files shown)


illuhad:
Hmm... I wonder whether it might be better to just discover the bitcode libraries in cmake. This might also integrate better with what we are discussing in #1819. We could still merge this as this is still an improvement over the old way.

Oblomov:
I personally have a preference for runtime dependency discovery so I'd prefer the current mechanism (with the improvements herein), but thanks for pointing me to the other issue, which is definitely relevant. I've added my two cents to that discussion as well.

illuhad:
mhmm... I'm a bit concerned about pure runtime dependency discovery.
- what if we pick up the wrong components? In general, picking the wrong bitcode libraries (say, from a different CUDA or ROCm version that also happens to be installed on the system) can break the compiler.
   - Runtime detection as in your case can potentially *silently* pick the wrong components, and there might not be a way for us to even figure out which component is the right one if there are several choices.
- How can the user even force AdaptiveCpp JIT to use a specific component if it *does* make the wrong choice? (assuming that the user would realize that there is a problem in the first place and be able to trace it back to bitcode library selection)

All of these issues would be solved by discoving bitcode libraries at cmake time.
- cmake could by default just look in well known subdirs of CUDA/ROCm. If it can't find the needed files there, it asks the user to manually provide the path
- User can set the right path using a cmake variable if desired/needed for specific setups.
- No backend-specific logic needed to find components at JIT-time. All backends can follow the same patterns:
   - first, check some redistributable path relative to the runtime library location
   - otherwise check the predefined path that was obtained from cmake. No other cases need to be handled.

src/compiler/llvm-to-backend/ptx/LLVMToPtx.cpp
```diff
@@ -69,6 +69,17 @@ class LibdevicePath {
     std::string CUDAPath = HIPSYCL_CUDA_PATH;
     std::vector<std::string> SubDir {"nvvm", "libdevice"};
     std::string BitcodeDir = common::filesystem::join_path(CUDAPath, SubDir);
+    // When CUDA is installed via system packages in /usr, the nvvm/libdevice structure
+    // will be under /usr/lib/cuda. Generalize this by checking first if the currently assumed
+    // BitcodeDir exists, and if not, trying under CUDAPath/lib/cuda/SubDir
+    if(!common::filesystem::exists(BitcodeDir)) {
+      std::vector<std::string> libCuda { "lib", "cuda" };
```
illuhad, line 76: Will it always be `lib/cuda`, or could it also be `lib/cuda-$VERSION`? For example, on Ubuntu I seem to be having a `/usr/local/lib/cuda-11.0` path. We really don't want to pick the wrong bitcode library if there are multiple CUDA versions available.

Sidenote: Even clang does not fully support such CUDA installations: https://llvm.org/docs/CompileCudaWithLLVM.html#id3
So perhaps we should not support them as a priority either?

Oblomov: Are you sure it's under `/usr/local/lib/cuda-$VERSION` and not directly under `/usr/local/cuda-$VERSION`, which would be caught by the existing code before this callback? AFAIK, no distro package installs under `/usr/local/`, but the third-party NVIDIA‌ packages from NVIDIA do:

```
~$ dpkg -S $(find /usr/local/ -name libdevice.10.bc)
cuda-nvvm-12-8: /usr/local/cuda-12.8/nvvm/libdevice/libdevice.10.bc
cuda-nvvm-12-5: /usr/local/cuda-12.5/nvvm/libdevice/libdevice.10.bc
```

The fallback this patch introduces is useful to manage distributions that have their own packages for the NVIDIA‌ stuff (e.g. in  Debian, `libdevice.10.bc` is installed by the `nvidia-cuda-toolkit` package which is not versioned.)

On the test system where I tried, for example, I have CUDA 12.2 from distro packages and CUDA 11.4 installed via NVIDIA tarballs under `/usr/local/cuda-11.4`. If I build acpp without specifying a CUDA‌ path, the “system” CUDA is found, but `libdevice.10.bc` isn't found at runtime (this patch is needed). If I build acpp specifying the `/usr/local/cuda-11.4` path, then `libdevice.10.bc` from `/usr/local/cuda-11.4/nvvm/libdevice` is used, with or without this patch.

illuhad: Ah... could be that that installation was from the official CUDA installer then. I've no idea where it came from and don't usually use that installation, I just noticed it when I looked at what I have in my filesystem. You're right, it's actually in `/usr/local/cuda-$VERSION`. My bad, sorry!

I agree that the existing code should handle this particular installation well.

But now I'm curious: As I said clang itself also has trouble with some CUDA distribution packages. Some distributions apparently have solved this by adding some shim to emulate the normal CUDA directory layout. See e.g. https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=882505
Similarly for Ubuntu: https://bugs.launchpad.net/ubuntu/+source/nvidia-cuda-toolkit/+bug/1706326

So why does it not work for us when building AdaptiveCpp and setting the CUDA path to `/usr/lib/cuda`? Is the shim incomplete for libdevice.bc (in this case we should perhaps report this as a bug to debian/ubuntu?)?

Oblomov: AFAICS in my system there's nothing of the sort and the “NVIDIA layout” isn't reproduced in my system anywhere except the `/usr/local/` installation. If I remove it, `clang` (v19) will just use the system installation without issues.  Maybe the shim was only needed on very old versions of llvm/clang and support for these installations has been fixed there too?

Oblomov:
As explained in the comments of #1839 the situation that required this patch was very peculiar (configuring under the current `master` branch and then reconfiguring and building under `develop` without clearing the CMakeCache), so I believe this patch can be considered obsolete/unnecessary. The changes being discussed in #1839 will probably be sufficient even to cover this unusual flow.

PR #1830, opened 2025-05-20
[SSCP][llvm-to-ptx] Use llc instead of clang during JIT

illuhad:
Use llc instead of clang for code generation at JIT-time to reduce dependency surface as discussed in #1819.

~I'm currently still observing 10% regression with this, so presumably, we need to tweak the llc flags a little more to achieve parity with what clang does. Draft until then.~


illuhad:
Regression was fixed by also running `opt -O3` before `llc -O3`. This should be more similar to invoking `clang -O3` as we did previously.
I also tried running a second O3 pipeline inside our own code without invoking `opt`, but curiously, that only solved around half of the regression. This is something worth investigating in the future.

As it stands, it looks like we'll also need `opt` as an AdaptiveCpp dependency (which is fine - these tools are all smaller than 1MB on my system).

Nevertheless, we should make more performance regression tests with this PR.

PR #1831, opened 2025-05-21
[stdpar][algorithms] Add support for `unique_copy` and `unique`

normallytangent:
This PR adds automatic offloading support for [unique_copy](https://en.cppreference.com/w/cpp/algorithm/unique_copy) and [unique](https://en.cppreference.com/w/cpp/algorithm/unique).

-  Does not use `copy_if` interface, although unique internally implements the same logic.  This is because the lambda 'pred' needs to operate upon the iterator. Since in `copy_if` the predicate is called with a value, the interface is not readily usable for `unique_copy` and `unique` algorithms.
- Overloads with the binary predicate are implemented and tested.


normallytangent:
@illuhad updated and ready 👍

include/hipSYCL/algorithms/algorithm.hpp
```diff
@@ -1368,6 +1368,98 @@ sycl::event merge(sycl::queue& q,
...
+                    ForwardIt1 first, ForwardIt1 last, ForwardIt2 d_first,
+                    BinaryPredicate p = {},
+                    std::size_t *num_elements_copied = nullptr,
+                    const std::vector<sycl::event> &deps = {}) {
+  if (first == last)
+    return sycl::event{};
+
+  auto pred = [first, p](auto x) {
+                auto input = x;
+                --input;
```
illuhad, line 1382: Is this well-defined if `input` is the first element?

normallytangent: I think so, the if-condition on the next line checks for that case.

illuhad: Yes, but that's my question: Is it always well-defined to decrement any iterator below begin? Maybe the iterator behaves like an unsigned int and then encounters some underflow behavior or so which may be UB.

PR #1835, opened 2025-05-22
[SSCP][llvm-to-amdgpu] Collect bitcode libraries ourselves and eliminate JIT-time dependency on hipcc

illuhad:
Currently we invoke `hipcc` and parse the output at JIT-time to figure out which bitcode libraries we need to link.

This has the disadvantage that it pulls in hipcc as a runtime dependency. This PR therefore adds support to determine the needed libraries ourselves, thus avoiding the hipcc dependency.

Bitcode libraries are first looked for in a directory relative to the AdaptiveCpp install root, and if that fails, in the ROCm bitcode directory that was discovered at cmake time.

@al42and - I think you have experience with AMD GPUs that support wavefronts of size 32, right? Do you know what users currently get as default when compiling with AdaptiveCpp for such targets? Currently this PR uses 64 as default for all GPUs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -233,19 +233,7 @@ if(WITH_CUDA_BACKEND)
     message(SEND_ERROR "CUDA was not found")
   endif()
 endif()
-if(WITH_ROCM_BACKEND)
-  if(NOT ROCM_FOUND)
-    #  message(SEND_ERROR "hipcc was not found")
 
-    # User has requested ROCm, but we could not find hipcc.
-    # this is not necessarily a reason to abort,
-    # since we only need libhip_hcc, the HIP includes,
-    # and the ROCm device headers. It could be that we
-    # are faced with a minimal/incomplete ROCm installation
-    # that could still be enough for us.
-    # Let's assume the user knows what he/she is doing.
-  endif()
-endif()
 
 if(WITH_OPENCL_BACKEND)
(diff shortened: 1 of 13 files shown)


al42and:
> @al42and - I think you have experience with AMD GPUs that support wavefronts of size 32, right? Do you know what users currently get as default when compiling with AdaptiveCpp for such targets? Currently this PR uses 64 as default for all GPUs.

All Wave32 devices (i.e, all RDNA GPUs so far) should support Wave64 as well. They default to Wave32 when using native HIP, ACpp in SMCP mode, or DPC++. Wave64 is, [by some accounts](https://github.com/ROCm/ROCm/issues/4121#issuecomment-2741913224), "not intended for production use."

We use Wave64 mode via native OpenCL on RDNA2 GPUs in our CI, and there's been no problems with it (with Wave64, that is; we had some unrelated fun with other aspects of this setup). I haven't used RDNA+Wave64 extensively via SYCL past "will it compile at all".

illuhad:
I see, thanks!
Then I will have to change this PR to query the warp size in the runtime, and then take it into account during JIT. Currently we follow whatever `hipcc` is doing, but the PR in its current form changes that.

According to this: https://github.com/ROCm/hip/issues/3198#issuecomment-1492090330
RDNA will not support some important builtins in wavefront64 mode, which would probably break our group algorithm library.

PR #1839, opened 2025-05-23
[SSCP][llvm-to-ptx] Look for libdevice.bc also in relative directory

illuhad:
In analogy to what we now do for AMD in #1835, this changes how we look for libdevice on CUDA:

1. First, find `libdevice` at cmake time in the CUDA SDK
   1.1. For CUDA installations in `/usr`, also take into account unusual libdevice locations as proposed in #1826 
2. At JIT-time, first try to find libdevice in a relative location if we have a redistributable package
3. Otherwise, look at the location we have from step 1.

This aligns with what we do on AMD.

@Oblomov, are you happy with that change? Can you check whether it works for your use case?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -474,6 +474,27 @@ if(WITH_ROCM_BACKEND)
   endif()
 endif()
 
+
+if(WITH_CUDA_BACKEND)
+  find_path(CUDA_DEVICE_LIBS_PATH libdevice.10.bc HINTS
+      ${CUDA_TOOLKIT_ROOT_DIR}/nvvm/libdevice
+      DOC "Path containing the CUDA bitcode libraries. Typically, $CUDA_PATH/nvvm/libdevice")
+  
+  if(NOT CUDA_DEVICE_LIBS_PATH AND CUDA_TOOLKIT_ROOT_DIR STREQUAL "/usr")
+    find_path(CUDA_DEVICE_LIBS_PATH libdevice.10.bc HINTS
+      /usr/lib/cuda/nvvm/libdevice)
+  endif()
+
+  if(CUDA_DEVICE_LIBS_PATH)
+    message(STATUS "Found CUDA bitcode library path: ${CUDA_DEVICE_LIBS_PATH}")
+  else()
+    message(SEND_ERROR "CUDA device library path not found; please provide the path containing the CUDA bitcode libraries (libdevice.10.bc) manually using -DCUDA_DEVICE_LIBS_PATH")
+  endif()
+endif()
+
+
+
+
 #add_compile_definitions(HIPSYCL_DEBUG_LEVEL="${HIPSYCL_DEBUG_LEVEL}")
 #Use add_definitions for now for older cmake versions
 cmake_policy(SET CMP0005 NEW)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -479,11 +479,6 @@ if(WITH_CUDA_BACKEND)
   find_path(CUDA_DEVICE_LIBS_PATH libdevice.10.bc HINTS
       ${CUDA_TOOLKIT_ROOT_DIR}/nvvm/libdevice
       DOC "Path containing the CUDA bitcode libraries. Typically, $CUDA_PATH/nvvm/libdevice")
-  
-  if(NOT CUDA_DEVICE_LIBS_PATH AND CUDA_TOOLKIT_ROOT_DIR STREQUAL "/usr")
-    find_path(CUDA_DEVICE_LIBS_PATH libdevice.10.bc HINTS
-      /usr/lib/cuda/nvvm/libdevice)
-  endif()
 
   if(CUDA_DEVICE_LIBS_PATH)
     message(STATUS "Found CUDA bitcode library path: ${CUDA_DEVICE_LIBS_PATH}")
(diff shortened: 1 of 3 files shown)


Oblomov:
~~I've tested it on my system and it seems to build and run correctly using my system CUDA installation.~~

EDIT: there was something leftover from my previous tests, and cannot even build correctly now, let me double-check a few things.

Oblomov:
OK, I tried a clean build specifying only `-DWITH_CUDA_BACKEND=ON` (no hint for the CUDA‌ path) and I get the following warning that later triggers an error because I'm forcing the CUDA‌ backend to ON:

```
CMake Warning at cmake/FindCUDA.cmake:46 (message):
  CUDAToolkit_LIBRARY_ROOT /usr/lib does not point to the correct directory,
  try setting it manually.  Detected CUDA installation cannot be used.
  CMakeLists.txt:129 (find_package) 
```

So in my configuration I still have to set `CUDAToolkit_LIBRARY_ROOT` manually to `/usr/lib/cuda`. With this manually set, it works. To make this fully transparent, we might want to check for this case in `cmake/FindCUDA.cmake` too.

illuhad:
@Oblomov Wait, I thought that for your configuration the CUDA root dir should be effectively `/usr` (not `/usr/lib/cuda`).
This PR currently special cases `libdevice` detection if the CUDA path is `/usr`. Do we need to change this logic?
I don't quite understand how it worked in your case without activating the special code path for libdevice detection.

Oblomov:
@illuhad I'm honestly confused myself. I even tried rolling back to the commit I based my patchset on and can't reproduce the fact that CUDA was found by CMake but libdevice wasn't, which was quite definitely a thing at some point in the past. I suspect some other part of the build system was upgraded in the mean time (possibly cmake) but the old config wasn't cleaned up, so some older detection logic was involved.

I'm not sure what to make of this.

Oblomov:
OK after some more experimentation, I have managed to reproduce.

The issue stems from the fact that when I first checked out `AdaptiveCpp`, I was on the `master` branch, that uses the old cmake `find_package(CUDA)` to find CUDA, and this actually succeeds on my system. After this, I switched to the `develop` branch, and didn't wipe out my build directory, so it was still happy with the system CUDA version that was found by the `master` branch build attempt, even if it couldn't find the libdevice (hence why I made the PR to look into the subdirectories).

So I have now done some additional testing with fully clean builds off the current tip of the `develop` branch, with neither this nor my PR. What happens is the following:

* a simple `cmake ..` finds my system CUDA installation but complains that it's not usable:

```
CMake Warning at cmake/FindCUDA.cmake:46 (message):
  CUDAToolkit_LIBRARY_ROOT /usr/lib does not point to the correct directory,
  try setting it manually.  Detected CUDA installation cannot be used.
  CMakeLists.txt:129 (find_package)
```

* manually specifying `cmake .. -DCUDAToolkit_LIBRARY_ROOT=/usr/lib/cuda/` results in CUDA being properly detected:

```
-- Found CUDA version 12.2.140 in /usr/lib/cuda/
```

and from the quick testing I've made, it works correctly without additional changes to the search path.

So the reason why this PR isn't sufficient presently is that the check for `nvvm` still fails. The second half of the patch is fine (and will work even on build system inherited from `master`), but if we want a setup like mine to be detected without user intervention, `cmake/FindCUDA.cmake` also needs to be adjusted.

I've tested this patch and it seems to do the job. Note that it's a bit more «aggressive» since rather than checking for the CUDA root path to be `/usr/lib` it just extends the search to the `cuda/` subdir of the root path in both `CMakeLists.txt` and `cmake/FindCUDA.cmake`.

```patch
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 5b8295ee..0810b03d 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -478,12 +478,8 @@ endif()
 if(WITH_CUDA_BACKEND)
   find_path(CUDA_DEVICE_LIBS_PATH libdevice.10.bc HINTS
       ${CUDA_TOOLKIT_ROOT_DIR}/nvvm/libdevice
+      ${CUDA_TOOLKIT_ROOT_DIR}/cuda/nvvm/libdevice
       DOC "Path containing the CUDA bitcode libraries. Typically, $CUDA_PATH/nvvm/libdevice")
-  
-  if(NOT CUDA_DEVICE_LIBS_PATH AND CUDA_TOOLKIT_ROOT_DIR STREQUAL "/usr")
-    find_path(CUDA_DEVICE_LIBS_PATH libdevice.10.bc HINTS
-      /usr/lib/cuda/nvvm/libdevice)
-  endif()
 
   if(CUDA_DEVICE_LIBS_PATH)
     message(STATUS "Found CUDA bitcode library path: ${CUDA_DEVICE_LIBS_PATH}")
diff --git a/cmake/FindCUDA.cmake b/cmake/FindCUDA.cmake
index f81999a1..02f941dc 100644
--- a/cmake/FindCUDA.cmake
+++ b/cmake/FindCUDA.cmake
@@ -42,7 +42,8 @@ else()
             get_filename_component(CUDAToolkit_LIBRARY_ROOT ${CUDAToolkit_LIBRARY_ROOT} DIRECTORY)
          endif()
 
-         if (NOT EXISTS "${CUDAToolkit_LIBRARY_ROOT}/nvvm")
+         # If CUDA is a system installation in /usr/lib, nvvm is under /usr/lib/cuda
+         if (NOT EXISTS "${CUDAToolkit_LIBRARY_ROOT}/nvvm" AND NOT EXISTS "${CUDAToolkit_LIBRARY_ROOT}/cuda/nvvm")
             message(WARNING "CUDAToolkit_LIBRARY_ROOT ${CUDAToolkit_LIBRARY_ROOT} does not point to the correct directory, try setting it manually. Detected CUDA installation cannot be used.")
             set(CUDAToolkit_FOUND FALSE)
          endif()
```

illuhad:
@Oblomov, okay, I think it might be better then to decouple this from this PR. I'll remove the special casing that I've added (which seems to be insufficient), and then we can properly solve your use case in a dedicated PR.

PR #1842, opened 2025-05-27
Using C++ modules with AdaptiveCpp

### LINKED ISSUE #1722: Support using SYCL in C++ module
C++ modules are gaining more and more support from compilers and CMake. Currently trying to use AdaptiveCpp in C++ module leads to a link error with LLVM 19 and AdaptiveCpp 24.10.0+git.3ac3c8fb.20250210.branch.develop (see example project below):

```
undefined reference to `__acpp_reflection_init_registered_function_pointers'
```

It would be nice to be able to modularize codebase that uses AdaptiveCpp.

**Example project**

main.cpp:

```cpp
import examplemodule;

int main() {
    ls_platforms();
}

```

examplemodule.cpp:

```cpp
module;

#include <print>

#include "sycl/sycl.hpp"

export module examplemodule;

export void
ls_platforms() {
    // Loop through available platforms
    for (auto const& platform : sycl::platform::get_platforms()) {
        std::println("Found platform: {}", platform.get_info<sycl::info::platform::name>());
        // Loop through available devices in this platform
        for (auto const& device : platform.get_devices()) {
            std::println("With device: ", device.get_info<sycl::info::device::name>());
        }
        std::println("");
    }
}

```

CMakeLists.txt:

```
cmake_minimum_required(VERSION 3.30 FATAL_ERROR)

project(
    sycl-adaptivecpp-example
    VERSION 0.0.1
    DESCRIPTION ""
    HOMEPAGE_URL "test@foo.com"
    LANGUAGES CXX
)

find_package(AdaptiveCpp REQUIRED)

add_library(examplemodule STATIC)
target_sources(examplemodule
    PUBLIC FILE_SET CXX_MODULES FILES 
    examplemodule.cpp
)

add_sycl_to_target(TARGET examplemodule)

# Tell CMake we're using C++26 and that C++26 is needed to consume it.
target_compile_features(examplemodule
    PRIVATE
    cxx_std_26
    INTERFACE
    cxx_std_26
)

# Make an executable.
add_executable(main)

target_sources(main
  PRIVATE
    main.cpp
)

target_link_libraries(main PRIVATE examplemodule)
```

Build with: `cmake -GNinja -Bbuild -DCMAKE_CXX_COMPILER=clang++ && ninja -C build`

EkremA:
As brought up in [Issue#1722](https://github.com/AdaptiveCpp/AdaptiveCpp/issues/1722), trying to use AdaptiveCpp with C++ modules currently leads to a linker error with LLVM 19. An examination of the emitted LLVM IR shows that the function declaration of `__acpp_reflection_associate_function_pointer` is lost during compilation. 

This MR proposes a possible solution by declaring the function in `src/compiler/reflection/FunctionNameExtractionPass.cpp` *explicitly* and *leaving out* the [inline definition](https://github.com/AdaptiveCpp/AdaptiveCpp/blob/develop/include/hipSYCL/glue/reflection.hpp#L77) in `include/hipSYCL/glue/reflection.hpp`.

### DIFF
--- a/src/runtime/support/reflection.cpp
+++ b/src/runtime/support/reflection.cpp
@@ -18,3 +18,10 @@ namespace hipsycl::rt::support {
     return si;
   }
 }
+
+// Compiler will generate calls to this function to register functions
+void __acpp_reflection_associate_function_pointer(const void *func_ptr,
+                                           const char *func_name) {
+  hipsycl::rt::support::symbol_information::get()
+      .register_function_symbol(func_ptr, func_name);
+}
--- a/src/runtime/support/reflection.cpp
+++ b/src/runtime/support/reflection.cpp
@@ -19,10 +19,9 @@ namespace hipsycl::rt::support {
   }
 }
 
-//avoid cpp name mangling
-extern "C" void __acpp_reflection_associate_function_pointer(const void*, const char*);
+
 // Compiler will generate calls to this function to register functions
-void __acpp_reflection_associate_function_pointer(const void *func_ptr,
+extern "C" void __acpp_reflection_associate_function_pointer(const void *func_ptr,
                                            const char *func_name) {
   hipsycl::rt::support::symbol_information::get()
       .register_function_symbol(func_ptr, func_name);
--- a/src/runtime/support/reflection.cpp
+++ b/src/runtime/support/reflection.cpp
@@ -19,6 +19,8 @@ namespace hipsycl::rt::support {
   }
 }
 
+//avoid cpp name mangling
+extern "C" void __acpp_reflection_associate_function_pointer(const void*, const char*);
 // Compiler will generate calls to this function to register functions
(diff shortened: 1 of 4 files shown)


src/runtime/support/reflection.cpp
```diff
@@ -18,3 +18,12 @@ namespace hipsycl::rt::support {
     return si;
   }
 }
+
+//avoid cpp name mangling
+extern "C" void __acpp_reflection_associate_function_pointer(const void*, const char*);
+// Compiler will generate calls to this function to register functions
+void __acpp_reflection_associate_function_pointer(const void *func_ptr,
```
illuhad: We don't need the extra declaration in line 23, just put `extern "C"` in front of the definition here (line 25) :)

illuhad (approved):
Thanks :)

PR #1844, opened 2025-05-29
Use Minicoro instead of Boost.Fiber

al42and:
It's header-only and Unlicense / MIT-0, so can be easily bundled.

Quick testing on Linux with a [simple matrix transpose](https://gist.github.com/al42and/eda0d7c3d1667c94d038c7dfd5abfa9c) shows that performance is better than with Boost, except when ucontext fallback is used (measures throughput, more is better):

CPU | Compiler | Minicoro backend | Boost (GB/s) | Minicoro (GB/s)
-- | -- | -- | -- | --
Intel i9-12900 | Clang 19 | X86_64 asm | 0.387 | 0.600
POWER9 02CY089 | GCC 12.9 | ucontext | 0.134 | 0.056
Neoverse-N1 (4 shared cores) | Clang 18 | Aarch64 asm | 0.084 | 0.165
SiFive Freedom U740 | GCC 11.4 | RV64 asm | 0.007 | 0.013
Apple M1 | AppleClang 14 | Aarch64 asm | 0.076 | 0.098

Still orders of magnitude worse than accelerated CPU, but the main advantage is that we don't need boost anymore: neither for ACpp itself nor for applications it builds. CMake is also simplified.


include/hipSYCL/glue/generic/host/collective_execution_engine.hpp
```diff
@@ -45,118 +43,196 @@ enum class group_execution_iteration {
   sequential
 };
 
+namespace yield_kind {
+// Some odd value not easily confused with pointers
+static void* spawn = reinterpret_cast<void*>(0xff07);
+static void* barrier = reinterpret_cast<void*>(0xff09);
+static void* next_item = reinterpret_cast<void*>(0xff11);
+}
+static constexpr size_t fiber_stack_size = 256*1024;
```
illuhad: How does this compare to what `boost.fiber` uses? It seems a little small.

al42and:
Passes edge-cases of (a) single kernel with two work-groups, only one using work-group barriers; (b) using a mix of "no-barrier" (one fiber for all work-items) and "barrier" (one fiber per work-item) kernels.

include/hipSYCL/glue/generic/host/collective_execution_engine.hpp
```diff
@@ -14,20 +14,18 @@
...
+ * Allow disabling fibers; and don't try using them in device pass.
  */
-#if !defined(HIPSYCL_NO_FIBERS) && !defined(SYCL_DEVICE_ONLY)
-#define HIPSYCL_HAS_FIBERS
+#if !defined(ACPP_NO_FIBERS) && !defined(SYCL_DEVICE_ONLY)
+#define ACPP_USE_FIBERS
 #endif
 
-#ifdef HIPSYCL_HAS_FIBERS
+#ifdef ACPP_USE_FIBERS
```
illuhad, line 23: Perhaps we should consider removing this macro in the future. Not sure if after this PR there's still a use case for not having fibers available?

illuhad (commented):
Thank you :) This looks good as far as I can see, but I'm not an expert on minicoro patterns.

Do you think it would be possible to move the minicoro stuff into some class (say, `fiber` or `coroutine`) to abstract away the spawning and barrier semantics so that it could be used similarly to the `boost::fiber` class we have used so far?

It's a bit difficult to follow our execution-model-switching algorithm with the minicoro-specifics in between. But if you say that such an abstraction is a lot of work, I wouldn't want to hold up the PR because of that.

Note also my question regarding stack size :)

illuhad:
@al42and What are your plans here?

al42and:
> Do you think it would be possible to move the minicoro stuff into some class (say, fiber or coroutine) to abstract away the spawning and barrier semantics so that it could be used similarly to the boost::fiber class we have used so far?

Are you interested specifically in boost::fiber-like API, or anything to further isolate the implementation details? Shoving the code into a separate file is easy; if you want to have a nice abstraction layer to simplify replacement with a different engine: more complex and would be nicer to have an idea of what targets are in mind.

include/hipSYCL/glue/generic/host/collective_execution_engine.hpp
al42and: > Not sure if after this PR there's still a use case for not having fibers available?

Some weirdo Windows or microcontroller (#1720) build?

include/hipSYCL/glue/generic/host/collective_execution_engine.hpp
al42and: [The default for Boost.Fiber is 128KiB](https://github.com/boostorg/context/blob/boost-1.89.0/src/posix/stack_traits.cpp#L73-L76), so it's actually twice that

al42and:
BTW, do you think the `yield_kind` is helpful here? It's totally optional and removing it will simplify the code. But it allows the threads to signal _why_ they are yielding, which could help with debugging.

include/hipSYCL/glue/generic/host/collective_execution_engine.hpp
illuhad: Ah, I see, thanks!

include/hipSYCL/glue/generic/host/collective_execution_engine.hpp
illuhad: Ah, right!

illuhad:
> Are you interested specifically in boost::fiber-like API, or anything to further isolate the implementation details? Shoving the code into a separate file is easy; if you want to have a nice abstraction layer to simplify replacement with a different engine: more complex and would be nicer to have an idea of what targets are in mind.

I don't think we need to replicate boost::fiber specifically, or necessarily need to have the flexibility to switch out fiber implementations.
I was mainly asking about just hiding more of the implementation details, so that we have a better separation between the execution engine and the optimization it performs (delayed on-demand fiber spawning etc) and how fibers are created and synchronize with each other.

> BTW, do you think the yield_kind is helpful here? It's totally optional and removing it will simplify the code. But it allows the threads to signal why they are yielding, which could help with debugging.

I agree, could be helpful. If we can encapsulate it a bit further it probably also wouldn't be distracting from the rest of the code.

al42and:
> I was mainly asking about just hiding more of the implementation details,

Done. The abstraction is still a bit leaky since we need to keep track of the currently active fiber (can implement a pool to hide it too, but that seems like just moving 90% of `collective_execution_engine` functionality to a different file). On the upside, now the user applications don't require linking to minicoro at all, it's all in the ACpp runtime library.

include/hipSYCL/glue/generic/host/minicoro_wrapper.hpp
```diff
@@ -0,0 +1,62 @@
+#ifndef HIPSYCL_COROUTINE_WRAPPER_HPP
+#define HIPSYCL_COROUTINE_WRAPPER_HPP
+
```
illuhad: AdaptiveCpp copyright header?

src/runtime/support/minicoro_wrapper.cpp
```diff
@@ -0,0 +1,108 @@
+/*
```
illuhad, line 1: Would it make sense to move this to `src/runtime/support`?

include/hipSYCL/glue/generic/host/minicoro_wrapper.hpp
```diff
@@ -0,0 +1,62 @@
+#ifndef HIPSYCL_COROUTINE_WRAPPER_HPP
```
illuhad: Would it make sense to also move this file to `support`?

include/hipSYCL/glue/generic/host/minicoro_wrapper.hpp
```diff
@@ -0,0 +1,62 @@
...
+#define HIPSYCL_COROUTINE_WRAPPER_HPP
+
+#include <functional>
+
+struct mco_coro;
+
+namespace hipsycl {
+namespace glue {
+namespace host {
```
illuhad: If moving to support, namespace should be `rt::support`. Perhaps with `host` sub-namespace, but I don't have a strong opinion about that.

src/runtime/minicoro_wrapper.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+template<typename arguments_type>
+void fiber<arguments_type>::entry_point(mco_coro* co) {
+  auto* self = static_cast<fiber*>(mco_get_user_data(co));
+  self->_function(self);
+}
+}
+
+// Not beautiful, but helps keep it all isolated
+#include "hipSYCL/glue/generic/host/collective_execution_engine.hpp"
```
illuhad: Is there a way to get rid of this include? We generally try to avoid including `glue` headers into runtime code, since those might have specific compiler requirements etc.

illuhad (commented):
Almost there :)

include/hipSYCL/glue/generic/host/minicoro_wrapper.hpp
al42and: done

src/runtime/support/minicoro_wrapper.cpp
al42and: done

include/hipSYCL/glue/generic/host/minicoro_wrapper.hpp
al42and: done

include/hipSYCL/glue/generic/host/minicoro_wrapper.hpp
al42and: done. No `host ` namespace due to personal preference :)

src/runtime/minicoro_wrapper.cpp
al42and: Ok, replaced with a void pointer. Also replaced `std::vector<unique_ptr>` with `std::deque` to have less indirection while maintaining stable pointers.

al42and:
Curiously, on a GH200 machine (probably any other high core-count machine), Minicoro suffers a bit from larger fiber creation overheads (`mco_create`). E.g., in the matrix transpose test linked above, it outperfroms Boost up to ~56 threads, after which the performance tanks:

image

If we increase the problem size, the switchover happens later:

image

I don't think it's a problem for a backend that's intended primarily for debugging, but FYI.

illuhad (approved):
Awesome, thanks! Thanks for checking performance - I agree that, while interesting, this is not really cause for concern.

PR #1846, opened 2025-05-30
Respect user-defined HIPSYCL_DEBUG_LEVEL and remove redundant local settings

blinkfrog:
This PR makes the following improvements:

* Moves the default-setting logic for `HIPSYCL_DEBUG_LEVEL` to the top-level `CMakeLists.txt` only

* Removes redundant and conflicting `set(... CACHE ...)` call from subdirectory `src/compiler`

* Ensures that user-defined values via -DHIPSYCL_DEBUG_LEVEL=... or presets are respected (removes FORCE)

* Adds a validation in submodules to ensure `HIPSYCL_DEBUG_LEVEL` is set

* Fixes the issue where the macro was not propagated to compiler command lines, avoiding the need to patch debug.hpp manually

* Replaces `if(NOT HIPSYCL_DEBUG_LEVEL)` with `if(NOT DEFINED HIPSYCL_DEBUG_LEVEL)` to ensure that explicitly setting -DHIPSYCL_DEBUG_LEVEL=0 is respected. The previous check would incorrectly treat 0 as undefined and override it with the default.

This improves developer experience by adding a robust way to set default `HIPSYCL_DEBUG_LEVEL` during build configuring

I have tested these changes with older AdaptiveCpp, not with the latest, as I have some problems compiling it on Windows with LLVM Clang currently, but I expect these changes to work, because modified parts weren't changed since the version I tested. Configuring the latest AdaptiveCpp still works correctly, as expected, with `HIPSYCL_DEBUG_LEVEL` correctly set in `build.ninja`.

I know,  `HIPSYCL_DEBUG_LEVEL` should be renamed to `ACPP_DEBUG_LEVEL`, but this is better to do in a separate PR anyway.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -106,11 +106,11 @@ if(NOT HIPSYCL_DEBUG_LEVEL)
   if(CMAKE_BUILD_TYPE MATCHES "Debug")
     set(HIPSYCL_DEBUG_LEVEL 3 CACHE STRING
       "Choose the debug level, options are: 0 (no debug), 1 (print errors), 2 (also print warnings), 3 (also print general information)"
-FORCE)
+  )
   else()
     set(HIPSYCL_DEBUG_LEVEL 2 CACHE STRING
       "Choose the debug level, options are: 0 (no debug), 1 (print errors), 2 (also print warnings), 3 (also print general information)"
-FORCE)
+  )
   endif()
 endif()
 
--- a/src/compiler/CMakeLists.txt
+++ b/src/compiler/CMakeLists.txt
@@ -12,8 +12,8 @@ else()
   add_compile_definitions(ACPP_LLVM_COMPONENT)
 endif()
 
-if(NOT HIPSYCL_DEBUG_LEVEL)
-  set(HIPSYCL_DEBUG_LEVEL 2 CACHE INTEGER)
+if(NOT DEFINED HIPSYCL_DEBUG_LEVEL)
+  message(FATAL_ERROR "HIPSYCL_DEBUG_LEVEL was not defined by the top-level CMakeLists.txt.")
 endif()
 
 if(NOT ACPP_LLVM_COMPONENT)


illuhad (approved):
I'm not opposed to merging these changes as it fixes something that has been broken, but we should perhaps also consider the overall gameplan here.
I don't think it's ideal that users like you have to rebuild AdaptiveCpp at all to configure the logging behavior. Ideally, this could be user-configurable with a command line flag for `acpp` or so.
The main challenge is with logging that happens early on when the runtime starts up. The information would need to be propagated to the AdaptiveCpp runtime before it starts :thinking: 

Perhaps the runtime could look for a configuration file in the same directory as `libacpp-rt`, and pull the default settings from there?

blinkfrog:
Thank you for considering merging this PR.

I completely agree that having to rebuild AdaptiveCpp just to change the logging behavior isn’t ideal. I’ve also been thinking about alternatives, such as command-line flags, a configuration file (which I think is a great idea!), or even runtime control through API calls (though that has obvious drawbacks).

That said, I believe the current mechanism - using an environment variable to control the logging level - is already a solid solution. The only issue, in my opinion, is the default logging level itself. I would suggest setting the default to 0 (NONE), so that logging is completely disabled unless explicitly enabled. Logging is mostly useful for debugging and development, not for end users running final applications. Developers can easily override it through the environment when needed.

illuhad:
> That said, I believe the current mechanism - using an environment variable to control the logging level - is already a solid solution. The only issue, in my opinion, is the default logging level itself. I would suggest setting the default to 0 (NONE), so that logging is completely disabled unless explicitly enabled. Logging is mostly useful for debugging and development, not for end users running final applications. Developers can easily override it through the environment when needed.


I think this might be a case where different user groups might want different behaviors. I can see that in your case, you wouldn't want logging to be enabled by default. But in general, the warnings we print (and even more so, errors) are highly relevant, although it might be debatable specifically whether failure to load a backend should trigger a warning. Warnings prevent mistakes, and in many cases (e.g. HPC/academia) the software developer and end user might not be as distinct as in your case.
For example, we have a warning that tells the user that the performance of the application was affected by JIT overheads and that they should rerun the application to benefit from the now-populated JIT cache. This type of warning might prevent a researcher running benchmarks from accidentally publishing results that are affected by JIT and thus not representative.

Similarly, a Linux user who is a SYCL beginner and just wants to use their GPU should be informed of any issues with their setup (which will also lead to better bug reports), but they might be overwhelmed by the AdaptiveCpp documentation and not be aware that there are environment variables to turn on logging.

Of course, ideally everyone would read the documentation throroughly and enable logging in such cases... but experience has shown that the world does not work that way :-)

So it seems a bit more attractive to leave the default log level as is, but add the configuration file mechanism.

We do have several environment variables that can be used for performance tuning (`ACPP_ADAPTIVITY_LEVEL`, `ACPP_ALLOCATION_TRACKING`, `ACPP_STDPAR_PREFETCH_MODE` etc). So, the application developer might want to use those for fine-tuning AdaptiveCpp for the specific application, and then ship that configuration to users.
So, a configuration file I think would be beneficial for more cases than just logging :-)

blinkfrog:
I agree, these concerns are valid.

Also, additional flexibility is always good. May be, though, configuration file isn't needed, and simpler approach would work well enough - as you have mentioned already, compile flags for acpp when compiling SYCL app - they could set default values for all these parameters.

illuhad:
> Also, additional flexibility is always good. May be, though, configuration file isn't needed, and simpler approach would work well enough - as you have mentioned already, compile flags for acpp when compiling SYCL app - they could set default values for all these parameters.

The thing is, it's not really clear whether compile flags would be easier:
- The settings need to be passed to the runtime before it initializes. This means we need a mechanism that does not depend on initialization order of globals, and it's not really clear how to achieve that.
- What should happen if different components of the application are compiled with different flags?

The configuration file avoids both problems. So it might even be simpler in the end :-)

PR #1848, opened 2025-06-03
Add support for config files for runtime settings

illuhad:
From the documentation:

---
All environment variables for the runtime (not JIT compiler) can also be set in a configuration file. This configuration file must be placed in the same directory as your program.

1. First, AdaptiveCpp will attempt to read `acpp-config.cfg`.
2. Next, it reads `acpp-config-<app-name>.cfg`, where `app-name` is the full filename of your program (including file extension if you are on a platform that uses an extension for programs). If some variable was already set from `acpp-config.cfg`, it will be overwritten.

This allows users to specify a global behavior for all programs located in a directory, and special-case the behavior for individual programs using `acpp-config-<app-name>.cfg` files.

When there are conflicting environment variables and entries in the config file, environment variables take precedence.

The content of the configuration file is comprised of `<environment-variable>=<value>` entries, one by line.

Example:
```
ACPP_DEBUG_LEVEL=0
ACPP_ADAPTIVITY_LEVEL=2
```
----

@blinkfrog this PR includes a Windows-specific code path to determine the path to the running program that I could not test (`get_this_executable_path()` in `filesystem.cpp`). It would be good if you could check whether it works as intended.

### DIFF
--- a/bin/acpp
+++ b/bin/acpp
@@ -1924,7 +1924,8 @@ class compiler:
   def common_linker_flags(self):
     linker_args = [
       "-L"+self._acpp_lib_path,
-      "-lacpp-rt"
+      "-lacpp-rt",
+      "-lacpp-common"
     ]
 
     if sys.platform == "darwin":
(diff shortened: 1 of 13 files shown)


blinkfrog:
Thanks, this works, and is very useful! Tested both: usual `acpp-config.cfg` and `acpp-config-<app-name>.cfg` - the latter successfully overrides the former.

PR #1850, opened 2025-06-03
[SSCP][CUDA] Improve ptx version and target selection

illuhad:


src/runtime/cuda/cuda_queue.cpp
```diff
@@ -47,12 +47,61 @@
...
+       {86, 71},
+       {87, 74},
+       {89, 78},
+       {90, 78},
+       // some variants of 100, 101 need 8.6, some 8.8
+       // so require 8.8 for all to be on the safe side.
+       {100, 88},
+       {101, 88},
+       {103, 88},
+       {120, 87},
```
al42and: If we care about 120f (which does not make much sense to choose for JIT, I guess, but for consistency with 100f):

```suggestion
       {120, 88},
```

src/runtime/cuda/cuda_queue.cpp
```diff
@@ -47,12 +47,61 @@
...
+  // based on data from
+  // https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#ptx-module-directives-target
+  static std::vector<std::array<unsigned, 2>> sm_min_ptx_version =
+      {{72, 61},
+       {75, 63},
+       {80, 70},
+       {86, 71},
+       {87, 74},
+       {89, 78},
+       {90, 78},
```
al42and: If we care about 90a (which looks like a good choice for JIT, since we don't care about future-compatibility of PTX):

```suggestion
       {90, 80},
```

al42and (approved):

src/runtime/cuda/cuda_queue.cpp
illuhad: Thanks! feels like NVIDIA learnt from AMD here with this SM variant stuff with different requirements :/

PR #1853, opened 2025-06-08
Fix OpenMP flag handling for omp target in LLVM + AdaptiveCpp integrated build on Windows (MSVC/clang-cl)

blinkfrog:
This PR makes minimal changes, sufficient to make OpenMP backend to work in `generic` target. `MSVC` variable is used to check if `clang-cl` or `msvc` compiler is used to compile LLVM + AdaptiveCpp, thus, if possibility to build AdaptiveCpp with `clang` (GNU-style) on Windows is restored, it still will work correctly.


illuhad (approved):
Thanks, this looks like a great solution :)

PR #1852, opened 2025-06-07
[SSCP][llvm-to-host] Use opt; llc; lld for JIT compilation instead of clang

illuhad:
Currently, we use clang when JIT compiling for the host CPU, which creates a large dependency surface, potentially hindering deployment.

This PR changes host JIT to instead rely on LLVM tools `opt`, `llc`, `lld` which are better suited for redistribution than `clang`.
Note that we've done similar changes already to other backends (#1830). Once this PR is merged, no backends need clang anymore at JIT time.

@blinkfrog CI is happy, but you might want to test that is reliable on Windows. On Windows different linking logic is used during JIT. Also note that we now also need `lld` in `LLVM_ENABLE_PROJECTS`.

PR #1854, opened 2025-06-10
[SSCP] ranged parallel_for: On CPU, check whether WGs are in the domain only if they are at the edges

illuhad:
As found out in #1852, there can be a noticable performance hit in `parallel_for(range)` compared to `parallel_for(nd_range)` because `parallel_for(range)` needs to also support cases where the `range` is not divisible by the work group size, in which case the last work groups will only partially participate in the computation.

This requires an additional if statement to check whether items are still part of the parallel domain. This if statement can negatively affect the generated code.

This PR changes the logic, such that on CPU, we only do this check for the last work groups in each dimension. This allows us to have most of the work groups without the check (and thus be more efficient). This approach is possible because only the last work groups in each dimension can potentially exceed the iteration space.

@blinkfrog With this change, perf improves from ~78ms to 58ms for your original benchmark code for me! Also, OpenCL on CPU improves from 225 to ~200ms.

### DIFF
--- a/include/hipSYCL/glue/llvm-sscp/sscp_kernel_launcher.hpp
+++ b/include/hipSYCL/glue/llvm-sscp/sscp_kernel_launcher.hpp
@@ -31,11 +31,14 @@
 #include "ir_constants.hpp"
 
 #include "../kernel_launcher_data.hpp"
+#include "s1_ir_constants.hpp"
 
 #include <array>
 #include <atomic>
 #include <string_view>
 
+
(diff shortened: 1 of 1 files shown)


mikaoP:
Is something blocking this?

illuhad:
> Is something blocking this?

No, I just forgot pushing it :)

@sbalint98, can you review? This PR has potential performance impact for the SSCP->CPU path.

sbalint98 (approved):
LGTM, sorry for the delay

PR #1855, opened 2025-06-12
[SSCP][llvm-to-amdgpu] Fix discovery of default ABI version if LLVM module is not tagged with ABI version

illuhad:
When AdaptiveCpp is built against older LLVM versions (e.g. 15), then our bitcode libraries don't come  pre-tagged with a target ROCm ABI version. In this case, we need to manually figure out the target ABI so that we can link the appropriate `oclc_ABI_version_*.bc` file.

Currently, this fallback code path tries to list files in the ROCm bitcode library path, and then attempts to parse the filenames to obtain available ABI versions. However, it assumes that `list_regular_files` returns filenames, when in fact it returns the full paths.

This can cause the wrong or no `oclc_ABI_version_*.bc` to be linked. This PR fixes that.

@al42and I think this might be related to your issue.


al42and:
> @al42and I think this might be related to your issue.

Yes, it was! Thanks for the quick fix!

PR #1857, opened 2025-06-12
[stdpar][algorithms] Drastically improve performance of early-exit-based algorithms (any_of, all_of, none_of)

illuhad:
The `any_of`, `all_of`, `none_of` algorithms rely on early-exit-optimizations. This can substantially improve performance for inputs where the exit condition is reached early (e.g. `any_of` where the first elements already meet the condition).

However, the current implementation relies on excessive use of atomics, which can significantly limit performance in cases where the exit condition is only met late (or never) in the data.

This PR substantially optimizes this, by
- processing multiple data elements per work item to saturate memory bandwidth
- optimizing memory access patterns for CPU backend
- only performing atomics in the first work item per work group, and exchanging data within a work group using group algorithms.
- Using `device` instead of `host` memory to store the output of the flag.

With this PR, I reach close(~85% on GPU) to peak memory bandwidth for the case where the exit condition is not met while still achieving an effective memory bandwidth that is higher than peak when the exit condition is met early.
Similar numbers are achieved on CPU as well.

Some additional minor speedup on NVIDIA GPUs specifically can be achieved with the changes in #1856.

@normallytangent This should fix your performance issues :) Note that for best effect, you will also have to change your algorithms to use `device` memory for the output flag, and then copy the result back to the host using an explicit `memcpy` as I did in this PR for `any_of`/`all_of`/`none_of`.

EDIT: CI errors are unrelated :( Looks like there is currently an issue with CI.

### DIFF
--- a/include/hipSYCL/algorithms/algorithm.hpp
+++ b/include/hipSYCL/algorithms/algorithm.hpp
@@ -500,28 +500,10 @@ sycl::event early_exit_for_each(sycl::queue &q, std::size_t problem_size,
   std::size_t dispatched_global_size = streamer.get_required_global_size();
 
   auto kernel = [=](sycl::nd_item<1> idx) {
-      const std::size_t item_id = idx.get_global_id(0);
-  
-      util::abortable_data_streamer::run(problem_size, idx, [&](sycl::id<1> idx){
-        
-        if (sycl::detail::__acpp_atomic_load<
-                sycl::access::address_space::global_space>(
-                output_has_exited_early, sycl::memory_order_relaxed,
-                sycl::memory_scope_device)) {
-          return true;
-        }
-
-        if (should_exit(idx)) {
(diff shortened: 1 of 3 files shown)


normallytangent:
Thank you so much! Just tested this for `is_sorted` and can confirm the significant improvement in performance on GPU and CPU, and incorporated changes from #1856 in my tests for NVIDIA GPUs.  I get around 70-75% of peak bandwidth on GPU for the "worst case" data. The missing ~15% gain compared to `any_of` can be attributed to the additional accessing of neighbouring elements for the pairwise comparisions in `is_sorted`.

illuhad:
Nice :) I suppose there's still room for a bit more optimization and fine tuning of this PR (e.g. playing around with the amount of work that is assigned to each work item), but I think it should be good enough for now :)

PR #1860, opened 2025-06-17
Fixing fast-math flag transmission to JIT Host compiler

thomasRoglin:
This PR activate flags transmission to the JIT Host compiler. 
Currently, using the ffast-math flag at compile time, does not toggle `IsFastMath`. 
This prevents `libkernel-sscp-host-fast-full.bc` to be loaded (libkernel-sscp-host-full.bc is loaded instead). 
And the [condition](https://github.com/AdaptiveCpp/AdaptiveCpp/blob/799002752fc1ecf49a347de048d3680bdf00190d/src/compiler/llvm-to-backend/host/LLVMToHost.cpp#L253) that adds fast-math flags to the llc invocation is never true.


illuhad (approved):
Thanks, great catch!

PR #1862, opened 2025-06-18
Fix implicit conversion for accessors

### LINKED ISSUE #1735: Missing implicit conversion between  read-write local accessor to read-only
Standard [Read-only local accessors](https://github.khronos.org/SYCL_Reference/iface/local_accessor.html#read-only-local-accessors) says about `sycl::local_accessor<T>` and  `sycl::local_accessor<const T>`

>  [t]here is an implicit conversion from the read-write specialization to the read-only specialization, provided that all other template parameters are the same.

**To Reproduce**
```cpp
#include <sycl/sycl.hpp>

void kernel_function(sycl::local_accessor<const float> data){
	// do nothing
};

int main(){
	sycl::queue my_queue;

	my_queue.submit([](sycl::handler & cgh){
		const sycl::local_accessor<float> data{10, cgh};

		cgh.single_task(
			[=](){
				kernel_function(data);
			}
		);
	}).wait();
}
```

Passing a read-write local_accesor to the read only version fails to compile
```txt
/opt/AdaptiveCpp/bin/acpp const_correctness_local_accessor.cpp 
acpp warning: No optimization flag was given, optimizations are disabled by default. Performance may be degraded. Compile with e.g. -O2/-O3 to enable optimizations.
const_correctness_local_accessor.cpp:15:5: error: no matching function for call to 'kernel_function'
   15 |                                 kernel_function(data);
      |                                 ^~~~~~~~~~~~~~~
const_correctness_local_accessor.cpp:3:6: note: candidate function not viable: no known conversion from 'const local_accessor<float>' to 'local_accessor<const float>' for 1st argument
    3 | void kernel_function(sycl::local_accessor<const float> data){
      |      ^               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

**Expected behavior**

Should compile, i.e. compiles successfully with oneAPI.

**Describe your setup**

```
acpp [AdaptiveCpp compilation driver], Copyright (C) 2018-2024 Aksel Alpay and the AdaptiveCpp project
  AdaptiveCpp version: 24.10.0+git.7677cf6e.20241220.branch.HEAD
  Installation root: /opt/AdaptiveCpp
  Plugin LLVM version: 17, can accelerate CPU: True
  Available runtime backends:
     librt-backend-omp.so
     librt-backend-ocl.so
     librt-backend-cuda.so
```

**Additional context**

I enjoy const correctness therefore would be very much pleased if this would compile :)

ath0mos:
I modified the given implicit-conversion operator and added also a copy constructor to allow for more types of initialization/conversion


include/hipSYCL/sycl/libkernel/accessor.hpp
```diff
@@ -2231,6 +2231,23 @@ class local_accessor
 
   local_accessor() = default;
 
+  template <typename U,
+  typename = std::enable_if_t<
+  std::is_const_v<dataT> &&
+  std::is_same_v<U, std::remove_const_t<dataT>>>>
```
illuhad: Can you please indent code that is part of template parenthesis? (see other template functions in this class as examples) :)

include/hipSYCL/sycl/libkernel/accessor.hpp
```diff
@@ -2231,6 +2231,23 @@ class local_accessor
...
+  local_accessor(const local_accessor<U, dimensions> &other) :
+    _addr(other.getAddr()), _num_elements(other.getNumElem()) {
+  }
+
+  template <typename U,
+  typename = std::enable_if_t<
+  std::is_const_v<dataT> &&
+  std::is_same_v<U, std::remove_const_t<dataT>>>>
```
illuhad: Here too

illuhad (approved):
Thanks!

include/hipSYCL/sycl/libkernel/accessor.hpp
```diff
@@ -2250,6 +2267,11 @@ class local_accessor
       _num_elements{allocationSize}
   {}
 
+  specialized<address> getAddr() const { return _addr; }
```
illuhad:
```suggestion
  specialized<address> AdaptiveCpp_get_addr() const { return _addr; }
```
Public functions that are not part of the spec are effectively AdaptiveCpp extensions. We usually prefix those with `AdaptiveCpp` :)

include/hipSYCL/sycl/libkernel/accessor.hpp
```diff
@@ -2250,6 +2267,11 @@ class local_accessor
...
+
+  specialized<range<dimensions>> getNumElem() const { return _num_elements; }
```
illuhad:
```suggestion
  specialized<range<dimensions>> AdaptiveCpp_get_num_elem() const { return _num_elements; }
```
same here.
(an alternative to these getter functions might be to make every local accessor `friend` with every other one?)

illuhad (approved):
Thanks!

PR #1865, opened 2025-06-25
Enable direct peer access for CUDA and HIP

MarkusBuettner:
Following the recent discussion in https://github.com/AdaptiveCpp/AdaptiveCpp/discussions/1849, this patch adds direct peer-to-peer access for NVIDIA and AMD GPUs.

### DIFF
--- a/src/runtime/cuda/cuda_hardware_manager.cpp
+++ b/src/runtime/cuda/cuda_hardware_manager.cpp
@@ -11,6 +11,7 @@
 #include "hipSYCL/runtime/cuda/cuda_hardware_manager.hpp"
 #include "hipSYCL/runtime/cuda/cuda_event_pool.hpp"
 #include "hipSYCL/runtime/cuda/cuda_allocator.hpp"
+#include "hipSYCL/runtime/cuda/cuda_device_manager.hpp"
 #include "hipSYCL/runtime/device_id.hpp"
 #include "hipSYCL/runtime/error.hpp"
 #include "hipSYCL/runtime/hardware.hpp"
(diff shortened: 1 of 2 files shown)


src/runtime/cuda/cuda_hardware_manager.cpp
```diff
@@ -56,6 +56,27 @@ cuda_hardware_manager::cuda_hardware_manager(hardware_platform hw_platform)
     _devices.emplace_back(dev);
   }
 
+  for (int dev = 0; dev < num_devices; ++dev) {
+    cudaSetDevice(dev);
```
illuhad: Maybe it might be cleaner to use `hip/cuda_device_manager::get().activate_device(dev);` here?

This is the pattern we usually use to set the active device. It caches the active device and only sets it if needed. Setting it directly might be a little error prone because it might lead to mismatches between what the device manager thinks is active, and what is actually active. Although I suspect it's currently fine since the `hip/cuda_device_manager` is likely only first used (and thus initialized) after this point.

MarkusBuettner: That's a fair point. I didn't encounter any issues in my tests on Nvidia and AMD GPUs, but I'll change it.

illuhad (approved):
Thanks!

fodinabor:
Not sure if this really matters in practice, but I'll just comment that there is indeed a few machines out there, that pose as supporting 10 (single-slot) GPUs in a chassis: https://www.supermicro.com/en/products/system/gpu/5u/sys-521ge-tnrt
So, potentially, we might want to limit doing this to 8 GPUs? Or do we just want to take the error code and emit a warning if it failed?

illuhad:
But do they all support P2P to everyone? If there's some switching in between, maybe it could be that `cudaDeviceCanAccessPeer` would return false for some combinations, in which case we'd be okay?

PR #1869, opened 2025-07-03
[NFC] llvm-to-backend: remove duplicate AlwaysInliner include

bv2k4:
Remove a duplicated `#include <llvm/Transforms/IPO/AlwaysInliner.h>` in  
`src/compiler/llvm-to-backend/LLVMToBackend.cpp`.


illuhad (approved):
Thanks!

PR #1871, opened 2025-07-05
[CBS] fix missing erasure of replaced load instructions

bv2k4:
In `src/compiler/cbs/IRUtils.cpp`, `replaceUsesOfGVWith`, after replacing uses of `LoadInst` with new value, the original load instructions were not being erased. This change adds each replaced `LoadInst` to the `ToErase` vector to ensure proper cleanup.


fodinabor (approved):
Lgtm, thanks!

PR #1877, opened 2025-07-16
Use ockl functions for subgroup collpredicates for SSCP

sbalint98:
These functions should be more performant than our basic reductions, as they rely on intrinsics


illuhad (approved):
Thanks!

PR #1878, opened 2025-07-18
changed namespace cl::sycl into sycl

ath0mos:


tests/compile_commands.json
```diff
@@ -0,0 +1,212 @@
+[
```
illuhad: I think this file probably should not be part of the PR, right? Can you remove it?

tests/compiler/cbs/accumulator_for.cpp
```diff
@@ -10,31 +10,31 @@
 #include <array>
 #include <iostream>
 
-#include <CL/sycl.hpp>
+#include <SYCL/sycl.hpp>
```
illuhad: Please use lower-case `<sycl/sycl.hpp>`.

Early versions of the SYCL 2020 specification used `<SYCL/sycl.hpp>`, which is why AdaptiveCpp supports both - but current SYCL 2020 only has `<sycl/sycl.hpp>`.

Please change this everywhere :)

illuhad (approved):
Thanks!

illuhad (commented):
There are some CI issues, so it looks like these changes have broken some of the test cases. Can you look into it?

illuhad (approved):
Thanks!

PR #1882, opened 2025-07-23
Removing warnings/errors when compiling with headers 

ath0mos:
removed deprecated functions, which caused warnings. 
Mostly i removed 

1. Commit:
  - vector_class -> std::vector
  - string_class -> std::string

2. Commit:
  - Removed a few conversion errors, by adding/changing static_cast


illuhad (approved):
Thanks :)

PR #1883, opened 2025-07-25
Fix kernel with a range greater than INTMAX

thomasRoglin:
Hello, This pull request fixes the bug preventing the generic backend from correctly invoking kernels whose globalSize is greater than INTMAX.  ( [which should be allowed ](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#nditem-class))

This should make [this test ](https://github.com/KhronosGroup/SYCL-CTS/blob/97daa051bf99d3e39f96dcba7f92614c4a5816e8/tests/range/range_api.cpp#L274C9-L274C36) from the SYCL-CTS work. 

Here is a reproducer : 
(⚠️This will use more than 50G of memory )

```
#include <sycl/sycl.hpp>
#include <omp.h>

#define WG (size_t) 16

int main(int argc, char* argv[])
{

  if (argc != 2) {
    printf("Usage: %s <size>\n", argv[0]);
    return 1;
  }
  const size_t size = std::stoull(argv[1]);

  sycl::queue q(sycl::cpu_selector_v, sycl::property::queue::in_order());

  size_t *a_d = sycl::malloc_device<size_t>(size, q);
  size_t *a = sycl::malloc_host<size_t>(size, q);

  size_t wg_cnt = (size + WG - 1) / WG;
  sycl::range<1> gws ( wg_cnt * WG );
  sycl::range<1> lws (WG);

  q.submit([&] (sycl::handler &cgh) {
    cgh.parallel_for<class fwd>(sycl::nd_range<1>(gws, lws), [=] (sycl::nd_item<1> item) {

      size_t idx = item.get_global_id(0);
      if(idx < size)
        a_d[idx] = idx;

    });
  });
  q.memcpy(a, a_d, size*sizeof(size_t));
  q.wait();

  int err = 0;
#pragma omp parallel for reduction(+:err)
  for(size_t i = 0; i<size; i++){
    if( a[i] != i )
      err += 1;
  }
  printf("verif: %s \n", err ? "FAIL" : "PASS");

  return 0;
}

```

`acpp --acpp-targets=generic -Ofast -fopenmp main.cpp -o main &&  ACPP_VISIBILITY_MASK=omp ./main 4294967296`

### DIFF
--- a/include/hipSYCL/sycl/libkernel/detail/thread_hierarchy.hpp
+++ b/include/hipSYCL/sycl/libkernel/detail/thread_hierarchy.hpp
@@ -25,7 +25,7 @@ namespace detail {
 #define HIPSYCL_DEFINE_BUILTIN_VARIABLE_QUERY(                                 \
     name, cuda_variable, hip_variable, sscp_variable, host_variable)           \
   ACPP_KERNEL_TARGET                                                        \
-  inline int name() {                                                          \
+  inline size_t name() {                                                       \
     __acpp_backend_switch(return 0, return sscp_variable(),                    \
                                  return cuda_variable, return hip_variable)    \
   }


illuhad:
I would like to look into some performance regression tests for this one first. This PR widens the internal interface for *all* queries to 64-bit, including those that clearly never will be that large (such as local id).
I suspect that actually, for your use case, it would be to sufficient to only widen the number of groups and group id, right?

Note that your PR only fixes this issue on `omp` because `omp` happens to support an unlimited number of work groups, which is not the case for all other backends.
So, for full support of this test case, we would need to insert a grid stride loop in the `sscp_kernel_launcher` here: 
https://github.com/AdaptiveCpp/AdaptiveCpp/blob/7b4368004b1cfd9f4f292084c994e5477fd5c04e/include/hipSYCL/glue/llvm-sscp/sscp_kernel_launcher.hpp#L150

So, something like this:
```
int my_group_id = ...
int my_local_id = ...
int my_global_id = ...
if(needs_grid_stride) {
  for(int effective_global_id = my_global_id; my_global_id < requested_size; my_global_id += local_size * num_groups) {
    // (a local barrier here should be considered as it might help performance on CPU with such loops)
    kernel(effective_global_id);
  }
} else {
  kernel(my_global_id);
}
```
Then query the maximum number of supported work groups of the backend, and if we need more, only submit the maximum possible number and let the grid stride handle it. Some JIT magic could be used to remove the overhead of the `if`.

Such an approach could be an alternative solution if we don't want to widen the queries, and would also solve the problem for all other backends. Sooner or later, we will probably need to make this  change anyway to support the CTS test in question universally.

Note also that the specification only guarantees ranges greater than `INTMAX` to work with `range<>` kernels, not with `nd_range<>` kernels as in your example.

thomasRoglin:
Thanks for your reply, 
For this PR, I did not take into account the limitation on the maximum number of work groups. When testing on PoCL and Omp, this limitation was not restrictive.

>This PR widens the internal interface for _all_ queries to 64-bit, including those that clearly never will be that large (such as local id)

I do not think this PR widden internal interface for all queries, It seems all query are already 64-bit. 
Here is some .ll that I obtain with `ACPP_S2_IR_DUMP_ALL=1`, without this patch.  


First, `__acpp_sscp_get_local_id_x` return an 64-bit ([here](https://github.com/AdaptiveCpp/AdaptiveCpp/blob/7b4368004b1cfd9f4f292084c994e5477fd5c04e/src/libkernel/sscp/host/core.cpp#L30))

Since this [function ](https://github.com/AdaptiveCpp/AdaptiveCpp/blob/7b4368004b1cfd9f4f292084c994e5477fd5c04e/include/hipSYCL/sycl/libkernel/detail/thread_hierarchy.hpp#L28 ) return an int,  the value returned by `__acpp_sscp_get_local_id_x` is cast into an i32 and returned by `_ZN7hipsycl4sycl6detail16__acpp_get_lid_xEv`.

`_ZN7hipsycl4sycl6detail15get_global_id_xEv` retrieve this i32 value, compute the `_global_id_x` and recast it again into an i64. 


Finally, we start and end in i64. The intermediate transition to i32 only results in a loss of information.

```cpp
 ; Function Attrs: inlinehint mustprogress nounwind
 define linkonce_odr dso_local noundef i32 @_ZN7hipsycl4sycl6detail16__acpp_get_lid_xEv() local_unnamed_addr #3 comdat  {
   br label %1

 1:                                                ; preds = %0
   %2 = call i64 @__acpp_sscp_get_local_id_x() #5
   %3 = trunc i64 %2 to i32
   br label %4

 4:                                                ; preds = %1
   ret i32 %3
 }
```

```cpp
 ; Function Attrs: inlinehint mustprogress nounwind
 define linkonce_odr dso_local noundef i64 @_ZN7hipsycl4sycl6detail15get_global_id_xEv() local_unnamed_addr #3 comdat  {
   %1 = call noundef i32 @_ZN7hipsycl4sycl6detail16__acpp_get_gid_xEv() #5
   %2 = call noundef i32 @_ZN7hipsycl4sycl6detail18__acpp_get_lsize_xEv() #5
   %3 = mul nsw i32 %1, %2
   %4 = call noundef i32 @_ZN7hipsycl4sycl6detail16__acpp_get_lid_xEv() #5
   %5 = add nsw i32 %3, %4
   %6 = sext i32 %5 to i64
   ret i64 %6
}
```


> Note that your PR only fixes this issue on `omp`

I don't know the limits for cuda/hip/ze, but for the OpenCL backend with PoCL, the accepted workgroup size is greater than intmax, so this patch will also be beneficial for that backend.


> Note also that the specification only guarantees ranges greater than `INTMAX` to work with `range<>` kernels, not with `nd_range<>` kernels as in your example.

Ok, I did not know that, i assumed that the limitation for `nd_range<>` would be the same than `range<>`.

illuhad:
> Finally, we start and end in i64. The intermediate transition to i32 only results in a loss of information.

This might be misleading. Even if in the end the query is widened again to 64-bit, the compiler might capitalize on the knowledge that the intermediate value cannot have been larger than 32-bit `INTMAX` in the generated optimized code. Depending on how the query is used, this can result e.g. in 32-bit instructions being generated in the optimized code.

There is also a JIT-time optimization that we do that specifically tries to trigger this behavior, see https://dl.acm.org/doi/10.1145/3731125.3731127#sec-3-2 (section 3.2).

Can you check whether there is a change in the generated code for the example mentioned in this paper before and after your PR?

> I don't know the limits for cuda/hip/ze, but for the OpenCL backend with PoCL, the accepted workgroup size is greater than intmax, so this patch will also be beneficial for that backend.

I think OpenCL implementations for CPU in general might behave similar to our OpenMP backend, but I would already expect pocl's CUDA backend to mirror the limitations of CUDA.

> Ok, I did not know that, i assumed that the limitation for nd_range<> would be the same than range<>.

The general idea is that `nd_range<>` is supposed to be a low-level model that interfaces directly with backends. As such, users are responsible for handling backend limitations with `nd_range` - maximum work group sizes, maximum number of work groups, available amount of local memory etc.
`range<>` on the other hand is intended as a high-level abstraction that "just works" :)

thomasRoglin:
I rerun the code from the example of the paper with targetting a RTX 3060 GPU. 
There doesn't seem to be any difference between before and after my PR:

Before my PR
```asm
//
// Generated by LLVM NVPTX Back-End
//

.version 8.6
.target sm_86
.address_size 64

        // .globl       _Z18__acpp_sscp_kernelIN7hipsycl4glue15__sscp_dispatch20ndrange_parallel_forIZZ4mainENK3$_0clERNS0_4sycl7handlerEEUlNS5_7nd_itemILi3EEEE_Li3EEEEvRKT_ // -- Begin function _Z18__acpp_sscp_kernelIN7hipsycl4glue15__sscp_dispatch20ndrange_parallel_forIZZ4mainENK3$_0clERNS0_4sycl7handlerEEUlNS5_7nd_itemILi3EEEE_Li3EEEEvRKT_
                                        // @"_Z18__acpp_sscp_kernelIN7hipsycl4glue15__sscp_dispatch20ndrange_parallel_forIZZ4mainENK3$_0clERNS0_4sycl7handlerEEUlNS5_7nd_itemILi3EEEE_Li3EEEEvRKT_"
.visible .entry _Z18__acpp_sscp_kernelIN7hipsycl4glue15__sscp_dispatch20ndrange_parallel_forIZZ4mainENK3$_0clERNS0_4sycl7handlerEEUlNS5_7nd_itemILi3EEEE_Li3EEEEvRKT_(
        .param .u64 .ptr .global .align 1 _Z18__acpp_sscp_kernelIN7hipsycl4glue15__sscp_dispatch20ndrange_parallel_forIZZ4mainENK3$_0clERNS0_4sycl7handlerEEUlNS5_7nd_itemILi3EEEE_Li3EEEEvRKT__param_0
)
.maxntid 128, 1, 1
{
        .reg .b32       %r<13>;
        .reg .b64       %rd<4>;

// %bb.0:
        ld.param.u64    %rd1, [_Z18__acpp_sscp_kernelIN7hipsycl4glue15__sscp_dispatch20ndrange_parallel_forIZZ4mainENK3$_0clERNS0_4sycl7handlerEEUlNS5_7nd_itemILi3EEEE_Li3EEEEvRKT__param_0];
        mov.u32         %r1, %ctaid.x;
        mov.u32         %r2, %tid.x;
        mov.u32         %r3, %nctaid.x;
        mov.u32         %r4, %ctaid.y;
        mov.u32         %r5, %nctaid.y;
        mov.u32         %r6, %ctaid.z;
        shl.b32         %r7, %r1, 7;
        or.b32          %r8, %r7, %r2;
        mad.lo.s32      %r9, %r6, %r5, %r4;
        mul.lo.s32      %r10, %r3, %r9;
        shl.b32         %r11, %r10, 7;
        add.s32         %r12, %r8, %r11;
        mul.wide.u32    %rd2, %r12, 4;
        add.s64         %rd3, %rd1, %rd2;
        st.global.u32   [%rd3], %r12;
        ret;
                                        // -- End function
}

```

After my PR:


```asm
//
// Generated by LLVM NVPTX Back-End
//

.version 8.6
.target sm_86
.address_size 64

        // .globl       _Z18__acpp_sscp_kernelIN7hipsycl4glue15__sscp_dispatch20ndrange_parallel_forIZZ4mainENK3$_0clERNS0_4sycl7handlerEEUlNS5_7nd_itemILi3EEEE_Li3EEEEvRKT_ // -- Begin function _Z18__acpp_sscp_kernelIN7hipsycl4glue15__sscp_dispatch20ndrange_parallel_forIZZ4mainENK3$_0clERNS0_4sycl7handlerEEUlNS5_7nd_itemILi3EEEE_Li3EEEEvRKT_
                                        // @"_Z18__acpp_sscp_kernelIN7hipsycl4glue15__sscp_dispatch20ndrange_parallel_forIZZ4mainENK3$_0clERNS0_4sycl7handlerEEUlNS5_7nd_itemILi3EEEE_Li3EEEEvRKT_"
.visible .entry _Z18__acpp_sscp_kernelIN7hipsycl4glue15__sscp_dispatch20ndrange_parallel_forIZZ4mainENK3$_0clERNS0_4sycl7handlerEEUlNS5_7nd_itemILi3EEEE_Li3EEEEvRKT_(
        .param .u64 .ptr .global .align 1 _Z18__acpp_sscp_kernelIN7hipsycl4glue15__sscp_dispatch20ndrange_parallel_forIZZ4mainENK3$_0clERNS0_4sycl7handlerEEUlNS5_7nd_itemILi3EEEE_Li3EEEEvRKT__param_0
)
.maxntid 128, 1, 1
{
        .reg .b32       %r<13>;
        .reg .b64       %rd<4>;

// %bb.0:
        ld.param.u64    %rd1, [_Z18__acpp_sscp_kernelIN7hipsycl4glue15__sscp_dispatch20ndrange_parallel_forIZZ4mainENK3$_0clERNS0_4sycl7handlerEEUlNS5_7nd_itemILi3EEEE_Li3EEEEvRKT__param_0];
        mov.u32         %r1, %ctaid.x;
        mov.u32         %r2, %tid.x;
        mov.u32         %r3, %nctaid.x;
        mov.u32         %r4, %ctaid.y;
        mov.u32         %r5, %nctaid.y;
        mov.u32         %r6, %ctaid.z;
        shl.b32         %r7, %r1, 7;
        or.b32          %r8, %r7, %r2;
        mad.lo.s32      %r9, %r6, %r5, %r4;
        mul.lo.s32      %r10, %r3, %r9;
        shl.b32         %r11, %r10, 7;
        add.s32         %r12, %r8, %r11;
        mul.wide.u32    %rd2, %r12, 4;
        add.s64         %rd3, %rd1, %rd2;
        st.global.u32   [%rd3], %r12;
        ret;
                                        // -- End function
}

```

PR #1884, opened 2025-07-25
KnownGroupSize type should be size_t instead of int

thomasRoglin:
[According to the standard](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#api:info-device-max-work-group-size), the type of groupSize should be size_t. This pull request changes the type of KnownGroupSize variables from int to size_t. 

The max_group_size of any existing device should always be way smaller than UINTMAX+1, but since AdaptiveCpp is currently not verifying that the requested group size is lesser than the max_group_size of the device, this PR should prevent unexpected behavior before AdaptiveCpp transmits the kernel to the device. ( This should also remove this crash  that occur when requesting a group size > INTMAX  :

```
terminate called after throwing an instance of 'std::out_of_range'
  what():  stoi
```
)

### DIFF
--- a/include/hipSYCL/compiler/llvm-to-backend/GlobalSizesFitInI32OptPass.hpp
+++ b/include/hipSYCL/compiler/llvm-to-backend/GlobalSizesFitInI32OptPass.hpp
@@ -19,14 +19,14 @@ namespace compiler {
 
 class GlobalSizesFitInI32OptPass : public llvm::PassInfoMixin<GlobalSizesFitInI32OptPass> {
 public:
-  GlobalSizesFitInI32OptPass(bool GlobalSizesFitInInt, int KnownGroupSizeX = -1,
-                             int KnownGroupSizeY = -1, int KnownGroupSizeZ = -1);
+  GlobalSizesFitInI32OptPass(bool GlobalSizesFitInInt, size_t KnownGroupSizeX = 0,
+                             size_t KnownGroupSizeY = 0, size_t KnownGroupSizeZ = 0);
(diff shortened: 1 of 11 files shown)


illuhad:
I'm not sure understand I understand the motivation behind this change. It's true that the SYCL APIs return `size_t` , but the standard makes no guarantee on how large group sizes that devices support can actually be. So isn't the type of `KnownGroupSize` in the JIT compiler purely an implementation detail?

No backend that we have currently supports a group size group size larger than UINT_MAX, or even USHRT_MAX, and most cap out at around 1024, including the OpenMP backend.

For the error that you see, wouldn't a proper fix be to just add some validation logic for the group size at kernel submission time?

thomasRoglin:
The main motivation for this patch was to standardize the group size type in AdaptiveCpp and prevent any unwanted and undetectable type casting from size_t to int.

But I agree with you, If we want AdaptiveCpp to verify that the group size is legal, then int or size_t will make no difference because no acceptable group size can approach int max.

I'll look into how to implement this verification. Do you want me to revert this commit before ? ( I still think this modification has a purpose just to standardize the type of group size across AdaptiveCpp )

illuhad:
> I'll look into how to implement this verification. Do you want me to revert this commit before ? ( I still think this modification has a purpose just to standardize the type of group size across AdaptiveCpp )

I think I'm a bit concerned about the change from signed to unsigned, and thus denoting 0 as unknown group size instead of -1. Not sure if there are backends where a group size of 0 is legal, but it's less otherworldly than -1 :)
So I think I'd prefer just doing the validation.

PR #1885, opened 2025-08-01
Build LLVMSpirvTranslator in parallel

thomasRoglin:
The `LLVMSpirvTranslator` dependency is currently built sequentially, even though `AdaptiveCpp` is built in parallel.  
This PR enables parallel building of `LLVMSpirvTranslator`.


src/compiler/llvm-to-backend/CMakeLists.txt
```diff
@@ -196,7 +196,7 @@ if(WITH_SSCP_COMPILER)
         -DLLVM_DIR:PATH=${LLVM_DIR}
         -DCMAKE_INSTALL_PREFIX:PATH=${LLVMSPIRV_INSTALLDIR}
         -DCMAKE_CXX_FLAGS:STRING=${CMAKE_CXX_FLAGS}
-      BUILD_COMMAND ${CMAKE_COMMAND} --build <BINARY_DIR> --config Release
+      BUILD_COMMAND ${CMAKE_COMMAND} --build <BINARY_DIR> --config Release --parallel ${CMAKE_BUILD_PARALLEL_LEVEL}
```
illuhad: Could this result in oversubscription? If a system has N cores, and we build AdaptiveCpp with `make -jN`, then wouldn't this spawn N jobs for AdaptiveCpp, and another N for the translator, for up to 2N in total?

thomasRoglin: Yes, I didn't think of that. I don't think it's possible to prevent oversubscription. 
Otherwise, we could do as HemePure does, [here](https://github.com/UCL-CCS/HemePure/blob/ad16f548817c939e9a9e9412a6dd715309161e46/dep/CMakeLists.txt#L53C1-L53C55), and use a cmake variable set to 1 by default, that the user could increase when calling cmake to build llvmspirvtranslator in parallel.

illuhad: Okay, this sounds like a good solution. I suppose if oversubscription is a potential hazard, users may want to fine-tune the number of compilation processes anyway for best performance.

thomasRoglin:
I added the variable `ACPP_SUBPROJECT_PARALLEL_JOBS`. The variable name is generic even though it is only used for building LLVMSpirvTranslator.

illuhad (approved):
Thanks!

PR #1889, opened 2025-08-04
Remove double overloads from SYCL native builtins

carbotaniuman:
These are not in the SYCL standard, and can no longer even be called after the refactoring of the math builtins.

https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_native_precision_math_functions


illuhad (approved):
Thanks!

PR #1887, opened 2025-08-01
Change some LLVM API calls when using LLVM 21

thomasRoglin:
LLVM pre-release 21 is now out (pre-release available [here](https://github.com/llvm/llvm-project/releases/tag/llvmorg-21.1.0-rc1) ) and introduces some API changes.

This PR adds support for (yet to be released)  LLVM21 : 
 -  `setTargetTriple` now takes `llvm::Triple` instead of a string (https://github.com/llvm/llvm-project/pull/129868)  
 -  `Attribute::NoCapture` no longer exists, replace it with `llvm::CaptureInfo::none()`  (https://github.com/llvm/llvm-project/commit/29441e4f5fa5f5c7709f7cf180815ba97f611297)

PR #1892, opened 2025-08-07
[CUDA] Add support for CUDA 13

al42and:
Some runtime functions got removed.

Requires LLVM 22 (and thus #1887).

Release notes: https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html


illuhad (approved):
Thanks!

> SM101 has been renumbered as SM110 from this release.

Should we also take this into account for our sm/ptx version map in `cuda_queue.cpp`?

al42and:
Good catch. The support for sm_88 and sm_110 was added in PTX 9.0, so I'm setting that. Don't think we want to add any automatic target renaming.

illuhad (approved):
Thanks!

PR #1895, opened 2025-08-11
[Windows] Fix GNU-style Clang build by guarding /EHs with MSVC instead of CMAKE_CXX_SIMULATE_ID

blinkfrog:
This PR fixes build failure when using GNU-style clang on Windows.
Previously `/EHs` flags were added based on `CMAKE_CXX_SIMULATE_ID`, which can be "MSVC" even for GNU-style clang. This incorrectly applied MSVC-only flags and caused compile errors.

In the same time, `MSVC` variable is set only when the MSVC driver front-end is in use. This is the correct condition for `/EHs` support and avoids false positives with GNU-style clang.

I've tested on my system with various compilers, adding debug output for the next variables:

```
1. `clang.exe`:

-- MSVC:
-- CMAKE_CXX_COMPILER_ID: Clang
-- CMAKE_CXX_SIMULATE_ID: MSVC


2. `clang-cl.exe`:

-- MSVC: 1
-- CMAKE_CXX_COMPILER_ID: Clang
-- CMAKE_CXX_SIMULATE_ID: MSVC


3. `cl.exe` (MSVC):

-- MSVC: 1
-- CMAKE_CXX_COMPILER_ID: MSVC
-- CMAKE_CXX_SIMULATE_ID:
```

So, checking if `MSVC` is set seems more correct and works correctly in my test builds with different compilers.

(Newer CMake versions also provide `CMAKE_CXX_COMPILER_FRONTEND_VARIANT`, which would be ideal solution, however, we need to support CMake 3.10 as a minimum requirement)

Additional information. I actually was okwith building AdaptiveCpp with `clang-cl` instead of `clang`  (when building it as a plugin for Clang instead of building it as a part of LLVM), however, I discovered, that it links against `OpenMP.lib`, which belongs to MSVC, not to LLVM, and, in runtime, requires `libomp140.dll` instead of LLVM's `libopm.dll`. This was the last straw casing me to make this PR.


illuhad (approved):
Thanks!

PR #1896, opened 2025-08-11
Fix acpp-clang.dll install location when building as Clang plugin on Windows

blinkfrog:
This PR fixes an installation regression introduced when adding support for building AdaptiveCpp as part of LLVM for the SSCP compiler.

Before that change, `acpp-clang.dll` was built as a SHARED library in standalone Clang plugin builds, so CMake treated it as a RUNTIME artifact and installed it into `\bin\`.

After the change, the target was created as a MODULE, which on Windows makes CMake treat the DLL as a LIBRARY artifact and install it into `\lib\`, breaking runtime loading duting compiling unless manually moved to `\bin\`

Fixed by creating the target as SHARED instead of MODULE when building as a Clang plugin (`NOT ACPP_LLVM_COMPONENT`) on Windows.


illuhad (approved):
Thanks!

PR #1902, opened 2025-08-23
Make some fixes so CTS can build

carbotaniuman:
Probably best reviewed commit by commit.

### DIFF
--- a/include/hipSYCL/sycl/libkernel/accessor.hpp
+++ b/include/hipSYCL/sycl/libkernel/accessor.hpp
@@ -1111,7 +1111,25 @@ public:
   size_t get_count() const noexcept
   { return 1; }
 
-  /* void swap(accessor &other); */
+  void swap(accessor &other) {
+    std::swap(*this, other);
+  }
+
+  /* Available only when: (AccessMode != access_mode::atomic &&
+                           AccessMode != access_mode::read && Dimensions == 0) */
+  template<access::mode M = accessmode, int D = dimensions, std::enable_if_t<M != access_mode::atomic && M != access_mode::read && D == 0, bool> = false>
+  const accessor& operator=(const value_type& other) const {
+    *get_pointer() = other;
+    return *this;
+  }
+
+  /* Available only when: (AccessMode != access_mode::atomic &&
+                           AccessMode != access_mode::read && Dimensions == 0) */
+  template<access::mode M = accessmode, int D = dimensions, std::enable_if_t<M != access_mode::atomic && M != access_mode::read && D == 0, bool> = false>
+  const accessor& operator=(value_type&& other) const {
+    *get_pointer() = std::move(other);
+    return *this;
+  }
 
   template <bool IsAllowed = has_size_queries,
             std::enable_if_t<IsAllowed, int> = 0>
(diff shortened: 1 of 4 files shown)


include/hipSYCL/sycl/libkernel/accessor.hpp
```diff
@@ -990,11 +1003,11 @@ class HIPSYCL_EMPTY_BASES accessor
         detail::accessor::conditional_buffer_pointer_storage<
             has_buffer_pointer>{other},
         detail::accessor::conditional_access_range_storage<has_access_range,
-                                                           dimensions>{
-            detail::accessor::access_range<dimensions>{other.get_offset(),
-                                                       other.get_range()}},
+                                                           adj_dimensions>{
```
illuhad, line 1006: Is it correct to use `adj_dimensions` here? In 0-dim case, we shouldn't have to store any access range anyway, therefore, it might be better to always set `has_access_range` to false (or specialize `conditional_access_range_storage` for 0D, no?

include/hipSYCL/sycl/libkernel/accessor.hpp
```diff
@@ -1111,7 +1111,23 @@ class HIPSYCL_EMPTY_BASES accessor
...
+  const accessor& operator=(const value_type& other) const {
+    *(this->_ptr.get()) = other;
+  }
+
+  /* Available only when: (AccessMode != access_mode::atomic &&
+                           AccessMode != access_mode::read && Dimensions == 0) */
+  template<access::mode M = accessmode, int D = dimensions, std::enable_if_t<M != access_mode::atomic && M != access_mode::read && D == 0, bool> = false>
+  const accessor& operator=(value_type&& other) const {
+    *(this->_ptr.get()) = other;
+  }
```
illuhad, line 1132: It might be good to add a test case for this to add to CI. It's often quite easy to break things again and create ambiguities due to all the enable_if conditions, if we add other stuff in the future.

include/hipSYCL/sycl/libkernel/accessor.hpp
carbotaniuman: This is a converting constructor from read-write to read-only so it's called for all dimensions, we just have to avoid `get_offset` and `get_range` as they don't work for the 0-dim case.

illuhad: Okay.

tests/sycl/accessor.cpp
```diff
@@ -213,6 +219,40 @@ BOOST_AUTO_TEST_CASE(accessor_api) {
...
+  (void) (a3 != a3_copy);
+  a0.swap(a0_copy);
+  a1.swap(a1_copy);
+  a2.swap(a2_copy);
+  a3.swap(a3_copy);
+
+  s::buffer<int, 1> buf_single(1);
+  queue.submit([&](s::handler& cgh) {
+    cgh.single_task<class accessor_api_device_accessors_zero_dim>([&](){
+      s::accessor<int, 0> test{buf_single, cgh};
```
illuhad: My guess is that this line is (at least partly) responsible for CI failures. This attempts to create a new accessor inside device code, but this is only supported on the host side. So this line pulls in host code into device code (e.g. it captures the buffer object).
To fix, that line would have to be moved outside the `single_task`, but inside the `submit`.

tests/sycl/accessor.cpp
```diff
@@ -213,6 +219,40 @@ BOOST_AUTO_TEST_CASE(accessor_api) {
...
+  (void) (a2 != a2_copy);
+  (void) (a3 != a3_copy);
+  a0.swap(a0_copy);
+  a1.swap(a1_copy);
+  a2.swap(a2_copy);
+  a3.swap(a3_copy);
+
+  s::buffer<int, 1> buf_single(1);
+  queue.submit([&](s::handler& cgh) {
+    cgh.single_task<class accessor_api_device_accessors_zero_dim>([&](){
```
illuhad: And this lambda must be capture-by-value :)

carbotaniuman:
Ah, finally fixed the build failure.

illuhad (approved):
Thanks!

PR #1903, opened 2025-08-24
Refactor the group algorithms to work in-place

### LINKED ISSUE #1440: joint_exclusive_scan does not work in-place
**Bug summary**

The current implementation of `joint_exclusive_scan` (at least HIP-like; SSCP does not support them yet) does not seem to support in-place operations, even though the standard requires it ("Note that `first` may be equal to `result`.")

Also, it allocates a `__shared__` scratch storage for the operation (inside of `__hipsycl_inclusive_scan_over_group`) even if the output is `__shared__` too and can safely be used for scratch. Not a bug, just inefficiency

**To Reproduce**

```cpp
#include <sycl/sycl.hpp>

template <int workGroupSize, int nElements>
auto exclusivePrefixSumInplace(sycl::handler &cgh, int *gm_data) {
  static_assert(nElements % workGroupSize == 0,
                "This simple scan kernel does not handle padding");
  return [=](sycl::nd_item<1> itemIdx) {
    const int tid = itemIdx.get_local_id(0);
    const sycl::group<1> workGroup = itemIdx.get_group();
    sycl::joint_exclusive_scan(workGroup, gm_data, gm_data + nElements, gm_data,
                               0, sycl::plus<int>{});
  };
}

template <int workGroupSize, int nElements>
auto exclusivePrefixSumOutOfPlace(sycl::handler &cgh, int *gm_data) {
  static_assert(nElements % workGroupSize == 0,
                "This simple scan kernel does not handle padding");
  sycl::local_accessor<int, 1> sm_localBuf(nElements, cgh);
  return [=](sycl::nd_item<1> itemIdx) {
    const int tid = itemIdx.get_local_id(0);
    const sycl::group<1> workGroup = itemIdx.get_group();
    int *sm_localBufPtr = sm_localBuf.get_pointer();
    sycl::joint_exclusive_scan(workGroup, gm_data, gm_data + nElements,
                               sm_localBufPtr, 0, sycl::plus<int>{});
    sycl::group_barrier(workGroup);
    for (int elem = tid; elem < nElements; elem += workGroupSize) {
      gm_data[elem] = sm_localBufPtr[elem];
    }
  };
}

int main() {
  for (const auto &device : sycl::device::get_devices()) {
    // Creating SYCL queue
    sycl::queue q(device, {sycl::property::queue::in_order()});

    std::cout << "Running on " << device.get_info<sycl::info::device::name>()
              << "\n";

    constexpr size_t size = 8192;
    constexpr int wgSize = 256;
    int *dataD = sycl::malloc_device<int>(size, q);
    std::vector<int> dataH(size);

    std::cout << "Running inplace..." << std::endl;
    {
      q.fill<int>(dataD, 1, size);
        sycl::nd_range<1> range{wgSize, wgSize};
        cgh.parallel_for(range,
                         exclusivePrefixSumInplace<wgSize, size>(cgh, dataD));
      });
      q.copy<int>(dataD, dataH.data(), size).wait();

      int numFail = 0;
      for (size_t i = 0; i < size; i++) {
        numFail += (dataH[i] != i);
      }
      std::cout << "Got " << numFail << " failues!" << std::endl;
    }
    std::cout << "Running with a temp buffer..." << std::endl;
    {
      q.fill<int>(dataD, 1, size);
        sycl::nd_range<1> range{wgSize, wgSize};
        cgh.parallel_for(
            range, exclusivePrefixSumOutOfPlace<wgSize, size>(cgh, dataD));
      });
      q.copy<int>(dataD, dataH.data(), size).wait();

      int numFail = 0;
      for (size_t i = 0; i < size; i++) {
        numFail += (dataH[i] != i);
      }
      std::cout << "Got " << numFail << " failues!" << std::endl;
    }
  }
  return 0;
}
```

```console
$ acpp -O3 --acpp-targets='omp;cuda:sm_86;hip:gfx1034' test_scan.cpp && ACPP_VISIBILITY_MASK="cuda;hip" ./a.out 
[a bunch of "loop not vectorized" warnings]
4 warnings generated when compiling for host.
Running on NVIDIA GeForce RTX 3060
Running inplace...
Got 8191 failues!
Running with a temp buffer...
Got 0 failues!
Running on hipSYCL OpenMP host device
Running inplace...
Got 8191 failues!
Running with a temp buffer...
Got 0 failues!
Running on AMD Radeon RX 6400
Running inplace...
Got 8191 failues!
Running with a temp buffer...
Got 0 failues!
```

**Expected behavior**

**Describe your setup**

```
$ acpp --acpp-version
acpp [AdaptiveCpp compilation driver], Copyright (C) 2018-2023 Aksel Alpay and the AdaptiveCpp project
  AdaptiveCpp version: 24.02.0+git.0359cac9.20240401.branch.develop.dirty
  Installation root: /home/aland/local
  Plugin LLVM version: 17, can accelerate CPU: True
  Available runtime backends:
     librt-backend-cuda.so
     librt-backend-omp.so
     librt-backend-hip.so
     librt-backend-ocl.so


Full configuration [can be overridden using environment variables or command line arguments]:
    version-major: 24
    version-minor: 02
    version-patch: 0
    version-suffix: +git.0359cac9.20240401.branch.develop.dirty
    plugin-llvm-version-major: 17
    plugin-with-cpu-acceleration: true
    default-clang: /usr/lib/llvm-17/bin/clang++
    default-targets: generic
    default-cpu-cxx: /usr/bin/clang++-17
    default-rocm-path: /opt/rocm
    default-use-bootstrap-mode: false
    default-is-dryrun: false
    default-use-accelerated-cpu: true
    default-clang-include-path: /usr/lib/llvm-17/lib/clang/17/include/..
    default-sequential-link-line: -L/usr/lib/x86_64-linux-gnu -lboost_context -lboost_fiber -Wl,-rpath=/usr/lib/x86_64-linux-gnu
    default-sequential-cxx-flags: -I/usr/include -D_ENABLE_EXTENDED_ALIGNED_STORAGE
    default-omp-link-line: -L/usr/lib/x86_64-linux-gnu -lboost_context -lboost_fiber -Wl,-rpath=/usr/lib/x86_64-linux-gnu -fopenmp
    default-omp-cxx-flags: -I/usr/include -fopenmp -D_ENABLE_EXTENDED_ALIGNED_STORAGE
    default-is-explicit-multipass: false
    default-save-temps: false
    default-rocm-link-line: -Wl,-rpath=$HIPSYCL_ROCM_PATH/lib -Wl,-rpath=$HIPSYCL_ROCM_PATH/hip/lib -L/opt/rocm/lib -L/opt/rocm/hip/lib -lamdhip64
    default-rocm-cxx-flags: -isystem $HIPSYCL_PATH/include/AdaptiveCpp/hipSYCL/std/hiplike -isystem /usr/lib/llvm-17/lib/clang/17/include/.. -U__FLOAT128__ -U__SIZEOF_FLOAT128__ -I$HIPSYCL_ROCM_PATH/include -I$HIPSYCL_ROCM_PATH/include --rocm-device-lib-path=$HIPSYCL_ROCM_PATH/amdgcn/bitcode --rocm-path=$HIPSYCL_ROCM_PATH -fhip-new-launch-api -mllvm -amdgpu-early-inline-all=true -mllvm -amdgpu-function-calls=false -D__HIP_ROCclr__
    default-nvcxx: (unconfigured)
    default-cuda-path: /usr/local/cuda
    default-cuda-link-line: -Wl,-rpath=$HIPSYCL_CUDA_LIB_PATH -L$HIPSYCL_CUDA_LIB_PATH -lcudart
    default-cuda-cxx-flags: -U__FLOAT128__ -U__SIZEOF_FLOAT128__ -isystem $HIPSYCL_PATH/include/AdaptiveCpp/hipSYCL/std/hiplike
```

- ROCm 6.0.0, CUDA 12.3, Clang 17, Ubuntu 22.04

### LINKED ISSUE #1899: `sycl::joint_exclusive_scan` does not produce the correct result for a two-element array when run with a CPU queue using the generic compiler
**Bug summary**
`sycl::joint_exclusive_scan` does not produce the correct result for a two-element array when run with a CPU queue using the generic compiler. When executed on an array containing [x,0], the result is [0,0] instead of the expected [0,x]. I could only reproduce this behaviour when `first` and `result` are the same pointer.

**To Reproduce**
```cpp
constexpr auto length = 2;

void do_exclusive_scan(uint32_t* const data, sycl::queue& queue) {
	const auto workGroupSize = queue.get_device().get_info<sycl::info::device::max_work_group_size>();

	queue.parallel_for(sycl::nd_range<1>{workGroupSize,workGroupSize}, [=](const sycl::nd_item<1>& item) {
		sycl::joint_exclusive_scan(
			item.get_group(),
			data,
			data + length,
			data,
			uint32_t{0},
			sycl::plus<uint32_t>{}
		);
	});
}

void do_stdlib_exclusive_scan(uint32_t* const data) {
	std::exclusive_scan(
		data,
		data + length,
		data,
		0,
		std::plus<uint32_t>{}
	);
}

void print(const uint32_t* const data) {
	for (auto i = 0; i < length; ++i) {
		std::cout << data[i] << " ";
	}
	std::cout << "\n";
}

int main() {
	sycl::queue host_queue{sycl::cpu_selector_v};
	sycl::queue device_queue{sycl::gpu_selector_v};

	auto* const host_result = sycl::malloc_host<uint32_t>(length, host_queue);
	auto* const device_result = sycl::malloc_host<uint32_t>(length, device_queue);
	auto* const stdlib_result = new uint32_t[length];
	host_result[0] = 42;
	device_result[0] = 42;
	stdlib_result[0] = 42;

	do_exclusive_scan(host_result, host_queue);
	do_exclusive_scan(device_result, device_queue);
	do_stdlib_exclusive_scan(stdlib_result);

	host_queue.wait();
	device_queue.wait();

	std::cout << "host_result: "; print(host_result);
	std::cout << "device_result: "; print(device_result);
	std::cout << "stdlib_result: "; print(stdlib_result);

	sycl::free(host_result, host_queue);
	sycl::free(device_result, device_queue);
	delete[] stdlib_result;

	return 0;
}
```

**Expected behavior**
All `*_result` arrays contain [0,42] in the end. Actual encountered behaviour:
```
/tmp/exclusive_scan/cmake-build-debug-sycllod/exclusive_scan
host_result: 0 0 
device_result: 0 42 
stdlib_result: 0 42 

Process finished with exit code 0
```

**Describe your setup**
AdaptiveCpp 65e010622e37b25d75b70d41776ebb2685b2e4a8 (develop)
ROCm Docker containing Ubuntu 24.04 with copy-pasted setup from INSTALLATION.MD
ROCm 6.4.3
Clang 18
generic compilation flow for: Radeon 780M and Ryzen 7 7840HS

### LINKED ISSUE #1900: `sycl::joint_exclusive_scan` breaks when there are more data items than work items
**Bug summary**
`sycl::joint_exclusive_scan` breaks when there are more data items than work items in the work group. From my understanding of the example in the SYCL docs (https://github.khronos.org/SYCL_Reference/iface/group-algorithms-library.html#example), it is valid to process more data items than work items allocated.

**To Reproduce**
```cpp
#include <iostream>
#include <sycl/sycl.hpp>

constexpr auto numWorkItems = 5;
constexpr auto dataSize = 10;

void print(const uint32_t* const data) {
	for (auto i = 0; i < dataSize; i++) {
		std::cout << data[i] << ", ";
	}
	std::cout << "\n";
}

int main() {

	auto* const data = sycl::malloc_host<uint32_t>(dataSize, queue);
	queue.fill(data, uint32_t{1}, dataSize);

	queue.wait();

	std::cout << "Pre-scan Acpp: "; print(data);

	std::vector<uint32_t> test(dataSize);
	std::ranges::fill(test, 1);

	std::cout << "Pre-scan Stdlib: "; print(test.data());

	std::exclusive_scan(
		test.begin(),
		test.end(),
		test.begin(),
		0
	);

	queue.parallel_for(sycl::nd_range<1>{numWorkItems, numWorkItems},
		[=](const sycl::nd_item<1>& item) {
			sycl::joint_exclusive_scan(
				item.get_group(),
				data,
				data + dataSize,
				data,
				0,
				sycl::plus<uint32_t>{}
			);
		}
	);

	queue.wait();

	std::cout << "Post-scan Acpp: "; print(data);
	std::cout << "Post-scan Stdlib: "; print(test.data());

	sycl::free(data, queue);

	return 0;
}
```

**Expected behavior**
The same as the C++ stdlib. Actual behaviour:
```
/tmp/exclusive_scan/cmake-build-debug-sycllod/exclusive_scan
Pre-scan Acpp: 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 
Pre-scan Stdlib: 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 
Post-scan Acpp: 0, 1, 2, 3, 4, 5, 10, 11, 12, 13, 
Post-scan Stdlib: 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 

Process finished with exit code 0
```
This is observable on both a GPU queue, as well as a CPU queue.

**Describe your setup**
AdaptiveCpp 65e010622e37b25d75b70d41776ebb2685b2e4a8 (develop)
ROCm Docker containing Ubuntu 24.04 with copy-pasted setup from INSTALLATION.MD
ROCm 6.4.3
Clang 18
generic compilation flow for: Radeon RX 9070 and Ryzen 7 5800X

carbotaniuman:
This refactors the group algorithms, mainly `joint_exclusive_scan`, to work in-place. This change also fixes `inclusive_scan_over_group` signature to take the init parameter last, which is a breaking change.

Please let me know if I should add a compatibility wrapper for the old parameter order.

Fixes #1440, #1899, #1900

### DIFF
--- a/include/hipSYCL/sycl/libkernel/generic/hiplike/group_functions.hpp
+++ b/include/hipSYCL/sycl/libkernel/generic/hiplike/group_functions.hpp
@@ -218,9 +218,9 @@ __device__ T __acpp_shift_group_right(
   scratch[lid] = x;
   __acpp_group_barrier(g);
 
-  // checking for both larger and smaller in case 'group<Dim>::linear_id_type' is not unsigned
-  if (target_lid > g.get_local_range().size() || target_lid < 0)
+  if (target_lid > g.get_local_range().size()) {
     target_lid = 0;
+  }
 
   return scratch[target_lid];
 }
@@ -473,9 +473,9 @@ __device__ T __acpp_permute_group_by_xor(
   scratch[lid] = x;
   __acpp_group_barrier(g);
 
-  // checking for both larger and smaller in case 'group<Dim>::linear_id_type' is not unsigned
-  if (target_lid > g.get_local_range().size() || target_lid < 0)
+  if (target_lid > g.get_local_range().size()) {
     target_lid = 0;
+  }
 
   return scratch[target_lid];
 }
@@ -502,9 +502,9 @@ __device__ T __acpp_select_from_group(
   scratch[lid] = x;
   __acpp_group_barrier(g);
 
-  // checking for both larger and smaller in case 'group<Dim>::linear_id_type' is not unsigned
-  if (target_lid > g.get_local_range().size() || target_lid < 0)
+  if (target_lid > g.get_local_range().size()) {
     target_lid = 0;
+  }
 
   return scratch[target_lid];
 }
--- a/include/hipSYCL/sycl/libkernel/generic/hiplike/group_functions.hpp
+++ b/include/hipSYCL/sycl/libkernel/generic/hiplike/group_functions.hpp
@@ -15,6 +15,7 @@
 #include "../../backend.hpp"
 #include "../../detail/data_layout.hpp"
 #include "../../detail/thread_hierarchy.hpp"
+#include "../../functional.hpp"
 #include "../../id.hpp"
 #include "../../sub_group.hpp"
 #include "../../vec.hpp"
@@ -177,6 +178,58 @@ bool __acpp_leader_all_of(Group g, T *first, T *last, Predicate pred) {
 }
 }
 
+// shift_left
+template <int Dim, typename T>
+__device__ T __acpp_shift_group_left(
+    group<Dim> g, T x, typename group<Dim>::linear_id_type delta = 1) {
+  __shared__ std::aligned_storage_t<sizeof(T) * detail::max_group_size, alignof(T)>
+             scratch_storage;
+  T *        scratch = reinterpret_cast<T *>(&scratch_storage);
+
+  typename group<Dim>::linear_id_type lid        = g.get_local_linear_id();
+  typename group<Dim>::linear_id_type target_lid = lid + delta;
+
+  scratch[lid] = x;
+  __acpp_group_barrier(g);
+
+  if (target_lid > g.get_local_range().size())
+    target_lid = 0;
+
+  return scratch[target_lid];
+}
+
+template <typename T>
+__device__ T __acpp_shift_group_left(
+    sub_group g, T x, typename sub_group::linear_id_type delta = 1) {
+  return detail::__acpp_shuffle_down_impl(x, delta);
+}
+
+// shift_right
+template <int Dim, typename T>
+__device__ T __acpp_shift_group_right(
+    group<Dim> g, T x, typename group<Dim>::linear_id_type delta = 1) {
+  __shared__ std::aligned_storage_t<sizeof(T) * detail::max_group_size, alignof(T)>
+             scratch_storage;
+  T *        scratch = reinterpret_cast<T *>(&scratch_storage);
+
+  typename group<Dim>::linear_id_type lid        = g.get_local_linear_id();
+  typename group<Dim>::linear_id_type target_lid = lid - delta;
+
+  scratch[lid] = x;
+  __acpp_group_barrier(g);
+
+  // checking for both larger and smaller in case 'group<Dim>::linear_id_type' is not unsigned
+  if (target_lid > g.get_local_range().size() || target_lid < 0)
+    target_lid = 0;
+
+  return scratch[target_lid];
+}
+
+template <typename T>
+__device__ T __acpp_shift_group_right(
+    sub_group g, T x, typename sub_group::linear_id_type delta = 1) {
+  return detail::__acpp_shuffle_up_impl(x, delta);
+}
 
 // none_of
 template<int Dim>
@@ -352,102 +405,10 @@ __device__ T __acpp_inclusive_scan_over_group(group<Dim> g, T x,
 
 template <typename Group, typename V, typename T, typename BinaryOperation,
           std::enable_if_t<is_group_v<std::decay_t<Group>>, bool> = true>
-__device__ T __acpp_inclusive_scan_over_group(Group g, V x, T init,
-                                                 BinaryOperation binary_op) {
+__device__ T __acpp_inclusive_scan_over_group(Group g, V x,
+                                                 BinaryOperation binary_op, T init) {
   T scan = __acpp_inclusive_scan_over_group(g, T{x}, binary_op);
   return binary_op(scan, init);
-}
-
-template <typename Group, typename InPtr, typename OutPtr, typename T,
-          typename BinaryOperation,
-          std::enable_if_t<is_group_v<std::decay_t<Group>>, bool> = true>
-__device__ OutPtr __acpp_joint_inclusive_scan(Group g, InPtr first,
-                                                 InPtr last, OutPtr result,
-                                                 BinaryOperation binary_op,
-                                                 T init) {
-  using OutT = std::remove_reference_t<decltype(*result)>;
-
-  auto         lid          = g.get_local_linear_id();
-  auto         wid          = lid / __acpp_warp_size;
-  size_t       lrange       = g.get_local_range().size();
-  const size_t num_elements = last - first;
-  const size_t iterations   = (num_elements + lrange - 1) / lrange;
-
-  const size_t warp_size =
-      (wid == lrange / __acpp_warp_size && lrange % __acpp_warp_size != 0) ? lrange % __acpp_warp_size : __acpp_warp_size;
-
-  size_t offset     = lid;
-  OutT      carry_over = init;
-  OutT      local_x;
-
-  for (int i = 0; i < iterations; ++i) {
-    const size_t offset = lid + i * lrange;
-    local_x = (offset < num_elements) ? first[offset] : OutT{};
-    local_x =
-        __acpp_inclusive_scan_over_group(g, local_x, carry_over, binary_op);
-
-    if (offset < num_elements)
-      result[offset] = local_x;
-
-    carry_over = __acpp_group_broadcast(g, local_x, lrange - 1);
-  }
-
-  return result + num_elements;
-}
-
-template <typename Group, typename InPtr, typename OutPtr,
-          typename BinaryOperation,
-          std::enable_if_t<is_group_v<std::decay_t<Group>>, bool> = true>
-__device__ OutPtr __acpp_joint_inclusive_scan(Group g, InPtr first,
-                                                 InPtr last, OutPtr result,
-                                                 BinaryOperation binary_op) {
-  using OutT = std::remove_reference_t<decltype(*result)>;
-
-  auto         lid          = g.get_local_linear_id();
-  auto         wid          = lid / __acpp_warp_size;
-  size_t       lrange       = g.get_local_range().size();
-  const size_t num_elements = last - first;
-  const size_t iterations   = (num_elements + lrange - 1) / lrange;
-
-  OutT carry_over;
-  OutT local_x;
-
-  for (int i = 0; i < iterations; ++i) {
-    const size_t offset = lid + i * lrange;
-    local_x             = (offset < num_elements) ? first[offset] : OutT{};
-    if (i > 0) {
-      local_x =
-          __acpp_inclusive_scan_over_group(g, local_x, carry_over, binary_op);
-    } else {
-      local_x = __acpp_inclusive_scan_over_group(g, local_x, binary_op);
-    }
-
-    if (offset < num_elements)
-      result[offset] = local_x;
-
-    carry_over = __acpp_group_broadcast(g, local_x, lrange - 1);
-  }
-
-  return result + num_elements;
-}
-
-namespace detail { // until scoped-parallelism can be detected
-template <typename Group, typename V, typename T, typename BinaryOperation,
-          std::enable_if_t<is_group_v<std::decay_t<Group>>, bool> = true>
-__device__ T *
-__acpp_leader_inclusive_scan(Group g, V *first, V *last, T *result,
-                                BinaryOperation binary_op, T init) {
-  return __acpp_joint_inclusive_scan(g, first, last, result, binary_op,
-                                        init);
-}
-
-template<typename Group, typename V, typename T, typename BinaryOperation,
-          std::enable_if_t<is_group_v<std::decay_t<Group>>, bool> = true>
-__device__
-T *__acpp_leader_inclusive_scan(Group g, V *first, V *last, T *result,
-                         BinaryOperation binary_op) {
-  return __acpp_joint_inclusive_scan(g, first, last, result, binary_op);
-}
 }
 
 // exclusive_scan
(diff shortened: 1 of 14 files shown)


illuhad:
> Please let me know if I should add a compatibility wrapper for the old parameter order.

I don't think that's necessary, but it looks like there are some CI issues - presumably because test cases still use the old interface?

carbotaniuman:
I've finished testing this on CUDA + host on both SSCP and SMCP, using the in-place changes.

I've also chosen a new implementation option that hopefully is more maintainable and also maybe a bit faster on host. This is split into 2 commits now - the first one is some changes to `__acpp_if_target_{device,host}` to make it always an if statement.

The major change compared to previously is that the implementation of `joint_{exclusive,inclusive}_scan` is now in common instead of backend specific code. I've also added a sequential host implementation using `__acpp_if_target_host` to not regress SMCP host and probably give a performance increase to SSCP host.

Just also realized that instead of the else I could've used `__acpp_if_target_device` instead. Oh well, let me know if should back that change out.

illuhad:
Thanks for tackling this! (and sorry that I'm so far behind reviewing PRs at the moment...vacation and travel will do this :( )

> I've also chosen a new implementation option that hopefully is more maintainable and also maybe a bit faster on host. This is split into 2 commits now - the first one is some changes to __acpp_if_target_{device,host} to make it always an if statement.

This I'm not so sure about, because it would be a breaking change from current behavior in the SMCP case. It's not good style, but in theory people could be using `__acpp_if_target_host` to e.g. make data layout changes in a class. As long as the code is only compiled with SMCP, this would work fine.

Better client code would be using the macros for this (`ACPP_LIBKERNEL_IS_DEVICE_PASS_*`), and I agree that `__acpp_if_target` already kind of hints at control flow...

I'd be curious about some user opinions here, e.g. @al42and ?

al42and:
> I'd be curious about some user opinions here, e.g. @al42and ?

To clarify: is the question about making some (ab)uses of `__acpp_if_target*` no longer possible?

For us, we're not currently using `__acpp_if_target*` anywhere in `main` now. I don't think I would have attempted to use them in a non-_if_ context; agree that the way it's documented does not actively invite macro-like uses. Although, having different data layout for the same type in different compiler passes (as opposed to different builds) is something I'd prefer to avoid anyway, for my own sanity.

A quick GitHub [code search](https://github.com/search?q=__acpp_if_target_+NOT+repo%3AAdaptiveCpp%2FAdaptiveCpp+NOT+is%3Afork&type=code) does not bring up any "dangerous" uses either.

include/hipSYCL/sycl/libkernel/generic/hiplike/group_functions.hpp
```diff
@@ -177,6 +178,58 @@ bool __acpp_leader_all_of(Group g, T *first, T *last, Predicate pred) {
...
+  T *        scratch = reinterpret_cast<T *>(&scratch_storage);
+
+  typename group<Dim>::linear_id_type lid        = g.get_local_linear_id();
+  typename group<Dim>::linear_id_type target_lid = lid - delta;
+
+  scratch[lid] = x;
+  __acpp_group_barrier(g);
+
+  // checking for both larger and smaller in case 'group<Dim>::linear_id_type' is not unsigned
+  if (target_lid > g.get_local_range().size() || target_lid < 0)
```
illuhad: I don't know if this a new issue, or if this was already in the the original code: `target_lid` is of type `linear_id_type` which is typically unsigned, so the second part of the condition won't ever be triggered.

include/hipSYCL/sycl/libkernel/group_functions.hpp
```diff
@@ -310,6 +270,110 @@ T reduce_over_group(Group g, V x, T init, BinaryOperation binary_op) {
...
+
+  __acpp_if_target_host(
+    if (g.leader()) {
+      for (size_t i = 0; i < num_elements; i++) {
+        T next = first[i];
+        result[i] = carry_over;
+        carry_over = binary_op(carry_over, next);
+      }
+    }
+    group_barrier(g);
```
illuhad, line 293: Not necessary now, but we might also want to adopt a similar approach if we are in the device path, but the device happens to be the CPU device (this can happen if we are in the `generic` JIT compilation target, but we end up JITting for host). In that case, doing a big sequential for loop *may* be faster than an approach based on collective functions. IIRC this is also what we do for the scan in the `algorithms` library.

include/hipSYCL/sycl/libkernel/sscp/group_functions.hpp
```diff
@@ -285,20 +415,11 @@ struct sscp_binary_operation {};
...
 HIPSYCL_SSCP_MAP_GROUP_BINARY_OPERATION(sycl::minimum,
                                         __acpp_sscp_algorithm_op::min)
 HIPSYCL_SSCP_MAP_GROUP_BINARY_OPERATION(sycl::maximum,
                                         __acpp_sscp_algorithm_op::max)
-HIPSYCL_SSCP_MAP_GROUP_BINARY_OPERATION(sycl::bit_and,
-                                        __acpp_sscp_algorithm_op::bit_and)
-HIPSYCL_SSCP_MAP_GROUP_BINARY_OPERATION(sycl::bit_or,
-                                        __acpp_sscp_algorithm_op::bit_or)
-HIPSYCL_SSCP_MAP_GROUP_BINARY_OPERATION(sycl::bit_xor,
-                                        __acpp_sscp_algorithm_op::bit_xor)
```
illuhad, line 301: Can we add a comment that the code here assumes that bit_xor, bit_or, bit_and, plus, multiplies are aliases for `std::` functionals, and thus already handled by the code below? Otherwise someone reading the code might be wondering why those are missing.

illuhad (commented):
Thanks - great changes. Only some small comments :)

include/hipSYCL/sycl/libkernel/group_functions.hpp
carbotaniuman: Is there a way to ask the SSCP infrastructure if we're running on a CPU device?

carbotaniuman:
Alright, addressed review comments.

include/hipSYCL/sycl/libkernel/group_functions.hpp
illuhad: Yes, see https://github.com/AdaptiveCpp/AdaptiveCpp/blob/develop/doc/extensions.md#acpp_ext_jit_compile_if

An example usage that checks for CPU can be found in our algorithm library:
https://github.com/AdaptiveCpp/AdaptiveCpp/blob/75d5608939ffd3f5ed3126de6055fde84854593d/include/hipSYCL/algorithms/util/memory_streaming.hpp#L69

note that this does not handle the case where we JIT-compile for OpenCL, but the OpenCL device happens to be a CPU. This too could be checked using the `target_is_cpu` JIT reflection query from the linked extension. However, it's not always clear whether OpenCL CPU devices should better be treated as CPU-like or GPU-like, because different OpenCL implementations may behave differently. That's why e.g. the linked example does not currently special-case this.

illuhad:
There are currently some general CI issues, however, Windows CI reports an issue during JIT that might be related to this change:

```
 LLVM ERROR: Associative COMDAT symbol '?identity@?1???$joint_exclusive_scan@U?$group@$00@sycl@hipsycl@@PEAV?$vec@H$01V?$vec_storage@H$01@detail@sycl@hipsycl@@@23@PEAV423@V423@U?$plus@V?$vec@H$01V?$vec_storage@H$01@detail@sycl@hipsycl@@@sycl@hipsycl@@@std@@$00@sycl@hipsycl@@YAPEAV?$vec@H$01V?$vec_storage@H$01@detail@sycl@hipsycl@@@12@U?$group@$00@12@PEAV312@11V312@U?$plus@V?$vec@H$01V?$vec_storage@H$01@detail@sycl@hipsycl@@@sycl@hipsycl@@@std@@@Z@4V312@B' does not exist.
 ```

Perhaps there needs to be some Windows-style dllexport or something like that somewhere?

File: include/hipSYCL/sycl/libkernel/group_functions.hpp [hunk 4]
```diff
@@ -310,6 +270,110 @@ T reduce_over_group(Group g, V x, T init, BinaryOperation binary_op) {
...
 }
 
+// exclusive_scan
+template <typename Group, typename InPtr, typename OutPtr, typename T,
+          typename BinaryOperation,
+          std::enable_if_t<is_group_v<std::decay_t<Group>>, bool> = true>
+HIPSYCL_BUILTIN OutPtr joint_exclusive_scan(Group g, InPtr first, InPtr last, OutPtr result,
+                                                   T init, BinaryOperation binary_op) {
+  static_assert(std::is_same_v<decltype(binary_op(init, *first)), T>);
+  static constexpr auto identity = known_identity_v<BinaryOperation, T>;
```
illuhad: The failing symbol in Windows CI demangles to
```
class hipsycl::sycl::vec<int,2,class hipsycl::sycl::detail::vec_storage<int,2> > const `class hipsycl::sycl::vec<int,2,class hipsycl::sycl::detail::vec_storage<int,2> > * __ptr64 hipsycl::sycl::joint_exclusive_scan<struct hipsycl::sycl::group<1>,class hipsycl::sycl::vec<int,2,class hipsycl::sycl::detail::vec_storage<int,2> > * __ptr64,class hipsycl::sycl::vec<int,2,class hipsycl::sycl::detail::vec_storage<int,2> > * __ptr64,class hipsycl::sycl::vec<int,2,class hipsycl::sycl::detail::vec_storage<int,2> >,struct std::plus<class hipsycl::sycl::vec<int,2,class hipsycl::sycl::detail::vec_storage<int,2> > >,1>(struct hipsycl::sycl::group<1>,class hipsycl::sycl::vec<int,2,class hipsycl::sycl::detail::vec_storage<int,2> > * __ptr64,class hipsycl::sycl::vec<int,2,class hipsycl::sycl::detail::vec_storage<int,2> > * __ptr64,class hipsycl::sycl::vec<int,2,class hipsycl::sycl::detail::vec_storage<int,2> > * __ptr64,class hipsycl::sycl::vec<int,2,class hipsycl::sycl::detail::vec_storage<int,2> >,struct std::plus<class hipsycl::sycl::vec<int,2,class hipsycl::sycl::detail::vec_storage<int,2> > >)'::`2'::identity
```

So it looks like this line triggers it, when the scan is instantiated with int2 vector.

@fodinabor do you have an idea perhaps why this might cause "associative comdat symbol not found" on Windows for the `identity` defined here?

carbotaniuman:
I think it was the `static` in the `static constexpr`. I naively would've assumed the compiler wouldn't need to generate the variable, but it seems like just removing that fixed it.

illuhad:
> I think it was the static in the static constexpr. I naively would've assumed the compiler wouldn't need to generate the variable, but it seems like just removing that fixed it.

Awesome! I agree though that this should work. Even if a global is generated, AdaptiveCpp should be able to handle it. I'll open an issue to investigate this independently from this PR.

illuhad (approved):
Thanks! :)

PR #1904, opened 2025-08-25
Make Markdown bullet points render correctly in Material for MkDocs

marcosolanki:
Material for MkDocs appears to be more sensitive to empty lines between body text and lists than GitHub-flavoured Markdown. In particular, it will not correctly render lists when there is no empty line separating them from the body text immediately preceding them. This pull request adds empty lines to the Markdown documentation at appropriate positions to correct the rendering of previously incorrectly rendered lists in the Material for MkDocs-based documentation pages (see below for an example).

---

***Top:** Part of the compilation.md file rendered with Material for MkDocs prior to this pull request.*
***Bottom:** The same part of the compilation.md file rendered with Material for MkDocs after incorporation of this pull request.*

---

### DIFF
--- a/doc/accessor-variants.md
+++ b/doc/accessor-variants.md
@@ -3,6 +3,7 @@
 ## Terminology
 
 In the following, we will use the following terminology:
+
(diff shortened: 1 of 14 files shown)


tdavidcl (approved):
I don't know what is happening with the CI but this is fine for me.

illuhad (approved):
Thanks!

PR #1907, opened 2025-09-04
[stdpar] Add mqs infrastructure

illuhad:
This PR adds the multi-queue scheduling (mqs) infrastructure to stdpar, which allows automatic distribution of stdpar kernels across multiple queues or multiple devices from the same backend.

This makes AdaptiveCpp the first stdpar compiler able to automatically leverage multiple devices from a single thread in the stdpar model.


MarkusBuettner:
How does this relate to the multi-device queues that are already present? (https://github.com/AdaptiveCpp/AdaptiveCpp/blob/develop/doc/multi-device-queue.md)
It seems to me that there should be some overlap in terms of functionality/scheduling

illuhad:
> How does this relate to the multi-device queues that are already present? (https://github.com/AdaptiveCpp/AdaptiveCpp/blob/develop/doc/multi-device-queue.md)
> It seems to me that there should be some overlap in terms of functionality/scheduling

Not really the same thing.

Multi-device queues were an early API experiment that didn't really work out because the SYCL model assumes fundamentally that a queue is only associated with a single device. Lots of things break if you don't do that. Because of that, multi-device queues were never given a scheduler that actually does something more than trivial round-robin scheduling. We should remove it in the future, IMO. The MQS scheduler is much more advanced.

At a higher level, multi-device queues target SYCL queues, while MQS targets stdpar. Both have very different scheduling demands. In SYCL, you can have lengthy, complex task graphs.

In stdpar, even with AdaptiveCpp synchronization elision, you typically end up with smallish mini-DAGs. You'd construct a scheduler very differently because of that. Simultaneously, the need for multi-queue and multi-device scheduling for stdpar is even more pressing since there's no explicit way in this model to make use of multiple devices, or even e.g. multiple CUDA streams on a single device. MQS solves this gap.
MQS also is designed around specifics of the stdpar model, for example, because stdpar relies exclusively on USM, it relies on compiler support to prove that kernels are free of indirect access, and on the global allocation tracking that stdpar has which the SYCL layer currently only has as an optional feature (`ACPP_ALLOCATION_TRACKING=1`).

PR #1910, opened 2025-09-13
Change invalid CMake `INTEGER` type to regex-guarded `STRING`

marcosolanki:
I do not believe `INTEGER` is a valid CMake type, but instead is automatically reinterpreted as a `STRING` type. This causes CMake warnings of the form:

```
CMake Warning (dev) at CMakeLists.txt:97 (set):
```

To address this, I have changed `ACPP_SUBPROJECT_PARALLEL_JOBS` to be set to a `STRING` type instead. To guard against values of `ACPP_SUBPROJECT_PARALLEL_JOBS`, which are not positive integers, I have added a regex-based check. Notably, this check will reject values of `ACPP_SUBPROJECT_PARALLEL_JOBS` with leading zeros (e.g. as in `01`), but I do not think this should be a problem in any real-world scenarios.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,7 +94,10 @@ endif()
 
 set(ACPP_EXPERIMENTAL_LLVM OFF CACHE BOOL "Whether to allow building AdaptiveCpp against unsupported LLVM versions")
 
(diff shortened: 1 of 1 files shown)


thomasRoglin:
My mistake, I assumed that `INTEGER` was a CMake valid type and didn't see the warning. Thanks for noticing and fixing it.

illuhad (approved):
Thanks!

PR #1911, opened 2025-09-13
Refactor decoupled lookback scan

### LINKED ISSUE #1908: [Algorithms] Data race in In-Place Exclusive Scan Algorithm
**Bug summary**
When using `adaptivecpp::algorithms::exclusive_scan` i get on all backends either a segfault, either a corrupted output.

Here is an example (index, acpp output, expected one)
```
Debug: 16374 81309 81309                                                                                                                     [tmps][rank=0] 
Debug: 16375 81310 81310                                                                                                                     [tmps][rank=0] 
Debug: 16376 81317 81317                                                                                                                     [tmps][rank=0] 
Debug: 16377 81321 81321                                                                                                                     [tmps][rank=0] 
Debug: 16378 81330 81330                                                                                                                     [tmps][rank=0] 
Debug: 16379 81331 81331                                                                                                                     [tmps][rank=0] 
Debug: 16380 81337 81337                                                                                                                     [tmps][rank=0] 
Debug: 16381 81338 81338                                                                                                                     [tmps][rank=0] 
Debug: 16382 81338 81338                                                                                                                     [tmps][rank=0] 
Debug: 16383 81347 81347                                                                                                                     [tmps][rank=0] 
Debug: 16384 162694 81351                                                                                                                    [tmps][rank=0] 
Debug: 16385 162703 81360                                                                                                                    [tmps][rank=0] 
Debug: 16386 162712 81369                                                                                                                    [tmps][rank=0] 
Debug: 16387 162721 81378                                                                                                                    [tmps][rank=0] 
Debug: 16388 162727 81384                                                                                                                    [tmps][rank=0] 
Debug: 16389 162737 81394                                                                                                                    [tmps][rank=0] 
Debug: 16390 162738 81395                                                                                                                    [tmps][rank=0] 
Debug: 16391 162748 81405                                                                                                                    [tmps][rank=0] 
Debug: 16392 162753 81410                                                                                                                    [tmps][rank=0] 
Debug: 16393 162761 81418  
```

We suspect with @illuhad that the offset apply to read / write to perform an exclusive scan using the inclusive version result in a data race when used in place as the index of the read and the write aren't the same

**To Reproduce**
Run the following for in place scan

```c++

acpp::algorithms::util::allocation_cache cache{
            acpp::algorithms::util::allocation_type::device};
acpp::algorithms::util::allocation_group scratch{&cache, q.get_device()};

sycl::event e = adaptivecpp::algorithms::exclusive_scan(
            q, scratch, in_out_ptr, in_out_ptr + len, in_out_ptr, T{}, deps.get_events());
```

**Expected behavior**

no crash , no corruption

**Describe your setup**
MacOS - omp backend
Linux - omp backend
Linux - generic host
Linux - generic CUDA

carbotaniuman:
This does several refactorings, including changing how exclusive scans are implemented from reading one element before to writing the result element one position later. The last element in the work group is responsible for writing the global prefix, or the init element, if it is the first group.

This refactoring also removes `flat_group_scan_kernel` meant for non default constructible elements, opting instead for changing `scan_kernel` to handle non default constructible elements.

Closes #1908


carbotaniuman:
I think it would be slightly better for memory access patterns if we did a shuffle instead of changing index, but that is probably worse if we need to do it on the workgroup level

illuhad:
This is very performance sensitive code - do you have any benchmarks as to how the different code implementations of scan perform before/after the change?

~Also note my open comment from earlier.~ Just noticed the comment has been listed as "pending" for ages. Apparently I forgot to submit the review :(

include/hipSYCL/algorithms/scan/decoupled_lookback_scan.hpp
```diff
@@ -186,33 +201,44 @@ void iterate_and_inclusive_group_scan(
...
   
-
-  T current_exclusive_prefix;
-  T scan_result [WorkPerItem];
+  // Support types without default constructors
+  // There may be shenanigans with having to implictly start
+  // lifetimes or something that I am ignoring here. When
+  // trivial unions are implemented this will be fine.
+  Storage<T> current_exclusive_prefix;
+  Storage<T[WorkPerItem]> scan_result;
```
illuhad, line 209: Yes - if the type has no default constructor, chances are that it does something more complicated in constructor/destructor.

I suspect this is technically UB unless we use placement new/delete. Which we cannot do, because we don't know how to construct the type in general.

The fact that destructor calling is suppressed by the union might safe us in practice. 

Can we do a quick check whether other similar library (e.g. CUB, rocPRIM etc) use similar patterns? If so, we are fine.

illuhad:
Testing with https://github.com/gonzalobg/babelstream branch `read_write_scan`,
build as:
```
cd src/std
acpp -O3 -ffast-math *.cpp ../main.cpp -DSTD -DDATA17 -I . -I .. -ltbb --acpp-stdpar
```
run with
```
./a.out -o Scan -s 100000000
```

Results in:

| Device | This PR | `develop` |
| ----------|---------|----------|
|RTX A5000 | 272GB/s | 278 GB/s |
| Radeon Pro VII | 204GB/s | 374 GB/s |

So no change on NVIDIA, but significant regression on AMD -- this might be worth looking into. Do you have access to AMD hardware?

carbotaniuman:
Will take a look at the comment and will see if I missed something for performance. I unfortunately do not have access to AMD hardware.

illuhad:
Okay, let me know if you change something and need me to run some tests again :)
From experience, the decoupled lookback scan performance can be highly counter-intuitive. It relies effectively on a producer-consumer scheme, so performance can depend on the actual behavior of the warp scheduler.
I once found out that reducing the kernel's  resource usage actually worsened performance, because it caused more work groups to be mapped to one SM, and could thus allow for scheduling decisions that lead to weaker forward progress overall, if the active group waits for another group living on the same SM which is however not scheduled.

PR #1912, opened 2025-09-13
Bump clang to fix Windows builds

carbotaniuman:


illuhad:
Thanks - if I see it correctly, this allows us to get a little farther, as it no longer fails when building boost, but when building AdaptiveCpp, right?
@fodinabor do you perhaps have an idea what might still be missing?

fodinabor:
DBoost_COMPILER="clangw18"
Make sure to fix all clang version occurences :)
https://github.com/AdaptiveCpp/AdaptiveCpp/tree/ci/windows
Merging these in might fix it.. (Had boost build issues iirc and gave up before vacation)

carbotaniuman:
I got Windows to work, but the MacOS fix is a hack and it also is quite broken, so not sure if that's worth keeping.

src/libkernel/sscp/CMakeLists.txt
```diff
@@ -90,6 +90,17 @@ function(libkernel_generate_bitcode_target)
...
+if(APPLE)
+  if(NOT CMAKE_OSX_SYSROOT)
+    # for some reason this gets loss
+    execute_process(
+      COMMAND xcrun --show-sdk-path
+      OUTPUT_VARIABLE CMAKE_OSX_SYSROOT
+      OUTPUT_STRIP_TRAILING_WHITESPACE
+    )
+  endif()
+endif()
```
illuhad: What does this do? It seems it doesn't resolve the Mac issue, as it still fails in CI. (Perhaps we could split Windows and Mac fixes into separate PRs?)

illuhad (commented):
Thanks for digging through this! The Windows changes look good.

src/libkernel/sscp/CMakeLists.txt
carbotaniuman: Yeah, just removed the MacOS changes off of this.

illuhad (approved):
Thanks!

PR #1914, opened 2025-09-21
Added a function to return the CUDA subgroup size

ath0mos:
I added a function `hiplike_get_subgroup_size()` to return the actual subgroup size. The implementation follows the same logic as for the `__acpp_sscp_get_subgroup_size()`


tests/sycl/sub_group.cpp
```diff
@@ -15,7 +15,23 @@
...
+
+  auto* res = s::malloc_shared<uint32_t>(SG, q);
+  q.parallel_for(
+    s::nd_range<1>(SG, SG),
+    [=](s::nd_item<1> it) {
+      res[it.get_local_linear_id()] = it.get_sub_group().get_local_linear_range();
+    }).wait();
 
+  for (size_t i = 0; i < SG; i++)
+    BOOST_CHECK_EQUAL(res[i], SG);
```
illuhad: This is potentially problematic. On OpenCL devices (e.g. Intel GPUs), multiple subgroups may be supported at the same time. There's no guarantee which subgroup size the compiler will pick. If the subgroup size selected by the compiler is not the first in the vector that `get_info()` returns, then your test would fail.

A solution could be to check if the result matches any of the subgroup sizes returned by `get_info()`.

Perhaps I am misunderstanding your test case, but it does not seem to me that it really checks the most important code path that you have added: The case when a subgroup is incomplete. As I recall, this was the situation that prompted that change in the first place.
Wouldn't it be better to check a situation where the group size is not divisible by the subgroup size, such that you get an incomplete subgroup?

illuhad (approved):
Thanks!

PR #1821, opened 2025-05-12
Add support for parallel_for with 0-sized range

ath0mos:
Added a simple if statement inside of parallel_for which checks the size of numWorkItems

Issue #1764, opened 2025-03-28
24.10.0 Index space of size zero fails to run with generic backend.


will-saunders-ukaea:
**Bug summary**

SYCL2020 states "An index space of size zero is allowed" [1]. I interpret this as a parallel for with size zero is allowed?

[1] https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_sycl_kernel_execution_model

**To Reproduce**
```
$ acpp main.cpp --acpp-targets=generic

$ ACPP_VISIBILITY_MASK=omp ./a.out
[AdaptiveCpp Warning] kernel_cache: This application run has resulted in new binaries being JIT-compiled. This indicates that the runtime optimization process has not yet reached peak performance. You may want to run the application again until this warning no longer appears to achieve optimal performance.
0
zsh: floating point exception (core dumped)  ACPP_VISIBILITY_MASK=omp ./a.out
```
```
#include <iostream>
#include <sycl/sycl.hpp>

int main(int argc, char ** argv){

  sycl::queue q;
  const std::size_t N = 1024;
  
  auto d_y = static_cast<int *>(sycl::malloc_device(sizeof(int) * N, q));

  // runs
  q.parallel_for<>(
    sycl::range<1>(N),
    [=](sycl::id<1> idx) {
      d_y[idx] = static_cast<int>(idx);
    }
  ).wait_and_throw();

  // fails - has size 0
  std::cout << static_cast<std::size_t>(argc - 1) << std::endl;
  q.parallel_for<>(
    sycl::range<1>(static_cast<std::size_t>(argc - 1)),
    [=](sycl::id<1> idx) {
      d_y[idx] = static_cast<int>(idx);
    }
  ).wait_and_throw();

  std::vector<int> h_y(N);
  q.memcpy(h_y.data(), d_y, sizeof(int) * N).wait_and_throw();

  std::cout << h_y.at(0) << " " << h_y.at(N-1) << std::endl;
  sycl::free(d_y, q);
  return 0;
}
```

Relevant gdb output?
```
Thread 7 "test_particle_l" received signal SIGFPE, Arithmetic exception.
[Switching to Thread 0x15554ea006c0 (LWP 125500)]
0x00005555555f7f8a in hipsycl::glue::sscp_kernel_launcher::invoke (launch_config=..., node=0x555556bb9b60,
    kernel_config=..., launch_capabilities=..., backend_params=0x155540001660)
[...]

PR #1916, opened 2025-09-25
Add checks for zero-sized nd_range kernels

MarkusBuettner:
Issue #1764 and PR #1821 addressed submission of zero-sized kernels. A small oversight in the PR was that it only added checks for the standard range-based parallel_for, but not the nd_range parallel_for. 

This PR adds the remaining checks for nd_range parallel_for, as well as for hierarchical and scoped parallelism.

### DIFF
--- a/include/hipSYCL/sycl/handler.hpp
+++ b/include/hipSYCL/sycl/handler.hpp
@@ -378,7 +378,7 @@ public:
             typename... ReductionsAndKernel, int dimensions>
   void parallel_for(nd_range<dimensions> executionRange,
                     const ReductionsAndKernel &... redu_kernel) {
-    if (executionRange.get_global_range() == 0)
+    if (executionRange.get_global_range().size() == 0)
       AdaptiveCpp_enqueue_custom_operation([](auto&){});
 
     else {
@@ -413,7 +413,7 @@ public:
   void parallel_for_work_group(range<dimensions> numWorkGroups,
                                range<dimensions> workGroupSize,
                                const ReductionsAndKernel &... redu_kernel) {
-    if (numWorkGroups == 0)
+    if (numWorkGroups.size() == 0)
       AdaptiveCpp_enqueue_custom_operation([](auto&){});
 
     else {
@@ -435,7 +435,7 @@ public:
                 range<dimensions> workGroupSize,
                 const ReductionsAndKernel& ... redu_kernel)
   {
-    if (numWorkGroups == 0)
+    if (numWorkGroups.size() == 0)
       AdaptiveCpp_enqueue_custom_operation([](auto&){});
 
     else {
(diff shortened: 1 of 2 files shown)


illuhad:
I'm not sure if that was an oversight. The current specification does say that the index space can be zero, but I'm not sure how much sense that makes for `nd_range` kernels.

We also say that users of nd_range kernels do need to manually validate number of work groups and group sizes against what backends support.
So if you already need to check things, what is the benefit of being general for the 0-case? And what does that even mean - number of work groups being zero, group size being zero, or both, and how does that not contradict with the requirement of having to validate those against the backend?

I'm not sure if allowing the index space to be zero should not rather only apply to range-parallel for and whether the oversight is thus in the spec...We should perhaps clarify this in the SYCL WG.

MarkusBuettner:
I would interpret the sentence

> An index space of size zero is allowed. All aspects of kernel execution proceed as normal with the exception that the kernel function itself is not executed. Note this means the command queue will still schedule this kernel after satisfying the requirements and this satisfies requirements of any dependent enqueued kernels.

such that it applies for all types of kernels.

And at the moment, the behaviour is also inconsistent. Let's take this example kernel:
```cpp
   cgh.parallel_for(sycl::nd_range<1> { { N }, { 256 } }, [=](sycl::nd_item<1> i) {
        // whatever... 
   });
});
```

If `N` is zero, then at the moment I observe the following behaviour:
1. oneAPI does not throw errors and does not execute a kernel
2. AdaptiveCpp with ahead-of-time compilation (OpenMP, CUDA or ROCm, doesn't matter) do not execute the kernel, runs fine
3. AdaptiveCpp with SSCP and `ACPP_VISIBILITY_MASK=omp` or `=ocl` does not execute the kernel and runs fine
4. AdaptiveCpp with SSCP and `ACPP_VISIBILITY_MASK=cuda` throws an error: `cuda_queue: could not submit kernel from module (error code = CU:1)` 

Maybe then it is rather a bug in the CUDA backend (I actually didn't check ROCm yet), as it should just behave like CUDA SMCP flow -- if that is so I can see if I can fix this in the CUDA backend directly.

Otherwise, I would argue that it should only check the global size, and maybe not the group size.

Edit:  
HIP with SSCP throws an error `hip_queue: could not submit kernel from module (error code = HIP:9)`, while ahead-of-time runs without errors. Error code nine is invalid configuration:

> hipErrorInvalidConfiguration
> Invalid configuration argument. This error occurs when the configuration specified for a kernel launch or other 
> configurable operation contains invalid parameters. Common scenarios include:
>
> Block dimensions exceeding hardware limits (too many threads per block)
>
> Grid dimensions that are invalid (zero size or exceeding limits)
> 
> Invalid shared memory configuration
> 
> Incompatible combination of launch parameters
> 
> Block dimensions that don’t match kernel requirements
> 
> Attempting to use more resources per block than available on the device

PR #1920, opened 2025-10-10
Fix a few typos here and there

al42and:
Most notably, `CUDA_VISIBLE_DEVICES` in the device mask warning, since it is convenient to copy-paste it.


illuhad (approved):
Whoopsie, thank you!

PR #1921, opened 2025-10-10
Add info::device::priority_range

al42and:
While priority range (and even direction) is backend-defined, it is still nice to be able to query them.

Would it make more sense to use a custom struct instead of std::pair? Or, like with device_uint_properties, split the user-facing `device::info` struct into `priority_range_low` and `priority_range_high`?


include/hipSYCL/sycl/extensions.hpp
```diff
@@ -39,7 +39,7 @@
 #define HIPSYCL_EXT_QUEUE_WAIT_LIST
 #define HIPSYCL_EXT_MULTI_DEVICE_QUEUE
 #define HIPSYCL_EXT_COARSE_GRAINED_EVENTS
-#define HIPSYCL_EXT_QUEUE_PRIORITY
+#define HIPSYCL_EXT_QUEUE_PRIORITY 2
```
illuhad: Also change this for `ACPP_EXT_QUEUE_PRIORITY` below. `HIPSYCL_*` is just there for backwards compatibility.

We do have the slight issue that we didn't define those macros to 1 to begin with...I'd perhaps suggest to add a new feature test macro for priority range (i.e. treat it as a new extension, since it adds a new API), and then in a separate PR set all feature test macros for supported extensions to `1`.

include/hipSYCL/sycl/info/device.hpp
```diff
@@ -186,6 +186,18 @@ namespace device{
 }
 }
 }
+
+namespace ext::adaptivecpp {
```
illuhad: I don't think we currently use such namespaces. What we *do* however is to prefix extension APIs with `AdaptiveCpp_`. Perhaps that would be an option for `priority_range`?

src/runtime/cuda/cuda_hardware_manager.cpp
```diff
@@ -26,6 +26,22 @@
 namespace hipsycl {
 namespace rt {
 
+namespace
+{
+  int get_stream_priority_bound(bool get_least) {
```
illuhad: Why not just return an `std::pair` or so?

illuhad (commented):
Thanks - I think `std::pair` is fine here, but I do have some other comments.

EDIT: Please also add some documentation in `extensions.md` :)

include/hipSYCL/sycl/extensions.hpp
al42and: Okay `ACPP_EXT_QUEUE_PROPERTY_PRIORITY_RANGE` it is then. Added the docs for it, and for `ACPP_EXT_QUEUE_PRIORITY` too.

include/hipSYCL/sycl/info/device.hpp
al42and: Ok. I went by "Guidelines for portable extensions" from SYCL spec, but I guess consistency is more important. It's still in the same spirit.

src/runtime/cuda/cuda_hardware_manager.cpp
al42and: Ok

illuhad (approved):
Thank you!

PR #1922, opened 2025-10-10
Open bitcode files in binary

al42and:
Since LLVM 13 (c83cd8feef7eb8319131d968bb8c94fdc8dbb6a6), the second argument of `llvm::MemoryBuffer::getFile` was changed to `bool IsText = false` from `int64_t FileSize = -1`.

However, the `-1` was left in some invocations of this function, causing the file to be opened in text mode.

Harmless on Linux, but I guess it's not good


illuhad (approved):
Thanks, good catch!

PR #1923, opened 2025-10-20
Add queue priority support to OpenCL

al42and:


src/runtime/ocl/ocl_queue.cpp
```diff
@@ -108,18 +109,36 @@ result submit_ocl_kernel(cl::Kernel& kernel,
...
-  _queue = cl::CommandQueue{cl_ctx, cl_dev, props, &err};
+  if(priority != 0 && dev_ctx->has_cl_khr_priority_hints_extension()) {
+#ifdef CL_QUEUE_PRIORITY_KHR
+    cl_queue_properties clPriority = (priority < 0) ? CL_QUEUE_PRIORITY_HIGH_KHR : CL_QUEUE_PRIORITY_LOW_KHR;
+    cl_queue_properties props[] = {
+      CL_QUEUE_PRIORITY_KHR, clPriority,
+      0
+    };
+    cl_command_queue rawQueue = clCreateCommandQueueWithProperties(cl_ctx(), cl_dev(), props, &err);
+    _queue = cl::CommandQueue(rawQueue);
```
illuhad, line 130: Why are we first constructing a raw queue? As far as I can see, `cl::CommandQueue` already seems to have a constructor for `cl_queue_properties`, `context` and `device`. So can't we directly create the `cl::CommandQueue` as in the other branches?

illuhad (commented):
Thanks, this is fab, just one question :)

src/runtime/ocl/ocl_queue.cpp
al42and: I wish.

`cl::CommandQueue` indeed has such a ctor. But you [can't pass arbitrary properties there, it only accepts flags for `CL_QUEUE_PROPERTIES`](https://github.com/KhronosGroup/OpenCL-CLHPP/blob/main/include/CL/opencl.hpp#L7988-L8011).

There's an [open issue about it](https://github.com/KhronosGroup/OpenCL-CLHPP/issues/80), celebrating its 6th anniversary in a couple weeks.

illuhad: Gnaaaa....alright, ugly workaround it is :)

illuhad (approved):
Thanks!

PR #1925, opened 2025-10-23
[NFC][doc] Clarify that we universally need lld

illuhad:
We no longer only need lld when ROCm is used, we also need it for the other backends for `generic` target.


ferdymercury (approved):
Thanks a lot!

PR #1932, opened 2025-11-19
Replaced calloc for malloc in minicoro allocation approaches.

### LINKED ISSUE #1931: Performance regression in omp.library-only for 25.10.0 (seems very fixable)
Hello,

We use the `omp.library-only` compilation flow quite extensively as it is very easy to onboard new users with and get them to hello world as quickly and painlessly as possible (and then use a different compilation flow when for performance critical runs or once users have their environment working with LLVM). 

With 25.02.0, 1 OpenMP threads and 16 MPI ranks our test suite takes ~ 106s. With the 25.10.0 release this becomes ~1100s. Profiling the tests places the extra time in the `mco_create` function, specifically calls to `__memset_avx512_unaligned_erms`.

Image

If I replace the `calloc` here [1] with `malloc` then the time to run our test suite improves to ~84s and we achieve, and improve on, the original 25.02.0 performance. The other memory allocation approaches in this file do not seem to zero the memory, hence I assume this substitution is valid. 

```
Summary of times:
OMP_NUM_THREADS=1, 16 MPI ranks:
25.10.0 release calloc : 1100021 ms
25.02.0 release boost  : 105758 ms
25.10.0 modified malloc: 83996 ms
```

I can make a PR if you would like?

[1] https://github.com/AdaptiveCpp/AdaptiveCpp/blob/9f842c701a599107cc6d117d3539f971036363a1/src/runtime/support/minicoro.h#L517

will-saunders-ukaea:
Implements the replacement of calloc for malloc as discussed in issue 1931 [1].

[1] https://github.com/AdaptiveCpp/AdaptiveCpp/issues/1931
Closes #1931

### DIFF
--- a/src/runtime/support/minicoro_wrapper.cpp
+++ b/src/runtime/support/minicoro_wrapper.cpp
@@ -13,6 +13,16 @@
 
 #include "hipSYCL/runtime/support/minicoro_wrapper.hpp"
 
+/**
+ * Defines MCO_ALLOC as malloc to avoid performance issues when the compiler
+ * issues a malloc + memset for a calloc. The coroutines allocate more space
+ * than they touch. Hence if the compiler issues a malloc + memset then all the
+ * requested space written to and hence allocated in physical memory which is
+ * slow. See AdaptiveCpp GitHub issue 1931. @edubart advises that this
+ * substitution is safe.
+ */
+#include <stdlib.h>
+#define MCO_ALLOC(size) malloc(size)
 #define MINICORO_IMPL
 #include "minicoro.h"
 
--- a/src/runtime/support/minicoro_wrapper.cpp
+++ b/src/runtime/support/minicoro_wrapper.cpp
@@ -23,6 +23,7 @@
  */
 #include <stdlib.h>
 #define MCO_ALLOC(size) malloc(size)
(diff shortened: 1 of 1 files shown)


will-saunders-ukaea:
2025-11-19-124603_1708x1130_scrot
Is this expected? Fatal looking error but the CI run succeeded?

https://github.com/AdaptiveCpp/AdaptiveCpp/actions/runs/19499827231/job/55813440353

illuhad:
@will-saunders-ukaea Yes, this is okay. This particular test is for the NUMA memory placement extension, which also tests that the call fails when invalid arguments are passed in.

There's also currently an issue with the MacOS CI that is independent from your change.

illuhad (commented):
Thanks! Looks good so far. Since it was pointed out that `calloc` is a memory optimization, I wonder whether you noticed any change in memory consumption for a typical kernel with this PR?

will-saunders-ukaea:
I think I actually see lower memory consumption with malloc. If my system is replacing calloc with malloc+memset then I think this expected if we assume that malloc only allocates physical memory when entries are written to the address.

will-saunders-ukaea:
Updated PR to implement suggestion from @al42and .

illuhad (approved):
LGTM, thanks!

PR #1939, opened 2025-12-08
[llvm-to-backend] Need to set eh/rtti depending on LLVM's build.

fodinabor:
No clue why it worked without this so far... :)
My local build with LLVM 22-dev definitely didn't like it.


illuhad (approved):
Thanks!

PR #1943, opened 2025-12-14
Implement GROMACS specific builtins for the SSCP flow

sbalint98:
This MR exposes low level AMD device specific functionality that is used in GROMACS. It is important to use these functions for obtaining good performance on MI210 and MI300X GPUs.  

The current design is not idiomatic SYCL. However, using these a significant performance improvement can be reached, therefore I believe it might be useful. 

Includes:
* Common set of DPP instructions
* read_first_lane intrinsic
* unsafe atomic adds for MI210 GPUs
* fractf intrinsic


include/hipSYCL/sycl/libkernel/sscp/builtins/amdgpu_general.hpp
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright The AdaptiveCpp Contributors
+ *
+ * AdaptiveCpp is released under the BSD 2-Clause "Simplified" License.
+ * See file LICENSE in the project root for full license details.
+ */
+
+#include "builtin_config.hpp"
+
+HIPSYCL_SSCP_BUILTIN __acpp_int32 __acpp_sscp_read_first_lane(__acpp_int32 input);
+HIPSYCL_SSCP_BUILTIN __acpp_f32 __acpp_sscp_fractf(__acpp_f32 input);
```
illuhad, line 14: We already have `__acpp_sscp_fract_f32/64` as part of the math builtins, which is mapped to `__ocml_fract_f32`. Why is this not sufficient?

(In any case, the name would need to be changed to avoid confusion with the existing builtin).

I'd also suggest to rename `__acpp_sscp_read_first_lane` to `__acpp_sscp_read_first_lane_i32` to align with LLVM.
And since it's a backend specific thing, should probably be prefixed with `__acpp_sscp_amdgcn_` or so.

include/hipSYCL/sycl/libkernel/sscp/builtins/atomic.hpp
```diff
@@ -202,6 +202,15 @@ HIPSYCL_SSCP_BUILTIN __acpp_f64 __acpp_sscp_atomic_fetch_add_f64(
     __acpp_sscp_address_space as, __acpp_sscp_memory_order order,
     __acpp_sscp_memory_scope scope, __acpp_f64 *ptr, __acpp_f64 x);
 
+// Only has an effect when targeting gfx90a amdgcn
+// For other targets it will just call the regulare atomic_fetch
+HIPSYCL_SSCP_BUILTIN __acpp_f32 __acpp_sscp_unsafe_atomic_fetch_add_f32(
```
illuhad, line 207: Maybe better choose a name that makes it clear *what* is unsafe, especially if it's positioned as a general purpose builtin, not just something for amd.
`__acpp_sscp_atomic_fetch_only_device_mem`, `no_shared_mem` or so.

include/hipSYCL/sycl/libkernel/target_specific_builtins.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+HIPSYCL_BUILTIN int acpp_sscp_amdgcn_readfirstlane(int value) noexcept {
+    __acpp_backend_switch(return noop(), return __internal_acpp_sscp_read_first_lane(value) , return noop(), return noop());
+}
+
+HIPSYCL_BUILTIN float acpp_sscp_amdgcn_fractf(float value) noexcept {
+    __acpp_backend_switch(return noop(), return __internal_acpp_sscp_fractf(value) , return noop(), return noop());
+}
+
+}
```
illuhad, line 82: Should this file live in an amd-specific subdirectory? Why have code paths that just fail silently when used from the wrong backend or compilation flow?

src/libkernel/sscp/amdgpu/atomic_gfx90a.cpp
```diff
@@ -0,0 +1,24 @@
...
+HIPSYCL_SSCP_BUILTIN __acpp_f32 __acpp_sscp_unsafe_atomic_fetch_add_f32(
+  __acpp_sscp_address_space as, __acpp_sscp_memory_order order,
+  __acpp_sscp_memory_scope scope, __acpp_f32 *ptr, __acpp_f32 x) {
+  return __builtin_amdgcn_global_atomic_fadd_f32( (const __attribute__((address_space(1))) float *)ptr, x);
+}
+
+HIPSYCL_SSCP_BUILTIN __acpp_f64 __acpp_sscp_unsafe_atomic_fetch_add_f64(
+  __acpp_sscp_address_space as, __acpp_sscp_memory_order order,
+  __acpp_sscp_memory_scope scope, __acpp_f64 *ptr, __acpp_f64 x) {
+  return __builtin_amdgcn_global_atomic_fadd_f64( (const __attribute__((address_space(1))) float *)ptr, x);
```
illuhad, line 23: I don't like this extra file - this will not scale (see ROCm device libs...). Why can we not solve this with reflection in the existing file? I assume because clang only generates the right IR for the builtin when it sees the right target?
Inline assembly + reflection might still be easier...

src/libkernel/sscp/amdgpu/atomic_gfx90a_fallback.cpp
```diff
@@ -0,0 +1,79 @@
...
+    __acpp_sscp_memory_scope scope, __acpp_f32 *ptr, __acpp_f32 x) {
+    return __hip_atomic_fetch_add(ptr, x, builtin_memory_order(order),
+                                builtin_memory_scope(scope));(ptr, x);
+}
+
+HIPSYCL_SSCP_BUILTIN __acpp_f64 __acpp_sscp_unsafe_atomic_fetch_add_f64(
+    __acpp_sscp_address_space as, __acpp_sscp_memory_order order,
+    __acpp_sscp_memory_scope scope, __acpp_f64 *ptr, __acpp_f64 x) {
+    return __hip_atomic_fetch_add(ptr, x, builtin_memory_order(order),
+                                builtin_memory_scope(scope));
```
illuhad, line 77: Same here, I'd like this stuff to be an if/else branch.

src/libkernel/sscp/amdgpu/general_builtins.cpp
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Copyright The AdaptiveCpp Contributors
+ *
+ * AdaptiveCpp is released under the BSD 2-Clause "Simplified" License.
+ * See file LICENSE in the project root for full license details.
+ */
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include "hipSYCL/sycl/libkernel/sscp/builtins/builtin_config.hpp"
+#include "hipSYCL/sycl/libkernel/sscp/builtins/amdgpu_general.hpp"
```
illuhad, line 13: Not sure "general" is the right name here. Assuming we don't really need fractf, isn't read first lane related to broadcasts, shuffles etc? So maybe put there?

PR #1944, opened 2025-12-15
Use SMCP equivalent memory scope for barriers

sbalint98:
In case of HIP a `__syncthreads()` call will generate the follwing LLVM IR [godbolt](https://godbolt.org/z/16a9aP64r):

```
  fence syncscope("workgroup") release
  call void @llvm.amdgcn.s.barrier()
  fence syncscope("workgroup") acquire
```
Since in case of SMCP we call sync-threads directly, this PR makes SSCP also generate the same IR. This change has significantly improves the performance of the SSCP flow.

### DIFF
--- a/include/hipSYCL/sycl/libkernel/detail/device_barrier.hpp
+++ b/include/hipSYCL/sycl/libkernel/detail/device_barrier.hpp
@@ -31,7 +31,7 @@ inline void sscp_barrier(access::fence_space space) {
     __acpp_sscp_work_group_barrier(memory_scope::work_group,
                                       memory_order::seq_cst);
   } else {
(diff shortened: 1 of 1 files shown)


illuhad (approved):
Thanks! It seems this change is also in line with what DPC++ does: https://github.com/intel/llvm/blob/98636646c85d76cf6e066b6fa12b63c99ce91133/sycl/include/sycl/nd_item.hpp#L199

Maybe we also need to think about whether `seq_cst` is really needed here...Your example snippet indicates that `__syncthreads` on AMD might rather do an `acq_rel`. But we can address this in another PR :)

PR #1946, opened 2025-12-18
Add tracing framework

jabruniessner:
This branch implements a tracing utility for the SYCL-API of AdaptiveCpp. The utility is inspired by the OpenMP callback mechanism, as we felt that such a utility would also be useful to have in SYCL.

With this extension, it is possible to develop tooling, such as performance tracing for submission, event and queue creation (Task graph visualization), and memory leak detectors. I added documentation for its usage in the file:

`doc/sycl_callback_tooling.md`

The utility is based on hooks which runs before each submission over all currently present tools. I also benchmarked a simple program, which repeatedly sends empty kernels (in this case 1,000,000 times) and measures the time  with and without the modification and with an empty callback registered. I found the following benchmark values for an NVidia RTX 4090 and received the following values in s and deviation in percentile:

|Function| AdaptiveCpp original                    | AdatpiveCpp with tracing ability | AdaptiveCpp with tracing and empty tracer |
|------------------------------------------|---------------------|------------------------------------------------|----------------------------------------------------------------|
| `memcpy`                                 |`3.93364`|`3.93028` (-0.08%)| `3.93059` (-0.078%) |
| `memset`                                  |`2.56096`|`2.56249` (0.06%) |`2.56175` (0.03%)  |
| `wait`                                         |`0.12299`|`0.115619` (-5.9%)|`0.121551` (-1.17%)|
| `fill` and `copy`                         |`4.85937 `|`4.9052` (0.94%)  |`4.92104`  (1.27%)  |
| `single task` + `depends_on` |`2.80093`|`2.77507` (0.92%) |`2.835` (1.21%)  |
| `parallel_for`                             |`2.75662`|`2.769` (-0.45%)   |`2.83542` (2.84%)    |

I ran the same code with a CPU selector and the OpenMP backend. The results were

|Function| AdaptiveCpp original                    | AdatpiveCpp with tracing ability | AdaptiveCpp with tracing and empty tracer |
|------------------------------------------|---------------------|------------------------------------------------|----------------------------------------------------------------|
| `memcpy`                                 |`1.10861` |`1.03888`  (-6.28%)    | `1.07098` (-3.4%)|
| `memset`                                  |`0.8385`   |`0.832079` (-0.76%)   | `0.781739`(-6.76%) |
| `wait`                                         |`0.0097`   |`0.0103016` (6.18%)  | `0.013816` (42%)|
| `fill` and `copy`                         |` 2.40101`|`2.20124`  (-8.32%)    |  `2.2123` (-7.85%)|
| `single task` + `depends_on` |`1.30419` |`1.26127` (-3.4%)       |  `1.27112`(-2.53%)|
| `parallel_for`                             |`1.07341`|`0.950777` (-11.42%) |  `1.8714 (74.34%)` |

If there is any further data required or desired, I would be happy to provide everything. Do not hesitate to ask


src/runtime/CMakeLists.txt
```diff
@@ -58,11 +63,13 @@ add_library(acpp-rt SHARED
...
   pcuda/pcuda_thread_state.cpp
-  pcuda/pcuda_event.cpp)
+  pcuda/pcuda_event.cpp
+ )
 
+target_link_libraries(sycl_tracer PUBLIC acpp-common)
 target_compile_options(acpp-rt PRIVATE ${HIPSYCL_RT_EXTRA_CXX_FLAGS})
 target_link_libraries(acpp-rt PRIVATE ${HIPSYCL_RT_EXTRA_LINKER_FLAGS} Threads::Threads)
-target_link_libraries(acpp-rt PUBLIC acpp-common)
+target_link_libraries(acpp-rt PUBLIC acpp-common sycl_tracer)
```
illuhad, line 72: Can we rename it to something that fits existing naming schemes for our libraries, e.g. `acpp-tracer`?

src/runtime/CMakeLists.txt
```diff
@@ -22,7 +22,12 @@ set(HIPSYCL_RT_EXTRA_LINKER_FLAGS ${HIPSYCL_RT_EXTRA_LINKER_FLAGS} ${HIPSYCL_STD
 
 set(CMAKE_INSTALL_RPATH ${base} ${base}/hipSYCL)
 
+add_library(sycl_tracer SHARED
+  tracer_utils.cpp
+  tracer_utils_internal.cpp)
```
illuhad, line 27: Maybe it makes sense to create a `tracer` subdirectory for tracer files under `src/runtime` and `include/runtime`?

src/runtime/CMakeLists.txt
```diff
@@ -22,7 +22,12 @@ set(HIPSYCL_RT_EXTRA_LINKER_FLAGS ${HIPSYCL_RT_EXTRA_LINKER_FLAGS} ${HIPSYCL_STD
 
 set(CMAKE_INSTALL_RPATH ${base} ${base}/hipSYCL)
 
+add_library(sycl_tracer SHARED
```
illuhad, line 25: Does it need to be a separate library?
I wonder whether it could not just be part of `acpp-rt`, like so:
- Initially, while it's experimental, it is disabled by default, and must be enabled at cmake time when building AdaptiveCpp. Even if linked into the runtime library, no overhead would incur because nothing ever calls into the tracer library if it's disabled at cmake time (I assume this is correct?).
- Later on, it could just be enabled by default.

If it's a separate library then additional considerations will have to be made e.g. w.r.t how it folds into the deployment model.

doc/sycl_callback_tooling.md
```diff
@@ -0,0 +1,94 @@
+# Sycl callback tooling
```
illuhad:
```suggestion
# SYCL callback tooling 
```

doc/sycl_callback_tooling.md
```diff
@@ -0,0 +1,94 @@
+# Sycl callback tooling 
+
+To enable tracing, benchmarking, and simple debugging of sycl applications, this fork of AdaptiveCpp experimentally supports a callback mechanism.
```
illuhad: This is no fork :) Please adapt documentation.

doc/sycl_callback_tooling.md
```diff
@@ -0,0 +1,94 @@
+# Sycl callback tooling 
+
+To enable tracing, benchmarking, and simple debugging of sycl applications, this fork of AdaptiveCpp experimentally supports a callback mechanism.  
+
+
+## Usage: 
+
+### Registering Callbacks (User side)
+The user can start tracers by setting the environment variable `SYCL_TOOL_LIBRARY` to the paths to tooling libraries separated by a colon `:`.
```
illuhad, line 5: I would prefer `ACPP_TOOL_LIBRARY` so that we can also leverage this later for the other programming language models we support without it sounding weird.
If you want to support `SYCL_TOOL_LIBRARY` for standardization purposes, perhaps we can support both names?

Also, more AdaptiveCpp-ish might be separation by semicolon instead of colon. Compare e.g. the syntax of `--acpp-targets` or `ACPP_VISIBILITY_MASK`.
I understand that on Linux, variables like `LD_LIBRARY_PATH` also rely on `:`, so it matches those, but for AdaptiveCpp I would have a preference for `;`.

doc/sycl_callback_tooling.md
```diff
@@ -0,0 +1,94 @@
...
+**Note**: It is recommended that the tracer state is allocated on the heap, i.e. using `new` or `malloc` as stack-allocated states may cause undefined behavior during shutdown of the program. The state can be created within the `init_register`-function. 
+
+Example of a simple tracer implementation:
+
+```cpp 
+#include <iostream>
+#include "hipSYCL/sycl/tracer_utils.hpp"
+    
+struct MyTracerState {
+    ...
```
illuhad, line 20: AdaptiveCpp generally uses 2-space indentation everywhere, including examples.

doc/sycl_callback_tooling.md
```diff
@@ -0,0 +1,94 @@
...
+| malloc_host_end               | `void(void* state, void* ptr)`                         | end of a malloc_host operation|
+| malloc_shared_start           | `void(void* state)`                                    | start of a malloc_shared operation|
+| malloc_shared_end             | `void(void* state, void* ptr)`                         | end of a malloc_shared operation|
+| free_start                    | `void(void* state)`                                    | start of a free operation|
+| free_end                      | `void(void* state, void* ptr)`                         | end of a free operation|
+| finalize                      | `void(void* state)`                                    | shutdown of the SYCL runtime|
+
+If multiple tooling libraries are loaded, the order of initialization is the same as the order of the path list in the `SYCL_TOOL_LIBRARY` environment variable. The order of finalization is the inverse order of initialization. 
+
+**Note**: In AdaptiveCpp, the SYCL runtime is a singleton associated with the existence of a sycl::queue, i.e. the runtime exists as long as there is at least one sycl::queue.
```
illuhad, line 88: No, this is not quite true.
- If `ACPP_PERSISTENT_RUNTIME=0` (default), it exists as long as there are living by-reference semantics objects, which includes `queue`, but also e.g. `event`, `device` or `context`.
- If `ACPP_PERSISTENT_RUNTIME=1`, it exists independently of these objects

doc/sycl_callback_tooling.md
```diff
@@ -0,0 +1,94 @@
...
+| free_start                    | `void(void* state)`                                    | start of a free operation|
+| free_end                      | `void(void* state, void* ptr)`                         | end of a free operation|
+| finalize                      | `void(void* state)`                                    | shutdown of the SYCL runtime|
+
+If multiple tooling libraries are loaded, the order of initialization is the same as the order of the path list in the `SYCL_TOOL_LIBRARY` environment variable. The order of finalization is the inverse order of initialization. 
+
+**Note**: In AdaptiveCpp, the SYCL runtime is a singleton associated with the existence of a sycl::queue, i.e. the runtime exists as long as there is at least one sycl::queue. 
+If no sycl::queue is created, the runtime is not initialized and thus the tooling library is not loaded. Furthermore, if there is a point in the program at which there is no sycl::queue, the runtime is finalized and the tracers are finalized as well. When a queue is created again, the runtime is reinitialized and so is the tracer.
```
illuhad, line 89: I don't think this is the place to explain AdaptiveCpp runtime behavior; maybe it is easier if we just refer to the documentation of `ACPP_PERSISTENT_RUNTIME`, and then say that tracer lifetime is bound to runtime lifetime, which may not span the entire program lifetime.

include/hipSYCL/sycl/event.hpp
```diff
@@ -56,15 +59,19 @@ class event {
     return std::vector<event>{};
   }
 
-  void wait()
-  {
-    if(this->_node){
-      if(!this->_node->is_submitted())
+  void wait() {
+
+    TRACER_FUNCTION1ARG(wait_event_start);
```
illuhad, line 64: These macros should be prefixed, ideally by `ACPP_TRACER_*`.

include/hipSYCL/sycl/khr_sycl_tools.hpp
```diff
@@ -0,0 +1,10 @@
+#ifndef HIPSYCL_KHR_SYCL_TOOLS
+#define HIPSYCL_KHR_SYCL_TOOLS
+
+#define SYCL_IMPLEMENTATION_HIPSYCL
+#define SYCL_IMPLEMENTATION_ACPP
```
illuhad, line 5: AdaptiveCpp already defines these in `sycl.hpp`. Redefining here can trigger macro redefinition warnings. Why do you need them here?

include/hipSYCL/sycl/tracer_macros.h
```diff
@@ -0,0 +1,78 @@
+#include <vector>
+
+#ifndef TRACER_MACROS_H
+#define TRACER_MACROS_H
```
illuhad, line 4:
```suggestion
#define ACPP_TRACER_MACROS_H
```
We generally prefix all macros we define.

include/hipSYCL/sycl/tracer_macros.h
```diff
@@ -0,0 +1,78 @@
...
+
+#define ALL_TYPES(MACRO)                                                                           \
```
illuhad, line 6: since this is a header pulled into the SYCL headers, this pollutes the global namespace with the `ALL_TYPES` macro, which seems sufficiently generic to risk name collisions.
Please prefix and/or `#undef` at the end of the file.

Also... it's not really all types, but rather, all tracing points, no?

include/hipSYCL/sycl/tracer_macros.h
```diff
@@ -0,0 +1,78 @@
...
+
+#define TRACER_FUNCTION1ARG_END(type) TRACER_FUNCTION_VA_ARGS_END(type)
+#define TRACER_FUNCTION2ARG_END(type, arg2) TRACER_FUNCTION_VA_ARGS_END(type, arg2)
+#define TRACER_FUNCTION3ARG_END(type, arg2, arg3) TRACER_FUNCTION_VA_ARGS_END(type, arg2, arg3)
+
+#ifndef _WIN32
+#define MYLIB_API
+#else
+#ifdef MYLIB_EXPORTS
+#define MYLIB_API __declspec(dllexport)
```
illuhad, line 72: Can we not just use `ACPP_COMMON_EXPORT` from `common/export.hpp`? It does the same thing.

include/hipSYCL/sycl/tracer_utils.hpp
```diff
@@ -0,0 +1,40 @@
+// #pragma once
+
```
illuhad, line 2: Does it make sense to have these files live as part of the SYCL headers? It does not define any SYCL APIs.
Maybe as part of the runtime, since the source files already live under `src/runtime`?

include/hipSYCL/sycl/tracer_utils.hpp
```diff
@@ -0,0 +1,40 @@
...
+template <typename T> struct TD;
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+using hashtype = decltype(std::hash<void *>{}(0));
+
+enum tracer_start_end { START = 0, END = 1 };
```
illuhad:
```suggestion
enum class tracer_start_end { start = 0, end = 1 };
```

Maybe better use `enum class`?
And we never capitalize enum entries ;)

include/hipSYCL/sycl/tracer_utils_internal.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+#include "hipSYCL/common/export.hpp"
+#include "tracer_utils.hpp"
+#include <chrono>
+#include <vector>
+
+#ifndef TRACER_UTILS_INTERNAL_H
+#define TRACER_UTILS_INTERNAL_H
+
+namespace Tracer_utils {
```
illuhad:
```suggestion
namespace tracer_utils {
```

include/hipSYCL/sycl/usm.hpp
```diff
@@ -36,226 +36,220 @@ namespace sycl {
...
-inline void *malloc_device(size_t num_bytes, const device &dev,
-                           const context &ctx,
+inline void *malloc_device(size_t num_bytes, const device &dev, const context &ctx,
                            const property_list &propList = {}) {
-  if (num_bytes == 0)
+
+  TRACER_FUNCTION1ARG(malloc_device_start);
+
+  if (num_bytes == 0) {
+    TRACER_FUNCTION2ARG_END(malloc_device_end, nullptr);
```
illuhad, line 80: This pattern makes me wonder whether it might not be better to have an RAII guard to trigger the tracer end hooks. Otherwise it might  become easy to miss some code paths.

include/hipSYCL/sycl/usm.hpp
```diff
@@ -36,226 +36,220 @@ namespace sycl {
...
 
 template <typename T>
-T *aligned_alloc_device(std::size_t alignment, std::size_t count,
-                        const device &dev, const context &ctx,
-                        const property_list &propList = {}) {
-  return static_cast<T *>(
-      aligned_alloc_device(alignment, count * sizeof(T), dev, ctx, propList));
+T *aligned_alloc_device(std::size_t alignment, std::size_t count, const device &dev,
+                        const context &ctx, const property_list &propList = {}) {
+  return static_cast<T *>(aligned_alloc_device(alignment, count * sizeof(T), dev, ctx, propList));
```
illuhad: There are a lot of changes in this file that are unrelated to the tracer and seem to just be reformatting of code. Can you remove those changes such that the file only contains the tracer changes?

.clang-format
```diff
@@ -0,0 +1,2 @@
+BasedOnStyle: LLVM
+ColumnLimit: 100
```
illuhad: This file is unrelated to the PR, please remove.

jabruniessner: Done!

doc/sycl_callback_tooling.md
jabruniessner: Done

doc/sycl_callback_tooling.md
jabruniessner: Done

include/hipSYCL/sycl/tracer_utils.hpp
jabruniessner: Thanks for finding, this was legacy, I removed the code

include/hipSYCL/sycl/tracer_utils_internal.hpp
jabruniessner: Done

include/hipSYCL/sycl/usm.hpp
jabruniessner: I think this would be a really cool think to explore, but there are multiple sites, where I do not see, how this can obviously be done.

Several end hooks need data only known at the actual end, and that data varies by exit path. Look at submit_secondary_end in queue.hpp — it's called from 3 separate branches (lines 520, 528, 537), each passing a different evt/result depending on which code path executed. A guard's destructor can't know on its own which branch ran or what value to report — you'd still need to explicitly tell the guard the result before it goes out of scope (guard.record(evt, ...)), at each of those 3 places. So the "boilerplate reduction" benefit mostly evaporates for these cases; you're replacing one explicit call with another explicit call, just wrapped differently.

I am pretty sure that we tested all possibilities through and therefore do not think it is necessary for now

include/hipSYCL/sycl/usm.hpp
jabruniessner: Fixed

src/runtime/CMakeLists.txt
jabruniessner: I made it part of the runtime library

src/runtime/CMakeLists.txt
jabruniessner: Removed it

src/runtime/CMakeLists.txt
jabruniessner: Done

PR #1947, opened 2025-12-23
Use aspect selector for host USM SYCL test

EwanC:
Rather than using the GPU device selector to enable a SYCL queue test that requires host USM, use a aspect selector for the asserting the host usm capability.

This change allows the test to run on USM enabled devices that aren't GPUs, in particular OpenCL CPU implementations with the  USM/SVM extension features.

### DIFF
--- a/tests/sycl/queue.cpp
+++ b/tests/sycl/queue.cpp
@@ -35,10 +35,8 @@ BOOST_AUTO_TEST_CASE(queue_wait) {
 }
 
 BOOST_AUTO_TEST_CASE(queue_memcpy_host_to_host) {
-  try {
-    sycl::queue q{sycl::aspect_selector(std::vector{sycl::aspect::usm_host_allocations}),
-                  sycl::property::queue::in_order{}};
-
+  sycl::queue q{sycl::property::queue::in_order{}};
+  if (q.get_device().has(sycl::aspect::usm_host_allocations)) {
     auto source = sycl::malloc_host(sizeof(int), q);
     auto dest = malloc(sizeof(int));
 
@@ -46,8 +44,6 @@ BOOST_AUTO_TEST_CASE(queue_memcpy_host_to_host) {
 
     sycl::free(source, q);
     free(dest);
-  } catch (sycl::exception e) {
-    BOOST_CHECK(true); // Skip the test if host USM not available
   }
 }
 
--- a/tests/sycl/queue.cpp
+++ b/tests/sycl/queue.cpp
@@ -36,7 +36,8 @@ BOOST_AUTO_TEST_CASE(queue_wait) {
 
 BOOST_AUTO_TEST_CASE(queue_memcpy_host_to_host) {
   try {
-    sycl::queue q{sycl::gpu_selector_v, sycl::property::queue::in_order{}};
+    sycl::queue q{sycl::aspect_selector(std::vector{sycl::aspect::usm_host_allocations}),
+                  sycl::property::queue::in_order{}};
 
     auto source = sycl::malloc_host(sizeof(int), q);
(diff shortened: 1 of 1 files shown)


illuhad (approved):
Thanks! It's not clear to my why we restricted this test to GPU in the first place. AdaptiveCpp mandates USM pretty much anywhere anyway. Perhaps because we thought that USM on host case was trivial?

Out of curiosity, why would an OpenCL CPU device not support host USM? It should be pretty trivial, no?

tests/sycl/queue.cpp
```diff
@@ -36,7 +36,8 @@ BOOST_AUTO_TEST_CASE(queue_wait) {
 
 BOOST_AUTO_TEST_CASE(queue_memcpy_host_to_host) {
   try {
-    sycl::queue q{sycl::gpu_selector_v, sycl::property::queue::in_order{}};
+    sycl::queue q{sycl::aspect_selector(std::vector{sycl::aspect::usm_host_allocations}),
+                  sycl::property::queue::in_order{}};
```
illuhad: Perhaps one remark: Maybe it might be even better to just default construct a queue here. Our testing framework relies extensively on the default device being the one that is tested (e.g. CI just sets environment variables that alter the default device).
So, maybe it might be better for consistency to just use the default device, and then surround the test with an `if` statement that checks the USM support of the device?

EwanC: Thanks for the info about how CI testing works, I've updated the test to use the default device. Interestingly if I use `sycl::queue q` and leave out the in-order queue property I see an error when setting `ACPP_VISIBILITY_MASK="ocl"` so I've left the in-order property in.

The bug comes from the fact that I have 2 devices available, omp and OpenCL cpu, and the assert here gets hit https://github.com/AdaptiveCpp/AdaptiveCpp/blob/a0a8f2ba13c951ba64ac1c9052ba02423030d531/src/runtime/dag_direct_scheduler.cpp#L394-L397

From a little digging this is because the`memory_location`  associated of with the `malloc_host()` defaults to an omp device https://github.com/AdaptiveCpp/AdaptiveCpp/blob/develop/include/hipSYCL/sycl/device.hpp#L42 but that's not the same device as the node.

Shall I create an Issue for this?

EwanC:
> Thanks! It's not clear to my why we restricted this test to GPU in the first place. AdaptiveCpp mandates USM pretty much anywhere anyway. Perhaps because we thought that USM on host case was trivial?
> 
> Out of curiosity, why would an OpenCL CPU device not support host USM? It should be pretty trivial, no?

Thanks for the review, I agree OpenCL CPU host USM is indeed technically straightforward. However, as it's not a core part of the OpenCL standard vendors may for whatever reason not prioritize it's implementation, so it's availability is not guaranteed.

tests/sycl/queue.cpp
illuhad: Interesting. Yes, it would be great if you could open an issue. I can also reproduce it here.

The `omp` device is generically used to refer to the host, so a memcpy to the host would be referred to as a memcpy to the `omp` device. However, it tries to determine what the source and destination devices are using USM pointer queries.

So I'm guessing that we determine that the `malloc_host` pointer lives on the host, and then assume it's a normal host pointer - and then do host-to-host data transfer, but on an OpenCL queue.

But because the data transfer is host-to-host, the scheduler logic tries to redistribute the task to the OpenMP device and ultimately runs it there, thus creating the issue. This logic is mainly intended for accessors where we can have things like host accessors that are not directly tied to a specific device.

In the case of USM and in-order queue, that scheduling logic is bypassed, so we don't hit the issue there.

illuhad (approved):
Thanks!

PR #1948, opened 2025-12-23
[Prototype] cl_khr_unified_svm implementation of SYCL SVM

EwanC:
The OpenCL WG is currently developing the [cl_khr_unified_svm](https://github.com/KhronosGroup/OpenCL-Docs/pull/1282) extension with an explicit goal of being sufficient to implement SYCL USM ontop of. The APIs based on the current spec draft are now available in the upstream OpenCL-Headers repo, hence the commit bump to fetch these.

This PR adds support for using cl_khr_unified_svm as a AdaptiveCpp OpenCL USM backend. It was developed by using [an emulation layer](https://github.com/bashbaug/SimpleOpenCLSamples/blob/cl_khr_unified_svm/layers/99_svmplusplus/emulate.cpp) on top of an OpenCL CPU implementations with Intel USM extension support. Tested by running the AdaptiveCpp `usm_tests*` and `queue_tests*` unittests

I am intending to keep this PR as draft while the OpenCL APIs are subject to change until the extension is finished, hence the WIP title. So it's currently purpose is mostly to stress the OpenCL extensions design goal of being compatible with SYCL (although a separate DPC++ prototype does [already exist](https://github.com/intel/llvm/compare/sycl...bashbaug:llvm:cl_khr_unified_svm)).

### DIFF
--- a/doc/install-ocl.md
+++ b/doc/install-ocl.md
@@ -5,5 +5,6 @@ You will need an OpenCL implementation, and the OpenCL icd loader. The OpenCL li
 In order to generate correct code, AdaptiveCpp needs to use its own fork of the Khronos LLVM-SPIRV translator hosted at https://github.com/AdaptiveCpp/SPIRV-LLVM-Translator. It will *not* work with the upstream translator. When building, AdaptiveCpp will automatically fetch and build the llvm-spirv translator for the right LLVM version.
 
 The OpenCL backend can be enabled using `cmake -DWITH_OPENCL_BACKEND=ON` when building AdaptiveCpp.
-In order to run code successfully on an OpenCL device, it must support SPIR-V ingestion and the Intel USM (unified shared memory) extension. In a degraded mode, devices supporting OpenCL fine-grained system SVM (shared virtual memory) may work as well.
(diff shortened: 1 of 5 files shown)


illuhad:
Awesome, thank you! This looks good as far as I can see (not having tried it with the emulation layer). Let me know once the API on the OpenCL side is sufficiently mature so that we can merge this!

(At that point, we should also look into testing this in CI somehow - either via the emulation layer, or by using updated the OpenCL drivers in CI that would then hopefully implement that feature)

EwanC:
This work came up in the OpenCL WG F2F session yesterday, so I've rebased the branch and run it through the SYCL-CTS. On my Intel(R) Core(TM) Ultra 7 155H OpenCL driver version 2026.20.1.0.12_160000 I don't see all the CTS tests pass anyway with the Intel USM extension anyway.

However, I noticed a couple of regressions in `marray` when running using the emulation layer. However, I think it's a thread safety issues in the emulation layer and I've propose a patch here which resolves the
regressions I saw https://github.com/bashbaug/SimpleOpenCLSamples/pull/165

illuhad:
Heads up: Once https://github.com/AdaptiveCpp/AdaptiveCpp/pull/2054 is merged, the `ocl_usm` interface will require two additional functions to be implemented. You'll need to add them also to your `ocl_usvm_khr` class.

PR #1949, opened 2026-01-12
Use math vector library to improve host vectorization

thomasRoglin:
Hello, This pull request aims to improve the vectorization when using the generic-host target by compiling SYCL kernel with math vector libraries.

This works by finding available vector math library during cmake invocation (SLEEF or LIBMVEC or ARMPL), and adding relevant flags during JIT compilation of sycl kernels.

## How to use

Usage of math vector library can be enabled by setting `MATH_VECLIB` to `<ENABLE / DISABLE / LIBMVEC / SLEEF / ARMPL >`
 - Using `ENABLE` will automatically choose a library. If no library is found, a warning will be displayed
 - Using `<LIBMVEC / SLEEF / ARMPL>` will force the usage of the selected library. not finding the library will result in a crash.

## Performance improvements 

Here are some benchmarks from [HeCBench](https://github.com/zjin-lcf/HeCBench) that benefit greatly from the use of the math vector library. 

On **Sapphire Rapids** :

| bench                                                                              | *no vector library* | *libmvec*      | execution command                                                                                 | compilation command                                                      |
| ---------------------------------------------------------------------------------- | ------------------- | -------------- | ------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------ |
| [haversine](https://github.com/zjin-lcf/HeCBench/tree/master/src/haversine-sycl)   | `0.019433 (s)`      | `0.010052 (s)` | `OMP_NUM_THREAD=40 OMP_PROC_BIND=TRUE numactl -C 0-39 ./main ../geodesic-cuda/locations.txt 1000` | `acpp *.cpp --acpp-targets=generic -Ofast -o main -I ../haversine-cuda/` |
| [projectile](https://github.com/zjin-lcf/HeCBench/tree/master/src/projectile-sycl) | `0.004510 (s)`      | `0.003332 (s)` | `OMP_NUM_THREAD=40 OMP_PROC_BIND=TRUE numactl -C 0-39 ./main 4000`                                | `acpp *.cpp --acpp-targets=generic -Ofast -o main`                       |
| [gabor](https://github.com/zjin-lcf/HeCBench/tree/master/src/haversine-sycl)       | `82521.0 (us)`      | `48472.0 (us)` | `OMP_NUM_THREAD=40 OMP_PROC_BIND=TRUE numactl -C 0-39 ./main 8640 15360 1000`                     | `acpp *.cpp --acpp-targets=generic -Ofast -o main`                       |

On **NVIDIA Grace** :

| bench                                                                              | *no vector library* | *armpl*        | sleef               | execution command                                                                                   | compilation command                                                      |
| ---------------------------------------------------------------------------------- | ------------------- | -------------- | ------------------- | --------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------ |
| [haversine](https://github.com/zjin-lcf/HeCBench/tree/master/src/haversine-sycl)   | `0.007726 (s)`      | `0.003287 (s)` | `0.006824 (s)`      | `OMP_NUM_THREAD=72 OMP_PROC_BIND=TRUE numactl -C 72-143 ./main ../geodesic-cuda/locations.txt 1000` | `acpp *.cpp --acpp-targets=generic -Ofast -o main -I ../haversine-cuda/` |
| [projectile](https://github.com/zjin-lcf/HeCBench/tree/master/src/projectile-sycl) | `0.001542 (s)`      | `0.000937 (s)` | `0.001223 (s)`      | `OMP_NUM_THREAD=72 OMP_PROC_BIND=TRUE numactl -C 72-143 ./main 4000`                                | `acpp *.cpp --acpp-targets=generic -Ofast -o main`                       |
| [gabor](https://github.com/zjin-lcf/HeCBench/tree/master/src/haversine-sycl)       | `28361.0 (us)`      | `11507.0 (us)` | `27609.0(us)` | `OMP_NUM_THREAD=72 OMP_PROC_BIND=TRUE numactl -C 72-143 ./main 8640 15360 1000`                     | `acpp *.cpp --acpp-targets=generic -Ofast -o main`                       |

## Implementation details 

The CMAKE part of the code is inspired by [what is done in POCL ](https://github.com/pocl/pocl/blob/41a99f3ecdc3338b80a32739a3247960dfe1ad40/CMakeLists.txt#L1060C1-L1077C1)

At JIT time, the chosen math vector library will be searched either at its original location (found during cmake invocation), or in `path/to/acpp/lib/`, where it should be after deployment. This is the same way that `opt/lld/llc` are found. 

Armpl and Sleef are only available on aarch64, libmvec is available on x86 and on aarch64 with a recent LLVM version (21+). Also the libmvec might not contains every math function if the GLIBC is too old, that why we are checking that recent functions are available (`_ZGVnN2v_tanf`). 

When generating default `host` builtins (not to be confused with `host-fast` builtins), `-fno-math-errno -ffp-contract=fast` flags are added : 
 - `ffp-contract=fast` is used by default since https://github.com/AdaptiveCpp/AdaptiveCpp/pull/1351, so it makes sense to generate default builtins with `ffp-contract=fast`
 - `errno` cannot be used in a parallel sycl kernel, so `-fno-math-errno` flag should be added in default builtins. 

## Potential improvements

I think it could be interesting to be able to choose the vector library at runtime with an environment variable. 
However, there is currently no system for reading environment variables in this part of the code. Should we duplicate `getEnvironmentVariable` from `src/compiler/llvm-to-backend/LLVMToBackend.cpp`


src/compiler/CMakeLists.txt
```diff
@@ -69,6 +69,131 @@ if(WITH_ACCELERATED_CPU)
...
+elseif(MATH_VECLIB STREQUAL "LIBMVEC")
+  set(MATH_VECLIB_LIBMVEC true)
+elseif(MATH_VECLIB STREQUAL "ARMPL")
+  set(MATH_VECLIB_AMATH true)
+elseif(MATH_VECLIB STREQUAL "SVML")
+  #TODO
+  message(WARNING "SVML is currently not supported by AdaptiveCpp")
+else()
+  message(SEND_ERROR "Invalid value for MATH_VECLIB: ${MATH_VECLIB}"
+                     "Valid options are: ENABLE, DISABLE, LIBMVEC, SLEEF, ARMPL, SVML.")
```
fodinabor: If SVML is not supported, maybe don't list it here :)

src/compiler/llvm-to-backend/Utils.cpp
```diff
@@ -104,6 +104,29 @@ std::string getOptPath() {
   return path;
 }
 
+#ifdef ACPP_HOST_ENABLE_VECTORIZE_MATH
+std::string getMathVecLibPath() {
+  static std::string path;
+  if(!path.empty())
+    return path;
+
+  const auto lib_path = common::filesystem::get_lib_directory();
```
fodinabor, line 112: Would it make sense to install / copy the veclib to the lib_path?
-> at least for packaging, we'd need this.

fodinabor (commented):
Cool, love that you're working on this!

thomasRoglin:
I just added the SVML math vector library support. It seems that `libsvml.so` is dynamically linked to `libintc.so`, so I am also adding a linkage to it (`LldInvocation.push_back("-lintlc");` ).

src/compiler/CMakeLists.txt
thomasRoglin: SVML is now supported, so I am resolving this

src/compiler/llvm-to-backend/Utils.cpp
thomasRoglin: If i understand you comment, this is what I am doing by adding `  \"${MATH_VECLIB_PATH}\" : \"/\"` into ACPP_CORE_DEPLOYMENT_MANIFEST. 
When calling `acpp-deploy`, the math vector lib will be copied into the deployment location, which seems to be `lib_path `.

fodinabor: Ah... okay.. I actually don't know much about that deploy infra.. sounds good :)
Thanks for the work... lgtm overall.. :)

thomasRoglin:
Thanks for the review @fodinabor 
Before marking the PR as ready, there are still a few things left to do:
 - Do we want to add tests for this feature? Are we testing every math vector library?
 - This feature should be properly documented.
 - I think being able to choose at runtime which math vector library we want would make performance testing much easier (it would be a pain to have to recompile Acpp every time we want to disable or change the math vector library).

illuhad:
> so I am also adding a linkage to it (LldInvocation.push_back("-lintlc"); ).

Should we then also add libintlc to the deployment manifest if SVML is used (I assume it is legally redistributable?)?

> Do we want to add tests for this feature? Are we testing every math vector library?

If it's reasonably easy, I suppose we could install math vector libraries in CI, and see if those work.
Looks like CI has some hiccups in general though at the moment that we/I should fix first.

> I think being able to choose at runtime which math vector library we want would make performance testing much easier (it would be a pain to have to recompile Acpp every time we want to disable or change the math vector library).

This sounds useful. To your earlier question about `getEnvironmentVariable`: I think we could perhaps move that function and `getEnvironmentVariableOrDefault` from the same file to `Utils.hpp` in the `llvm-to-backend` directory?

thomasRoglin:
I just forced pushed the modification that allow the user to choose the math vectlib at runtime. 

Now cmake will try to find SLEEF, Armpl, SVML, LIBMVEC, without stopping when one is found. Every lib found is added to the deployment manifest (I am not sure of the legality redistribution of svml, I will look into it).

The use can now fill the `ACPP_S2_MATH_VECTOR_LIBRARY` env variable to `0/NO/SLEEF/ARMPL/LIBMVEC/SVML`. If not specified, the default math veclib will be the last one found by the cmake invocation (I dont think its necessary to add a cmake option to force a default lib )

I choose to display warning instead of crashing the application when `ACPP_S2_MATH_VECTOR_LIBRARY` is not filled correctly. Do not hesitate to tell me if you want to change any of the following behaviors:
 - The user request something that does not exist 
 - The user requested a backend that is not enabled (ex : SVML on aarch64)
 - The user requested an activated backend, but the library could not be found at runtime for an unknown reason.


I cannot test this patch on Apple or Windows,  so I'm disabling it on both platforms.

thomasRoglin:
I just added some documentation for `ACPP_S2_VECTOR_MATH_LIBRARY`. 

I also switch the naming "math vector library" to "vector math library"

illuhad:
> The use can now fill the ACPP_S2_MATH_VECTOR_LIBRARY env variable to 0/NO/SLEEF/ARMPL/LIBMVEC/SVML. If not specified, the default math veclib will be the last one found by the cmake invocation (I dont think its necessary to add a cmake option to force a default lib )

One thing that just came to my mind: The kernel cache will not be aware of environment variables that are only used directly inside JIT. So if a user sets `ACPP_S2_MATH_VECTOR_LIBRARY=X`, and then later sets it to `Y`, the kernel will not be recompiled if it already is in the JIT cache.

Whether or not this is a problem might depend on whether we expect end users to regularly fiddle around with this flag. If it's mainly for experimentation/developers it might be fine as is.

If we *do* want the JIT cache to distinguish different vector library options:
- The environment variable is read during kernel submission in the OpenMP backend in the runtime;
- We add a build option for it, which the OpenMP backend then sets in `kernel_configuration`. This will cause a different kernel hash to be generated for different vector libraries, and thus allows us to distinguish them in the JIT cache.
- Build options are forwarded to the JIT compiler, so we can then just read that setting inside `LLVMToHost`.

> I cannot test this patch on Apple or Windows, so I'm disabling it on both platforms.

This is fine for me.

@blinkfrog - would you be interested in this on Windows? But I assume you don't need it because you cannot use `generic` target due to the redistributability issue, right?

thomasRoglin:
>Should we then also add libintlc to the deployment manifest if SVML is used (I assume it is legally redistributable?)?

About the legality of the redistribution of SLEEF, Armpl and SVML are legaly redistributable (see [Intel End User License Agreement (EULA)](https://www.intel.com/content/www/us/en/content-details/777700/intel-end-user-license-agreement-for-developer-tools.html), [Arm Performance Libraries End User License Agreement](https://developer.arm.com/documentation/109611/1-0/End-User-License-Agreement--EULA-?lang=en) and [Boost Software License, Version 1.0](https://www.boost.org/LICENSE_1_0.txt) ).  
I added some information inside `deployment.md` that points to their respective EULA. Please tell me if you want me to add them elsewhere. 

>One thing that just came to my mind: The kernel cache will not be aware of environment variables that are only used directly inside JIT. So if a user sets `ACPP_S2_MATH_VECTOR_LIBRARY=X`, and then later sets it to `Y`, the kernel will not be recompiled if it already is in the JIT cache.

I think this is acceptable behavior; however, if there is feedback requesting that the JIT distinguish between different vector library options, I would be happy to implement it.

On my side, the PR is ready, so I am removing the Draft tag.

illuhad:
> I added some information inside deployment.md that points to their respective EULA. Please tell me if you want me to add them elsewhere.

I think that's fine - thanks!

> I think this is acceptable behavior; however, if there is feedback requesting that the JIT distinguish between different vector library options, I would be happy to implement it.

I would be fine with it if we could fix it easily when a user complains - but I'm not sure whether there might not be some friction then.

If we wanted to transition in the future to a design where the runtime can distinguish differently configured kernels in the cache, then the runtime would need to manage the environment variable. But runtime environment variables have different naming conventions than environment variables purely for the stage 2 JIT - those have `_S2_` in the name.

So if we wanted to make that change in the future, we would inevitably end up with a deprecated `_S2_` environment variable. Because of that, I currently suspect that it might be better to do things "right" here - would it be okay for you to add the vector library configuration into the kernel hash, as discussed? Assuming it's not too much work.

This is a great PR and I don't want to hold it up for too long :)

blinkfrog:
> @blinkfrog - would you be interested in this on Windows? But I assume you don't need it because you cannot use `generic` target due to the redistributability issue, right?

In fact I think this would still be useful on Windows, even if for `omp`-target-only.
I think I need to make some tests with a simple OpenMP program and check if `-fveclib=` would work with SVML or SLEEF on Windows.

thomasRoglin:
I agree that adding a new environment variable that is likely to be deprecated is not appealing.
So here is a modified version of the patcht hat retriggers the JIT when the vector math library is changed. 

I added a new setting : `omp_vector_math_library` and a new kernel  configuration : `host_vector_math_library`. 
Env variable is now `ACPP_OMP_VECTOR_MATH_LIBRARY`. 

---

I am not entirely sure about the naming convention I used: `host_*` for the configuration and `omp_*` for the setting. Should they both use `omp` or both use `host`?

I also have some doubts about the environment variable name, should it include `JIT` to make it clear that it's for sscp-omp (`ACPP_JIT_OMP_VECTOR_MATH_LIBRARY`) ?

---

Otherwise, I ran into a few issues related to the default vector library, which is defined at AdaptiveCpp compile time via the `DEFAULT_VEC_MATH_LIB` macro.

`DEFAULT_VEC_MATH_LIB` now needs to be defined in `src/runtime` rather than in `src/compiler`. However, the default vector library is determined in `src/compiler/CMakeLists.txt`. To make this variable available to `src/runtime/CMakeLists.txt`, I chose to pass it through CMake’s global variable mechanism. (`src/compiler`  being processed before `src/runtime` makes this method works. ).

I also chose not to use the macro directly in `settings.hpp` (with
`get_configuration_or_default<setting::omp_vector_math_library>(DEFAULT_VEC_MATH_LIB);`) because that would require every CPP file including this header to have the `DEFAULT_VEC_MATH_LIB` macro defined.

An alternative would be to move the vector library detection logic to the root `CMakeLists.txt`, but I dont want to overload it more than it already is. (If you think that would have been a better solution, I can make the change.  )

src/runtime/omp/omp_queue.cpp
```diff
@@ -469,6 +469,13 @@ result omp_queue::submit_sscp_kernel_from_code_object(
   for(const auto& opt : kernel_info->get_compilation_options())
     _config.set_build_option(opt.first, opt.second);
 
+  std::string omp_vector_math_library =
+      application::get_settings().get<setting::omp_vector_math_library>();
+  if(omp_vector_math_library.empty())
+    omp_vector_math_library = DEFAULT_VEC_MATH_LIB;
+  _config.set_build_option(kernel_build_option::host_vector_math_library,
+      omp_vector_math_library);
```
illuhad: I would perhaps suggest to treat the math library setting as `int`/enum value instead of string (which would need a change here as well as in `settings` and potentially `LLVMToHost`). Otherwise, we need to construct a new string for every kernel launch to calculate the hash.

illuhad (commented):
> I am not entirely sure about the naming convention I used: host_* for the configuration and omp_* for the setting. Should they both use omp or both use host?

Hm. Normally we use `omp` to refer to one specific runtime backend, and `host` to refer to the backend of the JIT compiler. I would be inclined to think that this feature is more related to *targeting host* rather than doing something with OpenMP. So I'd lean toward using `host` for both.

> I also have some doubts about the environment variable name, should it include JIT to make it clear that it's for sscp-omp (ACPP_JIT_OMP_VECTOR_MATH_LIBRARY) ?

Generally the environment variables reflect the names in the setting enum (except that runtime configuration settings get an `rt_` added).

Looking at what we have currently:

```
  debug_level,
  scheduler_type,
  visibility_mask,
  dag_req_optimization_depth,
  mqe_lane_statistics_max_size,
  mqe_lane_statistics_decay_time_sec,
  default_selector_behavior,
  hcf_dump_directory,
  persistent_runtime,
  sscp_failed_ir_dump_directory,
  gc_trigger_batch_size,
  ocl_no_shared_context,
  ocl_show_all_devices,
  no_jit_cache_population,
  adaptivity_level,
  jitopt_iads_relative_threshold,
  jitopt_iads_relative_eviction_threshold,
  jitopt_iads_relative_threshold_min_data,
  enable_allocation_tracking,
  omp_vector_math_library
  ```
It seems that we're not particularly consistent:
`sscp_failed_ir_dump_directory`, `no_jit_cache_population`, `adaptivity_level` and `jitopt_iads_` all refer to the JIT infrastructure and its configuration. `enable_allocation_tracking` also impacts JIT code generation, although somewhat indirectly.

`jitopt_iads_` is framed as a specific option for the IADS optimization pass - perhaps something like this most closely resembles vector math library choice? In that case, the setting should be `jitopt_host_vector_math_library`, and the associated environment library `(ACPP_)JITOPT_HOST_VECTOR_MATH_LIBRARY`.

> DEFAULT_VEC_MATH_LIB now needs to be defined in src/runtime rather than in src/compiler. However, the default vector library is determined in src/compiler/CMakeLists.txt. To make this variable available to src/runtime/CMakeLists.txt, I chose to pass it through CMake’s global variable mechanism. (src/compiler being processed before src/runtime makes this method works. ).

I think we already have some dependencies on the order in which cmake files are processed. While not pretty, I think such a solution is fine for now until someone feels motivated to clean up cmake as a whole.

> I also chose not to use the macro directly in settings.hpp (with
get_configuration_or_default<setting::omp_vector_math_library>(DEFAULT_VEC_MATH_LIB);) because that would require every CPP file including this header to have the DEFAULT_VEC_MATH_LIB macro defined.

That's okay. One alternative could perhaps be to move the default library selection from `omp_queue` to `LLVMToHost`: If the user does not set a math library, just don't configure it in the kernel configuration, and let the JIT compiler do whatever it wants to do as default.

In theory, one drawback is that then we could in principle run again into kernel cache collisions if the user rebuilds AdaptiveCpp with a different default vector lib. But we already do not guarantee kernel cache compatibility when the AdaptiveCpp stack is modified - otherwise we'd also have to put e.g. LLVM versions into the kernel hash. So I don't think we'd lose something in practice with that approach.

File: bin/acpp [hunk 4]
```diff
@@ -1233,6 +1233,8 @@ class acpp_config:
     if opt_level > 1:
       if not has_fast_math and not has_ffp_contract:
         opt_args.append("-ffp-contract=fast")
+      if not has_fast_math:
+        opt_args.append("-fno-math-errno")
```
illuhad: Can you elaborate on this change? I understand that we can freely set `-fno-math-errno` for our builtins, since they are only used in kernels, and kernels cannot use errno.

However, these here are the default optimization flags for clang, which especially in the case of SSCP (since it's a unified host-device compiler) will be used used for the clang invocation as a whole, and thus also affect non-kernel code.

It seems to me that this change might break user host code, particularly for those users who do *not* want fast math? :D

thomasRoglin: You are correct, I was not really thinking of non-kernel code. I will revert that.

thomasRoglin:
>That's okay. One alternative could perhaps be to move the default library selection from omp_queue to LLVMToHost: If the user does not set a math library, just don't configure it in the kernel configuration, and let the JIT compiler do whatever it wants to do as default.

>In theory, one drawback is that then we could in principle run again into kernel cache collisions if the user rebuilds AdaptiveCpp with a different default vector lib. But we already do not guarantee kernel cache compatibility when the AdaptiveCpp stack is modified - otherwise we'd also have to put e.g. LLVM versions into the kernel hash. So I don't think we'd lose something in practice with that approach.

If this drawback is fine with you, I think I'll reimplement it this way. ( I also think that keeping the default library selection in omp_queue will not work when making the setting an `enum`.)

illuhad:
> If this drawback is fine with you, I think I'll reimplement it this way. ( I also think that keeping the default library selection in omp_queue will not work when making the setting an enum.)

Fine with me! :)

blinkfrog:
> In fact I think this would still be useful on Windows, even if for `omp`-target-only. I think I need to make some tests with a simple OpenMP program and check if `-fveclib=` would work with SVML or SLEEF on Windows.

I just have tested SVML with a simple OpenMP-program, and it works, autovectorizer calls this library. SLEEF doesn't work - it isn't supported by Clang for `x86-64`.

And, if I understand correctly, I don't need this PR for `omp` target. I just tried to add the needed flags and paths to a test AdaptiveCpp project, and it worked with SVML!

However, if this PR supported SVML on Windows, it would be great - not all users face that legal issue (with distributing apps), so they still could use `generic` target.

bin/acpp
thomasRoglin: This has now been removed.

thomasRoglin:
I just pushed a change to make the setting an enum.
I also renamed a few variables as agreed (omp -> host).

Would you like me to squash my commits when everything is ok ?

thomasRoglin:
I'm not sure why `generic` wouldn't work on Windows. 
The redistribution issue only applies to the deployment, but for a standard build without deployment, it should be possible (once support for Windows has been added) to use vectorized math libraries with the `generic` target on Windows.

include/hipSYCL/runtime/settings.hpp
```diff
@@ -24,6 +24,15 @@ namespace rt {
 
 enum class scheduler_type { direct, unbound };
 enum class default_selector_behavior { strict, multigpu, system };
+enum class jitopt_host_vector_math_library {
+  empty = 0,
+  invalid = 1,
+  none = 2,
```
illuhad: Do we need `empty`, `invalid`, as well as `none`?

As far as I can see, the code currently does not generate `invalid`. Why can't we remove it?

For `empty` and `none`, I assume the rationale here is that we need to represent the case when no vector math library was configured. Couldn't we just remove these two enum values and instead just store an `std::optional<host_vector_math_library>` if we need to represent the unconfigured case?

thomasRoglin:
Moving `getEnvironmentVariable` from `LLVMToBackend.cpp` to `Utils.cpp` is no longer necessary in the PR. However, is there a problem with keeping it in `Utils.cpp`? This function may be useful in the future. I will probably use it on my local builds to quickly test stuff.

include/hipSYCL/compiler/llvm-to-backend/host/LLVMToHost.hpp
```diff
@@ -20,6 +20,16 @@
 namespace hipsycl {
 namespace compiler {
 
+enum class host_vector_math_library {
+  empty = 0,
+  invalid = 1,
+  none = 2,
```
illuhad: See comment on `jitopt_host_vector_math_library` enum.

illuhad (commented):
Thanks @thomasRoglin, I have one last comment -- apart from that it looks great :-)

> Would you like me to squash my commits when everything is ok ?

You can do that if you like, otherwise I can also just conveniently press the squash-and-merge button :-)

> I'm not sure why generic wouldn't work on Windows.
The redistribution issue only applies to the deployment, but for a standard build without deployment, it should be possible (once support for Windows has been added) to use vectorized math libraries with the generic target on Windows.

I agree.

> Moving getEnvironmentVariable from LLVMToBackend.cpp to Utils.cpp is no longer necessary in the PR. However, is there a problem with keeping it in Utils.cpp? This function may be useful in the future. I will probably use it on my local builds to quickly test stuff.

Yes, this makes sense to me -- it's a general purpose utility function that might be useful also in other parts of the compiler outside of `LLVMToBackend.cpp`.

EDIT: @thomasRoglin, also, please rebase on current `develop`. We have committed some fixes for CI to `develop` that should remove the CI failures for this PR.

include/hipSYCL/runtime/settings.hpp
thomasRoglin: I think `none`, `empty` and `invalid` are all required. :
- `none` is produced when the user explicitly ask for no vectorization (`ACPP_JITOPT_HOST_VECTOR_MATH_LIBRARY=none`). 
- `empty` is the default setting that is produced when `ACPP_JITOPT_HOST_VECTOR_MATH_LIBRARY` is unset. 
  ```
  _jitopt_host_vector_math_library =
        get_configuration_or_default<setting::jitopt_host_vector_math_library>(
            jitopt_host_vector_math_library::empty);
  ```
  in `applyBuildOption` from `LLVMToHost.cpp`, if `VectorMathLibary==empty` is empty, I use the default veclib option defined in cmake. 
- `invalid` is produced in `stettings.cpp` the when converting `string` to `enum` when `ACPP_JITOPT_HOST_VECTOR_MATH_LIBRARY` is set to a wrong value. I could display display an error message here, but I think displaying a more fine tuned warning message in `LLVMToHost.cpp` is clearer.

So I'm not sure to see the advantage of using std::optional for `jitopt_host_vector_math_library`.

thomasRoglin:
some CI builds are failing with this error message :


```
2026-02-04T10:21:57.3626410Z /Users/runner/work/AdaptiveCpp/AdaptiveCpp/AdaptiveCpp/src/compiler/llvm-to-backend/host/LLVMToHost.cpp:539:52: error: expected unqualified-id
2026-02-04T10:21:57.3627000Z   539 |       VectorMathLibary = host_vector_math_library::DEFAULT_VEC_MATH_LIB;
2026-02-04T10:21:57.3627310Z       |                                                    ^
2026-02-04T10:21:57.3627590Z <command line>:12:30: note: expanded from macro 'DEFAULT_VEC_MATH_LIB'
2026-02-04T10:21:57.3627860Z    12 | #define DEFAULT_VEC_MATH_LIB "none"
2026-02-04T10:21:57.3628050Z       |                              
```

I will look into it. 

EDIT: It seems that the problem was the quotation marks. CI is fine now

include/hipSYCL/runtime/settings.hpp
illuhad: Ah, I can see how `none` is needed -- I forgot that users might to explicitly request no vector math library, even if support for one would be available.

> empty is the default setting that is produced when ACPP_JITOPT_HOST_VECTOR_MATH_LIBRARY is unset.

Okay, but couldn't you also store `std::optional<jitopt_host_vector_math_library>` in settings, and then
```
_jitopt_host_vector_math_library =
      get_configuration_or_default<setting::jitopt_host_vector_math_library>(
         std::optional<jitopt_host_vector_math_library>{/*empty optional*/});
```

Then, the default value in `settings` will be an optional with no value stored.

> invalid is produced in stettings.cpp the when converting string to enum when ACPP_JITOPT_HOST_VECTOR_MATH_LIBRARY is set to a wrong value. I could display display an error message here, but I think displaying a more fine tuned warning message in LLVMToHost.cpp is clearer.

Why delay the error to JIT?

I think for other environment variables, the behavior is rather that you get a warning during parsing, and then a default value is used -- or the setting is just ignored.

thomasRoglin: >Why delay the error to JIT? 

If I understand correctly from this comment from `src/runtime/settings.cpp`, it seems that I can't use `HIPSYCL_DEBUG_WARNING` here,  so the error message is not displayed as a warning. 
```
} else if (name == "ocl" || name == "opencl") {
      backend = rt::backend_id::ocl;
    } else {
      istr.setstate(std::ios_base::failbit);
      // Don't use HIPSYCL_DEBUG_WARNING, it will cause recursive init error.
      std::cout << "'" << name << "' is not a known backend name." << std::endl;
      break;
    }

```

It is also not possible to detect from here the available veclib, so I will not be able to do a fine tuned message like :
```
      HIPSYCL_DEBUG_WARNING << "LLVMToHost: Invalid option for ACPP_JITOPT_HOST_VECTOR_MATH_LIBRARY. "
                            << "Available options are: none"
#ifdef LIBMVEC_AVAILABLE
                            << ", libmvec"
#endif
#ifdef SVML_AVAILABLE
                            << ", svml"
#endif
#ifdef SLEEF_AVAILABLE
                            << ", sleef"
#endif
#ifdef ARMPL_AVAILABLE
                            << ", armpl"
#endif
                            << ".\n";
```

However, if you are okay with these two drawbacks, I will do the modification.

illuhad: > If I understand correctly from this comment from src/runtime/settings.cpp, it seems that I can't use HIPSYCL_DEBUG_WARNING here, so the error message is not displayed as a warning.

That's true. As you point out, there is prior work that does the same thing, such as the OpenCL device mask handling. So I don't see this as a problem.

> It is also not possible to detect from here the available veclib, so I will not be able to do a fine tuned message like

That's okay. Other error messages for incorrect environment variables are also not super helpful either :)
If users put in a veclib that's not available, they'll still get a JIT-time error.

> However, if you are okay with these two drawbacks, I will do the modification.

Thanks!

thomasRoglin: both done ! :)

illuhad (approved):
Merci beaucoup! :-)

illuhad:
@blinkfrog 

> However, if this PR supported SVML on Windows, it would be great - not all users face that legal issue (with distributing apps), so they still could use generic target.

I would like to merge now - Windows support could be a subsequent PR :-)

thomasRoglin:
Hello @illuhad , thanks for accepting the PR yesterday :) 
While retesting the vector math library, I found that the `amath` veclib option is broken because I mixed up amath/armpl in the names of some function/macro. 
Could you please also merge the  commit I just added? (I'm not sure if I should have opened a new PR.)

illuhad:
Please open new PR - I don't want to push directly to `develop` :)

PR #1951, opened 2026-01-27
[CI] Add force option to prime-docker action.

fodinabor:
@illuhad 
Merging this, should (enable) rebuilding the images for 2 reasons:
- we changed sth in create-ci-docker, which automatically triggers a rebuild
- You now have the option to `force` with the prime-docker action.


illuhad (approved):
Thanks! Let's see if it works :)

PR #1952, opened 2026-01-27
Don't use double computations in float-only code

resetius:
These functions emit double-float casts in llvm-ir. I want to get rid of this casts in backends that does not support of doubles (Metal backend)

### DIFF
--- a/include/hipSYCL/sycl/libkernel/sscp/builtins.hpp
+++ b/include/hipSYCL/sycl/libkernel/sscp/builtins.hpp
@@ -454,7 +454,7 @@ HIPSYCL_BUILTIN T __acpp_clamp(T x, T minval, T maxval) noexcept {
 
 template<class T>
 HIPSYCL_BUILTIN T __acpp_degrees(T x) noexcept {
-  return (180.f / M_PI) * x;
+  return static_cast<T>(180.0 / M_PI) * x;
 }
 
 // __acpp_max() and __acpp_min() are handled by the overloads from the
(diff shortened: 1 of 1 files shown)


include/hipSYCL/sycl/libkernel/host/builtins.hpp
```diff
@@ -693,7 +693,7 @@ HIPSYCL_BUILTIN T __acpp_mix(T x, T y, T a) noexcept {
 
 template<class T>
 HIPSYCL_BUILTIN T __acpp_radians(T x) noexcept {
-  return (M_PI / 180.f) * x;
+  return static_cast<T>(M_PI / 180) * x;
```
illuhad: This file is only for the host backend - this should not need change since double is always available on the host thanks to C++.

include/hipSYCL/sycl/libkernel/sscp/builtins.hpp
```diff
@@ -454,7 +454,7 @@ HIPSYCL_BUILTIN T __acpp_clamp(T x, T minval, T maxval) noexcept {
 
 template<class T>
 HIPSYCL_BUILTIN T __acpp_degrees(T x) noexcept {
-  return (180.f / M_PI) * x;
+  return static_cast<T>(180.0 / M_PI) * x;
```
illuhad, line 457: Is there any negative impact due to widening when T=float? Would it be better if we did something like:

```
return static_cast<T>(180) / static_cast<T>(M_PI)
```
?

include/hipSYCL/sycl/libkernel/host/builtins.hpp
resetius: agree

include/hipSYCL/sycl/libkernel/sscp/builtins.hpp
resetius: Thanks, but I don’t think casting both operands makes sense here. 180.0 / M_PI is a compile-time constant, so the division happens in double anyway, and we only need a single cast to T for the final factor. Casting the operands to T could actually reduce precision (especially for float / half-like types) by doing the division in T instead of double and then propagating that error into the result.

illuhad (approved):
Thanks!

PR #1953, opened 2026-01-28
Fix macos workflow

resetius:


resetius:
now compilation is working : https://github.com/resetius/AdaptiveCpp/pull/1

illuhad (approved):
Thanks!

PR #1956, opened 2026-01-29
[libkernel][host] Add glibc version check for fmaxmag/fminmag

cbyrohl:
fmaxmag and fminmag were added in glibc 2.25. On older systems (e.g., RHEL 7 with glibc 2.17), these functions are not available, causing build failures.

Add __GLIBC_PREREQ(2, 25) check to skip these builtins on older glibc versions.

### DIFF
--- a/src/libkernel/sscp/host/math.cpp
+++ b/src/libkernel/sscp/host/math.cpp
@@ -195,9 +195,11 @@ HIPSYCL_SSCP_MAP_HOST_FLOAT_BUILTIN(log1p)
 HIPSYCL_SSCP_MAP_HOST_FLOAT_BUILTIN(logb)
 HIPSYCL_SSCP_MAP_HOST_FLOAT_BUILTIN3_NAME(mad,fmaf,fma)
 
-#if !defined(_WIN32) && !defined(__APPLE__) && defined(__GLIBC__) && __GLIBC_PREREQ(2, 25)
+#if !defined(_WIN32) && !defined(__APPLE__)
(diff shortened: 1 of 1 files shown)


src/libkernel/sscp/host/math.cpp
```diff
@@ -195,7 +195,7 @@ HIPSYCL_SSCP_MAP_HOST_FLOAT_BUILTIN(log1p)
 HIPSYCL_SSCP_MAP_HOST_FLOAT_BUILTIN(logb)
 HIPSYCL_SSCP_MAP_HOST_FLOAT_BUILTIN3_NAME(mad,fmaf,fma)
 
-#if !defined(_WIN32) && !defined(__APPLE__)
+#if !defined(_WIN32) && !defined(__APPLE__) && defined(__GLIBC__) && __GLIBC_PREREQ(2, 25)
```
illuhad: Small note: Currently, this would also prevent the builtins from being available if something else than glibc is used.

I don't know if there are other libcs that users might want to use, which do have these builtins (?).
If so, we could do something like:
```
#if !defined(_WIN32) && !defined(__APPLE__)
#if !defined(__GLIBC__) || __GLIBC_PREREQ(2,25)
...
#endif
#endif
```
But if those hypothetical other libc implementations don't have them either, it's better to exclusively enable them for the configuration where we know it works (glibc > 2.25), as you do now.

cbyrohl: As suggested, I changed to your permissive approach, preserving previous behavior for non glib.

illuhad (approved):
Thanks!

PR #1959, opened 2026-01-30
[SSCP][MacOS] Fix MacOS dlopen failure by linking libSystem

resetius:
This PR fixes the following dlopen errors on MacOS:
```
[AdaptiveCpp Warning] [omp_sscp_executable_object] Could not load library: /Users/aozeritsky/.acpp/apps/global/jit-cache/9615315375232329766.10473829325427112559.jit.so (dlopen(/Users/aozeritsky/.acpp/apps/global/jit-cache/9615315375232329766.10473829325427112559.jit.so, 0x0002): tried: '/9615315375232329766.10473829325427112559.jit.so' (no such file), '/Users/aozeritsky/inst/lib/9615315375232329766.10473829325427112559.jit.so' (no such file), '/Users/aozeritsky/.acpp/apps/global/jit-cache/9615315375232329766.10473829325427112559.jit.so' (missing LC_LOAD_DYLIB (must link with at least libSystem.dylib)), '/System/Volumes/Preboot/Cryptexes/OS/Users/aozeritsky/.acpp/apps/global/jit-cache/9615315375232329766.10473829325427112559.jit.so' (no such file), '/Users/aozeritsky/.acpp/apps/global/jit-cache/9615315375232329766.10473829325427112559.jit.so' (missing LC_LOAD_DYLIB (must link with at least libSystem.dylib)))
```


illuhad (commented):
Awesome! @fodinabor and I discussed this very issue just today :D

Since this used to work in CI before the MacOS version was updated -- do you happen to know how portable this solution is across MacOS versions? Do we need to introduce a minimum supported version?

resetius:
Unfortunately my pr did not fix macOS workflow since you have 2 macOS workflow files, I missed the second one. I ll try to fix it again :)

fodinabor:
Merging this and that should give you (I hope) a working version?
https://github.com/fodinabor/hipSYCL/tree/ci/fix-macos
(Minus the debug commits at the end)

resetius:
Yes, It should help

fodinabor:
Sorry for the accidental initial merge :)
I just added the workflow change as well, so the AdaptiveCpp-LLVM macOS workflow _should_ now also pass, hopefully.

Thanks for the libSystem fix!

.github/workflows/macos-acppllvm.yml
```diff
@@ -40,6 +40,7 @@ jobs:
         cmake ../llvm/llvm -GNinja \
               -DCMAKE_BUILD_TYPE=Release \
               -DCMAKE_INSTALL_PREFIX=${GITHUB_WORKSPACE}/install  \
+              -DCMAKE_OSX_SYSROOT="$(xcrun --show-sdk-path)" \
```
illuhad, line 43: Is `xcrun` a new dependency? If so, we should also document it (and ideally check for it in cmake when we build `LLVMToHost` for Mac).

resetius: xcrun is part of the standard macos toolchain. It is installed together with Xcode / Command Line Tools, and without it the macos SDK is not available, so building on macOS doesn't really work at all.

If the toolchain is missing or broken, cmake will most likely fail earlier anyway.

If you think this still needs to be explicitly documented, please let me know where you'd like that documented and I can add it

illuhad: Alright - no, sounds good, thanks :-)

illuhad (approved):
Awesome, thanks!

Issue #864, opened 2022-11-13
Implementation of Metal backend


philipturner:
This is a progress tracker for implementation of the hipSYCL backend. I figured it's most appropriate to extract it into a separate thread, and not flood #460.

Preliminary Research:
- [x] Completing GROMACS optimizations unrelated to hipSYCL
- [x] Exploring [MoltenCL](https://github.com/philipturner/MoltenCL), decided not to pursue that path
- [x] Exploring [SPIR-V -> AIR transpiler](https://github.com/philipturner/air-v), requires investing too much time

Milestones:
- [x] Bare-bones hipSYCL backend
- [ ] Feature-complete hipSYCL backend

Planned Features:
- [ ] USM pointers (all types)
- [ ] [atomic64](https://github.com/KhronosGroup/SYCL-Docs/issues/326) emulation
- [ ] Optimized subgroup/workgroup reductions

Optional Features:
- [ ] [float64](https://github.com/philipturner/metal-float64) emulation
- [ ] Simdgroup matrix and shuffle-and-fill instructions
- [ ] Port to iOS

FCLC:
I was thinking about this topic whilst trying to get HipSYCL going via both brew and SPACK. 

Specifically on the topic of a metal transpiler, is it worth looking at the Khronos MoltenVK project? https://github.com/KhronosGroup/MoltenVK 

I'd briefly brought up Sylkan as more of a place holder than anything in #839, with the main issues being around Spir-V (amongst others).

As for FP64, during a discussion on Mastodon with Karol Herbst of RedHat on the topic of a Mesa SYCL implementation using RUSTICL and the problems he was also facing around FP64, I'd made the suggestion of using/adapting the MIT licensed code already available from Intel in their OpenCL compute runtime. (I'm having trouble finding the discussion, will edit it in when I come across it)

philipturner:
> MIT licensed code already available from Intel in their OpenCL compute runtime. (I'm having trouble finding the discussion, will edit it in when I come across it)

[...]

PR #1961, opened 2026-02-02
Metal Backend

resetius:
Metal backend. Includes SSCP compilation pipeline (LLVM IR to MSL with CFG structurization and control flow extraction) and runtime (device management, memory allocation, kernel dispatch, synchronization)

The core translation happens in llvm-to-backend/metal. HLExtractionPass.cpp structurizes LLVM IR into a tree of Loop/If/Sequence nodes containing basic blocks. Emitter.cpp translates basic block instructions into MSL with the following rules: integers are stored as unsigned with signed casts inserted at signed operations, all pointers are stored as void* with casts to concrete  types at load/store/GEP sites. Although an address space attribution pass runs, LLVM sometimes inserts addrspacecast which has no Metal equivalent - we ignore the attributed address space and infer the original one at use site instead. ExternalFunction.cpp maps LLVM intrinsics and ACPP builtins to their Metal equivalents

Out of scope for this PR: atomic64, fp64, execution timestamps, full USM semantics, complex data structures like arrays of structs with pointers. The goal is initial basic support and correctness - everything else will follow in subsequent PRs

https://github.com/AdaptiveCpp/AdaptiveCpp/issues/864


al42and:
To what extent is USM supposed to be broken yet? I'm getting a deadlock on an M1 machine with the code below (works fine when `AdaptiveCpp_coarse_grained_events` property is removed):

```
#include <cstring>
#include <sycl/sycl.hpp>

int main() {
  sycl::queue q1{sycl::gpu_selector_v, sycl::property::queue::in_order{}};
  sycl::queue q2{sycl::gpu_selector_v, sycl::property::queue::in_order{}};

  auto cg_props = sycl::property_list{
      sycl::property::command_group::AdaptiveCpp_coarse_grained_events()};

  constexpr size_t numBytes = 4; // memcpy

  // Host USM allocations (use q1 for context)
  char *host1 = sycl::malloc_host<char>(numBytes, q1);
  char *host1b = sycl::malloc_host<char>(numBytes, q1);
  char *host2 = sycl::malloc_host<char>(numBytes, q1);
  char *host5 = sycl::malloc_host<char>(numBytes, q1);

  // Device USM allocations
  char *dev1 = sycl::malloc_device<char>(numBytes, q1);
  char *dev1b = sycl::malloc_device<char>(numBytes, q1);
  char *dev2 = sycl::malloc_device<char>(numBytes, q1);
  char *dev3 = sycl::malloc_device<char>(numBytes, q2); // Different queue
  char *dev4 = sycl::malloc_device<char>(numBytes, q2); // Different queue
  char *dev5 = sycl::malloc_device<char>(numBytes, q1);

  // Initialize host data
  std::memset(host1, 1, numBytes);
  std::memset(host1b, 2, numBytes);
  std::memset(host2, 3, numBytes);
  std::memset(host5, 42, numBytes);

  // Memcpy host->device (840 bytes x2) - split across queues
  q1.submit(cg_props,
            [&](sycl::handler &h) { h.memcpy(dev1, host1, numBytes); });
  q2.submit(cg_props,
            [&](sycl::handler &h) { h.memcpy(dev1b, host1b, numBytes); });

  // Memcpy host->device (168 bytes) on q1
  q1.submit(cg_props,
            [&](sycl::handler &h) { h.memcpy(dev2, host2, numBytes); });

  // Memset operations - split across queues
  q2.submit(cg_props, [&](sycl::handler &h) { h.memset(dev3, 0, numBytes); });
  q1.submit(cg_props, [&](sycl::handler &h) { h.memset(dev4, 0, numBytes); });

  // Memcpy host->device (4 bytes) on q2
  q2.submit(cg_props,
            [&](sycl::handler &h) { h.memcpy(dev5, host5, numBytes); });

  q1.wait();
  q2.wait();

  // Cleanup
  sycl::free(host1, q1);
  sycl::free(host1b, q1);
  sycl::free(host2, q1);
  sycl::free(host5, q1);
  sycl::free(dev1, q1);
  sycl::free(dev1b, q1);
  sycl::free(dev2, q1);
  sycl::free(dev3, q2);
  sycl::free(dev4, q2);
  sycl::free(dev5, q1);

  return 0;
}
```

src/compiler/llvm-to-backend/metal/ExternalFunction.cpp
```diff
@@ -0,0 +1,1962 @@
...
+  return result;
+}
+
+std::vector<ExternalFunctionInfo> generateLLVMIntrinsics() {
+  const std::vector<std::tuple<std::string, std::string, int>> intrinsics = {
+    {"ctlz", "clz", 1},
+    {"cttz", "ctz", 1},
+    {"ctpop", "popcount", -1},
+    {"umin", "min", 2},
+    {"umax", "max", 2},
```
al42and:
```suggestion
    {"umax", "max", 2},
    {"maxnum", "fmax", 2},
    {"minnum", "fmin", 2},
```

tdavidcl:
Hi, if (or when) there is support for USM with out of order queues, I can try to run Shamrock (https://github.com/Shamrock-code/Shamrock) with it (with fp64 atomics and all disabled) to see what passes or not.

src/compiler/llvm-to-backend/metal/ExternalFunction.cpp
```diff
@@ -0,0 +1,1962 @@
...
+  // atan2(±0, −0) returns ±π
+  // atan2(±0, +0) returns ±0.
+  const std::vector<std::string> funcNames = {
+    "tan", "asin", "acos", "atan", "atan2",
+    "sinh", "cosh", "tanh",
+    "cos", "sin",
+    "exp", "exp2", "exp10",
+    "log", "log2", "log10",
+    "sqrt", "rsqrt",
+    "floor", "ceil", "round", "trunc",
```
al42and:
```suggestion
    "floor", "ceil", "round", "trunc", "rint",
```

resetius:
> To what extent is USM supposed to be broken yet? I'm getting a deadlock on an M1 machine with the code below (works fine when `AdaptiveCpp_coarse_grained_events` property is removed):
> 

Thanks for the report. I haven't tested this specific use case yet, but it should work. I'll investigate and try to fix it.

src/compiler/llvm-to-backend/metal/Emitter.cpp
```diff
@@ -0,0 +1,1452 @@
...
+    }
+  }
+
+  std::unordered_map<const StructType*, std::unordered_set<const StructType*>> deps;
+
+  auto addDeps = [&](const StructType *ST) {
+    for (unsigned i = 0; i < ST->getNumElements(); ++i) {
+      Type *elemTy = ST->getElementType(i);
+      if (auto *elemST = dyn_cast<StructType>(elemTy)) {
+        deps[ST].insert(elemST);
```
al42and, line 284: It seems that this code does not capture all types in case of arrays-of-structures. Here's a vibecoded solution that seems reasonable and solves the problem, but I am no expert on compiler guts:

```
  std::function<void(const StructType*, Type*)> collectDeps = [&](const StructType *ST, Type *elemTy) {
    if (auto *elemST = dyn_cast<StructType>(elemTy)) {
      deps[ST].insert(elemST);
    } else if (auto *AT = dyn_cast<ArrayType>(elemTy)) {
      collectDeps(ST, AT->getElementType());
    }
  };

  auto addDeps = [&](const StructType *ST) {
    for (unsigned i = 0; i < ST->getNumElements(); ++i) {
      collectDeps(ST, ST->getElementType(i));
    }
  };
```

EDIT: ok, noticed that it's explicitly out-of-scope. Alright then. I'll keep using this for my testing anyways :)

resetius: acpp does not have that many tests yet and a lot of edge cases are not covered.

I personally found many issues only by running non-trivial workloads instead of unit tests

In particular I used my arbitrary-precision mandelbrot computation as a stress test: https://github.com/resetius/fdm/blob/master/test/test_mandelbrot.cpp

bin/acpp
```diff
@@ -2056,7 +2056,7 @@ class compiler:
       elif backend == "hip.explicit-multipass":
         self._multipass_backends.append(
           hip_multipass_invocation(config, config.targets["hip.explicit-multipass"]))
-      elif backend == 'sscp' or backend == 'generic':
+      elif backend == 'sscp' or backend == 'generic' or backend == 'metal':
```
illuhad: I don't think we need this alias. Aliases can become difficult to maintain in the long run, and the ones we have are mostly there for historic reasons. None of the other backends have aliases for `generic` either.

include/hipSYCL/runtime/metal/metal_allocator.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+} // namespace MTL
+
+namespace hipsycl {
+namespace rt {
+
+class metal_allocator : public backend_allocator
+{
+public:
+  enum class UsmAllocType {
```
illuhad: `snake_case` inside runtime code.

(We use ~LLVM style for compiler stuff, `snake_case` everywhere else).

include/hipSYCL/runtime/metal/metal_event.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+  virtual bool is_complete() const override;
+  virtual void wait() override;
+
+  std::shared_ptr<signal_channel> get_signal_channel() const;
+
+  virtual std::shared_ptr<signal_channel> request_backend_event() override;
+
+private:
+  std::shared_ptr<signal_channel> _signal_channel;
```
illuhad, line 35: `signal_channel` relies on `std::promise/future`, which are okay for host backend, but usually far slower than true heterogeneous events. Does metal not have a native event type?

include/hipSYCL/sycl/device.hpp
```diff
@@ -99,8 +99,14 @@ class device {
       // fp16 is only partially supported in hipSYCL
       return false;
     } else if(asp == aspect::fp64) {
+      if (get_backend() == rt::backend_id::metal) {
+        return false;
```
illuhad, line 103: Now that we have a backend where this may not be supported, we should consider adding a proper field to `rt::device_support_aspect`, and properly query that, rather than special-casing metal in the user-facing SYCL headers.

I wouldn't mind though if this were done in a follow-up PR.

include/hipSYCL/sycl/device.hpp
```diff
@@ -99,8 +99,14 @@ class device {
...
+      }
       return true;
     } else if(asp == aspect::atomic64) {
+      if (get_backend() == rt::backend_id::metal) {
+        return false;
```
illuhad, line 108: Same here

include/hipSYCL/sycl/device.hpp
```diff
@@ -60,11 +60,11 @@ class device {
```
illuhad, line 60: I think there is at least one other file from `sycl/` directory that we need to modify:

`sycl/backend.hpp` would need to expose a macro `SYCL_EXT_ACPP_BACKEND_METAL`, in analogy to the other backends and as required by the SYCL spec.

(Long term, we might also consider adding backend interop support for metal, but definitely not required for now)

src/compiler/llvm-to-backend/metal/Emitter.cpp
illuhad: For other backends, we not only test with SYCL unit tests, but also stdpar and PCUDA tests. The stdpar tests in particular are valuable because they run non-trivial parallel algorithms.

However, stdpar requires shared USM and tight integration with the OS (it needs to intercept all allocations), so I'd expect that probably it doesn't work as-is on Mac.

For compiler features and edge cases specifically, we tend to test those in the LIT tests. Those are curently not run for Mac though.

src/compiler/llvm-to-backend/metal/ExternalFunction.cpp
```diff
@@ -0,0 +1,1964 @@
...
+ * This file is part of AdaptiveCpp, an implementation of SYCL and C++ standard
+ * parallelism for CPUs and GPUs.
+ *
+ * Copyright The AdaptiveCpp Contributors
+ *
+ * AdaptiveCpp is released under the BSD 2-Clause "Simplified" License.
+ * See file LICENSE in the project root for full license details.
+ */
+// SPDX-License-Identifier: BSD-2-Clause
+#include "ExternalFunction.hpp"
```
illuhad: For this file, it might be worth investigating whether we can implement the AdaptiveCpp builtins in a different way.

For all other backends, builtins are implemented as `libkernel` bitcode libraries. Implementing them in a completely separate fashion here means that it's more difficult to gauge how complete the implementation is, and we might forget updating it here if we change anything about the builtin interface.

I wonder whether it would be possible to still implement the builtin interface in bitcode libraries, but do something like:

```c++
// This would be in a libkernel bitcode library for metal:

__acpp_f32 __acpp_sscp_sin_f32(__acpp_f32 x) {
  return __acpp_sscp_inline_metal("sin($1)");
}
```

which would require a new LLVM pass to evaluate those builtin calls. This pass would then replace the builtin call with the string in its argument during metal code generation. This could perhaps also be more broadly interesting as a kind of "inline-assembly" for optimized code paths.
(perhaps there is something that can be done with metal, which cannot be done easily e.g .with SYCL?)

And it reduces the amount of code in strings :)

src/compiler/llvm-to-backend/metal/LLVMToMetal.cpp
```diff
@@ -0,0 +1,172 @@
...
+  FPM.addPass(llvm::LowerSwitchPass());
+  FPM.addPass(llvm::LoopSimplifyPass());
+  FPM.addPass(llvm::LCSSAPass());
+  FPM.addPass(llvm::DCEPass());
+  FPM.addPass(llvm::ADCEPass());
+  FPM.addPass(llvm::StructurizeCFGPass());
+  FPM.addPass(llvm::SimplifyCFGPass());
+  llvm::ModulePassManager MPM;
+  MPM.addPass(createModuleToFunctionPassAdaptor(std::move(FPM)));
+  MPM.run(FlavoredModule, MAM);
```
illuhad: We already have `withPassBuilder`/`withPassBuilderAndMAM` in `Utils.hpp`. Can we reuse those, or extend/add a FPM variant?

src/compiler/llvm-to-backend/metal/LLVMToMetal.cpp
```diff
@@ -0,0 +1,172 @@
...
+  ASMap[AddressSpace::Constant] = 4;
+  ASMap[AddressSpace::AllocaDefault] = 5;
+  ASMap[AddressSpace::GlobalVariableDefault] = 1;
+  ASMap[AddressSpace::ConstantGlobalVariableDefault] = 4;
+
+  return ASMap;
+}
+
+bool LLVMToMetalTranslator::isKernelAfterFlavoring(llvm::Function& F) {
+  return true;
```
illuhad: The `LLVMToBackend` infrastructure uses this function to test whether something is a kernel entrypoint. I think the way it's currently used, this is fine, but it might cause issues in the future.

The background is that the `LLVMToBackend` infrastructure may create wrapper functions as part of certain optimizations for the original kernels. This means that the kernel-ness of a function might shift.

I would recommend implementing this properly. 

Since metal code manipulation only really enters after LLVM is done, it shouldn't be too hard:
- Maintain a `vector<string>` (or `unordered_set`) for your actual kernel names (initialize with `KernelNames`)
- When `migrateKernelProperties` is called, remove matching function from vector, insert new one
- `isKernelAfterFlavoring` just checks in that vector.

illuhad (commented):
Awesome! I'm still far away from having looked at everything; here are some initial thoughts on the first files.

src/compiler/llvm-to-backend/metal/ExternalFunction.cpp
```diff
@@ -0,0 +1,1964 @@
...
+      out.parts[i] = simd_shuffle_xor(in.parts[i], mask);
+    }
+    return out.value;
+  }
+}
+)__",
+      .exactMatch = false,
+    },
+    {
+      .name = "__acpp_sscp_sub_group_permute",
```
al42and: Duplicate?

resetius: Yes. Thank you

include/hipSYCL/runtime/metal/metal_event.hpp
resetius: Metal does have native events. For now I just cut corners for development speed and reused the omp-style approach. This is a temporary solution and will need to be improved later

bin/acpp
fodinabor: Also, it suggests the OpenMP JIT variant would be disabled, which isn't the case

resetius:
> To what extent is USM supposed to be broken yet? I'm getting a deadlock on an M1 machine with the code below (works fine when `AdaptiveCpp_coarse_grained_events` property is removed):

USM should work correctly except for pointer indirection scenarios (i.e., passing structs to kernels that contain pointers to other structs).  The deadlock you're seeing is unrelated to USM functionality - it's caused by suboptimal event handling. That's out of scope for this PR. I've temporarily disabled coarse-grained events for Metal

separate PRs:
  - [ ] Full USM semantics (i.e., pointer indirections in kernel arguments ...)
  - [ ] Proper asynchronous event handling (replace blocking `waitUntilCompleted()` with completion handlers)
  - [ ] Add new fields to rt::device_support_aspect (atomic64, fp64)
  - [ ] Non-relaxed atomics
  - [ ] Metal logging

src/compiler/llvm-to-backend/metal/LLVMToMetal.cpp
resetius: Good point! Refactored to reuse withPassBuilder. Does this look better now?

illuhad: Yep, thanks! :)

al42and:
> That's out of scope for this PR. I've temporarily disabled coarse-grained events for Metal

Fair enough, thanks!

src/compiler/llvm-to-backend/metal/ExternalFunction.cpp
resetius: Like this? https://github.com/AdaptiveCpp/AdaptiveCpp/pull/1961/changes/9408243ec0c6e86e4b2ee1c2c9964f3306452c42#diff-3734a8798cbbbbfe2fae9670065ba34da5afbda72a288bad07e261316410f8d1

resetius:
> Hi, if (or when) there is support for USM with out of order queues, I can try to run Shamrock (https://github.com/Shamrock-code/Shamrock) with it (with fp64 atomics and all disabled) to see what passes or not.

Thanks, that sounds great 🙂 Shamrock looks like a very interesting project - I had a quick look, and hydrodynamics is definitely my kind of thing. I gave it a quick try: it builds fine, but currently fails at the link stage with a libc++-related issue.
I'll take another look and try again later when I have more time

tdavidcl:
Oh nice that's a good start, thanks for trying.

> It builds fine, but currently fails at the link stage with a libc++-

Does the metal backend add additional linkage to apple toolchains beside metal API ? I can try to take a look

src/compiler/llvm-to-backend/metal/ExternalFunction.cpp
illuhad: Yes, this is roughly what I had imagined. We should prefix the builtin with `__acpp_sscp_` though as we do with other SSCP-specific functionality.

Is there a particular reason why you only did this for the math functions, but not e.g. for `__acpp_sscp_get_local_id` or group algorithms? Or was this just because this is a proof-of-concept for now?

I imagine we could use this mechanism for all builtins, and implement the full bitcode library interface that the other backends have in that way.

resetius: Good point. There's no fundamental reason why this mechanism should be limited to math functions - in principle it can be extended to all SSCP builtins.

I started with math mostly because they are the simplest case, but the same approach could be applied to things like `__acpp_sscp_get_local_id` and group algorithms as well. The main open question for me is how we want to model the more complex cases.

For example, some builtins depend on implicitly declared Metal builtins (like the *_id variables), and others rely on hidden state such as threadgroup memory used by work-group algorithms. One option would be to encode this via custom LLVM attributes and interpret them during text emission. Another option would be to generate a fixed Metal "header" that defines the required builtins, variables, and helpers, and have the bitcode reference those symbols.

Both seem feasible, and I think it mostly comes down to which model we prefer long-term

resetius: I think I figured out how to do this properly. It should be possible to move everything into bitcode, except maybe the llvm intrinsics translation part

illuhad: Ok, sounds good - let me know if you need more guidance.

resetius: Next iteration: moved everything to bitcode (except llvm intrinsics)

src/compiler/sscp/StdBuiltinRemapperPass.cpp
```diff
@@ -58,10 +58,26 @@ static constexpr std::array math_builtins = {
...
 static constexpr std::array remapped_llvm_math_builtins = {
-  "sin", "cos", "sqrt"
+  "sin", "cos", "tan", "sqrt",
+  "asin", "acos", "atan", "atan2",
+  "sinh", "cosh", "tanh",
+  "log", "log2", "log10",
+  "exp", "exp2", "exp10",
+  "ldexp",
+  "fabs", "floor", "ceil",
+  "copysign"
```
resetius: Is it OK to add this to the remapper here?
I see that some backends (PTX) already does its own remapping, and it partially overlaps with this

src/compiler/llvm-to-backend/metal/ExternalFunction.cpp
resetius: Moved everything to bitcode

include/hipSYCL/runtime/metal/metal_backend.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+  virtual std::string get_name() const override;
+
+  virtual ~metal_backend();
+
+  std::unique_ptr<backend_executor>
+  create_inorder_executor(device_id dev, int priority) override;
+private:
+    mutable metal_hardware_manager _hw;
+    mutable lazily_constructed_executor<multi_queue_executor> _executor;
```
illuhad: Indentation looks a bit off here

src/libkernel/sscp/metal/atomic.cpp
```diff
@@ -0,0 +1,460 @@
...
+HIPSYCL_SSCP_BUILTIN void __acpp_sscp_atomic_store_i16(
+    __acpp_sscp_address_space as, __acpp_sscp_memory_order order,
+    __acpp_sscp_memory_scope scope, i16 *ptr, i16 x) {
+  __acpp_sscp_metal_atomic_store_i16("atomic_store_explicit(__atomic_pointer_cast<short>(%s), %s, memory_order_relaxed)", ptr, x);
+}
+
+HIPSYCL_SSCP_BUILTIN void __acpp_sscp_atomic_store_i32(
+    __acpp_sscp_address_space as, __acpp_sscp_memory_order order,
+    __acpp_sscp_memory_scope scope, i32 *ptr, i32 x) {
+  __acpp_sscp_metal_atomic_store_i32("atomic_store_explicit(__atomic_pointer_cast<int>(%s), %s, memory_order_relaxed)", ptr, x);
```
illuhad, line 95: This is going to be incorrect if `order` is something other than `relaxed`. Does metal not support other memory orders, or is this just out of the scope of the PR for now?

src/libkernel/sscp/metal/core.cpp
```diff
@@ -0,0 +1,61 @@
...
+ * AdaptiveCpp is released under the BSD 2-Clause "Simplified" License.
+ * See file LICENSE in the project root for full license details.
+ */
+// SPDX-License-Identifier: BSD-2-Clause
+#include "hipSYCL/sycl/libkernel/sscp/builtins/core.hpp"
+
+HIPSYCL_SSCP_BUILTIN __acpp_uint64 __acpp_sscp_metal_symbol_id(const char* s);
+
+HIPSYCL_SSCP_BUILTIN __acpp_uint64 __acpp_sscp_get_local_id_x() {
+  return __acpp_sscp_metal_symbol_id("__acpp_sscp_get_local_id.x");
```
illuhad: Maybe we should namespace `__acpp_sscp_get_local_id` and friends to something like `__acpp_sscp_metal_get_local_id` in case we want to have a universal 3D query in the future?

src/libkernel/sscp/metal/print.cpp
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * AdaptiveCpp is released under the BSD 2-Clause "Simplified" License.
+ * See file LICENSE in the project root for full license details.
+ */
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include "hipSYCL/sycl/libkernel/sscp/builtins/print.hpp"
+
+void __acpp_sscp_print(const char* msg) {
+}
```
illuhad, line 19: Not a strong opinion, just wondering whether it might be better to issue an assert fail/trap here if it's unsupported for now?

src/runtime/metal/metal_queue.cpp
```diff
@@ -0,0 +1,803 @@
...
+    MTL::Device* device,
+    metal_allocator* allocator,
+    void** args,
+    std::size_t* arg_sizes,
+    std::size_t num_args,
+    const std::vector<int>& isPointerArg
+) {
+  for (std::size_t i = 0; i < num_args; ++i) {
+    if (isPointerArg[i]) {
+        void* usm_ptr = *(void**)args[i];
```
illuhad: 2 spaces as indentation

src/libkernel/sscp/metal/barrier.cpp
```diff
@@ -0,0 +1,144 @@
...
+    case __acpp_sscp_memory_order::acquire:
+    case __acpp_sscp_memory_order::release:
+    case __acpp_sscp_memory_order::acq_rel:
+    case __acpp_sscp_memory_order::seq_cst:
+      if (fence_scope == __acpp_sscp_memory_scope::work_item) {
+        __acpp_sscp_metal_symbol_barrier("atomic_thread_fence(mem_flags::mem_none, memory_order_seq_cst, thread_scope_thread)");
+      } else if (fence_scope >= __acpp_sscp_memory_scope::device) {
+        __acpp_sscp_metal_symbol_barrier("atomic_thread_fence(mem_flags::mem_device, memory_order_seq_cst, thread_scope_device)");
+      } else {
+        __acpp_sscp_metal_symbol_barrier("atomic_thread_fence(mem_flags::mem_threadgroup, memory_order_seq_cst, thread_scope_threadgroup)");
```
illuhad: For `seq_cst` semantics, is it sufficient to have only a fence after the barrier? Often in such constructs a fence both before and after the barrier is necessary.

src/runtime/metal/metal_queue.cpp
```diff
@@ -0,0 +1,803 @@
...
+{
+  if (!library) {
+    return make_error(__acpp_here(),
+                      error_info{"metal: Library is null"});
+  }
+
+  //std::string entry = "my_kernel"; //
+  auto entry = std::string(kernel_name);
+  // Get the kernel function from library
+  NS::String* functionName = NS::String::string(entry.c_str(), NS::UTF8StringEncoding);
```
illuhad: Snake case `function_name`

src/runtime/metal/metal_queue.cpp
```diff
@@ -0,0 +1,803 @@
...
+          default:
+            isPointerArg[j] = 0;
+            break;
+        }
+      }
+      break;
+    } else if (arg->bufferPointerType()) {
+      isPointerArg[i] = 1;
+    } else {
+      isPointerArg[i] = 0;
```
illuhad: Why can't we just use the information that already is in `hcf_kernel_info` to determine whether arguments are pointers? We already store that information as part of the kernel information struct.

src/runtime/metal/metal_queue.cpp
```diff
@@ -0,0 +1,803 @@
...
+    hcf_object_id hcf_object,
+    const rt::range<3> &num_groups,
+    const rt::range<3> &group_size,
+    unsigned local_mem_size, void **args,
+    std::size_t *arg_sizes, std::size_t num_args,
+    std::string_view kernel_name,
+    const rt::hcf_kernel_info* kernel_info,
+    const kernel_configuration& config)
+{
+    return _queue->submit_sscp_kernel_from_code_object(
```
illuhad: 2 spaces indentation

illuhad (commented):
Thanks -- this looks mostly good now! :-)

The main two points I would like to see:

- We should list the backend as an experimental backend in documentation. And probably also describe what users need to do to enable it, what the most important limitations are etc.
- To what extent can we enable CI for it? I don't know what hardware the GitHub Mac runners provide, but if they don't give us Metal-compatible hardware, can we at least check if the backend compiles?

src/compiler/llvm-to-backend/metal/LLVMToMetal.cpp
```diff
@@ -0,0 +1,497 @@
...
+        }
+      } else if (auto* MS = llvm::dyn_cast<llvm::MemSetInst>(II)) {
+        llvm::expandMemSetAsLoop(MS);
+        II->eraseFromParent();
+        Changed = true;
+      } else if (auto* MSP = llvm::dyn_cast<llvm::MemSetPatternInst>(II)) {
+        llvm::expandMemSetPatternAsLoop(MSP);
+        II->eraseFromParent();
+        Changed = true;
+      } else if (auto* AMC = llvm::dyn_cast<llvm::AnyMemCpyInst>(II)) {
```
al42and, line 216: They changed the signature of `expandAtomicMemCpyAsLoop` in LLVM 21 (650dca5d89602), so, to support older versions, it should be (based on the function signature, at least) something like:


```suggestion
#if LLVM_VERSION_MAJOR >= 21
      } else if (auto* AMC = llvm::dyn_cast<llvm::AnyMemCpyInst>(II)) {
#else
      } else if (auto* AMC = llvm::dyn_cast<llvm::AtomicMemCpyInst>(II)) {
#endif
```

src/compiler/sscp/StdBuiltinRemapperPass.cpp
illuhad: The PTX backend only does this because at least in some versions of LLVM, some LLVM math intrinsics are not correctly handled by the LLVM NVPTX backend.

I don't remember why we handle `sin`/`cos`/`sqrt` here, but the point of `StdBuiltinRemapperPass` is a little different -- it enables `std::` math functions inside kernels by remapping function calls that clang might generate to `__acpp_sscp` builtins.

If there's a need to workaround missing handling of LLVM builtins in some backends (e.g. metal) then I suspect it might be better to work around this on a per-backend basis. There might (?) be drawbacks in hiding the actual LLVM intrinsics from LLVM.

src/runtime/metal/metal_queue.cpp
resetius: I just didn't understand how to properly use hcf_kernel_info, so I quickly used reflection instead. It was mostly for speed of implementation. I knew the information is already there, but I wasn't sure what the intended access pattern is and didn't want to block the progress on this change

resetius:
> * We should list the backend as an experimental backend in documentation. And probably also describe what users need to do to enable it, what the most important limitations are etc.

Ok

> * To what extent can we enable CI for it? I don't know what hardware the GitHub Mac runners provide, but if they don't give us Metal-compatible hardware, can we at least check if the backend compiles?

We can. I'll try to experiment with CI

src/libkernel/sscp/metal/barrier.cpp
resetius: You are right - we need 2 fences

src/libkernel/sscp/metal/atomic.cpp
resetius: Yes, good catch. For now I always emit memory_order_relaxed for metal atomics. Metal's atomic operations are basically "relaxed-only" and stronger ordering has to be built using atomic_thread_fence(..., memory_order_seq_cst, ...) (plus the right thread scope / mem_flags). I plan to add the non-relaxed mapping via fences in a follow-up PR

src/runtime/metal/metal_queue.cpp
resetius: Possibly I misused kernel_info - I was querying get_argument_type(i) without considering the _arg_mapper remapping. I'll rework this to rely on the remapped kernel info and drop the reflection

resetius:
Metal supported by github: https://github.com/resetius/AdaptiveCpp/actions/runs/22246464501/job/64361410712?pr=4
`Apple Paravirtual device`: all passed except 1 cpu-only test.
The test failed on ld64.lld. I thought we fixed this issue

resetius:
> Metal supported by github: https://github.com/resetius/AdaptiveCpp/actions/runs/22246464501/job/64361410712?pr=4 `Apple Paravirtual device`: all passed except 1 cpu-only test. The test failed on ld64.lld. I thought we fixed this issue

i did not install lld, tests are green now: https://github.com/resetius/AdaptiveCpp/actions/runs/22257233640/job/64389874885?pr=4

al42and:
A few issues observed when testing on my code (and pestering an LLM for self-contained reproducers):

1. `llvm.minnum` / `llvm.maxnum` also need explicit expanding

<details>

<summary>Reproducer</summary>

Compiled with `acpp ./sycl_min_max_test.cpp -O3 -ffast-math -o sycl_min_max_test` (ffast-math is important)

```cpp
#include <iostream>
#include <sycl/sycl.hpp>
#include <vector>
int main() {
    sycl::queue q;
    std::cout << "Running on device: " << q.get_device().get_info<sycl::info::device::name>() << std::endl;
    float a = 12.5f, b = 7.3f;
    std::vector<float> results(2, 0.0f);
    {
        sycl::buffer<float, 1> res_buf(results.data(), sycl::range<1>(2));
        q.submit([&](sycl::handler& h) {
            sycl::accessor res(res_buf, h, sycl::write_only);
            h.single_task([=]() {
                res[0] = sycl::max(a, b);
                res[1] = sycl::min(a, b);
            });
        });
    }
    std::cout << "Input a: " << a << ", Input b: " << b << std::endl;
    std::cout << "sycl::max(a, b): " << results[0] << " (Expected: 12.5)" << std::endl;
    std::cout << "sycl::min(a, b): " << results[1] << " (Expected: 7.3)" << std::endl;
    return 0;
}
```

</details>

2. Choosing the barrier scope based on a runtime condition triggers a cryptic `__acpp_sscp_metal: Argument must be a string constant` error.

<details>

<summary>Reproducer</summary>

Compiled with `acpp ./repro_barrier_select.cpp -O3 -o repro_barrier_select`
                                                                                                                                                                                            
```cpp
#include <sycl/sycl.hpp>
#include <cstdio>
int main() {
  constexpr int blockSize = 64;
  sycl::queue q;
  int subGroupSize = q.get_device().get_info<sycl::info::device::sub_group_sizes>()[0];

  auto* result = sycl::malloc_shared<int>(1, q);
  auto* input  = sycl::malloc_shared<int>(blockSize, q);
  for (int i = 0; i < blockSize; ++i) input[i] = 1;
  q.submit([&](sycl::handler& h) {
    sycl::local_accessor<int, 1> sm_buf(blockSize, h);
    int sgSize = subGroupSize;
    h.parallel_for(sycl::nd_range<1>{blockSize, blockSize},
                   [=](sycl::nd_item<1> itemIdx) {
      int threadIdx = (int)itemIdx.get_local_id(0);
      sm_buf[threadIdx] = input[threadIdx];
      itemIdx.barrier(sycl::access::fence_space::local_space);
      for (int divideBy = 2; divideBy <= blockSize; divideBy *= 2) {
        int dividedAt = blockSize / divideBy;
        if (threadIdx < dividedAt) {
          sm_buf[threadIdx] += sm_buf[threadIdx + dividedAt];
        }
        if (dividedAt > sgSize / 2) {
          sycl::group_barrier(itemIdx.get_group());
        } else {
          sycl::group_barrier(itemIdx.get_sub_group());
        }
      }
      if (threadIdx == 0)
        *result = sm_buf[0];
    });
  }).wait();

  printf("%s (result=%d, expected=%d)\n", *result == blockSize ? "PASS" : "FAIL", *result, blockSize);
  sycl::free(result, q);
  sycl::free(input, q);
  return 0;
}
```

doc/install-metal.md
```diff
@@ -0,0 +1,81 @@
...
+The Metal backend is experimental and has the following important limitations:
+
+* **No full USM pointer semantics.** Metal does not currently support arbitrary GPU-side pointer dereferencing across separate allocations. All buffers that a kernel needs must be passed explicitly as kernel arguments. Passing a struct that contains a pointer to another GPU buffer — for example:
+
+    ```cpp
+    struct Entity { double* data; };
+    // ... array of Entity passed to kernel ...
+    ```
+
+    and then dereferencing `entity.data` inside the kernel will cause the program to crash, because the nested pointer is not valid from the GPU's perspective. Only flat buffers passed directly as kernel arguments are safe to access.
```
illuhad, line 63: Why is this the case though? The SSCP compiler decomposes structs into their fundamental constituents. So, if you use `Entity` as a kernel argument, the compiler will generate a kernel with `ptr` argument which the runtime can see.

doc/install-metal.md
```diff
@@ -0,0 +1,81 @@
...
+```
+
+## Using the Metal backend
+
+The Metal backend is part of the `generic` compilation flow, so you use it the same way as any other `generic` target:
+
+```bash
+acpp --acpp-targets=generic -o my_program my_program.cpp
+./my_program
+```
```
illuhad, line 46: If AdaptiveCpp has been built with `generic` support, it will become the default compilation flow, unless `-DDEFAULT_TARGETS` is set explicitly to something else. So in practice, users will most likely be able to just run `acpp -o my_program my_program.cpp`.

.github/workflows/macos-metal.yml
```diff
@@ -0,0 +1,72 @@
...
+        cmake -GNinja \
+          -DACPP_TARGETS="generic" \
+          -DAdaptiveCpp_DIR=${GITHUB_WORKSPACE}/install/lib/cmake/AdaptiveCpp \
+          ${GITHUB_WORKSPACE}/tests
+        ninja
+
+    - name: run Metal tests
+      run: |
+        cd ${GITHUB_WORKSPACE}/build/tests-sscp
+        DYLD_LIBRARY_PATH=${GITHUB_WORKSPACE}/install/lib ./sycl_tests
```
illuhad, line 72: What about PCUDA test cases? (I assume we might not be able to support stdpar, at least not initially)

illuhad (commented):
Thanks! Only some small nitpicks.

There are some images in `doc/img` (made with draw.io; the xml/drawio files are the sources) where we'd also need to add the backend.

I don't have a strong opinion whether to do this now or once the backend has matured.

doc/install-metal.md
resetius: I mean indirections like this
```
#include <sycl/sycl.hpp>
#include <iostream>

struct Entity {
  float* ptr;
  int    len;
  int    off;
};

int main() {
  sycl::queue q{sycl::default_selector_v};

  constexpr int N = 8;
  constexpr int LEN = 16;
  constexpr int TOTAL = N * LEN;

  float* block = sycl::malloc_shared<float>(TOTAL, q);
  Entity* entities = sycl::malloc_shared<Entity>(N, q);

  for (int i = 0; i < TOTAL; ++i)
    block[i] = float(i);

  for (int i = 0; i < N; ++i) {
    entities[i].ptr = block + i * LEN;
    entities[i].len = LEN;
    entities[i].off = i * LEN;
  }

  q.submit([&](sycl::handler& h) {
    h.parallel_for(sycl::range<1>(N), [=](sycl::id<1> tid) {
      int i = (int)tid[0];
      float* p = entities[i].ptr;
      int len  = entities[i].len;

      for (int j = 0; j < len; ++j)
        p[j] = p[j] * 2.0f + float(i);

      (void)block;
    });
  }).wait();

  bool ok = true;
  for (int i = 0; i < N; ++i) {
    for (int j = 0; j < LEN; ++j) {
      float in  = float(i * LEN + j);
      float exp = in * 2.0f + float(i);
      float got = block[i * LEN + j];
      if (got != exp) {
        std::cerr << "Mismatch at i=" << i << " j=" << j
                  << " got=" << got << " exp=" << exp << "\n";
        ok = false;
        break;
      }
    }
    if (!ok) break;
  }

  std::cout << (ok ? "OK\n" : "FAIL\n");

  sycl::free(entities, q);
  sycl::free(block, q);
  return ok ? 0 : 1;
}
```

entity[i].ptr will be invalid

.github/workflows/macos-metal.yml
resetius: Unimplemented so far
```
std::unique_ptr<backend_executor>
metal_backend::create_inorder_executor(device_id dev, int priority) {
  return nullptr;
}
```

illuhad: Okay, makes sense. Perhaps we should say in the documentation that only the SYCL model is supported for now?

doc/install-metal.md
illuhad: Ah, right, I can see that. I was thinking of the case
```
Entity e = ...
  e.data[idx] = ...
});
```
which I would expect to work.

The whole detection-of-indirect-access is also relevant for other optimizations that we do, so we already have infrastructure for the compiler to detect such cases.

You can query that using the `is_freee_of_indirect_access` field of https://github.com/resetius/AdaptiveCpp/blob/e061edd269b0fc692a8d2c19d846449b9f0fa84c/include/hipSYCL/runtime/kernel_cache.hpp#L87

(of course, this is the compiler attempting to prove that a kernel is free of indirect access, so it might in some cases return `false` just because it cannot prove it -- so there might be false negatives)

You might want to consider whether you want to query that field and emit a warning to simplify diagnosing the issue when it occurs.

resetius:
> A few issues observed when testing on my code (and pestering an LLM for self-contained reproducers):
> 

Thank you! Should be ok now

doc/install-metal.md
resetius: Sure. And for `true` need to use @philipturner approach : https://github.com/philipturner/metal-usm

resetius:
> Thanks! Only some small nitpicks.
> 
> There are some images in `doc/img` (made with draw.io; the xml/drawio files are the sources) where we'd also need to add the backend.
> 
> I don't have a strong opinion whether to do this now or once the backend has matured.

I'd say this should be done a little bit later

doc/install-metal.md
resetius: Added warn message

illuhad (approved):
Lovely! Thank you :)

src/runtime/metal/metal_queue.cpp
```diff
@@ -0,0 +1,784 @@
...
+
+  auto* encoder = command_buffer->computeCommandEncoder();
+  if (!encoder) {
+    return make_error(__acpp_here(),
+                      error_info{"metal: Failed to create compute encoder"});
+  }
+
+  encoder->setComputePipelineState(pipeline_state.get());
+  auto user_local_mem_size = local_mem_size;
+  auto threads_in_group = num_groups[0] * num_groups[1] * num_groups[2];
```
al42and, line 167: `s/num_groups/group_size/g` ?

resetius: Fixed in https://github.com/AdaptiveCpp/AdaptiveCpp/pull/1996/files

src/libkernel/sscp/metal/math.cpp
```diff
@@ -0,0 +1,182 @@
...
+HIPSYCL_SSCP_BUILTIN i32 __acpp_sscp_isnormal_f32(f32 x) {
+  return __acpp_sscp_metal_math_i32_f32("isnormal", x);
+}
+
+HIPSYCL_SSCP_BUILTIN i32 __acpp_sscp_signbit_f32(f32 x) {
+  return __acpp_sscp_metal_math_i32_f32("signbit", x);
+}
+
+HIPSYCL_SSCP_BUILTIN f32 __acpp_sscp_ldexp_f32(f32 x, i32 k) {
+  return __acpp_sscp_metal_math_f32_f32_f32("ldexp", x, k);
```
al42and, line 95: ?

```suggestion
  return __acpp_sscp_metal_math_f32_f32_i32("ldexp", x, k);
```

resetius: Thank you

PR #1962, opened 2026-02-05
Fix some naming mismatch of amath/armpl introduced in #1949

thomasRoglin:
While retesting changes introduced in #1949, I found that the amath veclib option is broken because I mixed up amath/armpl in the names of some function/macro.
This PR fix the `getLibAmathDir` function and the `AMATH_AVAILABLE` macro


illuhad (approved):
Thanks!

Issue #1964, opened 2026-02-05
Support ROCm LLVM version 22.0.0 for GROMACS build


alexschroeter:
**Describe the motivation for the feature request**
I would like to build GROMACS with the current LLVM (22.0.0) provided by the latest ROCm (7.2.0) as recommended in the documentation. My understanding is that it is better to use the LLVM packaged with ROCm for the best experience.

**Describe the solution you'd like**

I believe the error during compile
```
In file included from /home/aschroet/src/investigate_memory_leak/GROMACS/AdaptiveCpp/src/compiler/frontend/AdaptiveCppClangPlugin.cpp:13:
In file included from /home/aschroet/src/investigate_memory_leak/GROMACS/AdaptiveCpp/include/hipSYCL/compiler/FrontendPlugin.hpp:14:
/home/aschroet/src/investigate_memory_leak/GROMACS/AdaptiveCpp/include/hipSYCL/compiler/Frontend.hpp:91:79: error: no member named
      'getDecl' in 'clang::RecordType'
   91 |             if(auto RecordDecl = llvm::dyn_cast<clang::CXXRecordDecl>(Record->getDecl()))
      |                                                                       ~~~~~~  ^
/home/aschroet/src/investigate_memory_leak/GROMACS/AdaptiveCpp/include/hipSYCL/compiler/Frontend.hpp:565:17: error: no member named
      'getDecl' in 'clang::RecordType'
  565 |     if(NameTag->getDecl()) {
      |        ~~~~~~~  ^
/home/aschroet/src/investigate_memory_leak/GROMACS/AdaptiveCpp/include/hipSYCL/compiler/Frontend.hpp:566:23: error: no member named
      'getDecl' in 'clang::RecordType'
  566 |       return NameTag->getDecl()->getQualifiedNameAsString() ==
      |              ~~~~~~~  ^
/home/aschroet/src/investigate_memory_leak/GROMACS/AdaptiveCpp/include/hipSYCL/compiler/Frontend.hpp:630:32: error: no member named
      'getDecl' in 'clang::RecordType'
  630 |         ErrorDecl = B->second->getDecl();
      |                     ~~~~~~~~~  ^
/home/aschroet/src/investigate_memory_leak/GROMACS/AdaptiveCpp/include/hipSYCL/compiler/Frontend.hpp:738:48: error: no member named
      'getDecl' in 'clang::RecordType'
  738 |                             KernelFunctorType->getDecl())
[...]

Issue #1958, opened 2026-01-30
make install error: no member named 'getDecl' in 'clang::RecordType'


sys-gellan:
Hey all, hope someone can help:

OS: Ubuntu 25.10
GFX: AMD 9060xt

I am trying to use adaptiveccp for gromacs, have followed installation instructions, installed rocm (including boost libraries). 

Steps:

`sudo cmake .. -DCMAKE_INSTALL_PREFIX=/usr/local -DCMAKE_C_COMPILER=/opt/rocm/llvm/bin/clang -DCMAKE_CXX_COMPILER=/opt/rocm/llvm/bin/clang++ -DLLVM_DIR=/opt/rocm/llvm/lib/cmake/llvm/ -DWITH_ROCM_BACKEND=ON -DWITH_SSCP_COMPILER=OFF -DWITH_OPENCL_BACKEND=OFF -DWITH_LEVEL_ZERO_BACKEND=OFF -DWITH_CUDA_BACKEND=OFF -DDEFAULT_TARGETS='hip:gfx1200' -DACPP_EXPERIMENTAL_LLVM=ON`

returns:
```
Your branch is up-to-date with 'origin/develop'.

nothing to commit, working tree clean
-- Performing Test HIP_CLANG_SUPPORTS_PARALLEL_JOBS
-- Performing Test HIP_CLANG_SUPPORTS_PARALLEL_JOBS - Success
CMake Deprecation Warning at CMakeLists.txt:144 (message):
  WITH_SSCP_COMPILER, WITH_STDPAR_COMPILER, WITH_ACCELERATED_CPU and
  WITH_CUDA_NVCXX_ONLY are deprecated and will be removed in a future
  version; please use ACPP_COMPILER_FEATURE_PROFILE instead.


CMake Deprecation Warning at CMakeLists.txt:192 (message):
  Configuring DEPRECATED custom compiler feature profile with SSCP=OFF,
  stdpar=true, accelerated cpu=true


-- Building with compiler feature profile: custom-deprecated
-- Building AdaptiveCpp against LLVM configured from /opt/rocm/llvm/lib/cmake/llvm/
-- Selecting clang: /opt/rocm-7.2.0/lib/llvm/bin/clang++
-- Found ROCm clang: /opt/rocm/llvm/bin/clang++
-- ROCm LLVM version: 22.0.0
-- Using clang include directory: /opt/rocm-7.2.0/lib/llvm/lib/clang/22/include/..
[...]

PR #1969, opened 2026-02-09
LLVM version compatibility after 'getDecl' rename to 'getOriginalDecl'

alexschroeter:
In LLVM > 21 the function getDecl() has temporarly been renamed to getOriginalDecl(). [Source](https://www.mail-archive.com/cfe-commits@lists.llvm.org/msg596147.html)

This pull requests introduces this change for those versions it applies to.
This change is related to the following Issues https://github.com/AdaptiveCpp/AdaptiveCpp/issues/1964 https://github.com/AdaptiveCpp/AdaptiveCpp/issues/1958

### DIFF
--- a/include/hipSYCL/compiler/Frontend.hpp
+++ b/include/hipSYCL/compiler/Frontend.hpp
@@ -88,7 +88,11 @@ class CompleteCallSet : public clang::RecursiveASTVisitor<CompleteCallSet> {
         // type that can be constructed on the GPU also can and will be destructed).
         if(auto Ptr = llvm::dyn_cast_or_null<clang::PointerType>(Callee->getThisType()->getCanonicalTypeUnqualified()))
           if(auto Record = llvm::dyn_cast<clang::RecordType>(Ptr->getPointeeType()))
+#if defined(ROCM_CLANG_VERSION_MAJOR) && ROCM_CLANG_VERSION_MAJOR == 7 && ROCM_CLANG_VERSION_MINOR == 2
+            if(auto RecordDecl = llvm::dyn_cast<clang::CXXRecordDecl>(Record->getOriginalDecl()))
+#else
             if(auto RecordDecl = llvm::dyn_cast<clang::CXXRecordDecl>(Record->getDecl()))
+#endif
               if(auto DtorDecl = RecordDecl->getDestructor())
                 TraverseDecl(DtorDecl);
       }
@@ -562,8 +566,14 @@ private:
       // If name tag is invalid, assume unnamed
       return true;
 
+#if defined(ROCM_CLANG_VERSION_MAJOR) && ROCM_CLANG_VERSION_MAJOR == 7 && ROCM_CLANG_VERSION_MINOR == 2
+    if(NameTag->getOriginalDecl()) {
+      return NameTag->getOriginalDecl()->getQualifiedNameAsString() ==
+#else
     if(NameTag->getDecl()) {
       return NameTag->getDecl()->getQualifiedNameAsString() ==
+#endif
+
              "__acpp_unnamed_kernel";
     }
 
@@ -627,7 +637,11 @@ private:
       clang::Decl* ErrorDecl = F;
 
       if(B != KernelBodies.end()) {
+#if defined(ROCM_CLANG_VERSION_MAJOR) && ROCM_CLANG_VERSION_MAJOR == 7 && ROCM_CLANG_VERSION_MINOR == 2
+        ErrorDecl = B->second->getOriginalDecl();
+#else
         ErrorDecl = B->second->getDecl();
+#endif
       }
 
       auto SL = ErrorDecl->getSourceRange().getBegin();
@@ -735,7 +749,11 @@ private:
             if (KernelFunctorType->getAsCXXRecordDecl() &&
                 KernelFunctorType->getAsCXXRecordDecl()->isLambda()) {
               auto SL = llvm::dyn_cast<clang::CXXRecordDecl>(
+#if defined(ROCM_CLANG_VERSION_MAJOR) && ROCM_CLANG_VERSION_MAJOR == 7 && ROCM_CLANG_VERSION_MINOR == 2
+                            KernelFunctorType->getOriginalDecl())
+#else
                             KernelFunctorType->getDecl())
+#endif
                             ->getSourceRange()
                             .getBegin();
               auto ID =


fodinabor:
In what sense "temporarily"? I.e. has it been changed back?

fodinabor:
If that fixes LLVM 21, we could add it to CI and raise the cmake experimental flag..

alexschroeter:
> In what sense "temporarily"? I.e. has it been changed back?

I only know what is written in the mail referenced in the source. There they say that this is temporary and yes I would assume that this means that they plan to revert it.

alexschroeter:
Is there an issue with the self hosted runner or with the commit? Or was it the github availability outage yesterday and the test should just be rerun?

illuhad:
> Is there an issue with the self hosted runner or with the commit? Or was it the github availability outage yesterday and the test should just be rerun?

Probably a github issue - I've restarted the check, and now the checkout seems to have worked correctly :)

> If that fixes LLVM 21, we could add it to CI and raise the cmake experimental flag..

I think this fixes LLVM 22, not 21. IIRC 21 was already reported to work, so we should definitely consider raising supported LLVM version. Should check JIT on AMD/NVIDIA though.

alexschroeter:
I found this PR https://github.com/llvm/llvm-project/pull/147835, which says

> PS: TagType::getDecl is renamed to getOriginalDecl in this patch, just for easier to rebasing. I plan to rename it back after this lands.

Since 22 is not officially released but ROCm has released their own with 7.2.0, it seems the situation is even worse. ROCm has released with the rename still there but clang has reverted the rename https://github.com/llvm/llvm-project/pull/147835#issuecomment-3399156873

Not quite sure what the right way to handle this is.

fodinabor:
And that's why we dont officially support ROCm clang :)

You can try and use the rocm version macros here: https://github.com/AdaptiveCpp/AdaptiveCpp/blob/0655cfc593e52672b441d0de0163eff3232ec249/src/compiler/CMakeLists.txt#L368
Or have a look at AMDs attempt of communicating ROCm llvm versions: https://github.com/ROCm/ROCm/issues/1680

illuhad:
Yeah, I agree with @fodinabor  -- we should probably try a ROCm-specific workaround here.

alexschroeter:
Ok, correct me if I am wrong but checking for ROCM_CLANG_VERSION_MAJOR won't work because it only tells which rocm version is linked. This however doesn't guarantee that the used clang is not some other version. And because AMD refuses to add some sort of flag that the compiler is not the original clang we don't really have chance to check this way.

So what about some cmake check that verifies if getDecl or getOriginalDecl compilation fails. This would have the benefit of also hitting on in-between development versions in llvm 21. I have never used such a test but here is what AI came up with. 

```
include(CheckCXXSourceCompiles)

set(CMAKE_REQUIRED_INCLUDES ${LLVM_INCLUDE_DIRS} ${CLANG_INCLUDE_DIRS})
set(CMAKE_REQUIRED_DEFINITIONS ${LLVM_DEFINITIONS})

check_cxx_source_compiles("
#include <clang/AST/Type.h>
void test(const clang::RecordType *RT) {
  (void)RT->getOriginalDecl();
}
int main() { return 0; }
" ACPP_HAS_GET_ORIGINAL_DECL)
```

and use this as the guard?

Let me know what you think. I don't feel good about this but I don't see another way to check if it is the ROCm clang.

illuhad:
> Ok, correct me if I am wrong but checking for ROCM_CLANG_VERSION_MAJOR won't work because it only tells which rocm version is linked. This however doesn't guarantee that the used clang is not some other version

I'm not sure I understand the issue. As far as I can see `ROCM_CLANG_VERSION_MAJOR` is only set if we are building against ROCm's clang/LLVM. I think we already use it to test for ROCm-specific LLVM issues.

alexschroeter:
> > Ok, correct me if I am wrong but checking for ROCM_CLANG_VERSION_MAJOR won't work because it only tells which rocm version is linked. This however doesn't guarantee that the used clang is not some other version
> 
> I'm not sure I understand the issue. As far as I can see `ROCM_CLANG_VERSION_MAJOR` is only set if we are building against ROCm's clang/LLVM. I think we already use it to test for ROCm-specific LLVM issues.

Ok, I didn't go far enough back in the CMakeLists, I didn't see that this is comming from the regex and assumed this is just the ROCm Version on the system.
I changed the commit to guard against clang from only rocm 7.2.0 since it seems reasonable this might be the only version affected by this. But we won't know until the next rocm is released.

include/hipSYCL/compiler/Frontend.hpp
```diff
@@ -88,7 +88,7 @@ class CompleteCallSet : public clang::RecursiveASTVisitor<CompleteCallSet> {
         // type that can be constructed on the GPU also can and will be destructed).
         if(auto Ptr = llvm::dyn_cast_or_null<clang::PointerType>(Callee->getThisType()->getCanonicalTypeUnqualified()))
           if(auto Record = llvm::dyn_cast<clang::RecordType>(Ptr->getPointeeType()))
-#if LLVM_VERSION_MAJOR > 21
+#if ROCM_CLANG_VERSION_MAJOR == 7 && ROCM_CLANG_VERSION_MINOR == 2 && ROCM_CLANG_VERSION_PATCH == 0
```
fodinabor: I believe this is a bit too strict. At least in the patch version ROCM doesn't change its compiler version AFAIK.
Minor.. I don't know.. (7.9 or 7.10 might also be affected btw..)..

alexschroeter: I agree that the patch version is unlikely. The minor has to be involved because 7.1.1 comes with 20, I believe.
Do you think it should be `ROCM_CLANG_VERSION_MINOR > 1` or would you involve LLVM_VERSION_MAJOR. I personally would guard tightly and change when it becomes relevant, but I don't have the level of experience you have. If you have a different suggestion I am all ears.

fodinabor: Then, let's just start with the following for now:
```suggestion
#if ROCM_CLANG_VERSION_MAJOR == 7 && ROCM_CLANG_VERSION_MINOR == 2
```

include/hipSYCL/compiler/Frontend.hpp
```diff
@@ -88,7 +88,11 @@ class CompleteCallSet : public clang::RecursiveASTVisitor<CompleteCallSet> {
         // type that can be constructed on the GPU also can and will be destructed).
         if(auto Ptr = llvm::dyn_cast_or_null<clang::PointerType>(Callee->getThisType()->getCanonicalTypeUnqualified()))
           if(auto Record = llvm::dyn_cast<clang::RecordType>(Ptr->getPointeeType()))
+#if ROCM_CLANG_VERSION_MAJOR == 7 && ROCM_CLANG_VERSION_MINOR == 2
```
illuhad: I think you need to check if `ROCM_CLANG_VERSION_MAJOR` is defined at all first. It's only going to be available when building against ROCm clang. But you can assume that if you have `ROCM_CLANG_VERSION_MAJOR`, you also have the others:

```
#if defined(ROCM_CLANG_VERSION_MAJOR) && ROCM_CLANG_VERSION_MAJOR ==7 && ROCM_CLANG_VERSION_MINOR == 2
```

alexschroeter: I think my version would still work because an undefined macros should default to 0 if I understand correctly and unless we are checking `== 0` this should still have the same effect. But I can add the is `defined` check to make it more explicit and I don't know if my variant might produce warnings about undefined macros, which are ugly.

illuhad (approved):
Thanks!

alexschroeter:
Do you want to add the versions tested in your CI/CD also, or do you want to do this separately?

illuhad:
> Do you want to add the versions tested in your CI/CD also, or do you want to do this separately?

We don't test AdaptiveCpp built against ROCm LLVM in CI. This configuration is only really supported on a best effort basis due to issues like the one you discovered. ROCm LLVM is not a stable target.

PR #1972, opened 2026-02-13
[KHR] Add KHR_PREFETCH_HOST

illuhad:
We already have this feature, but we have not yet exposed it with `khr_` prefix as proposed now: https://github.com/KhronosGroup/SYCL-Docs/pull/960


include/hipSYCL/sycl/handler.hpp
```diff
@@ -688,6 +688,10 @@ class handler {
     _command_group_nodes.push_back(node);
   }
 
+  void khr_prefetch_host(const void *ptr, std::size_t num_bytes) {
```
al42and: As you noted in https://github.com/KhronosGroup/SYCL-Docs/pull/960, `const void *ptr` makes more sense, but, since all of `queue::khr_prefetch_host` use non-const `ptr`, shall we do the same here, for consistency?

illuhad: Yeah, I was considering that. It's really a tradeoff between spec conformance and functionality/consistency. Ultimately, I decided to implement it exactly as in the spec, since we already have the AdaptiveCpp extension (without `khr_` prefix) that uses `const void*`.

But I really don't have a strong opinion here. Do you think we should use `const void*`?

al42and: > But I really don't have a strong opinion here. Do you think we should use const void*?

I don't foresee much practical difference, but I think following the spec is better. 
If someone _needs_ `const void*`, they already have the ACpp extension. If someone, for whatever reason, writes a spec-compliant code that uses a pointer of this function and cares about its type, and this code does not compile, that's bad. So,

```suggestion
  void khr_prefetch_host(void *ptr, std::size_t num_bytes) {
```

?

illuhad: Oh! I totally missed that even this PR was not consistent. The intent was to use `void*` everywhere for the KHR API. Thanks for catching that!

al42and (approved):

al42and (approved):

PR #1974, opened 2026-02-16
Improve AMDClang version parsing

al42and:
Some versions self-identify as "AMD clang version 22.0.0git", with a suffix.

Also, escape dots to only match dots.


fodinabor:
Potentially related? 

https://github.com/AdaptiveCpp/AdaptiveCpp/pull/807

illuhad (approved):
Thanks!

PR #1975, opened 2026-02-19
Warn by default if legacy compilation flows are used

illuhad:
I'm getting increasingly annoyed that people still publish results with the old `hip` and `cuda` flows, without a good reason to do so. Doing so typically:
- causes a more brittle setup because the set of LLVM versions compatible with a specific CUDA or ROCm version is narrower and less well-defined compared to `generic`
- may cause performance loss, because optimization for the last 3 years has focused on `generic`, and many optimizations (like the entire adaptivity stuff) is unavailable outside of `generic`
- some functionality like PCUDA is exclusively supported in `generic`

We already make it very clear in the documentation that people should just use `generic`. I don't want people to publish suboptimal results, or to have a poor experience.

So I think it's time to start emitting warning when people use `hip` or `cuda`. This PR causes `acpp` to emit a warning when `hip` or `cuda` targets are used.

We're not emitting warnings for `omp` yet because the use case there is still a bit wider (sanitizers, debugging, some clang features that are available in `omp.accelerated` but not necessarily in `generic` for now).

The warning can be deactivated using `--acpp-allow-legacy-flows`.

@al42and I've tried to be gentle with the Gromacs use case, and tried my best to not imply that Gromacs is doing something improperly. Please take a look if you are okay with it. Also, note that the warning is only active if AdaptiveCpp has been built with the generic compiler. So, in a typical Gromacs deployment where AdaptiveCpp is built against ROCm LLVM and SSCP is thus disabled, users should not even see it.

### DIFF
--- a/bin/acpp
+++ b/bin/acpp
@@ -607,7 +607,7 @@ class acpp_config:
 """  (Experimental) Treat all functions implicitly as SYCL_EXTERNAL. Only supported with generic target.
   This currently only works with translation units that include the sycl.hpp header."""),
       'allow-legacy-flows' : option("--acpp-allow-legacy-flows", "ACPP_ALLOW_LEGACY_FLOWS", "default-allow-legacy-flows",
-"""  Do not warn when the old 'hip' and 'cuda' targets are used, even though they might not result in best performance."""),
+"""  Do not warn when the old 'hip' and 'cuda' targets are used."""),
       'pcuda' : option("--acpp-pcuda", "ACPP_PCUDA", "default-pcuda",
 """  Enable AdaptiveCpp portable CUDA (PCUDA) support as input language.
   This is *exclusively* supported with --acpp-targets=generic."""),
--- a/bin/acpp
+++ b/bin/acpp
@@ -526,7 +526,7 @@ class acpp_config:
                                    Uses Minicoro for nd_range parallel_for support.
                - omp.accelerated: Uses clang as host compiler to enable compiler support
                                   for nd_range parallel_for (see --acpp-use-accelerated-cpu).
-      * cuda - CUDA backend
+      * cuda - Legacy CUDA interoperability backend
                Requires specification of targets of the form sm_XY,
                e.g. sm_70 for Volta, sm_60 for Pascal
                Backend Flavors:
@@ -536,7 +536,7 @@ class acpp_config:
                                            multipass mode.
       * cuda-nvcxx - CUDA backend with nvc++. Target specification is optional;
                if given requires the format ccXY.
-      * hip  - HIP backend
+      * hip  - Legacy HIP interoperability backend
                Requires specification of targets of the form gfxXYZ,
                e.g. gfx906 for Vega 20, gfx900 for Vega 10
                Backend Flavors:
@@ -606,6 +606,8 @@ class acpp_config:
       'is-export-all' : option("--acpp-export-all", "ACPP_EXPORT_ALL", "default-export-all",
 """  (Experimental) Treat all functions implicitly as SYCL_EXTERNAL. Only supported with generic target.
   This currently only works with translation units that include the sycl.hpp header."""),
+      'allow-legacy-flows' : option("--acpp-allow-legacy-flows", "ACPP_ALLOW_LEGACY_FLOWS", "default-allow-legacy-flows",
+"""  Do not warn when the old 'hip' and 'cuda' targets are used, even though they might not result in best performance."""),
       'pcuda' : option("--acpp-pcuda", "ACPP_PCUDA", "default-pcuda",
 """  Enable AdaptiveCpp portable CUDA (PCUDA) support as input language.
   This is *exclusively* supported with --acpp-targets=generic."""),
@@ -1119,6 +1121,13 @@ class acpp_config:
   def is_dryrun(self):
     try:
       return self._is_flag_set("is-dryrun")
+    except OptionNotSet:
+      return False
+    
+  @property
+  def are_legacy_flows_allowed(self):
+    try:
+      return self._is_flag_set("allow-legacy-flows")
     except OptionNotSet:
       return False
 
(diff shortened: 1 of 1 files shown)


bin/acpp
```diff
@@ -606,6 +606,8 @@ class acpp_config:
       'is-export-all' : option("--acpp-export-all", "ACPP_EXPORT_ALL", "default-export-all",
 """  (Experimental) Treat all functions implicitly as SYCL_EXTERNAL. Only supported with generic target.
   This currently only works with translation units that include the sycl.hpp header."""),
+      'allow-legacy-flows' : option("--acpp-allow-legacy-flows", "ACPP_ALLOW_LEGACY_FLOWS", "default-allow-legacy-flows",
+"""  Do not warn when the old 'hip' and 'cuda' targets are used, even though they might not result in best performance."""),
```
al42and: Nit: I don't think repeating the warning here is useful. If people want to silence it, I think they've read it :)

```suggestion
"""  Do not warn when the old 'hip' and 'cuda' targets are used."""),
```

bin/acpp
```diff
@@ -606,6 +606,8 @@ class acpp_config:
       'is-export-all' : option("--acpp-export-all", "ACPP_EXPORT_ALL", "default-export-all",
 """  (Experimental) Treat all functions implicitly as SYCL_EXTERNAL. Only supported with generic target.
   This currently only works with translation units that include the sycl.hpp header."""),
+      'allow-legacy-flows' : option("--acpp-allow-legacy-flows", "ACPP_ALLOW_LEGACY_FLOWS", "default-allow-legacy-flows",
```
al42and: How about `-Wno-acpp-legacy-flows`? For consistency with other compilation warnings, and also because, even without this flag, the legacy flows are still _allowed_. Unless you have plans to completely remove them in the future (in which case I ask you to reconsider)

bin/acpp
```diff
@@ -2004,6 +2013,17 @@ class compiler:
                       "simultaneously in non-explicit multipass; enabling explicit "
                       "multipass compilation.")
         self._is_explicit_multipass = True
+    
+    if "hip" in self._targets or "cuda" in self._targets:
+      if config.has_plugin_sscp_compiler:
+        if not config.are_legacy_flows_allowed:
+          print_warning("The 'hip' and 'cuda' compilation flows are not useful outside "
+                        "of specific interoperability use cases or applications that have "
+                        "been engineered specifically around them (e.g. Gromacs)."
```
al42and:
```suggestion
                        "been engineered specifically around them (e.g. GROMACS)."
```

bin/acpp
```diff
@@ -2004,6 +2013,17 @@ class compiler:
                       "simultaneously in non-explicit multipass; enabling explicit "
                       "multipass compilation.")
         self._is_explicit_multipass = True
+    
+    if "hip" in self._targets or "cuda" in self._targets:
+      if config.has_plugin_sscp_compiler:
+        if not config.are_legacy_flows_allowed:
+          print_warning("The 'hip' and 'cuda' compilation flows are not useful outside "
```
al42and: I'd argue there could still be some use (IIRC, in your earlier published results, SSCP was not _always_ faster, just most of the time)

```suggestion
          print_warning("The 'hip' and 'cuda' compilation flows are not recommended outside "
```

bin/acpp
illuhad: > How about -Wno-acpp-legacy-flows?

Mainly because all AdaptiveCpp-specific flags are prefixed with `--acpp-`. I'm not sure if somewhere, something assumes that this is always the case. But you're right that even with this PR, it's still allowed. So perhaps `--acpp-wno-legacy-flows` or `--acpp-Wno-legacy-flows` ?

> Unless you have plans to completely remove them in the future (in which case I ask you to reconsider)

No, this is not planned at all :)

bin/acpp
illuhad: > IIRC, in your earlier published results, SSCP was not always faster, just most of the time

I think the only place where I've published (in a paper) systematic comparisons between SSCP and SMCP is the original SSCP paper. However, the SSCP implementation back then was still early and partially immature, and pretty much all of the perf differences we've observed there have been understood and fixed afterward -- things like different builtin mapping (exact vs approximate), not leveraging `-ffast-math` properly yet...stuff like that.
Of course this was also before the adaptivity stuff, which can add ~10-30% performance on top.

So the expectation definitely is that SSCP is going to be faster, although there can always be edge cases where things are different.

`not recommended` is a little bit of a different statement semantically compared to `not useful`: One is our advice, the other one is focused on the experience of the user.
I'd like to continue focusing on the latter, but we could weaken it a little to `not typically useful` or so.

bin/acpp
al42and: > So perhaps `--acpp-wno-legacy-flows` or `--acpp-Wno-legacy-flows`?

Or `--acpp-no-warn-legacy-flows`? More consistent with other `--acpp-*` options, yet is not as confusing as "allow"

bin/acpp
al42and: > not typically useful

Sounds good

bin/acpp
tdavidcl: I'm also in favor of `--acpp-no-warn-legacy-flows` or `-Wno-acpp-legacy-flows`.
Also since acpp has some specific warnings sometimes (`loop not vectorized` for example which is not there by default in OpenMP) wouldn't it be good to have the -W / -Wno format then and keep `--acpp` flags for functional changes only or tools ?
Additionally, even without `--acpp-allow-legacy-flows` they are still allowed but you just get a warning

bin/acpp
tdavidcl: Last time I checked (about 9 or so months ago) on Lumi ROCM llvm + acpp SMCP was still the fastest for Shamrock by small margin (about 3-4 %). So there are still so edge cases in HPC centers where SMCP is favored.

| Config | Rate (part/sec) |
| --- | ----------- |
| Acpp SSCP | 33530835 |
| Acpp SMCP | 33630866 |
| Acpp SMCP Rom/LLVM | 34359364 |
| Intel LLVM v6 | 32455095 |

illuhad: Such a tiny difference in one application is not really meaningful when comparing the technological level of two compilers. You can easily end up with such differences simply by moving code around, changing code in a seemingly unrelated way etc.
More often than not, this is caused by
- people looking more often at the performance numbers of one compiler during development, thus creating an optimization bias
- Incomparable scenario. Maybe SSCP calculates one builtin exactly whereas SMCP approximates; maybe compiler flags are subtly different (SMCP shares host compiler flags, while for SSCP host/device compilation flags are strictly separate).

So this does not really say anything about whether to prefer SMCP or SSCP without knowing where exactly that difference is coming from.

bin/acpp
illuhad: > Also since acpp has some specific warnings sometimes (loop not vectorized for example which is not there by default in OpenMP) wouldn't it be good to have the -W / -Wno format then and keep --acpp flags for functional changes only or tools ?

This warning sometimes happen with the `omp.accelerated` flow whenever LLVM can vectorize across work items. So this is an expected behavior in many cases, and IIRC the plan was to just suppress that warning because it's in general not useful for end users.

I'm a bit concerned about `-W(no)` AdaptiveCpp flags because it's a path that potentially risks name collisions with whatever clang/LLVM is doing.

AdaptiveCpp-specific LLVM flags are passed as `-mllvm -acpp-*` flags, so we'd have to massage a `-W` flag anyway to fit into this format, so it seems like additional effort.

bin/acpp
illuhad: Changed to `not typically useful`.

bin/acpp
illuhad: Changed to `--acpp-no-warn-legacy-flows` as suggested.

PR #1977, opened 2026-02-24
usm: handle zero-sized allocations

al42and:
SYCL standard requires graceful handling of zero-sized USM allocations,
akin to `std::malloc`: return nullptr or a free'able pointer.
However, not all backends behave this way.
So, we now handle this case ourselves. Added tests too.

### DIFF
--- a/include/hipSYCL/sycl/usm.hpp
+++ b/include/hipSYCL/sycl/usm.hpp
@@ -74,6 +74,8 @@ namespace {
 inline void *malloc_device(size_t num_bytes, const device &dev,
                            const context &ctx,
                            const property_list &propList = {}) {
+  if (num_bytes == 0)
+    return nullptr;
   rt::allocation_hints hints = create_hints_from_proplist(propList);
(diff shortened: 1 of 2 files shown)


illuhad:
Thanks, but I wonder, would it perhaps be better to handle this at the SYCL level as part of the implementation of `malloc_device` and friends?
Then
- we'd have the same behavior for every backend
- also works for new backends (e.g. metal)
- SYCL-specific requirements remain in the SYCL layer, allowing other models to potentially have other behavior.


Also, CI indicates that AMD might be affected as well.

al42and:
> would it perhaps be better to handle this at the SYCL level as part of the implementation of malloc_device and friends?

Yeah, makes sense. I wanted to keep CUDA/HIP untouched since the `cudaMalloc` was behaving in line with how SYCL does things, but turns out `cudaMallocManaged` explicitly disallows `size == 0` :unamused:

illuhad (approved):
Thanks!

PR #1983, opened 2026-02-27
[Metal] Enable pcuda api

resetius:


illuhad (commented):
One nitpick, looks good apart from that :)

include/hipSYCL/pcuda/pcuda_runtime.hpp
```diff
@@ -165,6 +165,8 @@ struct pcudaDeviceProp {
   int memoryBusWidth;
   int l2CacheSize;
   int maxThreadsPerMultiProcessor;
+  bool hasFp64;
+  bool hasAtomic64;
```
illuhad: This is not in CUDA/HIP, right? Tht would make it a PCUDA extension. We generally prefix extensions that are only in PCUDA with `pcuda`, so in that case `pcudaHasFp64` etc will be better -- avoid potential name collisions in the future if NVIDIA/AMD introduce something similar but with somewhat different semantics.

illuhad (commented):
Oh, man, forgot to save my comment. My mistake.

include/hipSYCL/pcuda/pcuda_runtime.hpp
resetius: Do I understand correctly that you want to add fields with the pcuda prefix to the **pcuda**DeviceProp structure? So the code would look something like this
```
pcudaDeviceProp prop;
pcudaGetDeviceProperties(&prop, dev);
prop.pcudaHasFp64;
```

illuhad: Yes, that would be my suggestion.

We don't really have a precedent for fields in a struct that are extensions, but only for extension functions. Those carry `pcuda` prefix: https://github.com/AdaptiveCpp/AdaptiveCpp/blob/develop/doc/pcuda.md#extensions

If you have a better idea, let me know :)

illuhad (approved):
Thanks!

drongolex:
solid stuff, metal support in pcuda was long overdue

PR #1986, opened 2026-03-02
Support LLVM 22

al42and:
- PassPlugin moved in f54df0d09e19ec6b205cb0af45c7ecea2fd8aeff
- -enable-unsafe-fp-math removed in 3656f6f22652b01eb13fc2f39afcf36396d64c4f

Needs an updated branch in AdaptiveCpp/SPIRV-LLVM-Translator to work fully


al42and:
An automatically-updated SPIRV-LLVM-Translator repo setup at https://github.com/al42and/SPIRV-LLVM-Translator-ACpp (not super well-tested, but seems to work). I suggest doing the same in https://github.com/AdaptiveCpp/SPIRV-LLVM-Translator, since manual updates seem to be a hassle

illuhad:
> since manual updates seem to be a hassle

Automation is good -- but it's 1-2 git commands. Updating the translator is not the limiting factor when adding support for a new LLVM version, given that updating LLVM in general requires:
- fixing any API compatibility issues in the code
- validating all backends
- Adding new LLVM to CI
- Accepting new LLVM in cmake

Removal of `--enable-unsafe-fp-math` would also require performance testing (is it just removed without replacement in LLVM?).

al42and:
> Automation is good -- but it's 1-2 git commands. Updating the translator is not the limiting factor when adding support for a new LLVM version

There's also the question of getting bugfixes into older branches (although I haven't ever been bitten by those personally). And not having the SPIR-V Translator branch makes even local testing with `ACPP_EXPERIMENTAL_LLVM=ON` a hassle.

> is it just removed without replacement in LLVM?

As far as I understand, there's no direct replacement. It was removed as part of a refactoring to enable https://github.com/llvm/llvm-project/pull/105746. Related discussions: https://discourse.llvm.org/t/rfc-honor-pragmas-with-ffp-contract-fast/80797 and https://discourse.llvm.org/t/allowfpopfusion-vs-sdnodeflags-hasallowcontract/80909.

ACpp already enables `-ffp-contract=fast` at -O2 or `-ffast-math`. We set the denorm and other flags right below. That said, I'm no LLVM expert, so I primarily looked at how other projects (e.g., https://github.com/apache/tvm/pull/18818/changes) solved the problem.

CMakeLists.txt
```diff
@@ -436,7 +436,7 @@ if(BUILD_CLANG_PLUGIN)
     endif()
   endif()
   if(ENFORCE_COMPATIBLE_LLVM AND NOT ACPP_EXPERIMENTAL_LLVM)
-    if(${LLVM_VERSION_MAJOR} GREATER 20)
+    if(${LLVM_VERSION_MAJOR} GREATER 22)
       message(SEND_ERROR "LLVM versions greater than 20 are not yet tested/supported, use -DACPP_EXPERIMENTAL_LLVM=ON if you still wish to try your luck.")
```
alexschroeter: The message should increase the version as well right?

```suggestion
      message(SEND_ERROR "LLVM versions greater than 22 are not yet tested/supported, use -DACPP_EXPERIMENTAL_LLVM=ON if you still wish to try your luck.")
```

src/compiler/AdaptiveCppLlvmPasses.cpp
```diff
@@ -39,7 +39,11 @@
 
 #include "llvm/Pass.h"
 #include "llvm/Passes/PassBuilder.h"
+#if (LLVM_VERSION_MAJOR < 22) || (defined(ROCM_CLANG_VERSION_MAJOR) && (ROCM_CLANG_VERSION_MAJOR == 7) && (ROCM_CLANG_VERSION_MINOR == 2))
 #include "llvm/Passes/PassPlugin.h"
+#else
+#include "llvm/Plugins/PassPlugin.h"
+#endif
```
alexschroeter, line 46: This is a ROCm LLVM specific change that I believe will also be adopted into LLVM 23. 
See [here](https://github.com/AdaptiveCpp/AdaptiveCpp/issues/2041) specifically this [commit](https://github.com/llvm/llvm-project/commit/f54df0d09e19ec6b205cb0af45c7ecea2fd8aeff)

fodinabor: That change is already part of 22 (see tag list of the commit).
And this PR already includes upstream LLVM >= 22 here..

ferdymercury:
Rebase is needed due to some conflicts

Also, maybe https://github.com/AdaptiveCpp/AdaptiveCpp/commit/949075870c0e1fa1909ad3f3e6bf65c989bf1950 by @tdavidcl can be used as inspiration to see what equivalent changes are needed here

al42and:
> Removal of --enable-unsafe-fp-math would also require performance testing (is it just removed without replacement in LLVM?).

This affects LLVMToHost and LLVMToPTX passes. As Host is not a performance target, so I ran GROMACS with SSCP+CUDA, and seen **no** performance degradation from LLVM 20.1.8 (on develop) to LLVM 22.1.2 (on this PR) over a small set of benchmarks.

To get broader coverage, I also ran a subset of HeCBench tests.

Hardware: NVIDIA RTX3060 (sm_86) with CUDA 12.9; AMD RX6400 (gfx1034) with ROCm 7.2.4; Intel Arc A770 (Xe-HPG) with CR 26.05. Software: ACpp `develop` (ae89256486c1e470de5b77de5a1c430ff798ad03) + LLVM 20.1.8, ACpp `develop` + LLVM 21.1.8, and this PR + LLVM 22.1.2 

Results: https://docs.google.com/spreadsheets/d/17lC3RglTjfX5yeAlnUYnLM2AUR8takDHCFvnz1fpuD4/view?usp=sharing

No degradation of SSCP/PTX performance observed; some improvements even (`softmax` speedup is quite nice). SMCP also had no artifacts.

But with SSCP/GCN and SSCP/SPIR-V, things got more interesting. I vibecoded a few workarounds which restored performance, which I describe below but I don't understand them well enough to propose here :)

Remaining issues:
- Crash with libc++ due to [a problem with `small_vector`](https://github.com/KonanM/small_vector/issues/4); blocker for macOS.
- ROCm 7.2.0 (which identifies as LLVM 22) cannot read some IR  attributes from LLVM 22.1 ([SanitizeAllocToken](https://github.com/llvm/llvm-project/commit/224873d7acab430d29c978136418c40fa028a40d) and [NoCreateUndefOrPoison](https://github.com/llvm/llvm-project/commit/f037f413506af9e32898e102f391175a3a2852ef)), which are part of TheRock 7.11 but are not in ROCm 7.2.0-7.2.4, breaking SSCP/AMDGCN for many HeCBench tests.
- HeCBench `fft` test segfaults with SSCP/SPIR-V and SSCP/AMDGCN pathts due to a change in the number of arguments of LifetimeIntrinsic (https://github.com/llvm/llvm-project/commit/c23b4fbdbb70f04e637b488416d8e42449bfa1fb).

Out-of-scope:
- With both LLVM 21 and 22, sycl_tests with SSCP on Intel GPUs required using acq_rel ordering instead of relaxed in group collectives; this seems to make sense to me and does not harm performance on HeCBench, but is a whole new can of worms which is not triggered by LLVM 22

illuhad:
> As Host is not a performance target

Hu? Maybe it is not for you :)

Thanks for testing!

> ROCm 7.2.0 (which identifies as LLVM 22) cannot read some IR attributes from LLVM 22.1 ([SanitizeAllocToken](https://github.com/llvm/llvm-project/commit/224873d7acab430d29c978136418c40fa028a40d) and [NoCreateUndefOrPoison](https://github.com/llvm/llvm-project/commit/f037f413506af9e32898e102f391175a3a2852ef)), which are part of TheRock 7.11 but are not in ROCm 7.2.0-7.2.4, breaking SSCP/AMDGCN for many HeCBench tests.

Probably ROCm 7.2 is on LLVM 22.0 (not 22.1), so compatibility guarantee is broken. We'd need a newer ROCm version.

> With both LLVM 21 and 22, sycl_tests with SSCP on Intel GPUs required using acq_rel ordering instead of relaxed in group collectives; this seems to make sense to me and does not harm performance on HeCBench, but is a whole new can of worms which is not triggered by LLVM 22

Interesting. CI runs fine with LLVM 18 on Intel Arc B580.

al42and:
> Probably ROCm 7.2 is on LLVM 22.0 (not 22.1), so compatibility guarantee is broken.

Yes. But there is no newer ROCm released, so hacking together a solution seemed useful for testing purposes at least.

Whether in the main codebase we want to have a workaround or a warning is a question to you :) But at least a warning is needed; currently, CMake only checks `(LLVM|ROCM)_VERSION_MAJOR`, so no `ACPP_EXPERIMENTAL_LLVM` is needed to shoot oneself in the foot here.

> Interesting. CI runs fine with LLVM 18 on Intel Arc B580.

Tested with LLVM 18, same problem. I suspect it is more about the igc version. The way `__acpp_sscp_work_group_barrier` is lowered to SPIR-V should be rather straightforward. But reordering loads and stores around a `relaxed` barrier is very much up in the air.

illuhad:
> Whether in the main codebase we want to have a workaround or a warning is a question to you :) But at least a warning is needed; currently, CMake only checks (LLVM|ROCM)_VERSION_MAJOR, so no ACPP_EXPERIMENTAL_LLVM is needed to shoot oneself in the foot here.

Yes, we should also handle the case when both `ROCM_VERSION_MAJOR` and `LLVM_VERSION_MAJOR` are equal. I'm not sure if we can assume that ROCm is *always* going to be on `ROCM_VERSION_MINOR == 0` (instead of 1). If so, we'd only need to change the GREATER here to greater-equal.
https://github.com/AdaptiveCpp/AdaptiveCpp/blob/54c31cd4c5b29d8184fb87b6974ebc67fbb95bdf/CMakeLists.txt#L376
Otherwise we need a more generalized version comparison.

> Tested with LLVM 18, same problem. I suspect it is more about the igc version. The way __acpp_sscp_work_group_barrier is lowered to SPIR-V should be rather straightforward. But reordering loads and stores around a relaxed barrier is very much up in the air.

Which IGC version are you using? This is what we have in CI:
https://github.com/AdaptiveCpp/AdaptiveCpp/blob/54c31cd4c5b29d8184fb87b6974ebc67fbb95bdf/.github/workflows/runner.def#L19

al42and:
> Which IGC version are you using? This is what we have in CI:

Tested the same version, still fails. Weird. Perhaps [this](https://github.com/intel/intel-graphics-compiler/blob/ea4a76ab6f07e243e0f49b2c60e14059e30360f2/IGC/Compiler/CISACodeGen/Platform.hpp#L236) is related (A770 is DG2 / Xe-HPG)? Interestingly, the reductions work for smaller (`local_size=25`) but not larger (144 and 256) work-groups.

I'm still pretty sure the relaxed barriers are not correct for the algorithms, even if they happen to work. `sscp/ptx` uses `__syncthreads` which guarantees visibility; `sscp/amdgpu` explicitly does acquire-release, regardless of semantics requested -- so, the things work fine there.

illuhad:
> I'm still pretty sure the relaxed barriers are not correct for the algorithms, even if they happen to work. sscp/ptx uses __syncthreads which guarantees visibility; sscp/amdgpu explicitly does acquire-release, regardless of semantics requested -- so, the things work fine there.

Yes, I agree that probably acquire-release would be safer. Can you create a PR?

al42and:
Welp, CI does not seem to push the new llvm-22 images to the registry and so the llvm-22 jobs fail. But it seems to work locally on x86_64.

Limitations:
- The ~~latest~~ previous "stable" ROCm release, 7.2.4, is unusable with SSCP. Not really a bug (it's based on LLVM 22.0, so no guarantees it will work when paired with LLVM 22.1), but annoying. Can be filtered out of the IR after `opt` invocation if we want to have this workaround, but I lean towards not fixing it here.
- **UPD1**: The latest "stable" ROCm release, 10.0, is *also* unusable with SSCP because they moved/renamed device bitcode libraries; seem to require a fix on LLVM side? Using amdclang with "minimal" profile  works.

Two style questions:

- I added ROCm clang handling for header paths in SSCP and Metal code. I find it easier to have all preprocessor conditionals the same, but there we can never really encounter ROCm clang, so we can also just keep it shorter and simpler there, checking only LLVM_VERSION_MAJOR.
- Are CI resources unlimited, or should we consider dropping some intermediate LLVM versions? LLVM 16 is unlikely to break as long as we test LLVM 15 and 17.

Two issues remain:

- Haven't tested Vulkan backend yet :(
- libc++22 build is broken due to the `small_vector` issue, which completely breaks macOS/Metal build for me.
  - I have an workaround for that (https://github.com/KonanM/small_vector/pull/5), but then the ACpp's tests fail.
  - **UPD1**: Things work with #2198 + #2199, thanks @resetius!

@resetius, would you be able to look into the Metal problems? Grab this branch + small_vector.hpp from KonanM/small_vector#5, llvm@22 and lld@22 from Homebrew, and run the ACpp's `sycl_tests`. I see a few failures or IR->Metal generation around `vec<float, 4>` (`error: assigning to 'bool' from incompatible type 'bool
  __attribute__((ext_vector_type(4)))'` -- on the only IR line that changes between LLVM21 and LLVM22), and another with `fetch_min`/`fetch_max` in `atomic_tests/fetch_op` producing a wrong value. If you have trouble reproducing, I'd be happy to give more info.

resetius:
> `vec<float, 4>` (`error: assigning to 'bool' from incompatible type 'bool __attribute__((ext_vector_type(4)))'` -- on the only IR line that changes between LLVM21 and LLVM22), and another with `fetch_min`/`fetch_max` in `atomic_tests/fetch_op` producing a wrong value. If you have trouble reproducing, I'd be happy to give more info.

1. SimplifyCFGPass in LLVM-22 breaks structurized-CFG which was built by StructurizeCFGPass
2. Unexpected vector type. Need to replace ==/!= with all/any in generated code

resetius:
1. atomics: https://github.com/AdaptiveCpp/AdaptiveCpp/pull/2198
2. i128: https://github.com/AdaptiveCpp/AdaptiveCpp/pull/2199

Issue #1970, opened 2026-02-12
Investigate static constexpr in SSCP on Windows


illuhad:
There was some evidence that globals generated by `static constexpr` are not handled correctly by SSCP on Windows in #1903. We should investigate this.

PR #1987, opened 2026-03-02
Fix compilation error of vec::as<>() function

thomasRoglin:
as<>() function is const, so reinterpret_cast should cast to a `const asT` instead of `asT`

Also adding a regression tests for `vec_as` (mainly copy-pasted from the above `vec_convert` test).

### DIFF
--- a/include/hipSYCL/sycl/libkernel/vec.hpp
+++ b/include/hipSYCL/sycl/libkernel/vec.hpp
@@ -353,13 +353,7 @@ public:
     static_assert(std::is_same_v<VectorStorage, detail::vec_storage<T, N>>,
                   "Reinterpreting swizzled vectors directly is not supported");
 
-    asT result;
-    
-    auto in_ptr = reinterpret_cast<const typename asT::element_type*>(&_data[0]);
-    for(int i = 0; i < N; ++i)
-      result[i] = in_ptr[i];
-
-    return result;
+    return sycl::bit_cast<asT>(*this);
   }
 
   template<int... SwizzleIndices>
--- a/include/hipSYCL/sycl/libkernel/vec.hpp
+++ b/include/hipSYCL/sycl/libkernel/vec.hpp
@@ -14,6 +14,7 @@
 #include "backend.hpp"
 #include "half.hpp"
 #include "multi_ptr.hpp"
+#include "bit_cast.hpp"
 
 #include <cstdint>
 #include <type_traits>
--- a/include/hipSYCL/sycl/libkernel/vec.hpp
+++ b/include/hipSYCL/sycl/libkernel/vec.hpp
@@ -355,7 +355,7 @@ public:
 
     asT result;
     
-    auto in_ptr = reinterpret_cast<typename asT::element_type*>(&_data[0]);
+    auto in_ptr = reinterpret_cast<const typename asT::element_type*>(&_data[0]);
     for(int i = 0; i < N; ++i)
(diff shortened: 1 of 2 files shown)


thomasRoglin:
`vec::as()` new tests seems to be working except for the windows build..
Maybe linked to #1970 ?
Should we disable the test for windows until the problem is solved?

illuhad:
It does not look like it does anything related to `static constexpr` , which #1970 is about. Also, #1970 causes linker errors at JIT time, whereas this issue is about incorrect results.
Maybe it's because we violate strict aliasing rules here? Maybe we need to use `memcpy`/`bit_cast` here? (e.g. `HIPSYCL_INPLACE_BIT_CAST()` macro)

resetius:
in c++20 should work
```
  template<typename asT>
  ACPP_UNIVERSAL_TARGET
  asT as() const {
    static_assert(sizeof(asT) == sizeof(vec<T,N>),
                  "Reinterpreted vector must have same size");
    static_assert(std::is_same_v<VectorStorage, detail::vec_storage<T, N>>,
                  "Reinterpreting swizzled vectors directly is not supported");
   return std::bit_cast<asT>(*this);
  }
```

try sycl::bit_cast  ....

illuhad:
Yep, but we also need to still support C++17. And since we have library-only flows (`omp.library-only`, `cuda-nvcxx`) we cannot rely on `__builtin_bit_cast` either. `sycl::bit_cast` is fine as well as `HIPSYCL_INPLACE_BIT_CAST` which I suggested (which is what `sycl::bit_cast` uses under the hood).

resetius:
```
#define HIPSYCL_INPLACE_BIT_CAST(Tin, Tout, in, out)                           \
  {                                                                            \
    union {                                                                    \
      Tout union_out;                                                          \
      Tin union_in;                                                            \
    } u;                                                                       \
    u.union_in = in;                                                           \
    out = u.union_out;                                                         \
  }
#endif
```
Formally this is UB according to the C++ standard. A union has the notion of an active member: the member that was last written becomes active, and reading from a different member is technically undefined behavior

illuhad:
> Formally this is UB according to the C++ standard. A union has the notion of an active member: the member that was last written becomes active, and reading from a different member is technically undefined behavior

That's true. This code path is only taken for the nvc++ library-only mode which only supports C++17 (so no `std::bit_cast`) and does not support `__builtin_bit_cast` either. It's a workaround around limitations of a specific compiler.

All other compilation flows map this to `__builtin_bit_cast`.

include/hipSYCL/sycl/libkernel/vec.hpp
resetius: please try to replace this implementation with
```
return sycl::bit_cast<asT>(*this);
```

thomasRoglin:
Hello, sorry for the delay, and thanks for the suggested patch. I've pushed it.

illuhad:
Thanks -- looks like this fixed the Windows issue, great :-) I think you probably also need to include the `libkernel/bit_cast.hpp` header for `sycl::bit_cast` in `vec.hpp` to make the remaining error go away.

thomasRoglin:
Hmm, I'm not quite sure why `sycl::bit_cast` isn't recognized during the LINUX LIT CBS tests , but including `bit_cast.hpp` seems to be the solution

illuhad (approved):
Lovely,  thanks! :-)

PR #1989, opened 2026-03-02
[Metal] Use native events

resetius:


resetius:
@illuhad ready for review
i added a test but not sure if i placed it correctly

include/hipSYCL/runtime/metal/metal_queue.hpp
```diff
@@ -83,10 +86,21 @@ class metal_inorder_queue : public inorder_queue
...
+
   MTL::Device* _device = nullptr;
+  MTL::CommandQueue* _command_queue = nullptr;
+  MTL::SharedEvent* _shared_event = nullptr;
+  MTL::SharedEventListener* _event_listener = nullptr;
+  // Atomic because it may be read from arbitrary threads via wait() / query_status()
+  // concurrently with queue submissions on the DAG execution thread.
+  std::atomic<uint64_t> _event_counter{0};
+
+  // Only touched during queue submission on the DAG execution thread, so no atomics needed.
```
illuhad: Is this true? If you have in-order queues with instant submission mode (default), then AdaptiveCpp will submit eagerly directly to the `inorder_executor`. If your `sycl::queue` is then shared across multiple threads, then you can typically get submissions to the same queue from multiple threads.

tests/compiler/sscp/coarse_grained_events.cpp
```diff
@@ -0,0 +1,64 @@
+// RUN: %acpp %s -o %t --acpp-targets=generic
+// RUN: %t | FileCheck %s
+#include <cstring>
+#include <iostream>
+#include <sycl/sycl.hpp>
+
+int main() {
+  sycl::queue q1{sycl::gpu_selector_v, sycl::property::queue::in_order{}};
+  sycl::queue q2{sycl::gpu_selector_v, sycl::property::queue::in_order{}};
```
illuhad: Better use `default_selector_v`. Otherwise the test will fail when run in a scenario where there's no GPU (e.g. github CI).

tests/compiler/sscp/coarse_grained_events.cpp
```diff
@@ -0,0 +1,64 @@
...
+  char *dev4 = sycl::malloc_device<char>(numBytes, q2);  // q2!
+  char *dev5 = sycl::malloc_device<char>(numBytes, q1);
+
+  std::memset(host1, 1, numBytes);
+  std::memset(host1b, 2, numBytes);
+  std::memset(host2, 3, numBytes);
+  std::memset(host5, 42, numBytes);
+
+  // Critical: operations alternate AND use cross-queue allocations
+  q1.submit(cg_props, [&](sycl::handler &h) { h.memcpy(dev1, host1, numBytes); });
```
illuhad: You can also just `q.memcpy()` and construct the queues with `cg_props`, which will be less verbose.

tests/compiler/sscp/coarse_grained_events.cpp
```diff
@@ -0,0 +1,64 @@
+// RUN: %acpp %s -o %t --acpp-targets=generic
```
illuhad: We usually use LIT tests for compiler features, not for runtime/API tests. We also have a coarse-grained events test case in `sycl_tests` (`extensions.cpp`). Perhaps add it there?

include/hipSYCL/runtime/metal/metal_queue.hpp
resetius: _pending_cpu_event / _pending_gpu_event are protected by _impl->lock held in https://github.com/AdaptiveCpp/AdaptiveCpp/blob/develop/include/hipSYCL/sycl/queue.hpp#L398  for the entire duration of submit(). So atomics are not needed. The comment is misleading

illuhad: Okay, makes sense. We might want to just update the comment then.

illuhad (approved):
Thanks!

resetius:
@al42and this PR introduced a bug. I found it while running gromacs: in-order queue semantics violated for device->host->host->device transfer chains.
I have a fix. I'll try to submit a PR by Monday

resetius:
> @al42and this PR introduced a bug. I found it while running gromacs: in-order queue semantics violated for device->host->host->device transfer chains. I have a fix. I'll try to submit a PR by Monday

https://github.com/AdaptiveCpp/AdaptiveCpp/pull/2016

I also found a bug in gromacs itself. I think nobody run configuration with gpu pme compute + cpu fft so far.
I'll probably create an issue

al42and:
Thanks for testing!

> I also found a bug in gromacs itself. I think nobody run configuration with gpu pme compute + cpu fft so far.
I'll probably create an issue

Not quite "nobody", but yes, it's very niche. Please open an issue at https://gitlab.com/gromacs/gromacs/

resetius:
> Thanks for testing!
> 
> > I also found a bug in gromacs itself. I think nobody run configuration with gpu pme compute + cpu fft so far.
> > I'll probably create an issue
> 
> Not quite "nobody", but yes, it's very niche. Please open an issue at https://gitlab.com/gromacs/gromacs/

https://gitlab.com/gromacs/gromacs/-/work_items/5593

PR #1990, opened 2026-03-05
Fix host functions sinpi/cospi

resetius:
see https://github.khronos.org/SYCL_Reference/iface/math-functions.html#sycl-sinpi


illuhad:
Legacy flows are affected as well: https://github.com/AdaptiveCpp/AdaptiveCpp/blob/3bb37cc592da41461ddfd3acfe1f8b64cb288934/include/hipSYCL/sycl/libkernel/generic/hiplike/builtins.hpp#L65

I have a vague memory that earlier versions of the spec had ambiguous wording here, which would explain this.

Anyway -- do you want to focus only on host in this PR, or do you want to address the legacy flows as well?

resetius:
> Legacy flows are affected as well: https://github.com/AdaptiveCpp/AdaptiveCpp/blob/3bb37cc592da41461ddfd3acfe1f8b64cb288934/include/hipSYCL/sycl/libkernel/generic/hiplike/builtins.hpp#L65
> 
> 
> 
> I have a vague memory that earlier versions of the spec had ambiguous wording here, which would explain this.
> 
> 
> 
> Anyway -- do you want to focus only on host in this PR, or do you want to address the legacy flows as well?

I'll fix all sites. Thanks

resetius:
> return hiplike_builtins::__acpp_asin(x) / M_PI;

sinpi returns sin(pi x) (https://github.khronos.org/SYCL_Reference/iface/math-functions.html#sycl-sinpi)
asinpi returns asin / pi (https://github.khronos.org/SYCL_Reference/iface/math-functions.html#sycl-asinpi)

__acpp_sinpi calls sinpi/sinpif from math.h - this is correct


looks like only __acpp_sscp_sinpi_fXX/__acpp_sscp_cospi_fXX are incorrect

illuhad (approved):
Thanks!

PR #1993, opened 2026-03-06
Add lgamma_r implementation for macOS/Windows

resetius:
I'll add a test as a part of follow-up PR with metal math functions


illuhad (approved):
Thanks!

PR #1995, opened 2026-03-08
[SSCP] Fix typos in a few more math buiiltins

al42and:
* sscp/amdgpu: Avoid conversion to float in frexp_f64
* sscp/amdgpu: Fix atanh for float
* sscp/host: Fix minmag for float and double
* sscp/ptx: fix log2 and log10 for double


illuhad:
Thanks -- seems like nvc++ segfaults on of the math functions. Can we identify which one it is, and not test it in case we use nvc++?

al42and:
> Thanks -- seems like nvc++ segfaults on of the math functions. Can we identify which one it is, and not test it in case we use nvc++?

The compiler stacktrace was pointing to `std::ldexp`, which I definitely was not touching. Re-running the job fixed it somehow. Flaky ICE, yay :neutral_face:

illuhad (approved):
Alright, thanks...nvc++ is not a particularly good compiler :(

PR #1996, opened 2026-03-09
[Metal] Implement reduction functions

resetius:


src/libkernel/sscp/metal/scan_helpers.hpp
```diff
@@ -54,24 +64,44 @@ inline T __acpp_sscp_sub_group_inclusive_scan(T value) {
...
   }
 }
 
 template <__acpp_sscp_algorithm_op op, typename T>
-inline T __acpp_sscp_sub_group_exclusive_scan(T value) {
+inline T __acpp_sscp_sub_group_exclusive_scan(T value, T init) {
   if constexpr (op == __acpp_sscp_algorithm_op::plus) {
     return __acpp_sscp_metal_scan<T>("simd_prefix_exclusive_sum", value);
   } else if constexpr (op == __acpp_sscp_algorithm_op::multiply) {
     return __acpp_sscp_metal_scan<T>("simd_prefix_exclusive_product", value);
```
al42and, line 87: `init` ignored?

src/libkernel/sscp/metal/reduction.cpp
```diff
@@ -187,12 +210,47 @@ HIPSYCL_SSCP_CONVERGENT_BUILTIN type __acpp_sscp_work_group_reduce_##type(__acpp
...
+  case __acpp_sscp_algorithm_op::multiply: \
+    return __acpp_sscp_work_group_reduce<__acpp_sscp_algorithm_op::multiply>(value); \
+  case __acpp_sscp_algorithm_op::min: \
+    return __acpp_sscp_work_group_reduce<__acpp_sscp_algorithm_op::min>(value); \
+  case __acpp_sscp_algorithm_op::max: \
+    return __acpp_sscp_work_group_reduce<__acpp_sscp_algorithm_op::max>(value); \
+  case __acpp_sscp_algorithm_op::logical_and: \
+    return __acpp_sscp_work_group_reduce<__acpp_sscp_algorithm_op::logical_and>(value); \
+  case __acpp_sscp_algorithm_op::logical_or: \
+    return __acpp_sscp_work_group_reduce<__acpp_sscp_algorithm_op::logical_or>(value); \
```
al42and: `logical_and` and `logical_or` are only available for `bool`, so no need to define them for floating-point types

src/libkernel/sscp/metal/scan_helpers.hpp
resetius: For product it's always 1 for sum - 0. Metal does not have the initial value in API

src/libkernel/sscp/metal/reduction.cpp
resetius: I split it up: for float, only min/max/add/mul are available, for int*, everything else. In other backends it looks like logical ops are available for arbitrary integer types

src/libkernel/sscp/metal/broadcast.cpp
```diff
@@ -86,22 +87,7 @@ ACPP_SUBGROUP_BROADCAST(f32)
 template<typename T>
 inline T __acpp_sscp_work_group_broadcast(i32 local_id, T value) {
   __attribute__((address_space(3))) T* scratch = __acpp_sscp_get_typed_dynamic_local_memory<T>();
```
illuhad, line 89: An alternative might be to allocate local memory statically here (I assume metal can do this too?).

As I understand it, your implementation currently must pessimize and always assume that there might be a group algorithm called in the kernel.

If you allocate local memory statically, the compiler will likely optimize it away when the group algorithm is not invoked.

illuhad (approved):

src/libkernel/sscp/metal/broadcast.cpp
resetius: Agree, I'll try it's look like it's possible to allocate 1 cell of array statically here

resetius: ACPP_SHMEM_ATTRIBUTE generates a MSL which does not compile
Will try to do smth in a follow-up PR

illuhad: That attribute was primarily intended for CUDA/HIP/SPIR-V. We can diverge from it for Metal if necessary.
(But if the static address space 3 pattern doesn't give us local memory on Metal, we may have to do some additional compiler massaging for PCUDA to get `__shared__` syntax to work there)

PR #1997, opened 2026-03-09
[Metal] Add math functions

resetius:
This PR adds implementations of several missing math functions in the Metal backend
erf/erfc and lgamma/tgamma are implemented in a simple way using formulas from Wikipedia

erf/erfc work quite well, tgamma/lgamma are less accurate and may lose precision for some inputs 
The goal of this PR is not to provide a perfect implementation.

A fully robust implementation of tgamma/lgamma in math libraries is typically around ~300 lines of code, so this could be improved later if needed

### DIFF
--- a/include/hipSYCL/sycl/libkernel/builtin_interface.hpp
+++ b/include/hipSYCL/sycl/libkernel/builtin_interface.hpp
@@ -182,7 +182,7 @@ HIPSYCL_BUILTIN T __acpp_fmod(T x, T y) noexcept {
 }
 
 template<class T>
-T __acpp_fract(T x, T* ptr) noexcept {
+HIPSYCL_BUILTIN T __acpp_fract(T x, T* ptr) noexcept {
   HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_fract, x, ptr);
 }
 
--- a/include/hipSYCL/sycl/libkernel/generic/hiplike/builtins.hpp
+++ b/include/hipSYCL/sycl/libkernel/generic/hiplike/builtins.hpp
@@ -127,7 +127,8 @@ HIPSYCL_HIPLIKE_BUILTIN double __acpp_frexp(double x, int* y) noexcept {
 }
 
 HIPSYCL_DEFINE_HIPLIKE_MATH_BUILTIN2(__acpp_hypot, hypotf, hypot)
-HIPSYCL_DEFINE_HIPLIKE_MATH_BUILTIN(__acpp_ilogb, ilogbf, ilogb)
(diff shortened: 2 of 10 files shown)


tests/compiler/sscp/math_functions.cpp
```diff
@@ -0,0 +1,257 @@
+// RUN: %acpp %s -o %t --acpp-targets=generic
+// RUN: %t | FileCheck %s
+// RUN: %acpp %s -o %t --acpp-targets=generic -O3
+// RUN: %t | FileCheck %s
+
```
illuhad: Is this really a compiler check? We already have math functions tests in `math.cpp` in `sycl_tests`. Would it fit there?

resetius: Thank you. I'll move it to the sycl tests

resetius:
```
nvlink error   : Undefined reference to '_ZN7hipsycl4sycl6detail16hiplike_builtins12__acpp_fractIfEET_S4_PS4_' in 'CMakeFiles/sycl_tests.dir/sycl/math.cpp.o'
nvlink error   : Undefined reference to '_ZN7hipsycl4sycl6detail16hiplike_builtins12__acpp_fractIdEET_S4_PS4_' in 'CMakeFiles/sycl_tests.dir/sycl/math.cpp.o'
```

resetius:
Smth wrong with checks

illuhad:
Looks okay in principle, waiting for fixes to test cases. It seems something is not right there.

resetius:
> Looks okay in principle, waiting for fixes to test cases. It seems something is not right there.

cospi(0.5) = 0, -4.37e-8 is zero and 6.12e-17 is also zero
relative tol does not work well in this case

fixed

resetius:
`directory)
unknown location(0): fatal error: in "math_tests/math_unary_extra<float>": memory access violation at address: 0x77c51f8329cc: invalid permissions
/tmp/unihd-sycl-ci-fatboy-0/actions- `

illuhad:
Can we enable the tests for the new functions one by one so that we can figure out which causes problems? If we can pinpoint it to a specific function that causes problems it might be a problem in Intel OpenCL, or an incorrect SPIR-V mapping.

resetius:
it's likely ilogb ...

resetius:
YES
unknown location(0): fatal error: in "math_tests/math_ilogb<float>": memory access violation at address: 0x749db78329cc: invalid permissions

resetius:
wrong return type maybe

illuhad (approved):
Awesome, thank you for the metal functions and for drilling down and fixing the bug!

PR #1998, opened 2026-03-10
[Metal] Fix MSL with struct of arrays

resetius:
Fix topological sort of struct types: unwrap array element types when collecting struct dependencies


illuhad (approved):
Thanks!

PR #2001, opened 2026-03-12
implement missing relational SYCL operators (isequal, etc)

### LINKED ISSUE #1999: sycl::isequal not found for sycl-vecs
**Bug summary**
Hi, I've been using oneDPL with oneAPI compiler without problems.
Now, in one of my projects using acpp as compiler, I wanted to include a closed set of header-only files of oneDPL.
However, I am getting this weird error:

```
/opt/oneDPL/include/oneapi/dpl/internal/random_impl/normal_distribution.h:358:35: error: no member named 'isequal' in namespace 'sycl'; did you mean 'iseqsig'?
[build]   358 |                                   sycl::isequal(__u1, sycl::vec<scalar_type, __vec_size>{scalar_type{0.0}}));
```

**To Reproduce**

`cd /opt && git clone --depth=1 https://github.com/uxlfoundation/oneDPL`

and then in your CMakeLists.txt do:

target_include_directories(oneDPLrand INTERFACE /opt/oneDPL/include/oneapi/dpl/internal/random_impl/)

and in my code I have

```
#include "random_common.h"
#include "uniform_real_distribution.h"
#include "exponential_distribution.h"
#include "linear_congruential_engine.h"
#include "normal_distribution.h"
#include "philox_engine.h"
#include "weibull_distribution.h"
```

**Expected behavior**
No compilation error is printed. Or is isequal not supposed to work with sycl-vecs ?

My current workaround is to define before including these headers:

```
namespace sycl
{
  template <typename T, int N>
  inline bool isequal(const sycl::vec<T, N> &a, const sycl::vec<T, N> &b)
  {
    return a == b;
  }
} // namespace sycl
```

**Describe your setup**
Alma10, acpp cloned recently from default branch and built/installed in /opt

ferdymercury:
Fixes https://github.com/AdaptiveCpp/AdaptiveCpp/issues/1999

### DIFF
--- a/include/hipSYCL/sycl/libkernel/builtin_interface.hpp
+++ b/include/hipSYCL/sycl/libkernel/builtin_interface.hpp
@@ -659,62 +659,62 @@ HIPSYCL_BUILTIN int __acpp_signbit(T x) noexcept {
   HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_signbit, x);
 }
 
-// Relational binary functions below default always to host
+// Relational binary functions below are directly implemented here
 // Potential improvement: use HIPSYCL_RETURN_DISPATCH_BUILTIN and specialize for SPIRV as done in
 // https://github.com/intel/llvm/blob/sycl/sycl/include/sycl/detail/builtins/relational_functions.inc#L130
 
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_isequal(T x, T y) noexcept
 {
-    return host_builtins::__acpp_isequal(x, y);
+    return x == y;
 }
 
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_isnotequal(T x, T y) noexcept
 {
-    return host_builtins::__acpp_isnotequal(x, y);
+    return x != y;
 }
 
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_isgreater(T x, T y) noexcept
 {
-    return host_builtins::__acpp_isgreater(x, y);
+    return x > y;
 }
 
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_isgreaterequal(T x, T y) noexcept
 {
-    return host_builtins::__acpp_isgreaterequal(x, y);
+    return x >= y;
 }
 
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_isless(T x, T y) noexcept
 {
-    return host_builtins::__acpp_isless(x, y);
+    return x < y;
 }
 
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_islessequal(T x, T y) noexcept
 {
-    return host_builtins::__acpp_islessequal(x, y);
+    return x <= y;
 }
 
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_islessgreater(T x, T y) noexcept
 {
-    return host_builtins::__acpp_islessgreater(x, y);
+    return (x < y) || (x > y);
 }
 
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_isordered(T x, T y) noexcept
 {
-    return host_builtins::__acpp_isordered(x, y);
+    return __acpp_isequal(x, x) && __acpp_isequal(y, y);
 }
 
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_isunordered(T x, T y) noexcept
 {
-    return host_builtins::__acpp_isunordered(x, y);
+    return __acpp_isnan(x) || __acpp_isnan(y);
 }
 
 }
--- a/include/hipSYCL/sycl/libkernel/builtin_interface.hpp
+++ b/include/hipSYCL/sycl/libkernel/builtin_interface.hpp
@@ -708,7 +708,7 @@ HIPSYCL_BUILTIN bool __acpp_islessgreater(T x, T y) noexcept
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_isordered(T x, T y) noexcept
 {
-  return __acpp_isequal(x, x) && __acpp_isequal(y, y);
+  return !__acpp_isnan(x) && !__acpp_isnan(y);
 }
 
 template<class T>
--- a/include/hipSYCL/sycl/libkernel/builtin_interface.hpp
+++ b/include/hipSYCL/sycl/libkernel/builtin_interface.hpp
@@ -659,6 +659,60 @@ HIPSYCL_BUILTIN int __acpp_signbit(T x) noexcept {
   HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_signbit, x);
 }
 
+template<class T>
+HIPSYCL_BUILTIN bool __acpp_isequal(T x, T y) noexcept
+{
+    HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_isequal, x, y);
+}
+
+template<class T>
+HIPSYCL_BUILTIN bool __acpp_isnotequal(T x, T y) noexcept
+{
+    HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_isnotequal, x, y);
+}
+
+template<class T>
+HIPSYCL_BUILTIN bool __acpp_isgreater(T x, T y) noexcept
+{
+    HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_isgreater, x, y);
+}
+
+template<class T>
+HIPSYCL_BUILTIN bool __acpp_isgreaterequal(T x, T y) noexcept
+{
+    HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_isgreaterequal, x, y);
+}
+
+template<class T>
+HIPSYCL_BUILTIN bool __acpp_isless(T x, T y) noexcept
+{
+    HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_isless, x, y);
+}
+
+template<class T>
+HIPSYCL_BUILTIN bool __acpp_islessequal(T x, T y) noexcept
+{
+    HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_islessequal, x, y);
+}
+
+template<class T>
+HIPSYCL_BUILTIN bool __acpp_islessgreater(T x, T y) noexcept
+{
+    HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_islessgreater, x, y);
+}
+
+template<class T>
+HIPSYCL_BUILTIN bool __acpp_isordered(T x, T y) noexcept
+{
+    HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_isordered, x, y);
+}
+
+template<class T>
+HIPSYCL_BUILTIN bool __acpp_isunordered(T x, T y) noexcept
+{
+    HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_isunordered, x, y);
+}
+
 }
 }
 }
--- a/include/hipSYCL/sycl/libkernel/builtin_interface.hpp
+++ b/include/hipSYCL/sycl/libkernel/builtin_interface.hpp
@@ -666,55 +666,55 @@ HIPSYCL_BUILTIN int __acpp_signbit(T x) noexcept {
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_isequal(T x, T y) noexcept
 {
-    return x == y;
+  return x == y;
 }
 
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_isnotequal(T x, T y) noexcept
 {
-    return x != y;
+  return x != y;
 }
 
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_isgreater(T x, T y) noexcept
 {
-    return x > y;
+  return x > y;
 }
 
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_isgreaterequal(T x, T y) noexcept
 {
-    return x >= y;
+  return x >= y;
 }
 
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_isless(T x, T y) noexcept
 {
-    return x < y;
+  return x < y;
 }
 
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_islessequal(T x, T y) noexcept
 {
-    return x <= y;
+  return x <= y;
 }
 
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_islessgreater(T x, T y) noexcept
 {
-    return (x < y) || (x > y);
+  return (x < y) || (x > y);
 }
 
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_isordered(T x, T y) noexcept
 {
-    return __acpp_isequal(x, x) && __acpp_isequal(y, y);
+  return __acpp_isequal(x, x) && __acpp_isequal(y, y);
 }
 
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_isunordered(T x, T y) noexcept
 {
-    return __acpp_isnan(x) || __acpp_isnan(y);
+  return __acpp_isnan(x) || __acpp_isnan(y);
 }
 
 }
--- a/include/hipSYCL/sycl/libkernel/builtin_interface.hpp
+++ b/include/hipSYCL/sycl/libkernel/builtin_interface.hpp
@@ -659,58 +659,62 @@ HIPSYCL_BUILTIN int __acpp_signbit(T x) noexcept {
   HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_signbit, x);
 }
 
+// Relational binary functions below default always to host
+// Potential improvement: use HIPSYCL_RETURN_DISPATCH_BUILTIN and specialize for SPIRV as done in
+// https://github.com/intel/llvm/blob/sycl/sycl/include/sycl/detail/builtins/relational_functions.inc#L130
+
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_isequal(T x, T y) noexcept
 {
-    HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_isequal, x, y);
+    return host_builtins::__acpp_isequal(x, y);
 }
 
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_isnotequal(T x, T y) noexcept
 {
-    HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_isnotequal, x, y);
+    return host_builtins::__acpp_isnotequal(x, y);
 }
 
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_isgreater(T x, T y) noexcept
 {
-    HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_isgreater, x, y);
+    return host_builtins::__acpp_isgreater(x, y);
 }
 
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_isgreaterequal(T x, T y) noexcept
 {
-    HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_isgreaterequal, x, y);
+    return host_builtins::__acpp_isgreaterequal(x, y);
 }
 
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_isless(T x, T y) noexcept
 {
-    HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_isless, x, y);
+    return host_builtins::__acpp_isless(x, y);
 }
 
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_islessequal(T x, T y) noexcept
 {
-    HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_islessequal, x, y);
+    return host_builtins::__acpp_islessequal(x, y);
 }
 
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_islessgreater(T x, T y) noexcept
 {
-    HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_islessgreater, x, y);
+    return host_builtins::__acpp_islessgreater(x, y);
 }
 
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_isordered(T x, T y) noexcept
 {
-    HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_isordered, x, y);
+    return host_builtins::__acpp_isordered(x, y);
 }
 
 template<class T>
 HIPSYCL_BUILTIN bool __acpp_isunordered(T x, T y) noexcept
 {
-    HIPSYCL_RETURN_DISPATCH_BUILTIN(__acpp_isunordered, x, y);
+    return host_builtins::__acpp_isunordered(x, y);
 }
 
 }
--- a/include/hipSYCL/sycl/libkernel/builtins/relational_builtins.hpp
+++ b/include/hipSYCL/sycl/libkernel/builtins/relational_builtins.hpp
@@ -72,15 +72,14 @@ HIPSYCL_RELATIONAL_BUILTIN(UNARY_T_RET_BOOL, isnormal)
 HIPSYCL_RELATIONAL_BUILTIN(UNARY_T_RET_BOOL, signbit)
 
   template <typename T,
-            typename = std::enable_if_t<std::is_scalar<T>::value && std::is_floating_point_v<T>>>
+            typename = std::enable_if_t<std::is_scalar_v<T> && std::is_floating_point_v<T>>>
   bool isequal(T x, T y)
   {
     return x == y;
   }
 
-  template <typename T,
-            typename = std::enable_if_t<std::is_scalar<T>::value && std::is_floating_point_v<T>>,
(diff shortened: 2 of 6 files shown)


resetius:
I don’t think c++20 is allowed. Please use type traits and sfinae via enable_if

ferdymercury:
I fixed it to use type_traits.

I thought auto return type deducation via if-constexpr looked a bit nicer than sfinae

include/hipSYCL/sycl/libkernel/builtins/relational_builtins.hpp
```diff
@@ -69,5 +70,48 @@ HIPSYCL_RELATIONAL_BUILTIN(UNARY_T_RET_BOOL, isinf)
 HIPSYCL_RELATIONAL_BUILTIN(UNARY_T_RET_BOOL, isfinite)
 HIPSYCL_RELATIONAL_BUILTIN(UNARY_T_RET_BOOL, isnormal)
 HIPSYCL_RELATIONAL_BUILTIN(UNARY_T_RET_BOOL, signbit)
+
+  template <typename T,
+            typename = std::enable_if_t<std::is_scalar<T>::value && std::is_floating_point_v<T>>>
```
illuhad: Why not `is_scalar_v` ?

include/hipSYCL/sycl/libkernel/builtins/relational_builtins.hpp
```diff
@@ -69,5 +70,48 @@ HIPSYCL_RELATIONAL_BUILTIN(UNARY_T_RET_BOOL, isinf)
...
+  bool isequal(T x, T y)
+  {
+    return x == y;
+  }
+
+  template <typename T,
+            typename = std::enable_if_t<std::is_scalar<T>::value && std::is_floating_point_v<T>>,
```
illuhad: Can we move `enable_if`s to the end, so that we don't have dummy parameters interspersed with real ones?

include/hipSYCL/sycl/libkernel/builtins/relational_builtins.hpp
```diff
@@ -69,5 +70,48 @@ HIPSYCL_RELATIONAL_BUILTIN(UNARY_T_RET_BOOL, isinf)
...
+            long unsigned int N>
+  sycl::marray<bool, N> isequal(sycl::marray<T, N> x, sycl::marray<T, N> y)
+  {
+    return x == y;
+  }
+
+  namespace detail {
+  template <
+      typename R, typename = std::enable_if_t<std::is_scalar<R>::value && std::is_integral_v<R>>,
```
illuhad: same here

illuhad (commented):
Waiting for tests.

include/hipSYCL/sycl/libkernel/builtins/relational_builtins.hpp
```diff
@@ -69,5 +70,46 @@ HIPSYCL_RELATIONAL_BUILTIN(UNARY_T_RET_BOOL, isinf)
 HIPSYCL_RELATIONAL_BUILTIN(UNARY_T_RET_BOOL, isfinite)
 HIPSYCL_RELATIONAL_BUILTIN(UNARY_T_RET_BOOL, isnormal)
 HIPSYCL_RELATIONAL_BUILTIN(UNARY_T_RET_BOOL, signbit)
+
+  template <typename T,
+            typename = std::enable_if_t<std::is_scalar_v<T> && std::is_floating_point_v<T>>>
```
ferdymercury: I opted to be "independent" when checking for floating point types, since I am testing locally and I have no experience with the acpp project, but should I instead use:

is_valid_relational_scalar or is_valid_relational_nonscalar

that are defined above in detail namespace on the top of this header?

and if yes, how can I use it? Just `typename T, int N, typename = is_valid_relational_scalar<T>` ?

ferdymercury:
> Waiting for tests.

Could you give me an example here? You meant that I should include tests in https://github.com/intel/llvm/blob/sycl/sycl/test/basic_tests/builtins/relational_builtins.cpp but I am not sure how.

I was testing locally with:

```
sycl::vec<double, 3> x {1, 3, 6};
sycl::vec<double, 3> y {1, 2, 6};
sycl::vec<int64_t, 3> b{0, -1, 0};
assert(sycl::isequal(x,y), b);
```

Should I just copy paste that snippet into that file, or shall I somehow reuse those magic "CHECK" preprocessor that I am not familiar with?

Thanks.

Also on the TO-DO list:
- deal with __swizzled_vector__ (no idea yet how, if additional template overload with all permutations, or if I can "join" them together smartly)
- more operators (isgreater, etc)

resetius:
> Should I just copy paste that snippet into that file, or shall I somehow reuse those magic "CHECK" preprocessor that I am not familiar with?

I think you should implement them from scratch in tests/sycl/relational.cpp
intels tests only check return types

ferdymercury:
Sorry, I looked into the wrong link.

My question should have been:

should I use the _inputs_ and _outputs_ variable as done in https://github.com/AdaptiveCpp/AdaptiveCpp/blob/develop/tests/sycl/relational.cpp for isfinite, etc

or can I just write the test independently as I suggested above and only put at the end:
`BOOST_TEST(sycl::isequal(x,y) == b);`

resetius:
> Sorry, I looked into the wrong link.
> 
> My question should have been:
> 
> should I use the _inputs_ and _outputs_ variable as done in https://github.com/AdaptiveCpp/AdaptiveCpp/blob/develop/tests/sycl/relational.cpp for isfinite, etc
> 
> or can I just write the test independently as I suggested above and only put at the end: `BOOST_TEST(sycl::isequal(x,y) == b);`

I would copy BOOST_AUTO_TEST_CASE_TEMPLATE(rel_genfloat_unary, T, rel_test_genfloats), rename it to rel_genfloat_binary, create two input vectors inside, and run isequals and similar functions on them, checking the results with `BOOST_TEST(comp(outputs[i++], c) == isequal(comp(inputs1[0], c), comp(inputs2[0], c)));`. The input vectors can be defined as s::vec<DT, 16> v{...} - we just need to pick good numbers for the test

tests/sycl/relational.cpp
```diff
@@ -172,4 +172,84 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(rel_genfloat_unary, T,
...
+
+  constexpr int D = vector_length_v<T>;
+  using DT = vector_elem_t<T>;
+
+  namespace s = sycl;
+
+  using OutType = s::detail::builtin_input_boollike_t<T>;
+  using BoolType = s::detail::builtin_input_element_t<OutType>;
+
+  constexpr int FUN_COUNT = 5;
```
resetius:
```suggestion
  constexpr int FUN_COUNT = 7;
```

tests/sycl/relational.cpp
```diff
@@ -172,4 +172,84 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(rel_genfloat_unary, T,
...
+    auto outputs = out.get_host_access();
+    s::vec<DT, 16> v{NAN, INFINITY, INFINITY - INFINITY,
+		     0.0, 0.0/0.0, 1.0/0.0, sqrt(-1),
+		     std::numeric_limits<float>::min(),
+		     std::numeric_limits<float>::denorm_min(),
+		     std::numeric_limits<double>::min(),
+		     std::numeric_limits<double>::denorm_min(),
+		     -1.0, 17.0, -4.0, -2.0, 3.0};
+    inputs1[0] = get_subvector<DT, D>(v);
+	inputs2[0] = get_subvector<DT, D>(v);
```
resetius:
```suggestion
    s::vec<DT, 16> v1{
      DT(1.0), DT(1.0), DT(2.0), std::numeric_limits<DT>::quiet_NaN(),
      DT(1.0), DT(0.0), std::numeric_limits<DT>::infinity(),
      -std::numeric_limits<DT>::infinity(), std::numeric_limits<DT>::denorm_min(),
      DT(-1.0), DT(17.0), DT(-4.0), DT(-2.0), DT(3.0), DT(5.0), DT(-0.0)
    };
    s::vec<DT, 16> v2{
      DT(1.0), DT(2.0), DT(1.0), DT(1.0),
      std::numeric_limits<DT>::quiet_NaN(), DT(-0.0), std::numeric_limits<DT>::infinity(),
      std::numeric_limits<DT>::infinity(), DT(0.0),
      DT(-4.0), DT(17.0), DT(-1.0), DT(3.0), DT(-2.0), DT(5.0), DT(0.0)
    };
    inputs1[0] = get_subvector<DT, D>(v1);
	inputs2[0] = get_subvector<DT, D>(v2);
```

tests/sycl/relational.cpp
```diff
@@ -172,4 +172,84 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(rel_genfloat_unary, T,
...
+    }
+  }
+
+  // run functions
+
+  queue.submit([&](s::handler &cgh) {
+    auto inputs1  = in1.template get_access<s::access::mode::read>(cgh);
+	auto inputs2  = in2.template get_access<s::access::mode::read>(cgh);
+    auto outputs = out.template get_access<s::access::mode::write>(cgh);
+    cgh.single_task<kernel_name<class rel_unary, D, DT>>([=]() {
```
resetius:
```suggestion
    cgh.single_task<kernel_name<class rel_binary, D, DT>>([=]() {
```

tests/sycl/relational.cpp
```diff
@@ -172,4 +172,84 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(rel_genfloat_unary, T,
...
+	  outputs[i++] = s::isless(inputs1[0], inputs2[0]);
+      outputs[i++] = s::islessequal(inputs1[0], inputs2[0]);
+	  outputs[i++] = s::islessgreater(inputs1[0], inputs2[0]);
+    });
+  });
+
+  // check results
+
+  {
+    auto inputs  = in.get_host_access();
```
resetius:
```suggestion
    auto inputs1  = in1.get_host_access();
    auto inputs2  = in2.get_host_access();    
```

tests/sycl/relational.cpp
```diff
@@ -172,4 +172,84 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(rel_genfloat_unary, T,
...
+    auto outputs = out.get_host_access();
+
+    for(int c = 0; c < std::max(D,1); ++c) {
+      int i = 0;
+      BOOST_TEST(comp(outputs[i++], c) == std::isequal(comp(inputs1[0], c), comp(inputs2[0], c)));
```
resetius: have you run it? i think you need to define functions like this
```
  auto host_isequal = [](DT a, DT b) {
    return !std::isnan(a) && !std::isnan(b) && (a == b);
  };
  auto host_isnotequal = [](DT a, DT b) {
    return std::isnan(a) || std::isnan(b) || (a != b);
  };
  auto host_isgreater = [](DT a, DT b) {
    return !std::isnan(a) && !std::isnan(b) && (a > b);
  };
  auto host_isgreaterequal = [](DT a, DT b) {
    return !std::isnan(a) && !std::isnan(b) && (a >= b);
  };
  auto host_isless = [](DT a, DT b) {
    return !std::isnan(a) && !std::isnan(b) && (a < b);
  };
  auto host_islessequal = [](DT a, DT b) {
    return !std::isnan(a) && !std::isnan(b) && (a <= b);
  };
  auto host_islessgreater = [](DT a, DT b) {
    return !std::isnan(a) && !std::isnan(b) && ((a < b) || (a > b));
  };
```
and then use them instead of `std::` variants

ferdymercury: > have you run it?

no, not yet, sorry, i marked the PR as draft mode since I did not run it, will do next week and then mark as ready to review.

Thanks a lot though for all the comments and corrections in the meanwhile

include/hipSYCL/sycl/libkernel/host/builtins.hpp
```diff
@@ -822,6 +822,60 @@ HIPSYCL_BUILTIN int __acpp_signbit(T x) noexcept {
   return std::signbit(x);
 }
 
+template<class T>
+HIPSYCL_BUILTIN bool __acpp_isequal(T x, T y) noexcept
+{
+    return x == y;
```
ferdymercury: I guess we assume that the builtin equality operator handles correctly the NaNs, so we do not need to implement here the more verbose variant

`!std::isnan(x) && !std::isnan(y) && (x == y)`

(same stuff below).

include/hipSYCL/sycl/libkernel/host/builtins.hpp
```diff
@@ -822,6 +822,60 @@ HIPSYCL_BUILTIN int __acpp_signbit(T x) noexcept {
   return std::signbit(x);
 }
 
+template<class T>
```
illuhad: Why do you need to modify this file? This is just for host backend-specific builtin mapping. I don't think you need this for the relationals. I think you could just implement everything in `relational_builtins.hpp` since nothing is backend-dependent.

include/hipSYCL/sycl/libkernel/builtins/relational_builtins.hpp
```diff
@@ -50,6 +55,32 @@ namespace detail {
...
+      for (std::size_t i = 0; i < detail::builtin_input_num_elems_v<NonScalar1>; i++) {                       \
+        ret[i] = builtin_func(x[i], y[i]) ? -1 : 0;                                                           \
+      }                                                                                                       \
+    }                                                                                                         \
+    return ret;                                                                                               \
+  }
+
 #define HIPSYCL_RELATIONAL_BUILTIN_OVERLOADED_IMPL(h, template_h, builtin_name, builtin_func) \
   h(float, builtin_name, builtin_func)                                                        \
   h(double, builtin_name, builtin_func)                                                       \
```
illuhad, line 86: Also need half

illuhad (commented):
I'm not sure if we gain anything here from the macro layer. Wouldn't it be easier if we just implemented them as defined by the spec? (explicit definitions for float/double/half and template for non-scalars).

include/hipSYCL/sycl/libkernel/builtins/relational_builtins.hpp
ferdymercury: half is in the line below ?

include/hipSYCL/sycl/libkernel/host/builtins.hpp
ferdymercury: I did this for two reasons:
- consistency with the other relational operators that follow this pattern
- extendability: if in a next step one wants to add the SPIRV backend as done by Intel, one does not need to move things around

ferdymercury:
> I'm not sure if we gain anything here from the macro layer. Wouldn't it be easier if we just implemented them as defined by the spec? (explicit definitions for float/double/half and template for non-scalars).

I copy-pasted the macro layer that did (almost) exactly this job for copysign in math_builtins.hpp, doing the explicit definition is a lot of work and is very prone to errors. There are 9 operators, then all the three variants, so 27 functions, the enable_ifs, etc. So the macros do that automatically and they solve all the issues of checking that the input types are correct, and automatically deduces the return type. I just copied the binary_t_t template from math_builtins.hpp macro, and changed the return type. This way I also reduced the likelihood of introducing mistakes

include/hipSYCL/sycl/libkernel/builtins/relational_builtins.hpp
illuhad: sorry, missed this!

include/hipSYCL/sycl/libkernel/host/builtins.hpp
illuhad: > consistency with the other relational operators that follow this pattern

You only implemented for `host`... if you wanted to go this way, you'd also need a mapping for `hiplike` and an SSCP builtin bitcode library interface, and implementations of those builtins for every backend. Are you sure you want to go down that route? :)

>extendability: if in a next step one wants to add the SPIRV backend as done by Intel, one does not need to move things around

With 99% probability we don't need this.

illuhad:
As I've said, and as shown in CI, your approach does not work as is because you've only done the host mapping. Trust me when I say that you are likely easier off just avoiding the builtin interface :-)

include/hipSYCL/sycl/libkernel/host/builtins.hpp
ferdymercury: Sorry, I did not notice that mistake of mine, since I only tested on a local Linux build.

I think I fixed it now, do you think it's correct? I put for symmetry the implementation in the builtin_interface rather than on relational_builtins, but if you prefer, I can move that there, too, and not modify this file.

include/hipSYCL/sycl/libkernel/builtin_interface.hpp
```diff
@@ -654,6 +654,64 @@ HIPSYCL_BUILTIN int __acpp_signbit(T x) noexcept {
...
 }
 
+// Relational binary functions below are directly implemented here
+// Potential improvement: use HIPSYCL_RETURN_DISPATCH_BUILTIN and specialize for SPIRV as done in
+// https://github.com/intel/llvm/blob/sycl/sycl/include/sycl/detail/builtins/relational_functions.inc#L130
+
+template<class T>
+HIPSYCL_BUILTIN bool __acpp_isequal(T x, T y) noexcept
+{
+    return x == y;
```
ferdymercury: It seems that we have a failure on [macOS Metal backend / Metal backend [macOS, LLVM 20] (pull_request)](https://github.com/AdaptiveCpp/AdaptiveCpp/actions/runs/23348025811/job/67929820584?pr=2001)

My suspicion is that it does not treat the NaNs the same way as all the remainng platforms.

Does it mean that we should change here the equality operator to something more convoluted such as 
!std::isnan(x) && !std::isnan(x) && x == y ?

(The same for all functions below). If yes, with an ifdef only for this platform, or for all?

Or does it mean that there is something missing somewhere else on macos metal backend?

resetius: > Or does it mean that there is something missing somewhere else on macos metal backend?

let me check

tests/sycl/relational.cpp
```diff
@@ -172,4 +172,127 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(rel_genfloat_unary, T,
...
+  s::buffer<T> in2{{1}};
+  s::buffer<OutType> out{{FUN_COUNT}};
+  {
+    auto inputs1  = in1.get_host_access();
+	auto inputs2  = in2.get_host_access();
+    auto outputs = out.get_host_access();
+    s::vec<DT, 16> v1{
+      DT(1.0), DT(1.0), DT(2.0), std::numeric_limits<DT>::quiet_NaN(),
+      DT(1.0), DT(0.0), std::numeric_limits<DT>::infinity(),
+      -std::numeric_limits<DT>::infinity(), std::numeric_limits<DT>::denorm_min(),
```
resetius:
```suggestion
      -std::numeric_limits<DT>::infinity(), DT(0.0) /*std::numeric_limits<DT>::denorm_min()*/ /* metal: denorm == 0 => true, temporary disable input */
```

include/hipSYCL/sycl/libkernel/builtin_interface.hpp
resetius: apply my suggestion and also modify Emitter.cpp:
```
diff --git a/src/compiler/llvm-to-backend/metal/Emitter.cpp b/src/compiler/llvm-to-backend/metal/Emitter.cpp
index 1e04ec5f..89cdb07a 100644
--- a/src/compiler/llvm-to-backend/metal/Emitter.cpp
+++ b/src/compiler/llvm-to-backend/metal/Emitter.cpp
@@ -1127,7 +1127,7 @@ void MetalEmitter::emitFCmpInstruction(const FCmpInst* FC, const std::string& na
       os << indent(level) << name << " = (" << lhs << " == " << rhs << ");\n";
       break;
     case FCmpInst::FCMP_ONE:
-      os << indent(level) << name << " = (" << lhs << " != " << rhs << ");\n";
+      os << indent(level) << name << " = (isordered(" << lhs << ", " << rhs << ") && (" << lhs << " != " << rhs << "));\n";
       break;
     case FCmpInst::FCMP_OGT:
       os << indent(level) << name << " = (" << lhs << " > " << rhs << ");\n";
```

src/compiler/llvm-to-backend/metal/Emitter.cpp
```diff
@@ -1124,7 +1124,7 @@ void MetalEmitter::emitFCmpInstruction(const FCmpInst* FC, const std::string& na
 
   switch (FC->getPredicate()) {
     case FCmpInst::FCMP_OEQ:
-      os << indent(level) << name << " = (" << lhs << " == " << rhs << ");\n";
+      os << indent(level) << name << " = (isordered(" << lhs << ", " << rhs << ") && (" << lhs << " != " << rhs << "));\n";
```
resetius: wrong case, you need to modify `case FCmpInst::FCMP_ONE:`

include/hipSYCL/sycl/libkernel/builtin_interface.hpp
resetius: oh, sorry, I did not close bracket somewhere:
```
/Users/runner/work/AdaptiveCpp/AdaptiveCpp/tests/sycl/relational.cpp:208:7: error: expected '}'
  208 |       DT(-1.0), DT(17.0), DT(-4.0), DT(-2.0), DT(3.0), DT(5.0), DT(-0.0)
      |       ^
/Users/runner/work/AdaptiveCpp/AdaptiveCpp/tests/sycl/relational.cpp:204:22: note: to match this '{'
  204 |     s::vec<DT, 16> v1{
      |                      ^
```

src/compiler/llvm-to-backend/metal/Emitter.cpp
ferdymercury: Oops sorry fixed

tests/sycl/relational.cpp
```diff
@@ -172,4 +172,127 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(rel_genfloat_unary, T,
...
+  s::buffer<T> in2{{1}};
+  s::buffer<OutType> out{{FUN_COUNT}};
+  {
+    auto inputs1  = in1.get_host_access();
+	auto inputs2  = in2.get_host_access();
+    auto outputs = out.get_host_access();
+    s::vec<DT, 16> v1{
+      DT(1.0), DT(1.0), DT(2.0), std::numeric_limits<DT>::quiet_NaN(),
+      DT(1.0), DT(0.0), std::numeric_limits<DT>::infinity(),
+      -std::numeric_limits<DT>::infinity(), DT(0.0) /*std::numeric_limits<DT>::denorm_min()*/ /* metal: denorm == 0 => true, temporary disable input */
```
ferdymercury:
```suggestion
      -std::numeric_limits<DT>::infinity(), DT(0.0), /*std::numeric_limits<DT>::denorm_min()*/ /* metal: denorm == 0 => true, temporary disable input */
```

resetius:
It works

include/hipSYCL/sycl/libkernel/builtins/relational_builtins.hpp
```diff
@@ -33,6 +33,11 @@ namespace detail {
     return builtin_func(x);                                                                  \
   }
 
+#define HIPSYCL_RELATIONAL_BUILTIN_GENERATOR_BINARY_T_T_RET_BOOL(T, builtin_name, builtin_func) \
+HIPSYCL_BUILTIN bool builtin_name(T x, T y) noexcept {                                          \
+        return builtin_func(x, y);                                                              \
```
resetius: 2 spaces

include/hipSYCL/sycl/libkernel/builtin_interface.hpp
```diff
@@ -654,6 +654,64 @@ HIPSYCL_BUILTIN int __acpp_signbit(T x) noexcept {
...
+}
+
+template<class T>
+HIPSYCL_BUILTIN bool __acpp_isnotequal(T x, T y) noexcept
+{
+    return x != y;
```
resetius: 2 spaces

ferdymercury:
@illuhad I see some failures in the CI but they seem a connection issue? Do you need further changes on my side?
(Thanks a lot for the review.)
fyi @tdavidcl

ferdymercury:
All look green now, thanks for the rerun!
Shall I squash commits or you'd do that with the "Squash and merge" button?

ferdymercury:
@illuhad could you trigger a CI rerun? thanks!

include/hipSYCL/sycl/libkernel/builtin_interface.hpp
```diff
@@ -659,6 +659,64 @@ HIPSYCL_BUILTIN int __acpp_signbit(T x) noexcept {
...
+template<class T>
+HIPSYCL_BUILTIN bool __acpp_islessgreater(T x, T y) noexcept
+{
+  return (x < y) || (x > y);
+}
+
+template<class T>
+HIPSYCL_BUILTIN bool __acpp_isordered(T x, T y) noexcept
+{
+  return __acpp_isequal(x, x) && __acpp_isequal(y, y);
```
resetius:
```suggestion
  return !__acpp_isnan(x) && !__acpp_isnan(y);
```

resetius:
Minor nit before merging: __acpp_isordered uses the x == x nan trick which may not hold under -ffast-math / -ffinite-math-only on gpu backends 
Suggest mirroring __acpp_isunordered which already uses __acpp_isnan
This is my only concern. The metal FCMP_ONE fix alone makes this worth merging

resetius (approved):
LGTM

PR #2002, opened 2026-03-12
[Metal] Support atomic float fetch_add/sub for local_memory

resetius:


illuhad:
Needs investigation into why CI for the new test fails for host. The kernel looks correct to me; perhaps try changing the accessor's access mode to the more common `read_write`? Wouldn't be the first buffer-accessor bug...

resetius:
> Needs investigation into why CI for the new test fails for host. The kernel looks correct to me; perhaps try changing the accessor's access mode to the more common `read_write`? Wouldn't be the first buffer-accessor bug...

the atomic fetch add does not work for cuda and local memory in CI
I could not reproduce, locally it works, which is probably because I have cuda 12.4 installed

illuhad:
> the atomic fetch add does not work for cuda and local memory in CI
I could not reproduce, locally it works, which is probably because I have cuda 12.4 installed

Hang on -- the CUDA tests in CI pass. The failing CI is SSCP for CPU/host, not CUDA.

I think you might have been confused because the test name also has "CUDA" in its name. We also do some basic checks to see if things build with CUDA. But we cannot actually run any CUDA tests due to lack of GPUs in github CI. There, we only run the tests on CPU.

Actual CUDA tests are carried out in the self-hosted CI flows where we have our own GPUs.

resetius:
> > the atomic fetch add does not work for cuda and local memory in CI
> > I could not reproduce, locally it works, which is probably because I have cuda 12.4 installed
> 
> Hang on -- the CUDA tests in CI pass. The failing CI is SSCP for CPU/host, not CUDA.
> 
> I think you might have been confused because the test name also has "CUDA" in its name. We also do some basic checks to see if things build with CUDA. But we cannot actually run any CUDA tests due to lack of GPUs in github CI. There, we only run the tests on CPU.
> 
> Actual CUDA tests are carried out in the self-hosted CI flows where we have our own GPUs.

Is it just sscp+omp? Could not reproduce. Will try to install llvm15+ubuntu2204 in chroot

illuhad:
> Is it just sscp+omp? Could not reproduce. Will try to install llvm15+ubuntu2204 in chroot

Yes, I think so.

resetius:
> > Is it just sscp+omp? Could not reproduce. Will try to install llvm15+ubuntu2204 in chroot
> 
> Yes, I think so.

I managed to reproduce it in chroot !

tests/sycl/atomic.cpp
```diff
@@ -185,11 +185,13 @@ void atomic_local_reduction_test(AtomicOp op, Verifier v,
...
-        sycl::access::address_space::local_space>
-      local_ref{local_acc[0]};
+      {
+        sycl::atomic_ref<T, sycl::memory_order::relaxed, sycl::memory_scope::work_group,
+          sycl::access::address_space::local_space>
+        local_ref{local_acc[0]};
 
-      op(local_ref, per_item_val);
+        op(local_ref, per_item_val);
+      }
```
illuhad, line 194: Great that this fixed it. I don't see the UB though -- what was the problem with the old code?

resetius: Sycl atomic ref follows semantics of std atomic ref : The lifetime of an object must exceed the lifetime of all std::atomic_refs that references the object. While any std::atomic_ref instance referencing an object exists, the object must be exclusively accessed through these std::atomic_ref instances. No subobject of an object referenced by an std::atomic_ref object may be concurrently referenced by any other std::atomic_ref object

I added {} and limited the lifetime of atomic ref

resetius: Key part : While an atomic_ref to an object is alive, that object must be accessed exclusively through atomic_ref.

illuhad: Hmm.... this makes sense from a standard point of view. But in practice, our `atomic_ref` implementation is stateless, and directly mapped to SSCP builtin calls, which in the case of host, invoke `__atomic_store_n()` compiler builtins for the assignment.

So I don't see how the generated code would be different based on when the destructor is triggered.

I just noticed that the host backend's implementation of the SSCP `__acpp_sscp_work_group_barrier` builtin has a fixme because apparently it does not correctly implement the memory fence associated with the barrier. This could potentially be related too. Or there's a CBS compiler bug.

illuhad (approved):
Thanks!

PR #2003, opened 2026-03-13
[SSCP][host] Fix subgroup ops

al42and:
- `__acpp_sscp_get_subgroup_id` was a bit off
- `__acpp_sscp_sub_group_permute_i*` had inverted logic

I can add / expand tests, but not sure how useful that would be for these corner cases


illuhad (approved):
LGTM for subgroup.
@sbalint98 can you please review for the group algorithm change.

sbalint98 (approved):
Thanks LGTM

sbalint98 (approved):

PR #2004, opened 2026-03-15
[Metal] Fixes: fshl/fshr, bool sign expansion

resetius:
- Fix fshl/fshr intrinsic expansion: wrong operand order in fshr and wrong zero-shift result
  - Fix bool sign extension (sext i1): __as_signed(bool) now returns int with all-ones pattern

Started running acpp in CI for my project on apple hardware. Hit numerical mismatches traced back to incorrect fshl/fshr expansion and bool sign extension in the metal backend


illuhad (approved):
Thanks!

PR #2009, opened 2026-03-16
Experimental Vulkan backend

EwanC:
Proof-of-concept Vulkan backend which uses [clspv](https://github.com/google/clspv) to compile SSCP kernels. I'm opening this draft PR at a stage where the majority of sycl tests pass on my integrated GPU but before the work is mature enough for merging to prompt discussion from the community on if there's demand for this backend, and if so what the bar for readiness and ongoing support would be.

The Vulkan backend only supports device USM, see `doc/install-vulkan.md` for explanation, so this patch contains many changes to tests to prefer device USM over host/shared to improve test coverage of the backend. I would expect this may need to reworked to avoid dropping host/shared USM testing coverage, and would suggest the test changes be split off as their own PR if the community wanted to move forward with this.

This PR includes CI for the LLVMPipe CPU Vulkan implementation from Mesa, and `doc/install-vulkan.md` includes the current support status of both SYCL features and hardware, as well as the multitude of areas for future improvement.


illuhad:
Woaaaa!! nice! I'd recommend not only testing with `sycl_tests` but also with `pcuda_tests` and the LIT tests, as they stress different parts of the infrastructure.

illuhad:
>  if there's demand for this backend,

The question of SYCL for Vulkan has come up repeatedly. Also remember the big splash that the Sylkan paper caused when it was published. https://dl.acm.org/doi/fullHtml/10.1145/3456669.3456683

> and if so what the bar for readiness and ongoing support would be.

Working CI, clearly defined set of supported features (ideally everything that works currently). Initial PR does not need to support everything yet if follow up PRs address gaps (see the ongoing process for the new metal backend).

Someone would need to maintain the backend who is reasonably familiar with Vulkan (you?).

doc/install-vulkan.md
```diff
@@ -0,0 +1,270 @@
...
+* `bufferDeviceAddress`
+* `timelineSemaphore`
+* `shaderInt8`
+* `shaderFloat16`
+* `storagePushConstant8`
+* `variablePointers`
+* `variablePointersStorageBuffer`
+* `storagePushConstant16`
+* `shaderInt16`
+* `shaderInt64`
```
illuhad: I'm not very familiar with Vulkan. Do these capabilities also give us a generic address space? Without generic pointers, it's going to be very hard to support SYCL or similar C++ single-source models.

src/compiler/llvm-to-backend/clspv/AddrSpaceCastRemovalPass.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include <llvm/IR/Constants.h>
+#include <llvm/IR/Instructions.h>
+
+namespace hipsycl {
+namespace compiler {
+
+// Remove casting away the address space from arguments, as otherwise
+// clspv doesn't know how to lower the generic pointer to SPIRV
+llvm::PreservedAnalyses
+AddrSpaceCastRemovalPass::run(llvm::Function &F,
```
resetius, line 22: Maybe I need to use the similar thing in the metal backend instead of tracking the original addrspaces in Emitter.cpp

doc/install-vulkan.md
EwanC: Thanks for the comment I will extend this section to explain what these are doing.

Regarding generic address space, the majority of the compiler passes I've written to make the SSCP LLVM-IR to be consumable by clspv have been to try retain the address space information in pointers as otherwise clspv seems to struggle to lower the store instructions. That address space information is generally getting
lost by addrspace cast instructions or ptr2int/int2ptr idioms.

Saying that the way I've set up the address space map is that the private address space is `0` and generic is `4`, and in general I'm trying to restore the address space of default address space pointers rather than handling generic address space pointers. I do believe clspv has a pass to try
help with that to deal with OpenCL-C generic address space.

Do you know of concrete SYCL examples that stress the generic address space i could look at to give a better investigation/explanation of what's going on based on that? The multi-ptr tests have seemed to "just work" so far.

EwanC:
> Working CI, clearly defined set of supported features (ideally everything that works currently). Initial PR does not need to support everything yet if follow up PRs address gaps (see the ongoing process for the new metal backend).
> 
> Someone would need to maintain the backend who is reasonably familiar with Vulkan (you?).

That makes sense, thanks.  I'm doing this as a side project, so won't be able to work on it full time but happy to get a MVP upstream and then own maintaining and developing it further.

Looking at the CI, it seems that testing is mostly being done on CPU even when builds have CUDA/ROCm backend enabled. We can use llvmpipe to get some test coverage on CPU, and also moltenVK to still get some test coverage on the metal runner.. I'll look into putting some GitHub actions together to verify that, and also better document what's expected to currently work.

doc/install-vulkan.md
illuhad: > Saying that the way I've set up the address space map is that the private address space is 0 and generic is 4, and in general I'm trying to restore the address space of default address space pointers rather than handling generic address space pointers. I do believe clspv has a pass to try
help with that to deal with OpenCL-C generic address space.

This is the default SPIR address space convention in LLVM. When developing our existing SPIR-V backend, I found this to be quite problematic because AdaptiveCpp (due to being a single-pass compiler) starts out with the code being in AS 0 everywhere. Remapping that reliably is a lot of work. This is why we use a patched SPIRV-LLVM-Translator which swaps address space 0 and 4, so that 0 is generic as the gods intended.

> Do you know of concrete SYCL examples that stress the generic address space i could look at to give a better investigation/explanation of what's going on based on that? The multi-ptr tests have seemed to "just work" so far.

The `multi_ptr` test mainly just tests the surface API of `multi_ptr`, not really generic address space semantics.

I'd recommend trying to run some of the benchmarks that we routinely run: https://github.com/illuhad/benchmark-scripts/tree/pcuda-benchmarking

Generally without generic address space you can run into problems with:

- function calls with pointer arguments, which might need callgraph duplication to have both local and global address space function versions. It's possible that the SSCP compiler hides this problem to some extent because it inlines very aggressively and always does something akin to LTO for kernels.
- Pointer capture. When you need to store pointers to global memory and then load them later usually all address space inference breaks down. This will primarily affect pointer-based data structures.

illuhad:
> That makes sense, thanks. I'm doing this as a side project, so won't be able to work on it full time but happy to get a MVP upstream and then own maintaining and developing it further.

That sounds good :)

> Looking at the CI, it seems that testing is mostly being done on CPU even when builds have CUDA/ROCm backend enabled. We can use llvmpipe to get some test coverage on CPU, and also moltenVK to still get some test coverage on the metal runner.. I'll look into putting some GitHub actions together to verify that, and also better document what's expected to currently work.

We run on GPUs with our self-hosted CI workflows. So you could try to get GPU CI coverage there, if you can get Vulkan to work in a headless context.
Otherwise, llvmpipe on the github hosted runners also sounds like a good option :)

illuhad:
@EwanC I've invited you to the AdaptiveCpp github organization. This will make CI run automatically for your PRs which might be helpful in the development process :-)

src/compiler/llvm-to-backend/clspv/llvm-to-clspv.md
```diff
@@ -74,3 +92,8 @@ optimization until this point.
 Pass for fixing up atomic builtin calls that take an opaque ptr without an address
 space but pointer value for used for builtin parameter has an addrspace. Done
 by creating an address space cast for the relevant operand.
+
+## TODO
+
+* LLVM has an pass [llvm::InferAddressSpacesPasS](https://llvm.org/doxygen/InferAddressSpaces_8cpp_source.html)
```
illuhad: AdaptiveCpp already uses that in `AddressSpaceInferencePass` if you use that (all the other backends do). However, `InferAddressSpacesPass` is just an optimization and often bails out early.

doc/install-vulkan.md
EwanC: Ah thanks for the explanation about the address space numbering, that clarifies why I wasn't seeing any AS 4 when dumping IR.

I tried out the benchmarking scripts tests, which helpfully for pcuda use cudaMalloc rather than cudaMallocManaged for pcuda. 5 out of 10 successfully completed on both SYCL/PCUDA, with 2 benchmarks failing to verify and 3 failing to compile the kernel. The compilation fail which looks most related to generic pointers is Cloverleaf which has the same symptoms of a simple linked-list unit test I wrote.

From experimenting clspv is generally uncomfortable both storing and loading pointers to global memory, i.e. working with pointers to pointers, and has OpenCL-C restrictions in place to ignore that. There might be some workarounds we can do to mitigate things using the IR input, but I agree "Pointer Capture" will be a fundamental restriction in the backend.

EwanC:
@illuhad I've taken this PR out of draft for review now as it has what feels like a MVP amount of functionality. There is a whole lot more CI, bug-fixing, performance analysis to do but agree it will be better to continue to iterate in the upstream repo. I'm happy to split this PR up if it will help review/merging. The `doc/install-vulkan.md` file contains documentation on what features/devices are supported, and general status, let me know if it needs clarifications.

The CI I've added is for llvm-pipe on Linux. The Vulkan jobs on this PR fail because the docker container needs
rebuilt, but PRs [fallback](https://github.com/AdaptiveCpp/AdaptiveCpp/blob/develop/.github/workflows/create-ci-docker/action.yml#L162) to existing docker tags rather than creating a new images. However, my fork also runs CI on the same git branch and does build the Docker image. So can see the Vulkan jobs running and passing [there](https://github.com/EwanC/AdaptiveCpp/actions/runs/23755910640) for the same commit hash. I'm not sure there's a way to get it so that we can merge this without those red CI jobs, but let me know if you have ideas. I want to add the self hosted runner CI too, but decide to do that as a follow-up too.

**Edit**: I've split off the test changes as their own PR for review in https://github.com/AdaptiveCpp/AdaptiveCpp/pull/2027 so if we merge that first then I only need to add in the backend skips in this PR rather than also having semantic test changes.

**Edit 15/04/26**: Refactored Linux CI so it doesn't need docker changes, and can run/pass as part of this PR testing.

tests/sycl/smoke/queue.cpp
```diff
@@ -0,0 +1,517 @@
...
+  int *host;
+  q.memcpy(&host, ptr, sizeof(int *)).wait();
+
+  BOOST_CHECK(target_ptr == host);
+
+  sycl::free(ptr, q);
+  sycl::free(target_ptr, q);
+}
+
+BOOST_AUTO_TEST_CASE(queue_linked_list) {
```
resetius, line 474: It would be nice to have this test upstream now (before merging the vk backend), as I've started work on USM pointers for the Metal backend

tests/sycl/smoke/queue.cpp
```diff
@@ -0,0 +1,517 @@
...
+  sycl::queue q{sycl::property::queue::in_order{}};
+  if (q.get_device().get_backend() == sycl::backend::vk) {
```
resetius, line 476: The same is also true for backend::metal

tests/sycl/smoke/queue.cpp
EwanC: Sure, I'll splitting out all the sycl test changes as a separate PR since this is a new file. I don't know much about Metal, will you be able to support host/shared USM as well as device USM? Most of my changes to the test files are about preferring device USM which will also benefit the Metal backend if it also won't be able to support host/shared.

resetius: Pointers to host, shared, and device memory are already supported. The only remaining problem is that pointers seen on the host differ from the pointers seen inside the kernel by a constant offset. Because of this, indirection-based structures like linked lists and trees do not work inside kernels yet. We need to translate the address when loading a pointer from another pointer

EwanC: ah that's an interesting implementation detail, since the other test changes will be of less interest to you I just created a PR with this single test - https://github.com/AdaptiveCpp/AdaptiveCpp/pull/2026

doc/install-vulkan.md
```diff
@@ -0,0 +1,457 @@
...
+
+#### USM
+
+The use of the [VK_KHR_buffer_device_address](https://docs.vulkan.org/refpages/latest/refpages/source/VK_KHR_buffer_device_address.html)
+extension is a key part of the implementation as it allows us to implement
+device USM. By using `vkGetBufferDeviceAddress` we can get a device addressable to
+give to the user that they can use in a kernel. However, host/shared USM isn't
+supportable as this pointer isn't valid on host, the underlying Vulkan buffer needs
+memory mapped to a make that possible which is an intermediate step that the
+SYCL runtime performs when the user invokes a SYCL API using a device USM pointer.
```
illuhad: I think you're specifically talking about `sycl::buffer`? Otherwise I'm not sure what you refer to with the runtime's intermediate step.

doc/install-vulkan.md
```diff
@@ -0,0 +1,457 @@
...
+
+In future `VK_KHR_device_address_commands` could be used to implement
+`memcpy`.
+
+#### Compute Pipeline
+
+A compute pipeline is created for each kernel JIT compiled for a specific workgroup size.
+As we are JIT compiling kernels, each translation unit that the clspv compiler sees only
+contains a single kernel in it, and therefore only use a single descriptor set and binding
+when creating the kernels.
```
illuhad, line 218: This is true for the default adaptivity level 1. If a user sets `ACPP_ADAPTIVITY_LEVEL=0`, it will try to JIT-compile entire TUs. Might be worth to double check that against assumptions you might be making in the code :)

doc/install-vulkan.md
```diff
@@ -0,0 +1,457 @@
...
+is found in the cache with the JIT compiled SPIR-V kernel code. This will already
+have a `vk_kernel_object` attached representing the kernel function in that binary.
+If the workgroup size of the kernel enqueue is the same as previously then the
+old cached `vk_kernel_pipeline` instance will be used with the Vulkan compute pipeline created
+for that specific workgroup size. Otherwise, if the workgroup size is different, a
+new `vk_kernel_pipeline` object will be created to instantiate a Vulkan compute pipeline
+for that workgroup size. However, the AdaptiveCpp SSCP compiler seem to create a
+new executable for each work-group size with the information embedded in the required
+work-group size metadata so in practice a new `vk_executable_object` is created for
+each work-group size variant of a kernel.
```
illuhad, line 230: This is true at adaptivity level 1 (default). There are more cases for which it specializes kernels, not just work group size, see https://dl.acm.org/doi/10.1145/3731125.3731127

doc/install-vulkan.md
```diff
@@ -0,0 +1,457 @@
...
+and each argument is an offset into that uniform buffer.
+
+#### Queue
+
+Vulkan devices are conceptually split into physical devices and logical devices, where
+a logical devices is created by the user based on the available physical devices
+and their capabilities. During the creation of the logical device the SYCL runtime
+finds the first Vulkan queue family which supports compute, and creates a single
+single queue to tie to the logical device. All `sycl::queue` objects created map
+to that single queue, but each have control their own command-buffers to use
```
illuhad:
```suggestion
to that single queue, but each have control over their own command-buffers to use
```

doc/install-vulkan.md
```diff
@@ -0,0 +1,457 @@
...
+
+#### Specialization constants
+
+A specialization constant will be created by `clspv` if the required workgroup size attribute
+was not set in the kernel IR, which is part of the reflection information that the runtime
+can detect and set.
+
+#### libKernel
+
+To implement the kernel functions declared by `libKernel` many of them are mapped to
```
illuhad:
```suggestion
To implement the kernel functions declared by `libkernel` many of them are mapped to
```

doc/install-vulkan.md
```diff
@@ -0,0 +1,457 @@
...
+OpenCL-C builtin functions, which `clspv` then knows how to implement correctly for
+SPIR-V. However, clspv only supports OpenCL-C version 1.2 and some builtins (e.g.
+work-group collectives) require builtins from OpenCL 3.0. In such cases a more
+generic implementation is fallen back to.
```
illuhad, line 300: I wonder if there's a way to leverage the existing SPIR-V libkernel builtins...

include/hipSYCL/runtime/vk/vk_code_object.hpp
```diff
@@ -0,0 +1,225 @@
...
+
+struct range_hash {
+  size_t operator()(const rt::range<3> &key) const {
+    uint64_t result = uint16_t(key[0]);
+    result = (result << 16) + key[1];
+    result = (result << 16) + key[2];
+    static_assert(sizeof(size_t) == sizeof(uint64_t));
+    return (size_t)result;
+  }
+};
```
illuhad, line 118: Does this belong in this file?

include/hipSYCL/runtime/vk/vk_queue.hpp
```diff
@@ -0,0 +1,135 @@
...
+      hcf_object_id hcf_object, std::string_view kernel_name,
+      const rt::hcf_kernel_info *kernel_info, const rt::range<3> &num_groups,
+      const rt::range<3> &group_size, unsigned local_mem_size, void **args,
+      std::size_t *arg_sizes, std::size_t num_args,
+      const kernel_configuration &config) override;
+
+private:
+  /*
+   * Helper functions
+   */
```
illuhad: I don't think we use this comment style for function declarations

src/compiler/llvm-to-backend/clspv/AddrSpaceCastRemovalPass.cpp
```diff
@@ -0,0 +1,57 @@
...
+                              llvm::FunctionAnalysisManager &) {
+  llvm::SmallVector<llvm::Instruction *> ASCToRemove;
+  bool DidGEPTransform = false;
+  for (auto &BB : F) {
+    for (auto &I : BB) {
+      if (auto *ASC = llvm::dyn_cast<llvm::AddrSpaceCastInst>(&I)) {
+        if (ASC->getDestAddressSpace() == 0) {
+          llvm::Value *Op = ASC->getPointerOperand();
+          ASC->replaceAllUsesWith(Op);
```
illuhad, line 31: I this legal? Couldn't this generate invalid IR when client code expects ptr of a certain adress space, and you just pass in the operand prior to the cast (which might be of different address space and thus type)?

src/compiler/llvm-to-backend/clspv/llvm-to-clspv.md
```diff
@@ -0,0 +1,120 @@
...
+2) When you need to store pointers to global memory and then load them later, address space inference breaks down,
+   affecting pointer-based data structures.
+
+The current situation with these is:
+
+1) The aggressive inlining from the compiler mitigates this by the time we get
+   to optimizing the flavored IR stage of the pipeline.
+2) Not supported. To avoid this issue clspv will not compile OpenCL-C kernels that define structures which have pointers
+   members and does not support OpenCL-C generic address space feature. Additionally, OpenCL-C kernel arguments cannot
+   be pointer-to-pointer. Try to use clspv for SYCL kernels that do pointer chasing or try to exfiltrate pointers to
```
illuhad:
```suggestion
   be pointer-to-pointer. Trying to use clspv for SYCL kernels that do pointer chasing or trying to exfiltrate pointers to
```

src/libkernel/sscp/clspv/atomic.cpp
```diff
@@ -0,0 +1,1307 @@
...
+    return 3; // memory_scope_all_devices
+  }
+  return 0;
+}
+
+// ********************** atomic store ***************************
+
+extern "C" {
+// Address space(1), int32
+void _Z21atomic_store_explicitPU3AS1VU7_Atomicii12memory_order12memory_scope(
```
illuhad: Why do we not let name mangling do its job and pre-mangle names? Is this because of a portability concern? I think that the LLVM `spirv` target will probably automatically use Itanium target. Also, the mangled name does not match its arguments: It demangles to
```
atomic_store_explicit(int _Atomic volatile AS1*, int, memory_order, memory_scope)
```
while your signature lacks the `volatile` and `__attribute__((address_space(1)))`.

illuhad (commented):
Some thoughts. Will take me a while to get through everything :)

doc/install-vulkan.md
```diff
@@ -0,0 +1,457 @@
...
+* `relational_tests/rel_genfloat_unary`
+* `fill_tests/*`
+* `item_api_tests/*`
+* `extension_tests/buffer_page_size`
+* `explicit_copy/explicit_buffer_copy_host_ptr`
+* `group_functions_tests/group_x_of_local`
+
+### Issue 10
+
+> Issue only appears in RADV Phoenix
```
al42and: I also see it on `RADV NAVI24`

EwanC: Thanks for the feedback Andrey, can you share any more of your experience trying this backend out. Was it just the AdaptiveCpp sycl tests that you ran, or did you try run some GROMACS code as well?

doc/install-vulkan.md
EwanC: Yes that's the main time this runtime does this, I've rephrased this sentence.

src/libkernel/sscp/clspv/atomic.cpp
EwanC: This was an oversight because I thought there was a limitation of overloading C++ functions just based on the attribute, but seems this isn't the case for clang with address spaces. Updated now to the non-mangled names.

include/hipSYCL/runtime/vk/vk_code_object.hpp
EwanC: It's used to provide the hash function for `rt::range<3>` so we can use that object in a std map as a key later on in this header, i've added a comment to that effect but let me know if you still think this should be removed.

doc/install-vulkan.md
EwanC: Thanks for pointing me at this, I've removed the runtime assumptions about a single kernel per TU and ran the sycl test suite at the different adaptivity levels. I running level 2 until compilation stabilizes passes consistent with to the same correctness as adaptivity level 1.

On adaptivity level 0 some tests suites pass which gives me confidence the runtime can handle loading and setting arguments/work-group size with spv binaries containing multiple kernels. There are some fails though due to different kernel IR, e.g. dead arg elimination not running which results in https://github.com/google/clspv/issues/1593, but there are other compilation issues I need to look into as well. I'm thinking of treating getting adaptivity level 0 passing as future work now the core assumptions have been addressed.

jordy-van-appeven:
Is there any chance that physical devices exposing a driver API older than Vulkan 1.3 could be supported? My target platform is embedded Qualcomm hardware, non-Android/Snapdragon based devices. Looking at the QCS-, QRB-, and newly released IQ-series, Vulkan 1.3 support appears to be quite limited.

EwanC:
> Is there any chance that physical devices exposing a driver API older than Vulkan 1.3 could be supported? My target platform is embedded Qualcomm hardware, non-Android/Snapdragon based devices. Looking at the QCS-, QRB-, and newly released IQ-series, Vulkan 1.3 support appears to be quite limited.

@jordy-van-appeven Thanks for the interest and sharing your use case. I don't have any devices handy to test on that are less than Vulkan 1.3 which was why this was set. I've pushed a commit lowering the check to version 1.2 because we're not using any core version 1.3 features, but in version 1.2 we use the core features for timeline semaphores and device buffer address.

It might be possible to go down to version 1.1 as base requirement and interface through the extensions for these
features before they became core, but it would require a device exposing those extensions to make sure that development is correct.

src/compiler/llvm-to-backend/clspv/AddrSpaceCastRemovalPass.cpp
EwanC: Yes you're right. What I had been doing was fixing up these cases by writing later LLVM passes, however that leaves the IR in an invalid state between passes which is badly behaved. I've refactored this so the other passes are rolled into this pass, effectively doing multiple iterations over the function in a single pass but the IR at the end of the LLVM pass should be well formed.

jordy-van-appeven:
Haven’t tested this on any Qualcomm (Adreno) device yet, but I can confirm it works on an NXP i.MX SoM with a Mali-310 GPU and vendor support for Vulkan 1.2.

doc/install-vulkan.md
EwanC: I don't think there's a standard way to reprsent SPIR-V builtins in LLVM IR, the LLVM-SPIRV expects a particular format but that's not standardized so that clspv would understand it.  I've worked out that clspv represents spirv builtins in IR by using a function call to a mangled name prefixed with convention `_Z8spirv.op.<SPIRV OpCode>.<Mangled parameters>`. I've not tried to use this yet, but I think it could be and might at least align more with the spir-v libkernel implementation.

.github/workflows/linux-vulkan.yml
```diff
@@ -0,0 +1,163 @@
...
+        cmake -GNinja -S $GITHUB_WORKSPACE/clspv -B $GITHUB_WORKSPACE/clspv/build \
+              -DCMAKE_INSTALL_PREFIX=$GITHUB_WORKSPACE/clspv/build/install \
+              -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
+        cmake --build $GITHUB_WORKSPACE/clspv/build --target install
+
+    - name: clone LLVM
+      run: |
+        git clone --recurse-submodules https://github.com/llvm/llvm-project --depth 1 --single-branch -b release/${{matrix.clang_version}}.x llvm
+
+    - name: build Adaptive Cpp
```
illuhad:
```suggestion
    - name: build AdaptiveCpp
```

doc/install-vulkan.md
```diff
@@ -0,0 +1,496 @@
...
+| RADV Pheonix | Mesa 25.0.7       | AMD Integrated GPU   | Well supported                                                                        |
+| RADV MI100   | Mesa 23.2.1       | AMD Discrete GPU     | Well Supported                                                                        |
+| RADV MI210   | Mesa 23.2.1       | AMD Discrete GPU     | Many [Issues](#issue-9)                                                               |
+| Arc MTL      | Mesa 25.0.7       | Intel Integrated GPU | Issues with sub-groups                                                                |
+| RTX 500      | NVIDIA 580.95.5.0 | NVIDIA Discrete GPU  | Tests pass in isolation, but device stops being detected when running full sycl suite |
+| MoltenVK     | Khronos 1.4.1     | Apple Integrated GPU | CI testing with Macos 15.7.4, [some](#issue-11) tests disabled                        |
+| Swiftshader  | Google 5.0.0      | CPU                  | Poor support, no `Int64` or `VariablePointer` capabilities                            |
+
+Other devices are untested and support status unknown.
```
illuhad:
```suggestion
Other devices are untested and support status is unknown.
```

include/hipSYCL/runtime/vk/vk_code_object.hpp
illuhad: Wouldn't it make sense then to just move it to the definition of `rt::range`, so that the hash is available globally? This should be useful beyond Vulkan.

src/compiler/llvm-to-backend/clspv/LLVMToCLSPV.cpp
```diff
@@ -0,0 +1,452 @@
...
+  Invocation.push_back(InputFileName);
+
+  std::string ArgString;
+  for (const auto &S : Invocation) {
+    ArgString += S;
+    ArgString += " ";
+  }
+  HIPSYCL_DEBUG_INFO << "LLVMToCLSPV: Invoking " << ArgString << "\n";
+
+  int R = llvm::sys::ExecuteAndWait(CLSPV, Invocation);
```
illuhad: Please use the new `hipsycl::compiler::executeAndWait()`
(see #2094)

include/hipSYCL/common/debug.hpp
```diff
@@ -44,6 +45,11 @@ class output_stream {
   std::ostream &get_stream() const { return _output_stream; }
   int get_debug_level() const { return _debug_level; }
 
+  void atomic_print(const char *prefix, std::stringbuf *str_buf) {
+    std::lock_guard<std::mutex> lock(_mutex);
+    _output_stream << prefix << str_buf;
```
illuhad: Why do we need `HIPSYCL_DEBUG_ATOMIC`? Is the backend multi-threaded, and if so, in what way?

illuhad (commented):
Almost there :D

include/hipSYCL/runtime/vk/vk_code_object.hpp
EwanC: good point, I've moved this to `include/hipSYCL/runtime/util.hpp`

include/hipSYCL/common/debug.hpp
EwanC: Yes exactly, the backend is multi-threaded. Using this threadsafe logging improved the readability of the log output by preventing messages from being interleaved, but I wouldn't say it's essential if there are concerns about adding this logging.

The multi-threading is required to fill in functionality gaps in asynchronous commands between SYCL and Vulkan while keeping the asynchronous model rather than forcing a blocking host-wait in the implementation. The Vulkan timeline sempahores I use for synchronization can be signalled/waited-on by either the device or host code, so I can signal/wait on them as part of the host side thread to synchronize correctly.

Right now I'm doing this in 3 places with the `worker_thread` class from `generic/async_worker.hpp`.

1) In a Vulkan a memset command must be multiple of 4 bytes, while in SYCL there is no such requirement. Therefore I implement memset on host using `std::memset` wrapper as a host-task. I should probably add a faster path that prefers the Vulkan API when the size of a multiple of 4 bytes, and only fallack the the host thread impl when this isn't the case.
2) To implement memcopy asynchronously we use `vkCmdCopyBuffer` which takes Vulkan buffer objects rather than raw pointers, like trying to use `cl_mem` style `clEnqueueCopyBuffer` in CL. However, because a user can pass host allocated pointers rather than just pointers created from usm alloc APIs where we've already got a vkBuffer. Then if the host pointer is the source, we need to create a new backing VK buffer and copy the host pointer to into the buffer before the command can run. And if the host-pointer is the destination operand, then we need to copy the vk Buffer contents back to the host pointer before signally to the user that the command is complete.
3) To implement `submit_external_wait_for()` doing the synchronization with host side semaphore wait/signal calls, but I think instead we could probably submit an empty command buffer and use the VK command-buffer submit parameters for semaphore signal/wait to do the same thing.

illuhad: I see. I think I'd prefer if Vulkan didn't use a different logging mechanism compared to all other backends. I wonder if we should just make `DEBUG` messages mutex-protected globally -- if debug messages are generated, performance is probably not a prime concern anyway.
(Maybe warnings and errors could remain unprotected as they might also occur in production runs?)

EwanC: Having consistent logging makes sense from a maintenance  perspective, so I've reverted this part of the change so that the Vulkan backend using the existing logging mechanisms. Then if we want to change the logging to make it atomic across the backends we can do that as a dedicated PR which will be easier to review/revert as a standalone change.

illuhad: Okay, sounds good :)

illuhad (approved):
I guess it's  time -- thanks! :)

doc/install-vulkan.md
```diff
@@ -0,0 +1,493 @@
...
+Devices known to be well supported by the backend include:
+
+| Driver Name  | Driver Version    | Device Type          | Status
+| ------------ | ----------------- | -------------------- | ------------------------------------------------------------------------------------- |
+| llvmpipe     | Mesa 25.0.7       | CPU                  | Well supported                                                                        |
+| RADV Pheonix | Mesa 25.0.7       | AMD Integrated GPU   | Well supported                                                                        |
+| RADV MI100   | Mesa 23.2.1       | AMD Discrete GPU     | Well Supported                                                                        |
+| RADV MI210   | Mesa 23.2.1       | AMD Discrete GPU     | Many [Issues](#issue-9)                                                               |
+| Arc MTL      | Mesa 25.0.7       | Intel Integrated GPU | Issues with sub-groups                                                                |
+| RTX 500      | NVIDIA 580.95.5.0 | NVIDIA Discrete GPU  | Tests pass in isolation, but device stops being detected when running full sycl suite |
```
al42and, line 112: FYI, `ACPP_PERSISTENT_RUNTIME=1` fixes this issue for me. Not sure how robust this is, though: I get other failures, but they appear even when run individually.

PR #2016, opened 2026-03-21
[Metal] Fix in-order queue race in device->host->host->device transfer chains

resetius:
I ran GROMACS and found a regression introduced in #1989 
In-order queue semantics are violated for device->host->host->device transfer chains, where the CPU-side staging copy for host->device (and host->host) races  against the asynchronous completion callback of the preceding device->host transfer


tests/sycl/extensions.cpp
```diff
@@ -1409,4 +1409,73 @@ BOOST_AUTO_TEST_CASE(khr_queue_empty) {
 }
 #endif
 
+BOOST_AUTO_TEST_CASE(inorder_queue_d2h_h2d_ordering) {
```
illuhad: These tests do not really test SYCL extensions -- shouldn't they be better in `tests/sycl/queue.cpp`, which is supposed to focus on queue APIs?

resetius: Thank you. Moved

illuhad (approved):
Thanks!

PR #2021, opened 2026-03-25
implement native::sincos

ferdymercury:
Currently, there is no way to access from AdaptiveCpp/SYCL the fast sincos functions such as https://docs.nvidia.com/cuda/libdevice-users-guide/__nv_fast_sincosf.html

I tested that using sycl::sincos (which works fine with acpp) is much slower than using sycl::native::sin and sycl::native::cos one after the other.

This PR proposes to implement sycl::native::sincos to bridge this gap.

I don't know if I implemented it correctly, but it's building on my system, and at least now I get the same speed when using `sycl::native::sincos` compared to `sycl::native::sin` and `sycl::native::cos`. (Maybe I am not building correctly and it's defaulting to the generic target rather than the cuda target, which is why I see no difference in speed? Or maybe it's too marginal to be seen.)


include/hipSYCL/sycl/libkernel/builtins/math_builtins.hpp
```diff
@@ -396,6 +396,7 @@ HIPSYCL_NATIVE_MATH_BUILTIN(BINARY_T_T, powr)
 HIPSYCL_NATIVE_MATH_BUILTIN(UNARY_T, recip)
 HIPSYCL_NATIVE_MATH_BUILTIN(UNARY_T, rsqrt)
 HIPSYCL_NATIVE_MATH_BUILTIN(UNARY_T, sin)
+HIPSYCL_NATIVE_MATH_BUILTIN(BINARY_T_TPTR, sincos)
```
illuhad, line 399: SYCL does not have this function, so you are proposing an extension. Extensions are generally prefixed with `AdaptiveCpp_` in their user-facing API (and documented in `extensions.md` with feature-test macro in `extensions.hpp`).

include/hipSYCL/sycl/libkernel/sscp/builtins/amdgpu/ocml.hpp
```diff
@@ -360,6 +360,7 @@ extern "C" double __ocml_native_sqrt_f64 (double);
 extern "C" double __ocml_native_rsqrt_f64 (double);
 extern "C" double __ocml_native_sin_f64 (double);
 extern "C" double __ocml_native_cos_f64 (double);
+extern "C" double __ocml_native_sincos_f64 (double, double*);
```
illuhad, line 363: This function does not exist in OCML, please don't invent things that are not there :-)
https://github.com/ROCm/llvm-project/blob/amd-staging/amd/device-libs/ocml/inc/ocml.h

You could target `__ocml_sincos_*` (non-native), in which case this signature is likely still incorrect because the actual definition uses a pointer to private address space, see https://github.com/ROCm/llvm-project/blob/a4bbcf59ca52f331c3dc9d076f271fb2c3577ac6/amd/device-libs/ocml/inc/ocml.h#L353
so you might need `__amdgpu_private`/`__attribute__((address_space(5)))` and ensure in the implementation that the pointer passed in is exclusively a private pointer, e.g. by first passing in some temporary variable.

include/hipSYCL/sycl/libkernel/sscp/builtins/amdgpu/ocml.hpp
```diff
@@ -370,6 +371,7 @@ extern "C" float __ocml_native_sqrt_f32 (float);
 extern "C" float __ocml_native_rsqrt_f32 (float);
 extern "C" float __ocml_native_sin_f32 (float);
 extern "C" float __ocml_native_cos_f32 (float);
+extern "C" float __ocml_native_sincos_f32 (float, float*);
```
illuhad, line 374: same here

include/hipSYCL/sycl/libkernel/sscp/builtins/amdgpu/ocml.hpp
```diff
@@ -112,7 +112,7 @@ extern "C" double __ocml_sincos_f64 (double, double);
 extern "C" __acpp_f32_2 __ocml_ccos_f32 (__acpp_f32_2);
 extern "C" __acpp_f32_2 __ocml_ccosh_f32 (__acpp_f32_2);
 extern "C" __acpp_f32_2 __ocmlpriv_epexpep_f32 (__acpp_f32_2);
-extern "C" float __ocml_sincos_f32 (float, float);
+extern "C" float __ocml_sincos_f32 (float, float*);
```
illuhad, line 115: see remark further down regarding the signature and the address space in case you want to leverage this function.

src/libkernel/sscp/amdgpu/native.cpp
```diff
@@ -37,6 +37,8 @@ HIPSYCL_SSCP_BUILTIN float __acpp_sscp_native_rsqrt_f32(float x) { return __ocml
 
 HIPSYCL_SSCP_BUILTIN float __acpp_sscp_native_sin_f32(float x) { return __ocml_native_sin_f32(x); }
 
+HIPSYCL_SSCP_BUILTIN float __acpp_sscp_native_sincos_f32(float x, float* cosval) { *cosval = __ocml_native_cos_f32(x); return __ocml_native_sin_f32(x); }
```
illuhad, line 40: Why introduce the `sincos` builtins in OCML if you then don't even call them?

illuhad (commented):
This definitely needs some test cases to check if everything JIT-compiles correctly.

You might also want to provide a metal implementation (CC @resetius ), since we have this backend now as well.

The bigger question though might be whether there is benefit to having a universal `native_sincos`, if ultimately only one backend (CUDA) can map it to a fast path.

Note that you can already do something like this with generic JIT compiler (pseudo code -- you might have to forward-declare `__nv_fast_sincosf` and surround the whole thing with `__acpp_if_target_sscp()`):
```c++
namespace jit = sycl::AdaptiveCpp_jit;
bool is_nvidia = jit::reflect<jit::reflection_query::target_vendor_id>() == 
      jit::vendor_id::nvidia;
jit::compile_if(is_nvidia, [&](){
  // if branch
  result = __nv_fast_sincosf(...);
},
[&](){
  // else branch
  result = sycl::sincos(...);
});
```

Because the accuracy for `native` builtins is completely undefined, they generally aren't useful for portability anyway because backends may do something wildly differently. Unless you have zero requirements on accuracy and can accept in an extreme case if a backend just returns "0" as an approximation :-)

There might be a case though for correctly mapping non-native `sincos` instead of emulating it globally as we currently do: https://github.com/AdaptiveCpp/AdaptiveCpp/blob/c4c673ba10d44ed5829799f070cc52dc7fa06eb7/include/hipSYCL/sycl/libkernel/sscp/builtins.hpp#L209

Because `sincos` exists (without dedicated fast-path) in many backends, such a mapping might make sense on many backends.

src/libkernel/sscp/host/native.cpp
```diff
@@ -36,6 +36,8 @@ HIPSYCL_SSCP_BUILTIN float __acpp_sscp_native_rsqrt_f32(float x) { return __acpp
 
 HIPSYCL_SSCP_BUILTIN float __acpp_sscp_native_sin_f32(float x) { return __acpp_sscp_sin_f32(x); }
 
+HIPSYCL_SSCP_BUILTIN float __acpp_sscp_native_sincos_f32(float x, float* cosval) { return __acpp_sscp_sincos_f32(x, cosval); }
```
illuhad, line 39: I don't think the `__acpp_sscp_sincos` SSCP builtin exists currently because `sincos` is emulated globally.

src/libkernel/sscp/spirv/native.cpp
```diff
@@ -36,6 +36,8 @@ HIPSYCL_SSCP_BUILTIN float __acpp_sscp_native_rsqrt_f32(float x) { return __acpp
 
 HIPSYCL_SSCP_BUILTIN float __acpp_sscp_native_sin_f32(float x) { return __acpp_sscp_sin_f32(x); }
 
+HIPSYCL_SSCP_BUILTIN float __acpp_sscp_native_sincos_f32(float x, float* cosval) { return __acpp_sscp_sincos_f32(x, cosval); }
```
illuhad, line 39: Same here, I don't think this builtin currently exists.

include/hipSYCL/sycl/libkernel/builtins/math_builtins.hpp
ferdymercury: https://github.khronos.org/SYCL_Reference/iface/math-functions.html#sycl-sincos

I guess it does belong to the SYCL standard?

ferdymercury: Oh I see what you mean, it's not in the native subsection: https://github.khronos.org/SYCL_Reference/iface/native-prec-math-func.html

illuhad: Yep - we have `sycl::sincos` in the standard, but not `sycl::native::sincos`.

resetius:
for metal you can use
```
T sincos(T x, T &cosval) Compute sine and cosine of x. Return the
computed sine in the function return value, and
return the computed cosine in cosval.
```
where T = float or half

```
HIPSYCL_SSCP_BUILTIN f32 __acpp_sscp_sincos_f32(f32 x, f32* ptr) {
  return __acpp_sscp_metal_math_f32_f32_f32ptr("sincos(%s, *__pointer_cast<float>(%s))", x, ptr);
}
```

ferdymercury:
> Note that you can already do something like this with generic JIT compiler (pseudo code -- you might have to forward-declare `__nv_fast_sincosf` and surround the whole thing with `__acpp_if_target_sscp()`):

Thanks, this seemed to work:

```
extern "C" void __nv_fast_sincosf (float, float*, float*);

  inline sycl::vec<float, 3> sphericalAnglesToCartesianUnitDirection(const float theta, const float phi)
  {
    __acpp_if_target_sscp(
                          float cT, sT, cP, sP; namespace jit = sycl::AdaptiveCpp_jit;
                          bool is_nvidia =
                              jit::reflect<jit::reflection_query::target_vendor_id>() ==
                              jit::vendor_id::nvidia;
                          jit::compile_if_else(
                              is_nvidia,
                              [&]() {
                                // if branch
                                __nv_fast_sincosf(theta, &sT, &cT);
                                __nv_fast_sincosf(phi, &sP, &cP);
                              },
                              [&]() {
                                // else branch
                                cT = sycl::native::cos(theta);
                                sT = sycl::native::sin(theta);
                                cP = sycl::native::cos(phi);
                                sP = sycl::native::sin(phi);
                              });

                          return sycl::vec<float, 3>(sT * cP, sT * sP, cT);)

        const float cosTheta = sycl::native::cos(theta);
    const float sinTheta = sycl::native::sin(theta);
    const float cosPhi = sycl::native::cos(phi);
    const float sinPhi = sycl::native::sin(phi);
    return sycl::vec<float, 3>(sinTheta * cosPhi, sinTheta * sinPhi, cosTheta);
  }
```

ferdymercury:
Considering that there is a workaround for NVIDIA, I am leaning towards closing this PR since native sincos is not part of the SYCL standard. Thanks a lot for the help!

PR #2025, opened 2026-04-01
Propagate CMAKE_BUILD_TYPE to LLVMSpirvTranslator ExternalProject

blinkfrog:
This forwards `CMAKE_BUILD_TYPE` to the `LLVMSpirvTranslator` ExternalProject.

Without this, when building AdaptiveCpp with OpenCL backend on Windows, the translator sub-build is configured with a different build type than AdaptiveCpp build (Debug, while LLCM + AdaptiveCpp were built as Release), causing build to fail due to `_ITERATOR_DEBUG_LEVEL` mismatch.

The fix is simply to propagate:

`-DCMAKE_BUILD_TYPE:STRING=${CMAKE_BUILD_TYPE}`

to the ExternalProject configure step.

(Related: https://github.com/AdaptiveCpp/AdaptiveCpp/discussions/2014)


illuhad (approved):
Thanks!

PR #2026, opened 2026-04-01
Add USM linked list sycl test

EwanC:
This test was written as part of the Vulkan backend development PR, https://github.com/AdaptiveCpp/AdaptiveCpp/pull/2009#discussion_r3019004377, but as it is also useful for stressing the Metal backend development work I'm creating a standalone PR to add it.

This test is interesting as it stresses the generic address space semantics as the compiler can't infer the address space of the pointer inside the `Node` struct.

### DIFF
--- a/tests/sycl/usm.cpp
+++ b/tests/sycl/usm.cpp
@@ -467,38 +467,23 @@ BOOST_AUTO_TEST_CASE(allocation_zero_bytes) {
     sycl::free(aligned_shared_ptr, q);
 }
 
-BOOST_AUTO_TEST_CASE(linked_list) {
-  sycl::queue q{sycl::property::queue::in_order{}};
-  if (q.get_device().get_backend() == sycl::backend::metal) {
-    BOOST_TEST_MESSAGE("Not yet supported on Metal backend");
-    return;
-  }
+namespace linked_list {
 
-  struct Node {
-    int id;
-    Node *next;
-  };
+struct Node {
+  int id;
+  Node *next;
+};
 
-  const int num_nodes = 3;
-  Node *nodes = sycl::malloc_device<Node>(num_nodes, q);
-  Node nodeC{2, nullptr};
-  Node nodeB{1, nodes + 2};
-  Node nodeA{0, nodes + 1};
-
-  q.memcpy(nodes, &nodeA, sizeof(Node));
-  q.memcpy(nodes + 1, &nodeB, sizeof(Node));
-  q.memcpy(nodes + 2, &nodeC, sizeof(Node));
-  q.wait();
-
+std::pair<int, int>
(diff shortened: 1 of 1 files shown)


tests/sycl/usm.cpp
```diff
@@ -467,5 +467,47 @@ BOOST_AUTO_TEST_CASE(allocation_zero_bytes) {
...
+  q.memcpy(nodes, &nodeA, sizeof(Node));
+  q.memcpy(nodes + 1, &nodeB, sizeof(Node));
+  q.memcpy(nodes + 2, &nodeC, sizeof(Node));
+  q.wait();
+
+  int *ptr = sycl::malloc_device<int>(1, q);
+  q.submit([&](sycl::handler &cgh) {
+    cgh.single_task([=]() {
+      Node *curr = nodes;
+      while (curr != nullptr) {
```
resetius: Could you please add a max_iterations limit just in case to avoid hanging the kernel if there is some pointer or memory issue? Something like:
```
for (int i = 0; i < max_iterations && curr; ++i, curr = curr->next)
```
Also, please check the actual number of iterations as well

EwanC: Sure, I've given that a go and just set the max iterations to the number of nodes +1, since at that point we know things have gone wrong.

resetius (approved):
LGTM

illuhad (commented):
Thanks, we can merge it as is if you prefer, but I'm curious if you'd consider the following case also an interesting test case for this PR:

The idea would be to have the nodes point to different allocations, not just different offsets of the same allocation. This is the kind of indirection that would test what happens when the runtime cannot see all utilized allocations and is I think a pathological case for USM emulations (and I suppose probably does not work currently, but we could disable it for metal/vulkan for now)
```
// (pseudocode)
Node* node1 = malloc_device<Node>(1, ...);
Node* node2 = malloc_device<Node>(1, ...);
Node* node3 = malloc_device<Node>(1, ...);

Node n1 {val1, node2};
Node n2 {val2, node3};
Node n3 {val3, nullptr};
q.memcpy(node1, &n1, sizeof(Node));
//etc
```
Do you think there'd be value in this test?

resetius:
> 
> Do you think there'd be value in this test?

It would be nice to have, but please make it a separate test. I know it still will not work on Metal even if I implement address translation. Passing indirectly captured buffers to a Metal kernel is a known limitation and should be fixed separately

EwanC:
> It would be nice to have, but please make it a separate test. I know it still will not work on Metal even if I implement address translation. Passing indirectly captured buffers to a Metal kernel is a known limitation and should be fixed separately

I've added it as a separate test but made an effort to share the common code

illuhad (approved):
Thanks!

PR #2027, opened 2026-04-02
Prefer device USM in SYCL tests

EwanC:
This change is split off from the [PR adding the Vulkan backend](https://github.com/AdaptiveCpp/AdaptiveCpp/pull/2009) which only supports device USM and not host/shared USM. As a result this patch updates SYCL tests where possible to use device USM, and adds device aspect checks for the cases where tests are not suitable for updating.

Additionally, this introduces some tests from the Vulkan backend PR which I thought were useful and couldn't spot coverage for with the existing tests.


tests/sycl/accessor.cpp
```diff
@@ -958,4 +958,44 @@ BOOST_AUTO_TEST_CASE(zero_dim_accessor) {
...
+  {
+    sycl::buffer<int> a_buff{in_data.data(), size};
+    sycl::buffer<int> c_buff{size};
+    c_buff.set_final_data(out_data.data());
+
+    sycl::nd_range<1> b{sycl::range<1>(size), sycl::range<1>(4)};
+    q.submit([&](sycl::handler &cgh) {
+      auto const_acc =
+          a_buff.template get_access<sycl::access_mode::read,
+                                     sycl::target::constant_buffer>(cgh);
```
illuhad: `constant_buffer` is deprecated in SYCL 2020. We can add this test, but I would recommend to not focus effort on it.

tests/sycl/extensions.cpp
```diff
@@ -669,25 +673,32 @@ BOOST_AUTO_TEST_CASE(cg_property_prefer_execution_lane) {
...
+
   std::size_t test_size = 4096;
-  int *shared_mem = sycl::malloc_shared<int>(test_size, q);
+  int *ptr = sycl::malloc_shared<int>(test_size, q);
 
+  std::vector<int> host_data(test_size);
   for (std::size_t i = 0; i < test_size; ++i)
-    shared_mem[i] = i;
+    host_data[i] = i;
+  q.memcpy(ptr, host_data.data(), test_size * sizeof(int)).wait();
```
illuhad: Why is this necessary? `ptr` is still shared USM. I think apart from the early exit in case there's no shared USM, no change should be necessary to this test, or am I missing something?

tests/sycl/kernel.cpp
```diff
@@ -0,0 +1,84 @@
+/*
+ * This file is part of AdaptiveCpp, an implementation of SYCL and C++ standard
+ * parallelism for CPUs and GPUs.
+ *
+ * Copyright The AdaptiveCpp Contributors
+ *
+ * AdaptiveCpp is released under the BSD 2-Clause "Simplified" License.
+ * See file LICENSE in the project root for full license details.
+ */
+// SPDX-License-Identifier: BSD-2-Clause
```
illuhad: If this test is supposed to test compiler behavior, then it should rather go into the LIT tests suite, which focuses on that (`tests/compiler`, particulary `tests/compiler/sscp`).

tests/sycl/usm.cpp
```diff
@@ -444,28 +529,82 @@ BOOST_AUTO_TEST_CASE(allocation_zero_bytes) {
...
+  int *aligned_shared_ptr = shared_support ?
+      sycl::aligned_alloc_shared<int>(sizeof(int), 0, q) : nullptr;
   if (aligned_shared_ptr)
     sycl::free(aligned_shared_ptr, q);
 }
 
+BOOST_AUTO_TEST_CASE(private_address_store_to_global) {
+  sycl::queue q{sycl::property::queue::in_order{}};
+
+  // We just check the kernel compiles, there is no invalid
```
illuhad: LIT test case?

tests/sycl/usm.cpp
```diff
@@ -444,28 +529,82 @@ BOOST_AUTO_TEST_CASE(allocation_zero_bytes) {
...
+       int arr;
+       *ptr = &arr;
+     });
+   }).wait();
+
+  sycl::free(ptr, q);
+}
+
+BOOST_AUTO_TEST_CASE(local_address_store_to_global) {
+  sycl::queue q{sycl::property::queue::in_order{}};
```
illuhad: LIT test case?

tests/sycl/usm.cpp
```diff
@@ -444,28 +529,82 @@ BOOST_AUTO_TEST_CASE(allocation_zero_bytes) {
...
+  q.submit([&](sycl::handler &cgh) {
+     auto scratch = sycl::local_accessor<int, 1>{1, cgh};
+     cgh.single_task([=]() { *ptr = &scratch[0]; });
+   }).wait();
+
+  sycl::free(ptr, q);
+}
+
+BOOST_AUTO_TEST_CASE(global_address_store_to_global) {
+  sycl::queue q{sycl::property::queue::in_order{}};
```
illuhad: LIT test case?

tests/sycl/extensions.cpp
EwanC: Woops, this was an artifact from when I tried support this test with device USM before realizing the extension doc said "shared USM", will revert.

tests/sycl/accessor.cpp
EwanC: I don't think this test stressed anything in particular about the implementation, I just kept it around because I saw it was a gap. I didn't know this feature was depreciated so will just remove this test.

tests/sycl/kernel.cpp
EwanC: I've not given much attention to the lit tests but will look into them and port the tests you've highlighted over.

Although for the second tests in this file I did experience different behavior on different Vulkan devices so being able to run it was useful, but regardless it's the kernel code that's the interesting bit of the test so is more suitable for LIT.

illuhad: LIT tests are also executed, e.g. https://github.com/AdaptiveCpp/AdaptiveCpp/blob/2e93a5a84ffd1022f3fbd6ec0ab7783417acda92/tests/compiler/sscp/structs.cpp#L2
We have a JIT compiler, so generally we always want to run our test executables :)

tests/sycl/usm.cpp
EwanC: done

tests/sycl/usm.cpp
EwanC: done

tests/sycl/usm.cpp
EwanC: done

tests/sycl/accessor.cpp
EwanC: removed

tests/sycl/extensions.cpp
EwanC: reverted

tests/sycl/kernel.cpp
EwanC: Sounds good, I've updated to that now and also updated the LIT tests to prefer device USM over shared USM.

tests/compiler/sscp/structs.cpp
EwanC, line 1: @resetius The CI is showing up a CI fail on this test change for the metal backend due to an empty log, so I'd guess the test crashed at some point. Do you have any idea why that could be happening? To move forward I could create two shared/device USM versions of the test disable the device USM version on Metal. Thinking right now to run `acpp-info` as part of lit.cfg to workout the backend and any capabilities like shared/device USM then use that to set some LIT UNSUPPORTED/REQUIRED conditions on the test.

EwanC: Apologies, this was a bug in the PR that I've managed to figure out and push a fix for. Sorry for the noise.

tests/sycl/kernel.cpp
illuhad: Thanks, this looks good -- if you want your new LIT test to be also tested for Vulkan JIT, we'd need to also enable llvmpipe testing in the github CI runner for the SSCP LIT workflow. Otherwise it's only going to be tested for OpenCL/OpenMP.

EwanC: Yup, my plan is to get this PR merged, then rebase the Vulkan PR on-top and add the LIT checks to the workflows.

Now that I think about it, that might actually be quite tricky as I'll need to find a way to skip the LIT tests that don't pass yet, e.g. `// UNSUPPORTED: vulkan` based on a config feature set in the lit cfg. However, it's really the `get_queue` device selector which picks at runtime the device used and if you have multiple gpu devices then it's not clear that vulkan would get chosen over a HIP/CUDA/Level-Zero GPU device. 

My ideas so far is that if the lit cfg detects parses the backends set in the visibility mask env var then those could be reported as configs that are checked in the tests and can be used to mark as unsupported.

Now I write this out, rather than making that lit cfg change in the vulkan PR which is already hug it might be worth adding this test complexity to this PR where it will be easier to review in a smaller context.

illuhad: Okay -- sounds good. Then I'll wait for you to add the LIT infrastructure here before merging.

EwanC: cool, I've pushed that `lit.cfg.py` change now. I've not included the `// UNSUPPORTED: vk || vulkan` line in the relevant tests but checked that works and will add it in the Vulkan PR when I rebase on this.

illuhad (approved):
Thanks!

PR #2029, opened 2026-04-04
Add runtime install destination for OpenCL and ZE backends

blinkfrog:
This PR adds `RUNTIME DESTINATION bin/hipSYCL` to the install rules for `rt-backend-ocl` and `rt-backend-ze`.

These targets already specify `LIBRARY` and `ARCHIVE` destinations, which is sufficient on Linux, but on Windows the DLL is treated as a runtime artifact. Without `RUNTIME` destination, backend DLL is built but not installed.

This makes the install rules consistent with the other runtime backends which include `RUNTIME`.


illuhad (approved):
Thanks!

PR #2030, opened 2026-04-04
Use bracketed install(CODE) for LLVMSpirvTranslator install

blinkfrog:
This changes the `LLVMSpirvTranslator` install step from a quoted `install(CODE "...")` string to a bracketed `install(CODE [[ ... ]])` block.

In my testing on Windows, the original quoted form did not install `llvm-spirv.exe`, while the bracketed form does.

I think that the bracketed form should work just fine on Linux, so we don't need to add `WIN32` guard.


illuhad (approved):
Thanks!

PR #2033, opened 2026-04-07
Fix narrowing conversion in __acpp_rootn (host and hiplike backends)

### LINKED ISSUE #2032: sycl/libkernel/host/builtins.hpp: narrowing conversion in __acpp_rootn (hard error on Clang by default)
### Problem

`__acpp_rootn` at `include/hipSYCL/sycl/libkernel/host/builtins.hpp:305-307` uses brace-initialization `T{1}/T{y}`:

```cpp
template<class T>
HIPSYCL_BUILTIN T __acpp_rootn(T x, int y) noexcept {
  return std::pow(x, T{1}/T{y});
}
```

When instantiated with `T=float` and `y` an `int`, `T{y}` is a narrowing conversion (int → float can lose precision for values outside `[-2^24, 2^24]`). Clang rejects this as a **hard error by default** — `-Wc++11-narrowing` is promoted to error for non-constant-expression narrowing.

### Minimal reproducer

```cpp
// test.cpp
#include <hipSYCL/sycl/libkernel/host/builtins.hpp>

int main() {
    float x = 27.0f;
    int y = 3;
    float result = hipsycl::sycl::detail::host_builtins::__acpp_rootn(x, y);
    return static_cast<int>(result);
}
```

```console
$ clang++ --target=x86_64-pc-windows-msvc -c -std=c++17 -I /path/to/AdaptiveCpp/include test.cpp

In file included from test.cpp:2:
.../hipSYCL/sycl/libkernel/host/builtins.hpp:306:29: error: non-constant-expression
    cannot be narrowed from type 'int' to 'float' in initializer list [-Wc++11-narrowing]
  306 |   return std::pow(x, T{1}/T{y});
      |                             ^
test.cpp:6:58: note: in instantiation of function template specialization
    'hipsycl::sycl::detail::host_builtins::__acpp_rootn<float>' requested here
    6 |     float result = hipsycl::sycl::detail::host_builtins::__acpp_rootn(x, y);
      |                                                          ^
1 error generated.
```

No `-Werror`, `-Wall`, or any other flag is required. Reproduced against `develop` @ `75d56089` with Clang 21.

### Proposed fix

```diff
 template<class T>
 HIPSYCL_BUILTIN T __acpp_rootn(T x, int y) noexcept {
-  return std::pow(x, T{1}/T{y});
+  return std::pow(x, T(1)/static_cast<T>(y));
 }
```

Functional-cast `T(1)` and explicit `static_cast<T>(y)` permit the int-to-float conversion with no behavioral change.

### Why CI likely doesn't trip this

`__acpp_rootn` is a `HIPSYCL_BUILTIN` — it's only instantiated when a SYCL kernel actually calls `sycl::rootn` (or its C++ math counterpart) with `(float, int)` arguments. If no test kernel does so, the template body is never checked and the narrowing diagnostic never fires. Might be worth adding a small test that exercises `rootn<float>` so a future change doesn't reintroduce this failure mode.

### Environment

- AdaptiveCpp `develop` @ `75d56089`
- Clang 21
- Target: `x86_64-pc-windows-msvc`
- OS: Windows 11

daniel-jasinski:
Fixes #2032.

## Changes

- **`include/hipSYCL/sycl/libkernel/host/builtins.hpp`**: replace `T{y}`
  with `static_cast<T>(y)` in `__acpp_rootn`, keeping `T{1}` for the
  literal as you suggested.
- **`include/hipSYCL/sycl/libkernel/generic/hiplike/builtins.hpp`**:
  identical fix — the hiplike backend has the same construct. Although
  the narrowing diagnostic may not fire on the nvcc/hipcc paths, fixing
  both keeps the two backends consistent. Happy to drop this hunk if
  you'd rather leave the hiplike backend alone.
- **`tests/sycl/math.cpp`**: new `math_rootn` test case, inserted right
  after `math_genfloat_genint`. I noticed `rootn` is already touched
  indirectly from `math_genfloat_genint` (line 889 on develop), but a
  dedicated test makes regressions easier to diagnose, and the inputs
  are picked so that the expected results (2, 3, 5, 10, …) are visually
  checkable.

## Verification

Compiled the minimal reproducer from #2032 against this branch:

```console
$ clang++ --target=x86_64-pc-windows-msvc -c -std=c++17 \
    -I include issue2_rootn_narrowing.cpp
$ echo $?
0
```

Pre-fix, the same command on `origin/develop` fails with
`error: non-constant-expression cannot be narrowed from type 'int' to 'float'`
as described in the issue.

## Small observation (not blocking)

While preparing this, I noticed that the existing `math_genfloat_genint`
test already invokes `s::rootn` at line 889, yet the narrowing diagnostic
evidently slipped through. I couldn't determine from
`.github/workflows/windows-acppllvm.yml` why the Windows CI matrix
doesn't trip it on this existing call. Flagging in case it points to
something worth tightening elsewhere — no change from me, just an
observation.

### DIFF
--- a/tests/sycl/math.cpp
+++ b/tests/sycl/math.cpp
@@ -966,9 +966,16 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(math_rootn, T, math_test_genfloats) {
     auto outputs = out.get_host_access();
 
     for(int c = 0; c < std::max(D,1); ++c) {
+      // Reference is computed in double so the comparison happens at double
+      // precision. boost::unit_test::tolerance is type-keyed: tolerance(0.0001)
+      // registers a tolerance for `double` only, so a `float == float`
+      // comparison falls back to exact equality and any backend that returns
+      // a result differing by even one ULP from std::pow would fail. This
+      // matches the existing math_genfloat_genint pattern (line 906) and the
+      // DEFINE_*_MATH_TEST macros below.
       BOOST_TEST(comp(outputs[0], c) ==
-                 std::pow(comp(base_inputs[0], c),
-                          DT{1} / static_cast<DT>(comp(exp_inputs[0], c))));
+                 std::pow(static_cast<double>(comp(base_inputs[0], c)),
+                          1.0 / static_cast<double>(comp(exp_inputs[0], c))));
     }
   }
 }
--- a/tests/sycl/math.cpp
+++ b/tests/sycl/math.cpp
@@ -907,6 +907,72 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(math_genfloat_genint, T,
   }
 }
 
+// Focused test for rootn: positive bases with integer roots so the expected
+// results are exactly representable and visually verifiable (8^(1/3)=2,
+// 81^(1/4)=3, 125^(1/3)=5, 100^(1/2)=10, ...). Separate from
+// math_genfloat_genint so a rootn regression is immediately identifiable.
+BOOST_TEST_DECORATOR(*boost::unit_test::tolerance(0.0001))
+BOOST_AUTO_TEST_CASE_TEMPLATE(math_rootn, T, math_test_genfloats) {
+
+  constexpr int D = vector_length_v<T>;
+  using DT = vector_elem_t<T>;
+
+  namespace s = sycl;
+
+  using IntType = s::detail::builtin_input_intlike_t<T>;
+
+  s::queue queue;
+  if constexpr(std::is_same_v<DT, double>) {
+    if (!queue.get_device().has(sycl::aspect::fp64)) {
+      BOOST_TEST_MESSAGE("Skipping test for double since device has no fp64 support");
+      return;
+    }
+  }
+
+  // build inputs and allocate outputs
+
+  s::buffer<T> base_in{{1}};
+  s::buffer<IntType> exp_in{{1}};
+  s::buffer<T> out{{1}};
+  {
+    auto base_inputs = base_in.get_host_access();
+    auto exp_inputs = exp_in.get_host_access();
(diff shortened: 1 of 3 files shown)


tests/sycl/math.cpp
```diff
@@ -907,6 +907,72 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(math_genfloat_genint, T,
...
+
+  // check results
+
+  {
+    auto base_inputs = base_in.get_host_access();
+    auto exp_inputs = exp_in.get_host_access();
+    auto outputs = out.get_host_access();
+
+    for(int c = 0; c < std::max(D,1); ++c) {
+      BOOST_TEST(comp(outputs[0], c) ==
```
illuhad, line 976: Floating point comparisons...needs a tolerance since backends may have different precision for rootn than host. That's why tests currently fail.

daniel-jasinski: Thanks, you're right — and the cause was more specific than I expected. The test case already has `BOOST_TEST_DECORATOR(*tolerance(0.0001))`, but `boost::unit_test::tolerance` is type-keyed: writing `tolerance(0.0001)` only registers a tolerance for `double`, so when `DT` is `float` the `BOOST_TEST(float == float)` comparison can't find a tolerance entry and falls back silently to exact equality. The existing `math_genfloat_genint` test (and the `DEFINE_*_MATH_TEST` macros further down) work around this implicitly because their reference computation goes through `1./int` or `static_cast<double>(...)`, so the RHS is always `double` and the comparison happens at double precision.

Pushed 226e724 — casts both `std::pow` operands to `double` in the reference, matching the line-906 idiom. Should make the float instantiations stop failing.

illuhad (approved):
Thanks!

PR #2034, opened 2026-04-07
Use _WIN32 and lowercase <windows.h> in omp_queue.cpp

### LINKED ISSUE #2031: runtime/omp/omp_queue.cpp: preprocessor checks use WIN32 (not a predefined macro) instead of _WIN32
### Problem

`src/runtime/omp/omp_queue.cpp` guards its Windows-vs-POSIX code paths with `#ifndef WIN32` (three occurrences on `develop` @ `75d56089`: lines 37, 43, and 186). However, `WIN32` (no leading underscore) is **not** a predefined compiler macro on any Windows-targeting C++ compiler — the predefined macro is [`_WIN32`](https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros) with a leading underscore.

When the file is compiled by `clang++` targeting `x86_64-pc-windows-msvc` without an external `-DWIN32`, the preprocessor takes the POSIX branch, which `#include`s `<unistd.h>` and fails against the MSVC C runtime.

Line 46 also uses `<Windows.h>` with uppercase W. NTFS is case-insensitive so this happens to work on native Windows builds, but it breaks on case-sensitive filesystems (WSL cross-compile, case-sensitive mounts).

### Preprocessor probe

```console
$ clang++ --target=x86_64-pc-windows-msvc -E -dM -x c++ /dev/null | grep -iE '^#define _?WIN(32|64)'
#define _WIN32 1
#define _WIN64 1
```

Only `_WIN32` and `_WIN64` are defined. No `WIN32`. Same result with `clang-cl`.

### Minimal reproducer

```cpp
// test.cpp
#ifndef WIN32
#include <unistd.h>
#else
#include <windows.h>
#endif

int main() { return 0; }
```

```console
$ clang++ --target=x86_64-pc-windows-msvc -c -std=c++17 test.cpp
test.cpp:2:10: fatal error: 'unistd.h' file not found
    2 | #include <unistd.h>
      |          ^~~~~~~~~~
1 error generated.
```

No special flags needed. Reproduced against stock `src/runtime/omp/omp_queue.cpp` from `develop` @ `75d56089`: fails with `fatal error: 'unistd.h' file not found` at line 44, plus a spurious `#warning` at line 39 for the same reason (the `#ifndef WIN32` → `#ifndef _OPENMP` chain takes the wrong branch even when `_OPENMP` is in fact defined).

### Proposed fix

```diff
--- a/src/runtime/omp/omp_queue.cpp
+++ b/src/runtime/omp/omp_queue.cpp
@@ -34,16 +34,16 @@
 #include "hipSYCL/runtime/omp/omp_code_object.hpp"

-#ifndef WIN32 // MSVC might not have #warning?
+#if !defined(WIN32) && !defined(_WIN32) // MSVC might not have #warning?
 #ifndef _OPENMP
 #warning Building omp backend with JIT support, but OpenMP parallelization is not available - kernels will run sequentially! This points to an issue in the build system.
 #endif
 #endif

-#ifndef WIN32
+#if !defined(WIN32) && !defined(_WIN32)
 #include <unistd.h>
 #else
-#include <Windows.h>
+#include <windows.h>
 #endif

@@ -183,7 +183,7 @@
 std::size_t get_page_size() {
-#ifndef WIN32
+#if !defined(WIN32) && !defined(_WIN32)
   return static_cast<std::size_t>(sysconf(_SC_PAGESIZE));
 #else
   SYSTEM_INFO si;
```

I've kept `WIN32` alongside `_WIN32` to preserve compatibility with any build that defines `WIN32` externally.

### Question about CI

`.github/workflows/windows-acppllvm.yml` clearly builds this file successfully, which suggests *something* in the CI environment defines `WIN32`. I couldn't determine what from reading the workflow — possibly the `msvc-dev-cmd` action injects it via `CL`, or an LLVM-side compile flag does. If there's a preferred CMake-level fix (e.g., `add_compile_definitions(WIN32)` for Windows targets) I'm happy to submit that instead of the source-level change.

### Environment

- AdaptiveCpp `develop` @ `75d56089`
- Clang 21 (built as `ACPP_LLVM_COMPONENT=ON` on Windows per the CI workflow)
- Target: `x86_64-pc-windows-msvc`
- OS: Windows 11

daniel-jasinski:
Fixes #2031.

## Changes

Replace `WIN32` (no underscore, not a predefined compiler macro) with the
official `_WIN32` throughout `src/runtime/omp/omp_queue.cpp`, and switch
`<Windows.h>` to lowercase `<windows.h>` so case-sensitive filesystems
find the header.

Four edits in one file:

- Line 37: `#ifndef WIN32` → `#ifndef _WIN32` (the JIT/OpenMP `#warning` guard)
- Line 43: `#ifndef WIN32` → `#ifndef _WIN32` (the `<unistd.h>` vs `<windows.h>` guard)
- Line 46: `#include <Windows.h>` → `#include <windows.h>`
- Line 187: `#ifndef WIN32` → `#ifndef _WIN32` (inside `get_page_size()`)

## Verification

**Pre-fix** (clean `origin/develop` @ `21a8df24`), the minimal pattern
reproducer from #2031 fails as expected:

```console
$ clang++ --target=x86_64-pc-windows-msvc -c -std=c++17 test.cpp
test.cpp:2:10: fatal error: 'unistd.h' file not found
    2 | #include <unistd.h>
      |          ^~~~~~~~~~
1 error generated.
```

**Post-fix**, compiling the actual patched `omp_queue.cpp` with the
CMake-equivalent compile flags succeeds with zero errors:

```console
$ clang++ --target=x86_64-pc-windows-msvc -c -std=c++17 \
    -DHIPSYCL_WITH_SSCP_COMPILER -DHIPSYCL_DEBUG_LEVEL=1 \
    -DNOMINMAX -DWIN32_LEAN_AND_MEAN -D_USE_MATH_DEFINES \
    -I include -I src/runtime -I boost/include \
    src/runtime/omp/omp_queue.cpp
$ echo $?
0
```

## Not in scope

- Investigating why the Windows CI path doesn't trip this with the
  current `WIN32` check. You mentioned in the issue that you don't know
  either. Not something I'd want to expand the scope of this PR to
  chase — flagging it as an open question you may want to look into
  independently.
- Auditing other source files for the same `WIN32` vs `_WIN32` pattern.
  A quick `grep -rn 'ifn\?def WIN32' src/` suggests `omp_queue.cpp` is
  the only offender on `develop`, but I haven't verified exhaustively.
  Happy to extend the PR if you find more.

### DIFF
--- a/src/runtime/omp/omp_queue.cpp
+++ b/src/runtime/omp/omp_queue.cpp
@@ -34,16 +34,16 @@
 #include "hipSYCL/runtime/adaptivity_engine.hpp"
 #include "hipSYCL/runtime/omp/omp_code_object.hpp"
 
-#ifndef WIN32 // MSVC might not have #warning?
+#ifndef _WIN32 // MSVC might not have #warning?
 #ifndef _OPENMP
 #warning Building omp backend with JIT support, but OpenMP parallelization is not available - kernels will run sequentially! This points to an issue in the build system.
 #endif
 #endif
 
-#ifndef WIN32
+#ifndef _WIN32
 #include <unistd.h>
 #else
-#include <Windows.h>
+#include <windows.h>
 #endif
 #endif
 
@@ -184,7 +184,7 @@ private:
 #ifdef HIPSYCL_WITH_SSCP_COMPILER
 
 std::size_t get_page_size() {
-#ifndef WIN32
+#ifndef _WIN32
   return static_cast<std::size_t>(sysconf(_SC_PAGESIZE));
 #else
   SYSTEM_INFO si;


illuhad:
> Auditing other source files for the same WIN32 vs _WIN32 pattern.
A quick grep -rn 'ifn\?def WIN32' src/ suggests omp_queue.cpp is
the only offender on develop, but I haven't verified exhaustively.
Happy to extend the PR if you find more.

I have also checked this. Apart from your find, seems we only use `WIN32` in cmake where it's fine.

illuhad (approved):
Thanks!

PR #2037, opened 2026-04-09
[Windows] Honor ACPP_APPDB_DIR and use per-executable cache dir

### LINKED ISSUE #2035: [Windows] ACPP_APPDB_DIR is not honored; default JIT cache path is cwd-relative
**Version:** AdaptiveCpp 25.10.0 (verified at `75d56089`, also confirmed unchanged on current `origin/develop`).

## Summary

On Windows, `persistent_storage::persistent_storage()` takes a separate code path that does not call `try_retrieve_settings_variable("appdb_dir", ...)`, so `ACPP_APPDB_DIR` is silently ignored on Windows even though `doc/env_variables.md` documents it without a platform caveat. The Windows default is also `<cwd>/.acpp/apps/global/`, so each working directory the user runs the same executable from gets a disjoint cache.

## Code

`src/common/filesystem.cpp:208-249`:

```cpp
persistent_storage::persistent_storage() {
#ifndef _WIN32
  // ... XDG_DATA_HOME / HOME logic ...
  if(!settings::try_retrieve_settings_variable("appdb_dir", _base_dir)) {
    // ... fallback ...
  }
  std::string app_path = get_this_executable_path();
  _this_app_dir = generate_app_dir(app_path);
#else
  _base_dir     = (fs::current_path() / ".acpp").string();
  _this_app_dir = (fs::path{_base_dir} / "apps" / "global").string();
#endif
  _jit_cache_dir = (fs::path{_this_app_dir} / "jit-cache").string();
  ...
}
```

The Windows branch never calls `try_retrieve_settings_variable` and never calls `generate_app_dir(get_this_executable_path())`, even though `get_this_executable_path()` already has a working Windows implementation at lines 94-125 (using `GetModuleFileNameA`).

## Doc that promises the env var works

`doc/env_variables.md:30`:

> `ACPP_APPDB_DIR`: By default, AdaptiveCpp stores its application db (which in particular includes the per-app JIT cache) in `$HOME/.acpp`. This environment variable can be used to override the location.

## Practical effect

1. `ACPP_APPDB_DIR=C:\shared\acpp` has no observable effect on Windows. There is no warning. The same applies to the `acpp-config.cfg` / `acpp-config-<app-name>.cfg` mechanism added in #1848 — both env-var and config-file routes share the same `try_retrieve_settings_variable` entry point, which is unreachable on Windows.
2. Running the same executable from two different working directories produces two disjoint cache trees with no way to share, so users pay the full JIT warmup every time the cwd changes. The adaptivity engine's argument-value statistics in `app.v5.db` also reset.
3. Two different executables run from the same directory share the hardcoded `apps/global/` subdirectory and the same `app.v5.db`, since the per-executable hashing in `generate_app_dir()` is bypassed.

## Suggested direction

Pull the `appdb_dir` lookup out of the `#if` so both platforms honor it, and on Windows fall back to `%LOCALAPPDATA%\acpp` (via `std::getenv("LOCALAPPDATA")`) which is the standard per-user app-data location. The per-app subdirectory should also use the existing `generate_app_dir(get_this_executable_path())` path on Windows, so the same executable run from different cwds shares one cache and different executables get distinct trees — matching Linux behavior. No new Windows-specific code is required; the existing helpers already work.

Happy to send a PR if this direction looks acceptable.

daniel-jasinski:
Fixes #2035.

## Summary

On Windows, `persistent_storage::persistent_storage()` in `src/common/filesystem.cpp` took a separate code path that hardcoded `_base_dir` to `<cwd>/.acpp` and `_this_app_dir` to `<base>/apps/global`, silently ignoring `ACPP_APPDB_DIR` (and the `acpp-config.cfg` / `acpp-config-<app-name>.cfg` routes that share the same `try_retrieve_settings_variable` entry point).

As discussed in #2035, the Windows arm of `get_this_executable_path()` (`GetModuleFileNameA` loop, `filesystem.cpp:105–115`) already returns the same shape Linux returns, and `generate_app_dir()` / `generate_appdb_path()` (`filesystem.cpp:264–284`) are platform-independent — they only touch `fs::path` and `stable_running_hash` (raw bytes). Both already handle the empty-`app_path` case via the `"global"` fallback. So the Linux body works on Windows unchanged once a Windows fallback for `_base_dir` is provided.

## Changes

- Hoist `try_retrieve_settings_variable("appdb_dir", ...)` and the `get_this_executable_path()` / `generate_app_dir()` / `generate_appdb_path()` calls out of the `#ifndef _WIN32` so both platforms honor the env var and config files.
- Add a Windows arm to the existing `get_home` lambda that points at `%LOCALAPPDATA%\acpp` (the standard per-user app-data location), with the existing `cwd/.acpp` final fallback if `LOCALAPPDATA` is unset (e.g. service accounts).
- Drop the now-redundant Windows-only branches; the rest of the constructor is uniform across platforms.
- Update `doc/env_variables.md` to mention the Windows default location.

No new helpers, no new headers, `<windows.h>` / `<pwd.h>` stay where they were at the top of the file.

## Effects on Windows

- `ACPP_APPDB_DIR` and `acpp-config.cfg` / `acpp-config-<app-name>.cfg` start working.
- Default cache moves to `%LOCALAPPDATA%\acpp` instead of `<cwd>\.acpp`, so the same executable shares one cache across working directories.
- `_this_app_dir` becomes `<base>/apps/<exe-name>-<hash>`, so different executables get distinct app directories and `app.v5.db` files instead of clobbering a shared `apps/global/`.
- The adaptivity engine's per-app argument-value statistics in `app.v5.db` are no longer reset when the cwd changes, and no longer cross-contaminated between unrelated executables run from the same directory.

No behavioral change on Linux/macOS.

## Test plan

- [x] Builds on Windows (MSVC ABI, Clang 19)
- [x] `ACPP_APPDB_DIR=C:\some\path` is observed on Windows runs
- [x] Default cache lands at `%LOCALAPPDATA%\acpp\apps\<exe>-<hash>\jit-cache` on Windows
- [x] Same executable run from two different cwds reuses one cache
- [x] Two different executables run from the same cwd get distinct app directories
- [x] No regression on Linux (default still `$HOME/.acpp` / `$XDG_DATA_HOME/acpp`)

Verification evidence for the four Windows items in [this comment](https://github.com/AdaptiveCpp/AdaptiveCpp/pull/2037#issuecomment-4266417539). Linux confirmation from @illuhad above.


illuhad:
I can confirm that it still behaves as expected on Linux with this patch.
The other points from your test plan you would need to check yourself because I don't have a Windows test system.

daniel-jasinski:
Ran the four outstanding Windows test-plan items against a locally rebuilt `acpp-common.dll` from this branch (patched `filesystem.cpp` at 0d3564d1, otherwise identical AdaptiveCpp v19.1.4 install on Windows 11, MSVC ABI, Clang 19, CUDA sm_86 + OMP backends). Test harness is a trivial SYCL program that submits one `parallel_for` and exits, built as two byte-identical copies (`acpp-test1.exe`, `acpp-test2.exe`) so the paths differ only by filename.

### Results

```
=== T1: default, cwd = tests\, acpp-test1.exe ===
  [AdaptiveCpp Warning] kernel_cache: new binaries JIT-compiled   ← cold cache, expected
  result=42

=== T2: default, cwd = C:\tmp, acpp-test1.exe (same exe, different cwd) ===
  (no "new binaries JIT-compiled" warning)                         ← cache hit from T1
  result=42

=== T3: default, cwd = tests\, acpp-test2.exe (different exe, same cwd as T1) ===
  [AdaptiveCpp Warning] kernel_cache: new binaries JIT-compiled   ← cold cache (distinct app dir)
  result=42

=== T4: ACPP_APPDB_DIR=C:\acpp-custom-test, cwd = C:\tmp, acpp-test1.exe ===
  [AdaptiveCpp Warning] kernel_cache: new binaries JIT-compiled   ← cold cache (distinct base dir)
  result=42
```

### Filesystem state after the four runs

```
--- %LOCALAPPDATA%\acpp tree ---
C:\Users\Daniel\AppData\Local\acpp\apps
C:\Users\Daniel\AppData\Local\acpp\apps\acpp-test1.exe-18112716603755085612
C:\Users\Daniel\AppData\Local\acpp\apps\acpp-test1.exe-18112716603755085612\app.v5.db
C:\Users\Daniel\AppData\Local\acpp\apps\acpp-test1.exe-18112716603755085612\jit-cache
C:\Users\Daniel\AppData\Local\acpp\apps\acpp-test1.exe-18112716603755085612\jit-cache\3153096841371102686.6205151134146693852.jit
C:\Users\Daniel\AppData\Local\acpp\apps\acpp-test2.exe-16906605800620326874
C:\Users\Daniel\AppData\Local\acpp\apps\acpp-test2.exe-16906605800620326874\app.v5.db
C:\Users\Daniel\AppData\Local\acpp\apps\acpp-test2.exe-16906605800620326874\jit-cache
C:\Users\Daniel\AppData\Local\acpp\apps\acpp-test2.exe-16906605800620326874\jit-cache\3153096841371102686.6205151134146693852.jit

--- C:\acpp-custom-test tree ---
C:\acpp-custom-test\apps
C:\acpp-custom-test\apps\acpp-test1.exe-18112716603755085612
C:\acpp-custom-test\apps\acpp-test1.exe-18112716603755085612\app.v5.db
C:\acpp-custom-test\apps\acpp-test1.exe-18112716603755085612\jit-cache
C:\acpp-custom-test\apps\acpp-test1.exe-18112716603755085612\jit-cache\3153096841371102686.6205151134146693852.jit
```

### Verification against the test plan

- **`ACPP_APPDB_DIR=C:\some\path` honored on Windows** — T4: the full tree lands under `C:\acpp-custom-test\apps\...`, nothing under `%LOCALAPPDATA%\acpp` from that run.
- **Default lands at `%LOCALAPPDATA%\acpp\apps\<exe-name>-<hash>\jit-cache`** — T1: exactly `C:\Users\Daniel\AppData\Local\acpp\apps\acpp-test1.exe-18112716603755085612\jit-cache\`.
- **Same executable from two different cwds shares one cache** — T2 (run from `C:\tmp`) reused the `acpp-test1.exe-18112716603755085612` tree created by T1 (run from `tests\`); functionally confirmed by T2 *not* emitting the "new binaries JIT-compiled" warning.
- **Different executables from the same cwd get distinct app dirs** — T3 created a separate `acpp-test2.exe-16906605800620326874` directory (distinct hash, distinct `app.v5.db`, distinct `jit-cache`), and logged a cache-miss warning — the two executables' adaptivity stats do not collide.

The `.jit` filename (`3153096841371102686.6205151134146693852.jit`) is identical across all three cache directories because the kernels are identical — the segregation is purely at the directory level, as intended.

I'll check the "no regression on Linux" box on your behalf per your earlier confirmation. All six boxes now checked.

illuhad (approved):
Thanks!

daniel-jasinski:
Happy to help :)

PR #2038, opened 2026-04-09
Device compatibility for exception throwing code using assertions

EkremA:
Added compatibility for code possibly throwing exceptions on device. In LLVM C++ [thrown exceptions](https://llvm.org/docs/ExceptionHandling.html#llvm-code-generation) are handled with the functions `__cxa_throw`  and `__cxa_allocate_exception` in LLVM IR. However if we run such code on GPU devices, we may not have access to appropriate implementations of these exception functions. 

This PR suggests a compilation pass, which parses device IR for calls and invokes to `__cxa_throw` and replacing them with an assertion `__acpp_sscp_assert_fail`.  Then, the function  `__cxa_allocate_exception`  is removed altogether as any allocation of  exception structures would no longer be required.

Additionally, the fix of a typo in one of the CI workflow files is suggested.


include/hipSYCL/compiler/sscp/ExceptionToAssertionPass.hpp
```diff
@@ -0,0 +1,30 @@
...
+ * parallelism for CPUs and GPUs.
+ *
+ * Copyright The AdaptiveCpp Contributors
+ *
+ * AdaptiveCpp is released under the BSD 2-Clause "Simplified" License.
+ * See file LICENSE in the project root for full license details.
+ */
+// SPDX-License-Identifier: BSD-2-Clause
+#ifndef ACPP_REFLECTION_EXCEPTION_TO_ASSERTION_PASS_HPP
+#define ACPP_REFLECTION_EXCEPTION_TO_ASSERTION_PASS_HPP
```
illuhad, line 12: Why is it framed as part of reflection and in reflection subdirectory? I don't think it has anything to do with those. Probably it should go into `compiler/sscp`?

src/compiler/reflection/ExceptionToAssertionPass.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+namespace hipsycl {
+namespace compiler {
+
+
+llvm::PreservedAnalyses ExceptionToAssertionPass::run(llvm::Module &M, llvm::ModuleAnalysisManager &AM) {
+
+  static const char* CXAThrow = "__cxa_throw";
+  static const char* ACPPSSCPAssertFail = "__acpp_sscp_assert_fail";
+  //static const char* GlibcxxAssertFailBuiltinName = "__acpp_sscp_glibcxx_assert_fail";
```
illuhad: I don't think we need this -> remove

src/compiler/AdaptiveCppLlvmPasses.cpp
```diff
@@ -37,6 +37,11 @@
 #include "hipSYCL/compiler/reflection/FunctionNameExtractionPass.hpp"
 #endif
 
+#ifdef HIPSYCL_WITH_DEVICE_EXCEPTIONS
+#include "hipSYCL/compiler/reflection/ExceptionToAssertionPass.hpp"
+// #include "hipSYCL/compiler/reflection/FunctionNameExtractionPass.hpp"
+#endif
```
illuhad: Why do we need this include here? The pass is only used inside `TargetSeparationPass`, right?

src/compiler/CMakeLists.txt
```diff
@@ -12,8 +12,8 @@ else()
   add_compile_definitions(ACPP_LLVM_COMPONENT)
 endif()
 
-if(NOT DEFINED HIPSYCL_DEBUG_LEVEL)
-  message(FATAL_ERROR "HIPSYCL_DEBUG_LEVEL was not defined by the top-level CMakeLists.txt.")
+if(NOT HIPSYCL_DEBUG_LEVEL)
+  set(HIPSYCL_DEBUG_LEVEL 2 CACHE INTEGER)
```
illuhad: This seems like a change for debugging purposes - please remove from this PR!

src/compiler/CMakeLists.txt
```diff
@@ -286,13 +286,29 @@ else()
...
+    reflection/FunctionNameExtractionPass.cpp)
+  add_definitions(-DHIPSYCL_WITH_REFLECTION_BUILTINS)
+else()
+  set(REFLECTION_BUILTINS "")
+endif()
+
+
+set(DEVICE_EXCEPTIONS
+  reflection/ExceptionToAssertionPass.cpp)
+add_definitions(-DHIPSYCL_WITH_DEVICE_EXCEPTIONS)
```
illuhad: Just add it to the existing `SSCP_COMPILER` sources above. No need to create something new and even have a macro for it.

src/compiler/reflection/ExceptionToAssertionPass.cpp
```diff
@@ -0,0 +1,249 @@
...
+    }
+  }
+  
+  // create calls to __acpp_sscp_assert_fail and terminate with unreachable instructions in place of __cxa_throw invokes
+  llvm::SmallVector<llvm::UnreachableInst*> UnreachableForCXAInvokes;
+  for (int i=0; i<cxaThrowInvokes.size(); i++) {
+    // construct device assertion signature
+    const char *assertionStr = "Exception in Device Code";
+    const char *fileStr = M.getSourceFileName().c_str();
+    int lineNumber = -42;
```
illuhad: Better use something like `numeric_limits<int>::max()` in case the debug location cannot be retrieved.

src/compiler/reflection/ExceptionToAssertionPass.cpp
```diff
@@ -0,0 +1,249 @@
...
+llvm::PreservedAnalyses ExceptionToAssertionPass::run(llvm::Module &M, llvm::ModuleAnalysisManager &AM) {
+
+  static const char* CXAThrow = "__cxa_throw";
+  static const char* ACPPSSCPAssertFail = "__acpp_sscp_assert_fail";
+  //static const char* GlibcxxAssertFailBuiltinName = "__acpp_sscp_glibcxx_assert_fail";
+  static const char* CXAAllocExc = "__cxa_allocate_exception";
+
+  // declare __acpp_sscp_assert_fail
+  // detect if C++ throws occur in source code
+  llvm::Function* oldcxaThrow = nullptr;
```
illuhad: IIRC LLVM coding style typically capitalizes variable name initials.

src/compiler/reflection/ExceptionToAssertionPass.cpp
```diff
@@ -0,0 +1,249 @@
...
+  for(auto& F : M)
+    if(F.getName().contains(CXAThrow)) {
```
illuhad: Do we need `contains` or can we just compare for equality?

src/compiler/reflection/ExceptionToAssertionPass.cpp
```diff
@@ -0,0 +1,249 @@
...
+      oldcxaThrow = &F;
+    }
+  if (!oldcxaThrow) {
+    llvm::errs() << "Function " << CXAThrow << " not found. Assuming there is no use of exceptions in source code.\n";
```
illuhad: Error messages should be prefixed with the pass that produces them. E.g. `llvm::errs() << "ExceptionToAssertionPas: ..."`

Also, why would we even error in this case? Isn't it expected that in most cases there's no exception in device code?

src/compiler/reflection/ExceptionToAssertionPass.cpp
```diff
@@ -0,0 +1,249 @@
...
+  //static const char* GlibcxxAssertFailBuiltinName = "__acpp_sscp_glibcxx_assert_fail";
+  static const char* CXAAllocExc = "__cxa_allocate_exception";
+
+  // declare __acpp_sscp_assert_fail
+  // detect if C++ throws occur in source code
+  llvm::Function* oldcxaThrow = nullptr;
+  for(auto& F : M)
+    if(F.getName().contains(CXAThrow)) {
+      oldcxaThrow = &F;
+    }
```
illuhad: That whole segment I think can be replaced with `M.getFunction(CXAThrow)`

src/compiler/reflection/ExceptionToAssertionPass.cpp
```diff
@@ -0,0 +1,249 @@
...
+  }
+
+
+  // declare __acpp_sscp_assert_fail if not yet declared
+  llvm::Type* llvmCharType;
+  llvm::Function* ACPPSSCPAssertFailDeclaration;
+  #if LLVM_VERSION_MAJOR >= 16
+      llvmCharType = llvm::PointerType::get(M.getContext(), 0);
+  #else
+      llvmCharType = llvm::PointerType::get(llvm::Type::getInt8Ty(M.getContext()), 0);
```
illuhad: No indentation inside preprocessor if/else

src/compiler/sscp/ExceptionToAssertionPass.cpp
```diff
@@ -0,0 +1,249 @@
...
+  // declare __acpp_sscp_assert_fail if not yet declared
+  llvm::Type* llvmCharType;
+  llvm::Function* ACPPSSCPAssertFailDeclaration;
+  #if LLVM_VERSION_MAJOR >= 16
+      llvmCharType = llvm::PointerType::get(M.getContext(), 0);
+  #else
+      llvmCharType = llvm::PointerType::get(llvm::Type::getInt8Ty(M.getContext()), 0);
+  #endif
+  if(auto* F = M.getFunction(ACPPSSCPAssertFail))
+    ACPPSSCPAssertFailDeclaration=F;
```
illuhad, line 50: That `if` seems unnecessary.... why not `llvm::Function* ACPPSSCPAssertFailDeclaration = M.getFunction(...)` ?

illuhad (commented):
I've only read the beginning of the pass -- looks like this needs some general cleanups first :)
Let me know once it's ready to review.

illuhad:
Also, please add some LIT test cases for this feature :-)

EkremA:
Code has been slightly restructured. `getAssertFailBuiltin` is a static member function of `DeviceAssertPass` now. Added LIT tests which check for absence of function symbol `__cxa_throw` using the utility `acpp-hcf-tool`, see `hcfinspect.sh`. 

PR is ready for review.

src/compiler/sscp/ExceptionToAssertionPass.cpp
```diff
@@ -0,0 +1,209 @@
...
+      for (auto &Inst : BB) {
+        if(auto *Invoke = llvm::dyn_cast<llvm::InvokeInst>(&Inst)) {
+          if (Invoke->getCalledFunction()) {
+            if(Invoke->getCalledFunction()->getName() == CXAThrow) {
+                //count to the invoke instructions
+                cxaThrowInvokes.push_back(Invoke); 
+              }
+            }
+          }
+          if(auto *Call = llvm::dyn_cast<llvm::CallInst>(&Inst)) {
```
illuhad, line 53: This indentation is misleading. It looks like the `CallInst` check is inside the `InvokeInst` check (which would not make any sense), but I *think* it's actually outside.

src/compiler/sscp/ExceptionToAssertionPass.cpp
```diff
@@ -0,0 +1,209 @@
...
+    ACPPSSCPAssertFailArgs.push_back(builder.CreateGlobalStringPtr(functionName));
+  
+    auto *Call = llvm::CallInst::Create(DeviceAssertPass::getAssertFailBuiltin(M), ACPPSSCPAssertFailArgs);
+    Call->insertBefore(cxaThrowInvokes[i]);
+    auto *Unreachable = new llvm::UnreachableInst(M.getContext(), cxaThrowInvokes[i]); 
+    UnreachableForCXAInvokes.push_back(Unreachable);
+  }
+  // replace all uses of invoke @__cxa_throw with unreachable instruction and erase the invoke instruction
+  for (int i=0; i<cxaThrowInvokes.size(); i++) {
+    UnreachableForCXAInvokes[i]->replaceAllUsesWith(cxaThrowInvokes[i]);
```
illuhad, line 92: Isn't this the wrong way around? This looks like you are replacing the unreachables with the cxa throw invokes.

Also, it feels a little overcomplicated to store the unreachables in an array. Couldn't you just:
```c++
auto *Unreachable = new llvm::UnreachableInst(M.getContext(), cxaThrowInvokes[i]); 
cxaThrowInvokes[i]->replaceAllUsesWith(Unreachable);
```

src/compiler/sscp/ExceptionToAssertionPass.cpp
```diff
@@ -0,0 +1,209 @@
...
+    llvm::IRBuilder<> builder(cxaThrowCalls[i]);
+
+    llvm::SmallVector<llvm::Value*> ACPPSSCPAssertFailArgs;
+    ACPPSSCPAssertFailArgs.push_back(builder.CreateGlobalStringPtr(assertionStr));
+    ACPPSSCPAssertFailArgs.push_back(builder.CreateGlobalStringPtr(fileStr));
+    ACPPSSCPAssertFailArgs.push_back(builder.getInt32(lineNumber));
+    ACPPSSCPAssertFailArgs.push_back(builder.CreateGlobalStringPtr(functionName));
+  
+    auto *Call = llvm::CallInst::Create(DeviceAssertPass::getAssertFailBuiltin(M), ACPPSSCPAssertFailArgs);
+    Call->insertBefore(cxaThrowCalls[i]);
```
illuhad, line 118: This section is duplicated from the invoke case...can't you create a function with this code to share the code between both cases?

src/compiler/sscp/ExceptionToAssertionPass.cpp
```diff
@@ -0,0 +1,209 @@
...
+  }
+  // replace all uses of call @__cxa_throw with call to __acpp_sscp_assert_fail and clean up
+  for (int i=0; i<cxaThrowCalls.size(); i++) {
+    AssertForCXACalls[i]->replaceAllUsesWith(cxaThrowCalls[i]);
+    cxaThrowCalls[i]->eraseFromParent();
+  }
+
+
+
+
```
illuhad, line 129: Do we need all those empty lines? :D

src/compiler/sscp/ExceptionToAssertionPass.cpp
```diff
@@ -0,0 +1,209 @@
...
+              // retain amount of data to be allocated
+              // Assuming always first argument/llvm::Value is alloc size of the exception object AND that it is specified as i64
+              llvm::User::op_iterator cxaAllocExcIt;
+              llvm::SmallVector<llvm::Value*> cxaAllocExcArgs;
+              int i = 0;
+              for (cxaAllocExcIt=Call->arg_begin(); cxaAllocExcIt != Call->arg_end(); cxaAllocExcIt++) {
+                cxaAllocExcArgs.push_back(*cxaAllocExcIt);
+                i++;
+              }
+              cxaAllocExcCallsArgs.push_back(cxaAllocExcArgs);
```
illuhad, line 153: Instead of copying out all arguments, can't you just directly access the first argument with `callInst->getArgOperand(0)`?

src/compiler/sscp/ExceptionToAssertionPass.cpp
```diff
@@ -0,0 +1,209 @@
...
+              llvm::Value *allocSize;
+              if (auto *ConstInt = llvm::dyn_cast<llvm::ConstantInt>(cxaAllocExcArgs[0])) {
+                // Check if the type is i64
+                if (ConstInt->getType()->isIntegerTy(64)) {
+                  // Retrieve the zero-extended value and create array size value
+                  allocSize = llvm::ConstantInt::get(llvm::Type::getInt32Ty(M.getContext()), 
+                                                                  static_cast<int>(ConstInt->getZExtValue()));
+                }
+              }
+              auto* allocainstruction = new llvm::AllocaInst(cxaAllocExcArgs[0]->getType(), 0, allocSize, "exception_alloca", Call);
```
illuhad, line 167: This identifier name does not look like LLVM coding style to me.

src/compiler/sscp/ExceptionToAssertionPass.cpp
```diff
@@ -0,0 +1,209 @@
...
+          }
+        }
+      }
+    }
+  }
+
+  //remove all calls to __cxa_allocate_exception
+  for (auto *Call : cxaAllocExcCalls2) {
+      if(Call!=nullptr)
+      Call->eraseFromParent();
```
illuhad, line 196: Do we really need a separate iteration to erase? Can't we combine that with the RAUW step?

src/compiler/CMakeLists.txt
```diff
@@ -286,13 +287,25 @@ else()
...
 endif()
 
+if(WITH_REFLECTION_BUILTINS)
+  set(REFLECTION_BUILTINS
+    reflection/IntrospectStructPass.cpp
+    reflection/FunctionNameExtractionPass.cpp)
+  add_definitions(-DHIPSYCL_WITH_REFLECTION_BUILTINS)
+else()
+  set(REFLECTION_BUILTINS "")
+endif()
```
illuhad, line 297: This block seems to be copy-pasted from above....please remove!

src/compiler/CMakeLists.txt
```diff
@@ -286,13 +287,25 @@ else()
...
+
+
 set(COMPILER_LIBRARY_SOURCES
   AdaptiveCppLlvmPasses.cpp
   GlobalsPruningPass.cpp
   SMCPCompatPass.cpp
   ${SSCP_COMPILER}
   ${STDPAR_COMPILER}
   ${REFLECTION_BUILTINS}
+  ${DEVICE_EXCEPTIONS}
```
illuhad, line 308: No longer needed, right?

tests/compiler/sscp/kernel_stdoptional.cpp
```diff
@@ -0,0 +1,29 @@
...
+      std::optional<int> opt;
+      *data = opt.value();
+    }
+    catch(std::bad_optional_access& e){
+      *data = 424242;
+    }
+  });
+  q.wait();
+  std::cout << *data << std::endl;
+  //CHECK-NOT: CXATHROWHIT
```
illuhad, line 26: It would be good if we could also check for `ACPPASSERTFAILHIT` here, which I think you're not actually using anywhere?

PR #2043, opened 2026-04-11
[Metal] Move local memory ptr/size to global vars

resetius:
Small refactoring to simplify the Metal pipeline and prepare for upcoming address translation support needed for indirection structures. The address translation parameters will also be passed via globals, so moving these values to globals now helps keep the pipeline and calling convention simpler


illuhad (approved):
Thanks!

PR #2045, opened 2026-04-14
Remove wchar_size module flag during target separation

blinkfrog:
This removes the `wchar_size` module flag from device module during SSCP target separation.

As device IR is derived from host IR, it may inherit host-specific module flags. On Windows, this causes a later conflict involving `wchar_size` during IR->SPIR-V JIT compilation.

This implementation follows the earlier suggestion that host-specific flags likely should be erased. (https://github.com/AdaptiveCpp/AdaptiveCpp/discussions/2014)

This change is intentionally generic and is not guarded by `_WIN32` or `_MSC_VER`, since this appears to be a wider target-separation issue: host-specific module flags should not be preserved and should not leak into generic device IR.

Tested on Windows with OpenCL backend, where this resolves the conflict on my side.

If preferred, I can instead switch this to the backend-side workaround (in `LLVMToSpirv.cpp`) discussed earlier.

### DIFF
--- a/src/compiler/sscp/TargetSeparationPass.cpp
+++ b/src/compiler/sscp/TargetSeparationPass.cpp
@@ -216,6 +216,24 @@ void replaceInvalidMSABICharsInSymbolNames(llvm::Module &M) {
 }
 #endif
 
+void removeModuleFlag(llvm::Module& M, llvm::StringRef Name) {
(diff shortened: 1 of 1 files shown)


illuhad (approved):
Thanks!

PR #2047, opened 2026-04-17
Implement rounding_mode in sycl::vec::convert

ferdymercury:
Was marked as TODO
I made a very simple implementation.
At least it's better than silently ignoring the rounding parameter. Which can make users think that it's working fine but it's always doing a static cast.

Intel does a much more involved job:
https://github.com/intel/llvm/blob/sycl/sycl/include/sycl/detail/vector_convert.hpp

calling SPIRV and NVIDIA builtins but I think that's a more complicated story.

### DIFF
--- a/include/hipSYCL/sycl/libkernel/vec.hpp
+++ b/include/hipSYCL/sycl/libkernel/vec.hpp
@@ -339,7 +339,7 @@ public:
     vec<ConvertT, N> result;
 
     for(int i = 0; i < N; ++i) {
-// TODO handle float to float conversion. Here only float to short rounding is implemented
+// TODO handle float to float conversion. Here only float to integer rounding is implemented
       if constexpr (std::is_floating_point_v<T> &&
                        std::is_integral_v<ConvertT>) {
         if constexpr (RM == rounding_mode::rte || RM == rounding_mode::automatic) {
--- a/include/hipSYCL/sycl/libkernel/vec.hpp
+++ b/include/hipSYCL/sycl/libkernel/vec.hpp
@@ -339,8 +339,22 @@ public:
     vec<ConvertT, N> result;
 
     for(int i = 0; i < N; ++i) {
-      // TODO: Take rounding mode into account
-      result[i] = static_cast<ConvertT>(_data[i]);
+      if constexpr (std::is_floating_point_v<T> &&
+                       std::is_integral_v<ConvertT>) {
+        if constexpr (RM == rounding_mode::rte || RM == rounding_mode::automatic) {
+          result[i] = static_cast<ConvertT>(sycl::rint(_data[i]));          
+        } else if constexpr (RM == rounding_mode::rtz) {
+          result[i] = static_cast<ConvertT>(sycl::trunc(_data[i]));
+        } else if constexpr (RM == rounding_mode::rtp) {
+          result[i] = static_cast<ConvertT>(sycl::ceil(_data[i]));
+        } else if constexpr (RM == rounding_mode::rtn) {
+          result[i] = static_cast<ConvertT>(sycl::floor(_data[i]));
+        } else {
+          result[i] = static_cast<ConvertT>(_data[i]);
+        }
+      } else {
+        result[i] = static_cast<ConvertT>(_data[i]);
+      }
     }
 
     return result;
--- a/include/hipSYCL/sycl/libkernel/vec.hpp
+++ b/include/hipSYCL/sycl/libkernel/vec.hpp
@@ -339,6 +339,7 @@ public:
     vec<ConvertT, N> result;
 
     for(int i = 0; i < N; ++i) {
+// TODO handle float to float conversion. Here only float to short rounding is implemented
       if constexpr (std::is_floating_point_v<T> &&
                        std::is_integral_v<ConvertT>) {
         if constexpr (RM == rounding_mode::rte || RM == rounding_mode::automatic) {


tdavidcl (commented):
Yup that looks alright for `float -> int` conversions, maybe just add a comment that the automatic rounding mode is `rte` only if the source is a float and the dest an int, since float -> float will behave differently.

While this is much better than the silent rounding i think the dpcpp implementation is much more complex because of `float -> float` conversions. I'm not even sure if there is a builtin function to perform a rtp rounding for `f64 -> f32` for example. Maybe add a TODO for float float conversions since this correct the behavior only for float -> int ?

illuhad:
> While this is much better than the silent rounding i think the dpcpp implementation is much more complex because of float -> float conversions. I'm not even sure if there is a builtin function to perform a rtp rounding for f64 -> f32 for example

Yes. AFAIK OpenCL is the only backend that has arbitrary conversions from any floating point type to any other with any rounding mode. Not even CUDA has that. So we'd need to implement that manually with bit level manipulations if we wanted to have that...

tdavidcl:
> > While this is much better than the silent rounding i think the dpcpp implementation is much more complex because of float -> float conversions. I'm not even sure if there is a builtin function to perform a rtp rounding for f64 -> f32 for example
> 
> Yes. AFAIK OpenCL is the only backend that has arbitrary conversions from any floating point type to any other with any rounding mode. Not even CUDA has that. So we'd need to implement that manually with bit level manipulations if we wanted to have that...

This is one example where there could be a reference implementation provided by the working group that serves as a fallback when there is no builtin for a specific backend isn't it ? I know that the working group only deals with the standard but that's one of those cases where everyone that wants to be conformant will have to reinvent the same thing even though a fallback could be provided somewhere :sweat:

include/hipSYCL/sycl/libkernel/vec.hpp
```diff
@@ -339,8 +339,22 @@ class vec {
     vec<ConvertT, N> result;
 
     for(int i = 0; i < N; ++i) {
-      // TODO: Take rounding mode into account
-      result[i] = static_cast<ConvertT>(_data[i]);
+      if constexpr (std::is_floating_point_v<T> &&
```
ferdymercury, line 343:
```suggestion
// TODO handle float to float conversion. Here only float to short rounding is implemented
      if constexpr (std::is_floating_point_v<T> &&
```

include/hipSYCL/sycl/libkernel/vec.hpp
```diff
@@ -339,8 +339,23 @@ class vec {
     vec<ConvertT, N> result;
 
     for(int i = 0; i < N; ++i) {
-      // TODO: Take rounding mode into account
-      result[i] = static_cast<ConvertT>(_data[i]);
+// TODO handle float to float conversion. Here only float to short rounding is implemented
```
tdavidcl:
```suggestion
// TODO handle float to float conversion. Here only float to integer rounding is implemented
```
any type of integer actually

tdavidcl (approved):
LGTM 😄

illuhad:
>  This is one example where there could be a reference implementation provided by the working group that serves as a fallback when there is no builtin for a specific backend isn't it ? I know that the working group only deals with the standard but that's one of those cases where everyone that wants to be conformant will have to reinvent the same thing even though a fallback could be provided somewhere 😓

Yes. Or, alternatively, figure out whether we actually need to support all modes for all conversions between all types, and then perhaps slim down the spec a little. As I said, not even CUDA supports that.

illuhad (approved):
Thanks!

ferdymercury:
I was trying to implement this new functionality, and I am now getting a compilation error, well rather four, once for every function (ceil, floor, trunc and rint):

```
In file included from /opt/AdaptiveCpp/install/bin/../include/AdaptiveCpp/sycl/sycl.hpp:14:
In file included from /opt/AdaptiveCpp/install/bin/../include/AdaptiveCpp/sycl/../hipSYCL/sycl/sycl.hpp:50:
/opt/AdaptiveCpp/install/bin/../include/AdaptiveCpp/sycl/../hipSYCL/sycl/libkernel/vec.hpp:350:45: error: no member named 'ceil' in namespace 'hipsycl::sycl'; did you mean simply 'ceil'?
  350 |           result[i] = static_cast<ConvertT>(sycl::ceil(_data[i]));
      |                                             ^~~~~~
/usr/include/bits/mathcalls.h:159:14: note: 'ceil' declared here
  159 | __MATHCALLX (ceil,, (_Mdouble_ __x), (__const__));
      |              ^
```

Does this mean that these four functions are not implemented for SYCL? or is it just a namespace issue?

Removing `sycl::` seems to solve the issue.

illuhad:
I should have insisted on a test case :D

We do have `sycl::ceil`. Perhaps you're just missing the right include in `vec.hpp`? It's possible that `sycl.hpp` pulls it in before the math builtins.
Actually, that seems fairly likely because the math builtins also need to provide `vec` overloads, so there's a circular dependency... Try using the lower level `__acpp_ceil` directly.

tdavidcl:
oh yeah just noticed the same issue in Shamrock's CI :sweat_smile:

tdavidcl:
unless you are already on it i can rush to make a fix

illuhad:
@tdavidcl it's yours if you want it :-)

tdavidcl:
got it, leave convert unimplemented in vec.hpp and move the implementation to math builtins since they include vec anyway. I'm very suprised that the CI worked though since in Shamrock just including sycl.hpp made the error.

illuhad:
Yes, I'm also seeing CI fails now for independent functionality... don't know why.

> got it, leave convert unimplemented in vec.hpp and move the implementation to math builtins since they include vec anyway.

If you call the lower level `__acpp_ceil` and friends I think we could avoid hacking `vec` apart, since those live in a different header.

Issue #1982, opened 2026-02-27
Intel GPUs and fast-math: too fast, too relaxed


al42and:
When using SSCP mode to target Intel GPUs, `sycl::erf` gets pretty broken when `-ffast-math` is enabled: 1.5% error for `erf(0.1)`.

Interestingly, the same code works fine with `icpx -ffast-math` unless `-cl-fast-relaxed-math` is requested explicitly.

Of course, using `ffast-math` voids any warranties on floating point behavior, so not sure this is considered a bug. Besides, Intel GPUs are known for their creative implementation of math functions.

Since we're not even close to denormals or other edge cases here, the error is quite large, and behavior is different from DPC++ with equivalent flags (and I wasted enough time tracking this issue down), I think it's worth recording that things can go a bit too far.

```console
$ acpp --acpp-targets=generic -O2 -ffast-math -o test_sycl_erf test_sycl_erf.cpp
$ ACPP_VISIBILITY_MASK=opencl:1 ./test_sycl_erf
erf(+1.000000) = +0.842700779  expected +0.842700779  diff 0.00e+00  PASS
erf(+0.500000) = +0.520499885  expected +0.520499885  diff 0.00e+00  PASS
erf(+0.400000) = +0.426889002  expected +0.428392351  diff 1.50e-03  FAIL
erf(+0.300000) = +0.325401068  expected +0.328626782  diff 3.23e-03  FAIL
erf(+0.100000) = +0.114208192  expected +0.112462915  diff 1.75e-03  FAIL
erf(+0.010000) = +0.008815283  expected +0.011283415  diff 2.47e-03  FAIL
erf(-0.100000) = -0.114208192  expected -0.112462915  diff 1.75e-03  FAIL
erf(-0.010000) = -0.008815283  expected -0.011283415  diff 2.47e-03  FAIL
```

```console
$ acpp --acpp-targets=generic -O2 -ffast-math -o test_sycl_erf test_sycl_erf.cpp
$ ACPP_VISIBILITY_MASK=cuda  ./test_sycl_erf
erf(+1.000000) = +0.842700779  expected +0.842700779  diff 0.00e+00  PASS
erf(+0.500000) = +0.520499885  expected +0.520499885  diff 0.00e+00  PASS
erf(+0.400000) = +0.428392351  expected +0.428392351  diff 0.00e+00  PASS
erf(+0.300000) = +0.328626752  expected +0.328626782  diff 2.98e-08  PASS
[...]

PR #2049, opened 2026-04-19
[Cmake][CI] officially support llvm 21?

tdavidcl:
Hi,
Just a quick Sunday PR. I was about to upgrade Shamrock's CI to llvm21 when i just noticed that llvm 21 is not officially supported even though some use acpp with it (e.g. #1982) and apparently compatibility was fixed in August #1969.

This PR just add llvm 21 to CI and bump the max LLVM version in the CMake before the `please set -DACPP_EXPERIMENTAL_LLVM=ON` error.

I think those are the only required changed but i may have missed some since there is quite a lot of CMake and CI business there :smile:


tdavidcl:
Just primed the docker https://github.com/AdaptiveCpp/AdaptiveCpp/actions/runs/24625246534/job/72003000044

tdavidcl:
Oh the adding a new case fails if it is from a fork since the new docker image is created in my fork rather than the main repo :sweat_smile: I'll do another one from this repo directly

tdavidcl:
Close in favor of  #2050 which was made from the main repo to avoid docker related issues

PR #2050, opened 2026-04-19
[Cmake][CI] officially support llvm 21

tdavidcl:
Hi,
Just a quick Sunday PR. I was about to upgrade Shamrock's CI to llvm21 when i just noticed that llvm 21 is not officially supported even though some use acpp with it (e.g. https://github.com/AdaptiveCpp/AdaptiveCpp/issues/1982) and apparently compatibility was fixed in August https://github.com/AdaptiveCpp/AdaptiveCpp/pull/1969.

This PR just add llvm 21 to CI and bump the max LLVM version in the CMake before `the please set -DACPP_EXPERIMENTAL_LLVM=ON` error.

I think those are the only required changed but i may have missed some since there is quite a lot of CMake and CI business there 😄

PS: I was forced to make the PR directly from the main repo since the docker jobs were building the new docker containers in my fork rather than here. The old PR was #2049


illuhad (commented):
This looks good so far (our SPIRV-LLVM-Translator fork already supports LLVM 21). Please check the documentation, I suspect it says somewhere that the max LLVM version is 20.

We'd also need to test LLVM 21 with the JIT compiler on AMD/Intel/NVIDIA to ensure there are no issues in the JIT stage. I can do this this week, I hope (If you've already tested some backends, let me know, then I can skip them :-) ).

tdavidcl:
> This looks good so far (our SPIRV-LLVM-Translator fork already supports LLVM 21). Please check the documentation, I suspect it says somewhere that the max LLVM version is 20.

Oh right, I'll check 

> We'd also need to test LLVM 21 with the JIT compiler on AMD/Intel/NVIDIA to ensure there are no issues in the JIT stage. I can do this this week, I hope (If you've already tested some backends, let me know, then I can skip them :-) ).

Not yet, I was just trying to get the CI before attempting anywhere else.

tdavidcl:
I just went through the doc to update llvm 20 references to 21. The install llvm page can be kept with llvm 20 though. I've included it in the commit for now to list everywhere than was mentioning it (e.g. grep llvm 20, clang 20, and going through the giant list that is grep 20 :laughing:).

tdavidcl:
I'm starting testing on Nvidia (Ampere) & CPU with SSCP, SMCP, omp with acpp test suite & Shamrock

- [x] OpenMP
- [x] generic (Omp & CUDA)
- [ ] SMCP CUDA 13

New things:

i get a bunch of warnings like with SSCP + CUDA
```
warning: Linking two modules of different data layouts: '' is 'e-p6:32:32-i64:64-i128:128-v16:16-v32:32-n16:32:64' whereas '' is 'e-p:64:64:64-i1:8:8-i8:8:8-i16:16:16-i32:32:32-i64:64:64-i128:128:128-f32:32:32-f64:64:64-v16:16:16-v32:32:32-v64:64:64-v128:128:128-n16:32:64'
```

CUDA 13 + SMCP give an error, but i don't think we can do anything about it since it is on the LLVM side (`partially supported version 12.8`). Conveniently SSCP does not suffer from this issue and works just fine:
```
FAILED: [code=1] CMakeFiles/sycl_tests.dir/sycl/item.cpp.o 
/home/tdavidcl/Documents/shamrock-dev/Shamrock/build_3/.env/acpp-installdir/lib/cmake/AdaptiveCpp/syclcc-launcher --launcher-cxx-compiler=/usr/bin/c++ --launcher-syclcc=/home/tdavidcl/Documents/shamrock-dev/Shamrock/build_3/.env/acpp-installdir/bin/acpp  --acpp-targets="cuda:sm_60" /usr/bin/c++ -DACPP_ALLOW_INSTANT_SUBMISSION=1 -DHIPSYCL_DEBUG_LEVEL=2 -DLIB_NUMA_AVAILABLE -I/home/tdavidcl/Documents/shamrock-dev/Shamrock/build_3/.env/acpp-git/tests -isystem /home/tdavidcl/Documents/shamrock-dev/Shamrock/build_3/.env/acpp-installdir/include -isystem /home/tdavidcl/Documents/shamrock-dev/Shamrock/build_3/.env/acpp-installdir/include/AdaptiveCpp -fdiagnostics-color=always -O3 -DNDEBUG -MD -MT CMakeFiles/sycl_tests.dir/sycl/item.cpp.o -MF CMakeFiles/sycl_tests.dir/sycl/item.cpp.o.d -o CMakeFiles/sycl_tests.dir/sycl/item.cpp.o -c /home/tdavidcl/Documents/shamrock-dev/Shamrock/build_3/.env/acpp-git/tests/sycl/item.cpp
acpp warning: The 'hip' and 'cuda' compilation flows are not typically useful outside of specific interoperability use cases or applications that have been engineered specifically around them (e.g. GROMACS).

For all other use cases, they may not result in best performance!

Please use 'generic' target for better performance, portability, and usability!
(use --acpp-no-warn-legacy-flows to silence this warning)
clang++: warning: CUDA version is newer than the latest partially supported version 12.8 [-Wunknown-cuda-version]
In file included from <built-in>:1:
In file included from /usr/lib/llvm21/lib/clang/21/include/__clang_cuda_runtime_wrapper.h:209:
/opt/cuda/include/crt/math_functions.hpp:2987:40: error: expected function body after function declarator
 2987 | __func__(double rsqrt(const double a)) _NV_RSQRT_SPECIFIER
      |                                        ^
In file included from <built-in>:1:
/usr/lib/llvm21/lib/clang/21/include/__clang_cuda_runtime_wrapper.h:388:10: fatal error: 'texture_fetch_functions.h' file not found
  388 | #include "texture_fetch_functions.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~
2 errors generated when compiling for sm_60.
ninja: build stopped: subcommand failed.
```

illuhad:
> i get a bunch of warnings like with SSCP + CUDA

This is somewhat expected. The data layout string occasionally gets expanded/modified in newer LLVM versions. Not a functionality issue, but does not look nice. Can you change the data layout string to the shorter one from your error message in case LLVM > 20 here?
https://github.com/AdaptiveCpp/AdaptiveCpp/blob/43c29f989de655ade04d967c858db7b241483f88/src/compiler/llvm-to-backend/ptx/LLVMToPtx.cpp#L143
That should fix the warning.

> CUDA 13 + SMCP give an error, but i don't think we can do anything about it since it is on the LLVM side (partially supported version 12.8). Conveniently SSCP does not suffer from this issue and works just fine:

Yes, this is unrelated to supporting LLVM 21. This issue is related to the handling of CUDA headers, so SSCP is immune to those :)

tdavidcl:
> > i get a bunch of warnings like with SSCP + CUDA
> 
> This is somewhat expected. The data layout string occasionally gets expanded/modified in newer LLVM versions. Not a functionality issue, but does not look nice. Can you change the data layout string to the shorter one from your error message in case LLVM > 20 here?
> 
> https://github.com/AdaptiveCpp/AdaptiveCpp/blob/43c29f989de655ade04d967c858db7b241483f88/src/compiler/llvm-to-backend/ptx/LLVMToPtx.cpp#L143
> 
> That should fix the warning.

Yes it does but doing the following looks really off
```c++
#if LLVM_VERSION_MAJOR > 20
  std::string DataLayout =
      "e-p6:32:32-i64:64-i128:128-v16:16-v32:32-n16:32:64";
#else
  std::string DataLayout =
      "e-p:64:64:64-i1:8:8-i8:8:8-i16:16:16-i32:32:32-i64:64:64-i128:128:128-f32:32:32-"
      "f64:64:64-v16:16:16-v32:32:32-v64:64:64-v128:128:128-n16:32:64";
#endif
```

I would have to do a bit of reading to get a better understanding of what's going on there. While it does work on my GPU (3070) with Shamrock i have a feeling that this might depend on the target so i don't really like that change as is.

From what i can read MLIR peoples are rather doing something like (https://discourse.llvm.org/t/mandatory-data-layout-in-the-llvm-dialect/85875/32) to fetch the actual layout for that tripple
```c++
llvm::Target(triple)->createTargetMachine(Triple(triple), cpu, features)->createDataLayout()
```

We could try doing the same to automatically use the layout provided by llvm ? But i do not find any infos passed to that function about the actual PTX version that we are targeting which i assume is required to fetch the correct layout.

illuhad:
> I would have to do a bit of reading to get a better understanding of what's going on there. While it does work on my GPU (3070) with Shamrock i have a feeling that this might depend on the target so i don't really like that change as is.

We already do something similar on the AMD side:
https://github.com/AdaptiveCpp/AdaptiveCpp/blob/43c29f989de655ade04d967c858db7b241483f88/src/compiler/llvm-to-backend/amdgpu/LLVMToAmdgpu.cpp#L290


The target layout depends on the target triple `nvptx64-nvidia-cuda`, which does not include the actual GPU. In fact, the exact GPU cannot play a role because otherwise LLVM couldn't generate PTX that is forward-compatible to different GPU targets. PTX is intended as a portable IR between NVIDIA GPUs.

Also: As a single-source model supporting USM, SYCL requires that all data types must have same size and alignment on host and device. So in practice, the data layouts of our targets cannot differ apart from some details which mostly don't matter to us.

> We could try doing the same to automatically use the layout provided by llvm ? 

This might be more difficult because AdaptiveCpp does not explicitly generate the target machine as in your example. We reuse what LLVM gives us.


The main difference between the two data layout strings is:
New one introduces a new address space 6 with 32 bit pointers (no idea what this is, we don't use it. Perhaps some new CUDA or AI feature).
New one lets LLVM handle floats and integers implicitly via data layout defaults, while the old one fully defines every type. But I assume the LLVM defaults make it boil down to the same thing.
The first one also only defines small vectors.

As far as I can see clang trunk generates exactly the same string for sm_120, but also introduces 256-bit integers. https://godbolt.org/z/hWv5583Tq

tl;dr: I expect that we should be able to safely use the new data layout string.

EDIT: I should add that the conflict that you're observing is because there's a conflict between what clang generates for our bitcode libraries, and what we set explicitly in the JIT compiler. The bitcode library IR is implicitly generated for sm_60 because that's the oldest architecture that gives us the features we need. So what you're observing is that clang generates the same data layout for sm_60 and sm_120 (apart from the 256 bit int, but that could be a clang trunk vs clang 21 issue). So this is another data point that it does not depend on the GPU architecture beyond "it's nvptx64".

tdavidcl:
alright i'll go with that then.

Additionally why not allow passing an optional arch to this section or in general. Since SSCP jit at runtime it has the liberty of chosing the right arch anyway, right? (or ptx version in the case of NVidia since it is just an IR). Could there be some code paths that are specific to some PTX versions one day in that backend ?

illuhad:
> Additionally why not allow passing an optional arch to this section or in general. Since SSCP jit at runtime it has the liberty of chosing the right arch anyway, right? (or ptx version in the case of NVidia since it is just an IR). Could there be some code paths that are specific to some PTX versions one day in that backend ?

I'm not sure what you mean. SSCP JIT absolutely takes into account the SM version of the target. We don't generate the same PTX for every target. See e.g. https://github.com/AdaptiveCpp/AdaptiveCpp/blob/153efccd3acf94960d114632952a24e56e709f83/src/compiler/llvm-to-backend/ptx/LLVMToPtx.cpp#L249

It's just that for LLVM's target triple or data layout, the arch does not play a role.

tdavidcl:
> It's just that for LLVM's target triple or data layout, the arch does not play a role.

Indeed I meant for the layout since apparently the layout change depending on PTX or the clang version even though it seems to work if the "wrong" one is passed. why are they even bothering with having multiple one then 😅

illuhad:
> Indeed I meant for the layout since apparently the layout change depending on PTX or the clang version even though it seems to work if the "wrong" one is passed. why are they even bothering with having multiple one then 😅


The layout does not functionally change. The string is different, but that it is mainly due to different conventions as to what is spelled out explicitly versus what is defined by defaults. The only real new thing is that address space 6, which is probably some new feature as I said.

illuhad:
`sycl_tests` pass on Intel GPU with OpenCL/SPIR-V.

LIT tests to be done.

Looks like there's a similar warning on AMD as on CUDA:
```
warning: Linking two modules of different data layouts: '' is 'e-p:64:64-p1:64:64-p2:32:32-p3:32:32-p4:64:64-p5:32:32-p6:32:32-p7:160:256:256:32-p8:128:128:128:48-p9:192:256:256:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-v2048:2048-n32:64-S32-A5-G1-ni:7:8:9' whereas '' is 'e-p:64:64-p1:64:64-p2:32:32-p3:32:32-p4:64:64-p5:32:32-p6:32:32-p7:160:256:256:32-p8:128:128-p9:192:256:256:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-v2048:2048-n32:64-S32-A5-G1-ni:7:8:9'
```
On AMD, `sycl_tests` currently crash in `prefetch_host` test with nullptr access. Not sure if LLVM 21 issue or related to ROCm 7.2, to which I upgraded to get sufficiently new ROCm to target with SSCP LLVM 21.

illuhad:
Looks like ROCm 7.2 has an issue with `hipMallocManaged()` which causes the crashes. Have replaced with `hipMallocHost` for testing.

Now, AMD passes all `sycl_tests`.

OpenCL and OpenMP pass LIT tests.

However, CUDA and HIP fail on builtin remapping and dynamic function LIT tests. Needs more investigation.

illuhad:
LLVM 21 is fine with #2069 and #2070 on OpenMP, OpenCL, HIP and CUDA.

Not sure if @resetius or @ewanc want to do some testing for their backends as well?

@tdavidcl please add the new data layouts for AMD and NVIDIA to `LLVMToPtx` and `LLVMToAmdgpu`.

tdavidcl:
> @tdavidcl please add the new data layouts for AMD and NVIDIA to `LLVMToPtx` and `LLVMToAmdgpu`.

I'll do that, also should we enable llvm 21 on the self hosted GPU runner since it seems to be a more significant change than expected

tdavidcl:
> @tdavidcl please add the new data layouts for AMD and NVIDIA to `LLVMToPtx` and `LLVMToAmdgpu`.

Done

EwanC:
> Not sure if @resetius or @EwanC want to do some testing for their backends as well?

Thanks for the heads up, I'll take a look into LLVM 21 but don't holdup this PR waiting on the outcome of that. 
Edit: Tested this out in a branch on my fork that adds LLVM 21 to a Vulkan CI matrix and it [passed](https://github.com/EwanC/AdaptiveCpp/actions/runs/25216434717/job/73939202340)

illuhad:
> I'll do that, also should we enable llvm 21 on the self hosted GPU runner since it seems to be a more significant change than expected

@tdavidcl That would be good, however, we don't really have the resources currently to build and run the tests for all sorts of different LLVM versions. So unfortunately we currently can only run the self-hosted tests for one, especially on AMD and Intel. IIRC it's LLVM 18 at the moment.

resetius:
> LLVM 21 is fine with #2069 and #2070 on OpenMP, OpenCL, HIP and CUDA.
> 
> Not sure if @resetius or @EwanC want to do some testing for their backends as well?
> 
> @tdavidcl please add the new data layouts for AMD and NVIDIA to `LLVMToPtx` and `LLVMToAmdgpu`.

Metal backend supports llvm 21. I m testing it manually from time to time. Not sure if need CI job for this

illuhad (approved):
Thanks!

PR #2052, opened 2026-04-20
[Hotfix] fix declaration order for vec::convert

tdavidcl:


tdavidcl:
I just tested on Shamrock and it fixes the issue. I still have no idea why the bug is not triggered in the CI.
Wait is it a docker issue, as in we the CI used a old container instead of the one that was just rebuilt on develop ?

illuhad:
Thanks @tdavidcl ! Could you try leaving `vec` as one, and instead calling `__acpp_ceil`?

I'd like to avoid such splits because it will make it more annoying to implement the proposed includes KHR in the future: https://github.com/KhronosGroup/SYCL-Docs/pull/814

tdavidcl:
> `__acpp_ceil`

Yup agree this is not elegant :sweat_smile: it's just a quickfix.
I could but wouldn't __acpp_ceil suffer from the same issue ?

nevermind it's not the same header it should be fine

illuhad:
> I could but wouldn't __acpp_ceil suffer from the same issue ?

No, I don't think so -- the `__acpp_*` functions live in their own header and only exist as scalar overloads -> no vec dependency and thus no circular dependency.

They exist as lower-level scalar builtins that the `sycl::` math functions can call for `vec` and `marray`.

tdavidcl:
yup it seems ok for now with the builtin, I'm starting the tests on my side. I still really don't understand how the CI did pass in the original PR.

illuhad (approved):
Thanks - LGTM. Let's wait for CI and test it well this time :D

tdavidcl:
Something's weird in the CI... I just tried on Shamrock and the current patch works and passes all the tests

tdavidcl:
I'm getting in the CI
```
In file included from /__w/AdaptiveCpp/AdaptiveCpp/tests/sycl/vec.cpp:13:
In file included from /__w/AdaptiveCpp/AdaptiveCpp/tests/sycl/sycl_test_suite.hpp:24:
In file included from /__w/AdaptiveCpp/AdaptiveCpp/build/install/bin/../include/AdaptiveCpp/sycl/sycl.hpp:14:
In file included from /__w/AdaptiveCpp/AdaptiveCpp/build/install/bin/../include/AdaptiveCpp/sycl/../hipSYCL/sycl/sycl.hpp:50:
/__w/AdaptiveCpp/AdaptiveCpp/build/install/bin/../include/AdaptiveCpp/sycl/../hipSYCL/sycl/libkernel/vec.hpp:347:45: error: use of undeclared identifier '__acpp_rint'
  347 |           result[i] = static_cast<ConvertT>(__acpp_rint(_data[i]));
      |                                             ^
/__w/AdaptiveCpp/AdaptiveCpp/tests/sycl/vec.cpp:128:25: note: in instantiation of function template specialization 'hipsycl::sycl::vec<float, 4>::convert<int, hipsycl::sycl::rounding_mode::automatic>' requested here
  128 |   auto ints = floats_in.convert<int>();
      |                         ^
5 warnings and 1 error generated.
```

illuhad:
I think `detail::__acpp_rint` perhaps?

tdavidcl:
> I think `detail::__acpp_rint` perhaps?

Oh right i missed the additional namesapce

illuhad (approved):
Looks like we figured it out now - thanks :)

tdavidcl:
> Looks like we figured it out now - thanks :)

Yup and Shamrock's CI is back online too it worked there

PR #2053, opened 2026-04-20
[Metal] Introduce metal_mmap_region for USM allocations

resetius:
Reserves a large VA range via mmap and uses newBuffer(ptr, ...) to back all shared/host MTL::Buffer allocations at controlled CPU addresses. A calibration step determines the fixed offset _delta between CPU and GPU virtual addresses, and a correction loop enforces it for every subsequent allocation:                                                                                                              

``` 
buffer->gpuAddress() - (uintptr_t)buffer->contents() == _delta  // constant
```
 
This is a prerequisite for full USM support on Metal. The approach was proposed by Philip Turner in https://github.com/philipturner/metal-usm

### DIFF
--- a/src/runtime/metal/metal_allocator.cpp
+++ b/src/runtime/metal/metal_allocator.cpp
@@ -12,15 +12,190 @@
 #include "hipSYCL/runtime/metal/metal_allocator.hpp"
 
 #include <Metal/Metal.hpp>
+#include <sys/mman.h>
+#include <sys/sysctl.h>
+#include <unistd.h>
+
+#include <iostream>
 
 namespace hipsycl {
 namespace rt {
 
-metal_allocator::metal_allocator(MTL::Device* device, const device_id &id)
-  : _device{device}, _device_id{id}
-{}
+namespace {
+  uintptr_t align_up(uintptr_t v, size_t align) {
+    return (v + align - 1) & ~(uintptr_t)(align - 1);
+  }
 
-metal_allocator::~metal_allocator() = default;
+  size_t get_total_ram() {
+    uint64_t mem = 0;
+    std::size_t len = sizeof(mem);
+    if (sysctlbyname("hw.memsize", &mem, &len, nullptr, 0) == 0)
+      return static_cast<std::size_t>(mem);
+    return 8ULL << 30; // fallback: 8 GiB
+  }
+
+  // Metal allocates MTL::Buffer storage in page_size units and always appends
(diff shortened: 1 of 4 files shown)


resetius:
Rewrote this allocator three times. Third iteration finally came out the most concise and well-structured 
Previous ones were more complex and harder to follow

philipturner:
I prefer to not be pinged about this, as I am no longer involved in the effort. Thank you!

src/runtime/metal/metal_allocator.cpp
```diff
@@ -12,15 +12,184 @@
...
+    if (_base) {
+      munmap(_base, _mmap_size);
+    }
+  }
+
+  void* midpoint() {
+    return static_cast<void*>(_midpoint);
+  }
+
+  void* alloc(size_t size) {
```
illuhad, line 76: We already have some allocator-in-buffer logic for stdpar: https://github.com/AdaptiveCpp/AdaptiveCpp/blob/43c29f989de655ade04d967c858db7b241483f88/include/hipSYCL/std/stdpar/detail/allocation_map.hpp#L79

Note sure if it would make sense to share some code?

tests/sycl/usm.cpp
```diff
@@ -654,5 +654,48 @@ BOOST_AUTO_TEST_CASE(linked_list_separate_alloc) {
   }
 }
 
+BOOST_AUTO_TEST_CASE(usm_shared_ptr_gpu_delta_constant) {
+  sycl::queue q{sycl::property::queue::in_order{}};
```
illuhad, line 658: Should we disable the test in case the device does not support shared memory for the benefit of the Vulkan backend?

resetius: added sycl::aspect::usm_shared_allocations check

src/runtime/metal/metal_allocator.cpp
resetius: `free_space_map` uses a buddy allocator (power-of-two block sizes), which cannot allocate at an arbitrary address. We need `alloc_at()` to implement the calibration loop. Different algorithms, different requirements. Sharing would mean rewriting both

resetius:
Added MADV_FREE_REUSABLE/MADV_FREE_REUSE to reduce RSS
Free blocks are marked as reclaimable under memory pressure and reclaimed on the next allocation

illuhad (approved):
Thanks!

PR #2054, opened 2026-04-21
[OpenCL][llvm-to-spirv] Disable pointer kernel argument wrapping

illuhad:
Currently we wrap kernel pointer arguments in a struct in llvm-to-spirv.
This is because older L0 versions and some OpenCL SVM APIs strictly validate all pointer arguments, and error if a nullptr or host pointer is passed in. In SYCL (or stdpar or PCUDA) this is however legal as long as the pointer is not dereferenced on the device. The pointer wrapping hides the pointer from the driver, and we can just pass it in as a blob without additional validation.

A downside of this approach is that, since the kernel no longer has pointer arguments, optimizations like automatic noalias inference at JIT time no longer work, because those need pointers. It also poses additional challenges for address space inference.

This approach therefore relaxes pointer wrapping such that it is no longer done in APIs that do not perform this undesired validation (OpenCL+Intel USM extensions). In the future, we could also relax it for more cases (newer Level Zero versions?).

Pointer wrapping is still done in all other cases so that hopefully nothing breaks.


illuhad:
~Mhmm.... looks like a lot of Intel OpenCL runtimes are still on version 1.0 of the API which *does* validate and requires pointer wrapping :(~ Problem was mostly likely caused by bug in my code...

blinkfrog:
I've just tested it with buffers - works now!

illuhad:
Thanks for testing! :)

PR #2059, opened 2026-04-25
[Metal] Fix address space inference; align with other backends

resetius:
- extractStringConstant: handle AddrSpaceCast ConstantExpr around string literals
- getPhysicalPointerAddressSpace: fix infinite recursion
- LLVMToMetal: run KernelFunctionParameterRewriter and AddressSpaceInferencePass after linking (as PTX/AMDGPU do), add a second ASI pass post-optimization so address spaces are correctly annotated before pointer-translation passes

### DIFF
--- a/src/compiler/llvm-to-backend/metal/LLVMToMetal.cpp
+++ b/src/compiler/llvm-to-backend/metal/LLVMToMetal.cpp
@@ -482,8 +482,12 @@ bool LLVMToMetalTranslator::prepareBackendFlavor(llvm::Module& M) {
 bool LLVMToMetalTranslator::toBackendFlavor(llvm::Module &M, PassHandler& PH) {
   AddressSpaceMap ASMap = getAddressSpaceMap();
 
-  AddressSpaceInferencePass ASIPass{ASMap};
-  ASIPass.run(M, *PH.ModuleAnalysisManager);
+  KernelFunctionParameterRewriter ParamRewriter{
+      KernelFunctionParameterRewriter::ByValueArgAttribute::ByVal,
+      ASMap[AddressSpace::Generic],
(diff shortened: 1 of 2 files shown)


src/compiler/llvm-to-backend/metal/LLVMToMetal.cpp
```diff
@@ -543,9 +554,9 @@ bool LLVMToMetalTranslator::translateToBackendFormat(llvm::Module& FlavoredModul
 
   std::unordered_set<std::string> kernelNames(KernelNames.begin(), KernelNames.end());
 
-#ifdef ACPP_PRINT_IR_BEFORE_EMIT
-  FlavoredModule.print(llvm::errs(), nullptr);
-#endif
+  if (getenv("__ACPP_PRINT_IR_BEFORE_EMIT")) {
```
illuhad, line 557: Maybe we should move this function to a common file, so that we can use it here: https://github.com/AdaptiveCpp/AdaptiveCpp/blob/0932775fa11b2af7ad59ab207c8cb242a4375b56/src/compiler/llvm-to-backend/LLVMToBackend.cpp#L78

Once #2060 is merged, this would would also be useful for `llvm-to-host`.

resetius: Yes, I agree. But I think I'll just remove these changes from this PR and do it properly in a separate one. Here it looks unrelated. Also, merging this PR would unblock me for adding the new address translation passes

illuhad: Okay, sounds good.

illuhad (approved):
Thanks!

PR #2062, opened 2026-04-27
Remove broken interop_handle::get_native_queue overload

al42and:
The `glue::backend_interop::get_native_queue(_executor)` codepath was never used except in a single test.
Remove it to make the code easier to follow. Not part of any public API as far as I can tell.


illuhad (approved):
Thanks!

PR #2063, opened 2026-04-27
Make swizzled const access return a const-reference

al42and:
See Table 112 in SYCL 2020 rev 11:

DataT& XYZW_ACCESS()
const DataT& XYZW_ACCESS() const
DataT& RGBA_ACCESS()
const DataT& RGBA_ACCESS() const
DataT& INDEX_ACCESS()
const DataT& INDEX_ACCESS() const


illuhad (approved):
Thanks!

PR #2066, opened 2026-04-29
[WIP] [Metal] Allocator test fix

resetius:


resetius:
crash on allocator destruction
need to capture mmapped region as shared_ptr
able to reproduce in virtual machine

PR #2067, opened 2026-04-29
[Metal] Fix crash when MTL::Buffer outlives mmap region

resetius:
Tested manually in a GitHub CI-like environment: a Tart macOS Tahoe virtual machine with the Apple paravirtual Metal device


illuhad (approved):
Thanks!

PR #2070, opened 2026-04-30
[LLVM 21] More aggressive LLVM builtin remapping; introduce builtin remapping on amdgpu

illuhad:
Looks like in LLVM 21, LLVM intrinsics that previously worked fine in the nvptx backend now no longer work.
Similarly, previously there were no known cases of LLVM math intrinsics not working on amdgpu -- now there are.

This PR adds intrinsic remapping such that the `builtin_remapping.cpp` LIT test now passes with LLVM 21.

Currently, the PR has some code duplication, which is not nice. Also, since both nvptx and amdgpu now suffer from this (and LLVM as a whole seems unreliable, not just the nvptx backend), we should maybe think about whether we do want to address this globally after all -> draft PR.

### DIFF
--- a/include/hipSYCL/compiler/llvm-to-backend/Utils.hpp
+++ b/include/hipSYCL/compiler/llvm-to-backend/Utils.hpp
@@ -14,6 +14,7 @@
 #include <atomic>
 #include <optional>
 #include <sstream>
+#include <array>
 
 #include "hipSYCL/compiler/llvm-to-backend/LLVMToBackend.hpp"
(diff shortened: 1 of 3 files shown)


illuhad:
Now shares remapping code between amdgcn and ptx. I've decided against doing remapping globally for now, instead just providing the remapping as a util function.

@resetius Metal does something similar but has its own code paths. I think it does some additional stuff too. Not sure if it makes sense to unify things given all the other machinery that metal does. Just FYI in case you do want to reuse it -- should now be easier due to availability in utils header :-)

PR #2074, opened 2026-05-06
[Metal] Fix scratch buffer race in work-group reduce and predicates

resetius:
This should fix flaky metal reduction tests caused by a scratch buffer race between simd groups


illuhad (approved):
Thanks!

PR #2075, opened 2026-05-11
[Windows] Fix app-local redistributable path detection for acpp-common.dll

blinkfrog:
Windows did not propagate `HIPSYCL_COMMON_LIBRARY_OUTPUT_NAME` to the top-level CMake scope, causing `config.hpp` to generate an empty `HIPSYCL_COMMON_LIBRARY_NAME`. That caused `get_lib_directory()` to call `GetModuleHandleA(".dll")`, which failed, and it fell back to `HIPSYCL_INSTALL_PREFIX/lib`, breaking the lookup of JIT-time dependencies in app-local redistributable packages.

This PR fixes it by making Windows propagate the common library filename up to the top level, the same way Linux and macOS already do.


illuhad (approved):
Thanks!

PR #2076, opened 2026-05-11
Fix app-local llvm-spirv lookup for deployed AdaptiveCpp applications

blinkfrog:
This fixes llvm-spirv path resolution for deployed / app-local AdaptiveCpp applications.

Previously, llvm-to-spirv resolved llvm-spirv only via get_install_directory() + HIPSYCL_RELATIVE_LLVMSPIRV_PATH, which assumes an installation-style layout such as: `<prefix>/lib/hipSYCL/ext/llvm-spirv/bin/llvm-spirv`. This broke redistributable layouts where resources are placed relative to the application directory: `<app>/hipSYCL/ext/llvm-spirv/bin/llvm-spirv`

This patch checks the app-local redistributable layout via `get_lib_directory()`, and falls back to the existing installation layout.


src/compiler/llvm-to-backend/spirv/LLVMToSpirv.cpp
```diff
@@ -321,9 +321,15 @@ bool LLVMToSpirvTranslator::translateToBackendFormat(llvm::Module &FlavoredModul
     InputStream.flush();
   }
 
-  std::string LLVMSpirVTranslator = hipsycl::common::filesystem::join_path(
+  // Prefer the app-local redistributable layout used by deployment manifests.
+  // Fall back to the installation layout.
+  std::string local_llvm_spirv_translator = hipsycl::common::filesystem::join_path(
+      hipsycl::common::filesystem::get_lib_directory(),
+      std::vector<std::string>{"hipSYCL", "ext", "llvm-spirv", "bin", HIPSYCL_LLVMSPIRV_NAME});
+  std::string installed_llvm_spirv_translator = hipsycl::common::filesystem::join_path(
```
illuhad: Could you use camelCase here and for `local_llvm_spirv_translator`? That's what we generally use in LLVM code.

Looks good otherwise!

blinkfrog: Renamed to match the existing local variable naming style used in this file (PascalCase).

illuhad (approved):
Thanks!

Issue #2071, opened 2026-04-30
Destructor removes persistent JIT cache file, breaking parallel runs and cache reuse


cbyrohl:
**Bug summary**

Running multiple instances of a AdaptiveCpp-compiled binary in parallel that share the same `~/.acpp` directory (e.g. SLURM jobs on parallel `$HOME`, or several ranks on one node) is unsafe on cold cache: at least one process aborts with `Could not load library: .../<hash>.jit.so` followed by `omp_queue: Code object construction failed`.

The trace points at the destructor of `omp_sscp_executable_object` ([`src/runtime/omp/omp_code_object.cpp:82`](https://github.com/AdaptiveCpp/AdaptiveCpp/blob/develop/src/runtime/omp/omp_code_object.cpp#L82)), which unconditionally `fs::remove`s the *persistent* JIT cache file written by another (or the same) process — so a candidate fix is to drop that removal and let invalidation go through the existing `kernel_cache::clear_persistent_cache` paths. Happy to send a PR if you confirm this is a real bug and agree on the direction.

**To Reproduce**

Standalone CMake reproducer (single trivial `parallel_for`, no other deps). Drop the three files below into a directory and:

```bash
mkdir build && cd build
cmake -DCMAKE_PREFIX_PATH=<acpp-prefix> -DCMAKE_BUILD_TYPE=Release ..
cmake --build . -j
cd ..
./run_repro.sh 16        # N=16 concurrent processes
```

The driver wipes `~/.acpp/apps/jit_race_repro-*` so every invocation hits the cold-cache path, then launches N instances of the same kernel concurrently. The exit code is the count of failing processes. At N=16 on a single host, ≥1 process aborts on the first cold run.

<details>
<summary><b><code>main.cpp</code></b> (51 lines)</summary>

```cpp
// Minimal SYCL kernel to trigger AdaptiveCpp JIT compilation.
//
// Single trivial parallel_for so the cache hit/miss path is exercised on the
// very first call. The actual computation is unimportant — what matters is
// that running this from a cold cache forces AdaptiveCpp to write
// <hash>.jit.so to its on-disk cache. Run N copies concurrently from a cold
// cache and at least one should fail to dlopen a partial blob.
//
// Build with the sibling CMakeLists.txt; drive with run_repro.sh.

#include <sycl/sycl.hpp>
#include <unistd.h>

#include <cstdio>
#include <cstdlib>
#include <vector>

int main() {
    try {
[...]

PR #2077, opened 2026-05-13
runtime/omp: avoid deleting persistent JIT .so and use unique temp files in no-cache-population mode

crystarm:
Related to #2071

This PR fixes a race in OMP SSCP code object handling when multiple processes run concurrently and share the same `~/.acpp` cache directory.

Previously, `omp_sscp_executable_object` always removed `_kernel_cache_path` in its destructor. In normal mode, this path is derived from the persistent JIT cache id (`<id>.jit.so`), so one process could delete the file while another process still needed to load it.

## Root cause

The same file-path lifecycle was used for two different semantics:

1. A persistent cache artifact shared across processes.
2. Temporary file behavior needed when `ACPP_RT_NO_JIT_CACHE_POPULATION=1`.

Unconditional destructor cleanup is unsafe for the persistent-cache case, while non-unique naming is unsafe for temporary files under parallel runs.

## Changes in this PR

This PR adds mode-dependent path and cleanup behavior in `omp_sscp_executable_object`.

### Default mode

When `no_jit_cache_population == false`:

- keep using the persistent `<id>.jit.so` path;
- do not remove it in the destructor.

### `ACPP_RT_NO_JIT_CACHE_POPULATION=1` mode

When JIT cache population is disabled:

- create a unique temporary `.so` filename;
- make the name collision-resistant across processes and threads;
- remove only that temporary file in the destructor.

## Why this approach

This preserves persistent cache correctness across parallel processes while keeping temporary-file cleanup semantics for the no-cache-population mode.

It also addresses the naming and collision concern discussed in #2071.

## Local validation

I built `rt-backend-omp` in this environment, both with and without SSCP toolchain paths available.

I also ran reproducer-style parallel startup tests on CPU (`omp`) with:

- cold cache wipe + 16 concurrent processes;
- cold cache wipe + 32 concurrent processes;
- `ACPP_RT_NO_JIT_CACHE_POPULATION=1`.

All runs completed successfully in this environment.

## Note

The exact `generic`/SSCP runtime reproducer from #2071 could not be executed here because this environment is missing the LLVM/clang development components required for the full local toolchain setup.


src/runtime/omp/omp_code_object.cpp
```diff
@@ -25,12 +30,28 @@
...
 
+std::string make_unique_temp_cache_file_path(kernel_configuration::id_type id) {
+  std::string persistent_cache_file =
+      kernel_cache::get_persistent_cache_file(id);
+
+  const auto now = std::chrono::high_resolution_clock::now().time_since_epoch().count();
+  const std::size_t tid_hash = std::hash<std::thread::id>{}(std::this_thread::get_id());
+  const std::uint64_t random =
+      (static_cast<std::uint64_t>(std::random_device{}()) << 32) |
+      static_cast<std::uint64_t>(std::random_device{}());
```
illuhad: This does not seem like a good pattern to get a random number -- I think you should just to `random_device()` once as a static variable to generate a seed, and then use a random distribution.

Also not sure if the thread id on its own is going to help much -- we'd also need the process id, if we want to make it robust against collisions of the random number, which I think was the intent?

crystarm:
Implemented, thanks! I switched temp-name randomness to a static `mt19937_64` seeded once from `std::random_device`, and generate values via `uniform_int_distribution<uint64_t>`.  
I also added the process id (`getpid()` / `GetCurrentProcessId()`) to the temporary `.so` filename (alongside timestamp + thread id hash) to further reduce cross-process collision risk.

fodinabor (approved):
Generally looks good to me, thx!

I just have one more general question w.r.t the nulti-process runs, which doesnt have to be solved in this PR:
It sounds to me like we are having a race condition between the processes creating the cache files. On Linux I wouldn't worry too much, as the files would just get overridden (with at least semantically equivalent versions). In the Windows world, files are often locked. And therefore cannot be written once another process works with it.
How do we go in this case: process A does get a cache miss, process B a miss as well and then A generates the binary, B as well. Now A writes the binary and B tries to do it as well, but can't get a lock in the file -> process B has a IO error..?

illuhad:
@fodinabor Good question. Some thoughts:

Caches files are written using this `atomic_write` function: https://github.com/AdaptiveCpp/AdaptiveCpp/blob/f703b8115305c51d20c7819f001fdbfbeb6cb77b/src/common/filesystem.cpp#L177
In the scenario you are describing
- the write might not have proper atomic semantics (?) which I suppose might be possible with the current implementation on Windows. It's fine on Linux -> the function might need a Windows-specific code path.
- If both processes are trying to write the cache file, they must have also independently JIT-compiled the kernel before, and thus have already put it in the in-memory cache. Therefore, the program wouldn't actually care whether the files can be written successfully in this application run. Whether that case is handled gracefully (i.e. ignored) currently on Windows I don't know.

crystarm:
Thanks for the review!!

I do not have experience with Windows filesystem semantics, so for now I can't say much...    ( ╹ -╹)

Would you like me to open a separate issue about the possible Windows-specific persistent-cache population race? I could investigate the Windows behavior in more detail and try to address the problem, if there is one.

illuhad (approved):
Thanks, looks good -- I think the Windows issue is independent from this change here and could be addressed in a future PR.

PR #2079, opened 2026-05-18
[SSCP][SPIR-V]: Implement subgroup shifts using relative shuffles

tneval:
Background:
I've recently been optimizing subgroup shuffle operations in PoCL to enable autovectorization.
When testing shift_group_[left|right] via ACpp->PoCL, I noticed that autovectorization is blocked.
This is due to how shifts are implemented on ACpp: they are lowered to `OpGroupNonUniformShuffle`with some conditional logic to choose the valid indices.

In this PR, I propose implementing subgroup shift operations using relative shuffles, i.e. `OpGroupNonUniformShuffleUp` and `OpGroupNonUniformShuffleDown`. As these are specialized versions of 'regular' shuffle, it is reasonable to expect that implementations might have optimized versions that perform better. 

I have tested this patch with PoCL, and `shift_group_[left|right]` are autovectorized successfully.
Not sure if this change would be problematic for other SPIR-V consumers.

### DIFF
--- a/src/libkernel/sscp/spirv/shuffle.cpp
+++ b/src/libkernel/sscp/spirv/shuffle.cpp
@@ -23,7 +23,11 @@ template <typename dataT>
 dataT __spirv_SubgroupShuffleXorINTEL(dataT Data, __acpp_uint32 Value) noexcept;
 
 template <typename ValueT, typename IdT>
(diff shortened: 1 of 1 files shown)


sbalint98 (approved):
LGTM! thank you

illuhad (approved):
Thanks!

PR #2080, opened 2026-05-18
[CI] Disable ROCm version 6.3 for clang 17

illuhad:
Currently, we test ROCm 5.4 for LLVM < 21 and ROCm 6.3 for LLVM 21.

LLVM 17 is the outlier, because we seem to test it with both ROCm versions. I don't know what the background is, but it does seem like the ROCm 6.3 build with LLVM 17 causes hiccups in our CI pipeline (it seems like the wrong docker image is pulled?).

I don't have the time at the moment to figure out the exact problem, so I propose to just disable that test. It seems strange to me anyway that we have this special treatment for LLVM 17.

Note that both LLVM 17 and ROCm 6.3 continue to be tested, not just in that specific combination.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -21,12 +21,6 @@ jobs:
         cuda_patch_version: [2]
         rocm: [5.4.3]
         include:
-          - clang: 17
-            os: ubuntu-22.04
-            cuda_major_version: 11
-            cuda_minor_version: 0
-            cuda_patch_version: 2
-            rocm: 5.6.1
           - clang: 21
             os: ubuntu-22.04
             cuda_major_version: 11


tdavidcl:
looking at the blame it was @fodinabor that added Rocm 5.6.1 with clang 17 because of issues 3 years ago [Commit](https://github.com/AdaptiveCpp/AdaptiveCpp/commit/73334d18ffe4b7323af082df605f539f27da1dd1). And 17 was added back to the list latter (2 years ago) when the support for 18 in the CI was added [Commit](https://github.com/AdaptiveCpp/AdaptiveCpp/commit/9e7d291cb843596101cddd6804466eeeaf3fa03e).

We could maybe use 
 - rocm 5.4.3 for llvm 15, 16. 
 - rocm 5.6.1 for 17,18,19,20. 
 - rocm 6.3.1 from 21

But it's odd it was building fine during the llvm 21 PR.

illuhad:
I agree that splitting things up might be a good idea. Usually these kinds of issues happen when the containers need rebuilding. I triggered manual rebuilds, but it did not seem to help either...so let's first see if this PR does the trick.

EDIT: Since SSCP and SMCP have different compatibiltiy requirements, the ideal solution might run different configurations for both...

PR #2083, opened 2026-05-18
[CI] Leverage new Xe2 self-hosted runners and additional resources

illuhad:
We have a new dedicated machine with Intel Arc B580 (Xe2) that we can hook into CI. I've already set up the runner there.

This PR changes self-hosted workflow such that
- Testing on Intel with Level Zero/OpenCL is now done on the new Xe2 runner
- The new runner has more dedicated resources, so we can increase parallel build from 3 processes to 10.

The old runner is now exclusively used for AMD.
This should hopefully substantially improve our self-hosted CI throughput.

### DIFF
--- a/.github/workflows/linux-self-hosted.yml
+++ b/.github/workflows/linux-self-hosted.yml
@@ -188,7 +188,7 @@ jobs:
     - name: run PSTL Intel tests (L0)
       run: |
         cd ${GITHUB_WORKSPACE}/build/tests-sscp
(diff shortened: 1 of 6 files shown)


illuhad:
Mhm, looks like the new runner is not yet leveraging the discrete GPU, running on the iGPU instead. Perhaps we need to upgrade the OpenCL/L0 drivers in the container image.

illuhad:
Some failures found on new hardware:

- work group inclusive/exclusive scans fail tests for i8 type
- `inorder_queue_d2h_h2h_h2d_ordering` off by one failure

PR #2084, opened 2026-05-20
Increase score granularity in CPU device selector

EwanC:
The current CPU device selection logic for gives devices a score of either `1` or `-1` depending on if they are a CPU device or not.

However, if a user has multiple CPU backends in their build this does not allow for flexibility in selecting a specific CPU device with `ACPP_VISIBILITY_MASK` because the OpenMP device will also always be enabled and have the same score, leading to an undefined device being returned from the two devices with the same score. This patch favors the non-OpenMP device as is done in the default selector.

I experienced this when running hecbench executables compiled to use `sycl::cpu_selector_v` against an AdaptiveCpp library with 3 CPU backends enabled: OpenMP, OpenCL CPU, and Vulkan llvmpipe. In such a situation it's helpful to be able to use `ACPP_VISIBILITY_MASK` to specify which of 3 backhands to use.

### DIFF
--- a/include/hipSYCL/sycl/device_selector.hpp
+++ b/include/hipSYCL/sycl/device_selector.hpp
@@ -85,8 +85,15 @@ inline int select_accelerator(const device& dev) {
 }
 
 inline int select_cpu(const device& dev) {
-  if(dev.is_cpu())
-    return 1;
+  if(dev.is_cpu()) {
+    // Prefer non-OpenMP CPU device since the OpenMP backend cannot be disabled,
+    // so there would be no way to select e.g. an OpenCL CPU device
+    // using ACPP_VISIBILITY_MASK otherwise.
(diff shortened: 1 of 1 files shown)


EwanC:
Putting this into draft as it has implications for tests that I need to investigate

illuhad:
Don't mind the Intel failures, this is due to CI running now on Xe2 (Arc B580) since yesterday. You will need to rebase on #2083 once merged which fixes these issues.

EwanC:
> Don't mind the Intel failures, this is due to CI running now on Xe2 (Arc B580) since yesterday. You will need to rebase on #2083 once merged which fixes these issues.

ah thanks, rebasing  has indeed resolved the CI issues

illuhad (approved):

PR #2086, opened 2026-05-22
[Windows][SSCP][AMDGPU] Generate Windows AMDGPU libkernel bitcode with short wchar

blinkfrog:
On Windows, HIP device compilation emits LLVM IR with `wchar_size = 2`. AdaptiveCpp's generated AMDGPU libkernel bitcode used `wchar_size = 4`, which caused module flag conflicts during SSCP AMDGPU JIT linking.

This PR fixes this by generating AMDGPU libkernel bitcode on Windows with `-fwchar-type=short`.

### DIFF
--- a/src/libkernel/sscp/amdgpu/CMakeLists.txt
+++ b/src/libkernel/sscp/amdgpu/CMakeLists.txt
@@ -1,3 +1,10 @@
+set(amdgpu_libkernel_additional_args -nogpulib)
+
+if(WIN32)
+  list(APPEND amdgpu_libkernel_additional_args
+    -Xclang -fwchar-type=short)
+endif()
+
 if(WITH_LLVM_TO_AMDGPU_AMDHSA)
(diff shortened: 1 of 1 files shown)


src/libkernel/sscp/amdgpu/CMakeLists.txt
```diff
@@ -1,3 +1,10 @@
+set(amdgpu_libkernel_additional_args -nogpulib)
```
illuhad: Would it make sense to move all of the new code inside `if(WITH_LLVM_TO_AMDGPU_AMDHSA)`? As far as I can see, we don't need it outside of that if, right?

blinkfrog: Good point, you're right. I'll move it inside the `if(WITH_LLVM_TO_AMDGPU_AMDHSA)` block.

illuhad (approved):
Thanks!

PR #2087, opened 2026-05-22
[Windows][SSCP] Remove MSVC linker directives from generic device IR on Windows

blinkfrog:
On Windows, MSVC linker directives are injected into `llvm.linker.options` metadata in generic device IR. These directives are host-specific and can break SSCP JIT flows, for example causing HIP/ROCm bitcode linking to fail.

This PR fixes it by removing MSVC-style linker options from device IR during TargetSeparationPass.

This fix is currently applied only on Windows and removes linker options starting with `/`, matching MSVC linker directive style.

### DIFF
--- a/src/compiler/sscp/TargetSeparationPass.cpp
+++ b/src/compiler/sscp/TargetSeparationPass.cpp
@@ -234,7 +234,6 @@ void removeModuleFlag(llvm::Module& M, llvm::StringRef Name) {
   }
 }
 
-#ifdef _WIN32
 void removeLinkerOptionsByPrefixes(
     llvm::Module& M,
     llvm::ArrayRef<llvm::StringRef> Prefixes) {
@@ -278,7 +277,6 @@ void removeLinkerOptionsByPrefixes(
       NewNMD->addOperand(OptionNode);
   }
 }
-#endif
 
 std::unique_ptr<llvm::Module> generateDeviceIR(llvm::Module &M,
                                                const std::vector<std::string>& DynamicFunctions,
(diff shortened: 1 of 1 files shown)


src/compiler/sscp/TargetSeparationPass.cpp
```diff
@@ -234,6 +234,52 @@ void removeModuleFlag(llvm::Module& M, llvm::StringRef Name) {
...
+    llvm::Module& M,
+    llvm::ArrayRef<llvm::StringRef> Prefixes) {
+
+  llvm::NamedMDNode* NMD = M.getNamedMetadata("llvm.linker.options");
+  if(!NMD)
+    return;
+
+  auto StartsWithPrefix = [&](llvm::StringRef Option) {
+    for(auto Prefix : Prefixes) {
+      if(Option.starts_with(Prefix))
```
illuhad: better use `llvmutils::starts_with` from our `LLVMUtils.hpp` for compatibility (yes, I know that in practice we mainly target only new LLVM on Windows, but that's the pattern we have everywhere :) )

src/compiler/sscp/TargetSeparationPass.cpp
```diff
@@ -234,6 +234,52 @@ void removeModuleFlag(llvm::Module& M, llvm::StringRef Name) {
   }
 }
 
+#ifdef _WIN32
```
illuhad: I think we don't need to guard that function, since there's nothing Win32 specific here per se -- it's only the invocation that needs to be guarded.

illuhad (approved):
Thanks!

PR #2088, opened 2026-05-22
[SSCP][acpp] Auto-include hcf_registration.hpp for improved C++ modules support

### LINKED ISSUE #2085: [SSCP] "Could not obtain hcf kernel info" when instantiating exported template from a C++20 module
**Bug summary**
The following example program fails at runtime with this error message:
```
[AdaptiveCpp Error] from /home/zk/src/cxx/ext/AdaptiveCpp/src/runtime/cuda/cuda_queue.cpp:647 @ submit_sscp_kernel_from_code_object(): cuda_queue: Could not obtain hcf kernel info for kernel _Z18__acpp_sscp_kernelIN7hipsycl4glue15__sscp_dispatch11single_taskIZW13sycl_renderer1fIvEvvEUlvE_EEEvRKT_
```

If the explicit instantiation is uncommented it works as expect.

**To Reproduce**
`main.cxx`:
```C++
import mod;

int main()
{
  f<void>();
  return 0;
}
```

`mod.cxxm`:
```C++
module;

#include <sycl/sycl.hpp>

export module mod;

export template <typename>
void f()
{
  sycl::queue q{};
  q.single_task([]{});
  q.wait();
}

// export template void f<void>();
```

```
acpp -std=c++23 mod.cxxm --precompile -o mod.pcm
acpp -std=c++23 main.cxx -fmodule-file=mod=mod.pcm mod.pcm -o a.out
./a.out
```

**Setup**
How I installed AdaptiveCpp (branch develop, commit 223e9da3112b25a875448bf72b445011e231ecfd):
```
git clone https://github.com/AdaptiveCpp/AdaptiveCpp.git
cd AdaptiveCpp
mkdir build && cd build
cmake -GNinja -DCMAKE_BUILD_TYPE=Release -DLLVM_DIR=/usr/lib/llvm21/lib/cmake/llvm -DCMAKE_C_COMPILER=/usr/lib/llvm21/bin/clang -DCMAKE_CXX_COMPILER=/usr/lib/llvm21/bin/clang++ ..
ninja
sudo ninja install
```

LLVM 21 and clang 21 from official arch repositories:
```
❯ /usr/lib/llvm21/bin/clang --version
clang version 21.1.8
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/lib/llvm21/bin
```

CUDA 13.2.1

illuhad:
This is a workaround that fixes #2085

With C++ modules, it can happen that the instantiation site of a kernel is no longer guaranteed to have `hcf_registration.hpp` included. Without modules, this is the case because in order to submit a kernel, users must include `sycl.hpp`.
In the modules case, the missing include can cause the kernel not to be registered with the runtime.

This PR works around this problem by automatically including the registration header using `-include` for every source file.

Since the header is extremely small, this should not have measurable impact on compile times.

This can also fix issues with `--acpp-export-all`, which previously has also required that somewhere `sycl.hpp` was included. This should no longer be necessary.

### DIFF
--- a/bin/acpp
+++ b/bin/acpp
@@ -2334,11 +2334,17 @@ class compiler:
       file_replacement_map[s] = outfile
 
     result_args = []
+    consume_next = False
     for arg in args:
       if arg in file_replacement_map:
         result_args.append(file_replacement_map[arg])
       else:
-        result_args.append(arg)
+        if arg == "-include":
+          consume_next = True
+        elif consume_next:
+          consume_next = False
+        else:
+          result_args.append(arg)
     
     return result_args
 
--- a/bin/acpp
+++ b/bin/acpp
@@ -1947,6 +1947,8 @@ class llvm_sscp_invocation:
 
   def get_cxx_flags(self):
     flags = ["-D__ACPP_ENABLE_LLVM_SSCP_TARGET__",
+            "-include", os.path.join(self._config.acpp_include_path, 
+                "hipSYCL", "glue", "llvm-sscp", "hcf_registration.hpp"),
             "-Xclang", "-disable-O0-optnone", "-mllvm", "-acpp-sscp"]
     
     if self._config.is_export_all:
--- a/include/hipSYCL/glue/llvm-sscp/hcf_registration.hpp
+++ b/include/hipSYCL/glue/llvm-sscp/hcf_registration.hpp
@@ -9,26 +9,26 @@
  */
 // SPDX-License-Identifier: BSD-2-Clause
 
+// This file may be included early; avoid including std headers.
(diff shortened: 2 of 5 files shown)


Z0L1DK3K:
would it be possible to make the `add_sycl_to_target` CMake function also forcibly include the registration header?

illuhad:
What do you mean? This is implemented at the `acpp` level. Since `add_sycl_to_target` just boils down to an `acpp` invocation, you should also get it there. Do you see an issue?

Z0L1DK3K:
i see. no issues, i just didn't know that that's what `add_sycl_to_target` did.

PR #2094, opened 2026-05-28
[SSCP][Windows] Avoid console windows when launching JIT tools on Windows

### LINKED ISSUE #2089: Windows GUI applications briefly show console windows during SSCP/JIT compilation
When SSCP is used in a Windows GUI subsystem application (without an attached console), external tools (opt/llc/lld/llvm-spirv), invoked during JIT operations, visibly create console windows which are then closed.

The issue is not observable  when the host application itself is a console application, because child processes inherit the existng console.

In GUI applications, however, these windows become visible to the end user during runtime compilation, which is undesirable for shipped applications.

I experimented with replacing direct `llvm::sys::ExecuteAndWait()` calls with a small wrapper that launches processes using `CreateProcessW` with `CREATE_NO_WINDOW` on Windows, while still forwarding to `llvm::sys::ExecuteAndWait()` on non-Windows platforms.

This helped - I don't see these windows anymore.

I am not sure whether this direction would be considered acceptable, but if it is, I can open a PR. Or, are there any other solutions I could try?

blinkfrog:
This PR replaces direct `ExecuteAndWait()` calls in `llvm-to-backend` tool invocations with a small wrapper.

On Windows, this wrapper uses `ExecuteNoWait(..., DetachProcess=true)` + `Wait()`, which avoids visible console windows when AdaptiveCpp is used in GUI-subsystem applications.

For invocations without explicit redirects, stdout/stderr are redirected to temporary files and restored to `llvm::outs()` / `llvm::errs()` after completion. This preserves the original behaviour where tools still could still output to the parent console, while also providing valid standard streams for detached processes.

On non-Windows platforms, this wrapper forwards to `llvm::sys::ExecuteAndWait()`.

Fixes #2089

### DIFF
--- a/src/compiler/llvm-to-backend/Utils.cpp
+++ b/src/compiler/llvm-to-backend/Utils.cpp
@@ -258,7 +258,14 @@ int executeAndWait(
     std::optional<llvm::ArrayRef<llvm::StringRef>> Env,
     llvm::ArrayRef<std::optional<llvm::StringRef>> Redirects) {
 #if !defined(_WIN32) || LLVM_VERSION_MAJOR < 19
+#if LLVM_VERSION_MAJOR < 16
+  llvm::Optional<llvm::ArrayRef<llvm::StringRef>> EnvLegacy;
+  if(Env)
+    EnvLegacy = *Env;
+  return llvm::sys::ExecuteAndWait(Program, Args, EnvLegacy, Redirects);
+#else
   return llvm::sys::ExecuteAndWait(Program, Args, Env, Redirects);
+#endif
 #else
   std::string ErrMsg;
   bool ExecutionFailed = false;
--- a/src/compiler/llvm-to-backend/Utils.cpp
+++ b/src/compiler/llvm-to-backend/Utils.cpp
@@ -12,6 +12,12 @@
 #include "hipSYCL/compiler/llvm-to-backend/Utils.hpp"
 #include "hipSYCL/common/filesystem.hpp"
 
+#include <llvm/Support/Program.h>
+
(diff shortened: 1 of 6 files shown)


src/compiler/llvm-to-backend/Utils.cpp
```diff
@@ -246,5 +252,71 @@ std::string getRedistPackageBitcodePath(const std::string& backend) {
                                        std::vector<std::string>{"bitcode", backend});
 }
 
+int executeAndWait(
+    llvm::StringRef Program,
+    llvm::ArrayRef<llvm::StringRef> Args,
+    std::optional<llvm::ArrayRef<llvm::StringRef>> Env,
+    llvm::ArrayRef<std::optional<llvm::StringRef>> Redirects) {
+#ifndef _WIN32
```
illuhad: Maybe we should also have this depend on `LLVM_VERSION_MAJOR` given that the other code path requires a minimum LLVM version?

illuhad (commented):
Thanks, this looks good in principle. One remark/question :)

PS: I've invited you to the AdaptiveCpp github organization. This will make CI run automatically for your PRs, which might be useful.

src/compiler/llvm-to-backend/Utils.cpp
blinkfrog: Yes, it might be a good idea.
I checked this. The DetachProcess parameter is not present in LLVM 18, but is available in LLVM 19. I've added a version guard so this code path now requires minimum LLVM 19.

illuhad (approved):
Thanks!

PR #2099, opened 2026-06-10
Make CMake aware of bitcode library dependencies

al42and:
By default, `add_custom_command` is not aware of dependencies, and thus bitcode libraries don't get rebuilt if any of the includes is changed.


sbalint98:
Thank you for this, I wanted to make this change for a long time!

illuhad (approved):
Thanks!

PR #2101, opened 2026-06-12
CMake: Skip opt/llc/lld detection without SSCP

al42and:
Folks get creative with the way they install LLVM. But as long as we're building with the minimal profile, we can live without all these tools.


illuhad (approved):
Thanks!

PR #2102, opened 2026-06-13
Add .ixx as source file extension

M-Gjerde:
Adding  `.ixx` to the list of source file extensions recognized.
.ixx is default as a module interface unit in Jetbrain's CLION, and possibly also MSVC.


illuhad (approved):
Thanks!

PR #2105, opened 2026-06-17
[OpenMP] Improve NUMA behavior of memcpy and memset

illuhad:
Currently `queue::memcpy` and `queue::memset` are mapped to `memcpy()` and `memset()` calls in the worker thread. This can cause NUMA issues, because the copy/fills are potentially not processed in the same NUMA-distributed multi-threading environment that kernels use.

To address this, this PR maps the functions to `#pragma omp parallel for` sections which then call small `memcpy()` and `memset()`.

~Will test this later on the big machine to check if it actually improves NUMA behavior and behavior there, draft until then.~ Looks like it does what it's supposed to :) `queue.memcpy()` now performs just as well as a copy kernel on `omp` backend.

### DIFF
--- a/src/runtime/omp/omp_queue.cpp
+++ b/src/runtime/omp/omp_queue.cpp
@@ -57,6 +57,21 @@ namespace rt {
 
 namespace {
 
+std::size_t pick_default_group_size(std::size_t problem_size_in_largest_dim) {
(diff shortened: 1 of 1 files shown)


illuhad:
@TimThuering -- this PR might perhaps fix your AdaptiveCpp performance issue :)

PR #2106, opened 2026-06-18
Add free device memory query extension

blinkfrog:
This PR adds `ACPP_EXT_DEVICE_FREE_MEMORY`, an AdaptiveCpp extension for querying backend-reported free device memory.

The extension introduces:

- `sycl::aspect::AdaptiveCpp_free_memory`
- `sycl::info::device::AdaptiveCpp_free_memory`
- `ACPP_EXT_DEVICE_FREE_MEMORY`

Usage:

```
if(dev.has(sycl::aspect::AdaptiveCpp_free_memory)) {
  auto free_memory =
      dev.get_info<sycl::info::device::AdaptiveCpp_free_memory>();
}
```

The query currently maps to native backend APIs on CUDA and HIP.

The extension follows the existing SYCL pattern:

- device aspect for capability query
- `get_info()` for retrieving the value

Attempting to query `AdaptiveCpp_free_memory` on a device that does not support the query throws `sycl::errc::feature_not_supported`.

Despite I prefer a bit different naming (e. g. "available" instead of "free", and adding explicit "device" to distinguish from shared/host memory allocations), I decided to follow the existing Intel extension precedent (`free_memory`). This might be better for submitting extension idea as a KHR. However, I am open to renaming if there is a preferred direction.

The implementation intentionally does not use the existing device property infrastructure because:

- the value is dynamic rather than a static device property
- unsupported backends can simply inherit the default implementation without requiring changes to existing backend code

At the moment the query is implemented for CUDA and HIP only. OpenMP and OpenCL (on CPU) backends could potentially provide similar functionality, but their semantics are somewhat different and were intentionally left out of this initial implementation.


include/hipSYCL/runtime/hardware.hpp
```diff
@@ -139,6 +139,15 @@ class hardware_context
 
   virtual std::size_t get_platform_index() const= 0;
 
+  virtual bool supports_free_device_memory_query() const {
```
illuhad: Would it make sense to have this be a `device_support_aspect`? (see `has()` a couple lines above which we currently use to query capabilities).

include/hipSYCL/runtime/hardware.hpp
```diff
@@ -139,6 +139,15 @@ class hardware_context
...
+    return false;
+  }
+
+  virtual std::size_t get_free_device_memory() const {
+    assert(false && "free device memory query is not supported");
+    std::terminate();
```
illuhad: I don't think this should just `terminate()`. Maybe return `optional<>`?

Would it make sense to implement this as a `device_uint_property` (see `get_property()`)?

illuhad:
Thanks -- can you add a test case?

include/hipSYCL/runtime/hardware.hpp
blinkfrog: The idea was to avoid requiring every backend to explicitly handle an optional extension query. With the current approach unsupported backends automatically inherit the default behavior without terminating the program like it happens when `device_support_aspect` isn't handled in the backend. This seems convenient for existing and future backends. But I can switch this to `device_support_aspect` if the preference is to use the existing mechanisms.

include/hipSYCL/runtime/hardware.hpp
blinkfrog: Sure, changed this to use `optional<>`. As for `get_property` - the idea was similar - to avoid requiring existing unsupported backends to implement explicit handing of this query. If you prefer, I can switch this to use existing `get_property` mechanism.

blinkfrog:
>Thanks -- can you add a test case?

Sure, added a test.

include/hipSYCL/runtime/hardware.hpp
illuhad: I think we already have the situation though that not every query makes sense for all backends (e.g. OpenMP backend returns `numeric_limits<size_t>::max()` for some queries). I think the situation here is effectively the same:
A query does not make sense/is not supported for a certain device, so instead we return 0 or max (to be decided).

There are certainly advantages and tradeoffs with the strategy you use compared to what we do, but I'd prefer to fold everything into the existing mechanism if possible.

blinkfrog: Got it. I've updated the implementation to use the existing `device_support_aspect` and `device_uint_property` mechanisms as suggested. Unsupported backends (LevelZero, Metal, OpenCL, OpenMP) now explicitly report the query as unavailable. It looks like we can implement this query for Metal (`recommendedMaxWorkingSetSize` isn't exactly this but looks like a good approximation), and LevelZero, but I can't test these backends currently.

illuhad: That's alright, thank you! :)

include/hipSYCL/runtime/hardware.hpp
```diff
@@ -12,6 +12,7 @@
 #define HIPSYCL_HARDWARE_HPP
 
 #include <string>
+#include <optional>
```
illuhad: I think this is no longer needed?

include/hipSYCL/runtime/hardware.hpp
```diff
@@ -39,6 +40,7 @@ enum class device_support_aspect {
   work_item_independent_forward_progress,
   fp64,
   atomic64,
+  AdaptiveCpp_free_memory,
```
illuhad:
```suggestion
  free_memory,
```

This is an internal interface of the AdaptiveCpp runtime, we don't need to worry about name clashes with potential future SYCL APIs here (this is the main reason why extension APIs generally have `AdaptiveCpp_` prefix).

include/hipSYCL/runtime/hardware.hpp
```diff
@@ -104,7 +106,8 @@ enum class device_uint_property {
   backend_id,
 
   queue_priority_range_low,
-  queue_priority_range_high
+  queue_priority_range_high,
+  AdaptiveCpp_free_memory,
```
illuhad: Same here

illuhad (commented):
Thanks, only some nitpicks now :)

blinkfrog:
I also added support for the Vulkan backend: it now explicitly reports the extension as unsupported, like the other non-CUDA/HIP backends. The actual query implementation can be added later.

include/hipSYCL/runtime/hardware.hpp
blinkfrog: Sure, I fixed naming and removed `#include <optional>`.

illuhad (approved):
Thanks!

PR #2107, opened 2026-06-22
Make sure the ACPP_VISIBILITY_MASK in cbs tests is actually applied to the binary's environment.

fodinabor:
Without this, on my machine the tests acutally ended up running on my GPU :)


illuhad (approved):
Thank you! :)

PR #2113, opened 2026-06-23
[SSCP] Use correct barrier ordering

al42and:
A `relaxed` barrier does not actually guarantee visibility, so stronger semantics are needed.

Also add trailing barriers to some Metal operations, to avoid races between two scans one after another.

This was causing issues on Intel A770 with SPIR-V.

Potentially also affects Metal and Host.

CUDA, HIP, and Vulkan don't bother with barrier memory ordering, so there no impact is expected.

### DIFF
--- a/include/hipSYCL/pcuda/pcuda.hpp
+++ b/include/hipSYCL/pcuda/pcuda.hpp
@@ -161,12 +161,12 @@ inline int __pcuda_warp_size() {
 
 inline void __syncthreads() {
   PCUDA_BUILTIN_CALL(__acpp_sscp_work_group_barrier(
-      __acpp_sscp_memory_scope::work_group, __acpp_sscp_memory_order::relaxed));
+      __acpp_sscp_memory_scope::work_group, __acpp_sscp_memory_order::acq_rel));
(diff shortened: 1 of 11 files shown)


sbalint98:
Thank you, it looks good to me in general. I wonder if we should also change the behaviour for the pcuda `__syncthreads` `__syncwarp` calls as well?

illuhad:
> Thank you, it looks good to me in general. I wonder if we should also change the behaviour for the pcuda __syncthreads __syncwarp calls as well?

Good point. I think we probably should, yes.

resetius (approved):
LGTM

On Metal it's harmless but redundant: the barrier already fences threadgroup memory (MSL spec 6.9), so I'll drop the extra acq_rel fences in a follow-up PR. Not a blocker

resetius:
https://github.com/AdaptiveCpp/AdaptiveCpp/pull/2128

illuhad:
@sbalint98 I've added the change for `__syncthreads` and `__syncwarp`.

sbalint98 (approved):

PR #2128, opened 2026-06-24
[Metal] Drop redundant atomic_thread_fence around barriers

resetius:
A Metal barrier is already a full memory barrier for its address space, so wrapping threadgroup_barrier/simdgroup_barrier with extra atomic_thread_fence calls was pure overhead

Per MSL spec 6.9: "A barrier function acts as an execution and memory barrier."


illuhad (approved):
LGTM!

PR #2131, opened 2026-06-25
Propagate CMAKE_INSTALL_RPATH to llvm-spirv configuration

tneval:
Allows llvm-spirv configuration to utilize user-provided `CMAKE_INSTALL_RPATH`. This can be e.g. path to LLVM library if it is not visible to linker by default (custom LLVM builds).

I wonder why the `CMAKE_INSTALL_RPATH` is being manually overwritten in the modified CMakeLists.txt. Couldn't find it being used anywhere.


illuhad:
Thanks. What is the impact on deployment (`--acpp-deploy`) if llvm-spirv is allowed a custom rpath to a custom LLVM?

tneval:
Thanks for pointing this out -  I wasn't aware of the deployment feature. I'll check what the implications are.

tneval:
I'm afraid I dont' currently have enough time/knowledge to give a proper answer for this. 

Assuming we are running deployed app on another machine where it was compiled, I would assume that the `RPATH` gets simply ignored when it's not found. I'm not familiar how ACpp links to LLVM, but isn't same question valid for ACpp with custom LLVM + RPATH + deployment?

I'll get back to this at some point.

PR #2138, opened 2026-06-30
CMake: detect glibc 2.40 vectorized math symbols and Deploy sleefgnuabi library

thomasRoglin:
On aarch64, some tests rely on math functions that generate libmvec vectorized math symbols such as _ZGVnN2v_atanhf, which are only available starting with glibc 2.40.

This PR updates the CMake check to verify glibc >= 2.40 by searching for the _ZGVnN2v_atanhf symbol.

Additionally, CMake was searching for and deploying libsleef, while llvm-to-host links against libsleefgnuabi. This PR switches to find_library(sleefgnuabi) so the correct library is located and deployed.


illuhad (approved):
Thanks!

PR #2139, opened 2026-07-01
[VK] Re-fold instcombine byte-GEP chains before clspv

### LINKED ISSUE #2126: [VK] Fix array_index_subtract SSCP LIT test on Vulkan backend
The [array_index_subtract.cpp](https://github.com/AdaptiveCpp/AdaptiveCpp/blob/develop/tests/compiler/sscp/array_index_subtract.cpp) test f ails with either incorrect results or where device is lost during kernel execution.
   
The Same issue has also been observed using `clvk` with the following OpenCL kernel, and forcing physical addressing with `spir64` in the clspv invocation.
   
```c
 kernel void test_simple(global int* out, global int* in, int offset)
 { 
       size_t gid = get_global_id(0);
       out[gid] = in[offset-1];
 } 
 ```
   
Investigations so far:
    * Removing `-1` fixes the issue
    * Using `volatile*` to keep the load but remove the store fixes the issue.
    * Doing `out[id] = &in[offset-1] - &in` gives the correct offset
    * Generating logical memory model spirv from than physical addressing SPIRV gives the correct
      result in clvk

Further investigation needs done to root cause the issue so that it can either be fixed or the fundamental limitation documented.

### PR DESCRIPTION
**Pull request by crystarm**
Fixes #2126.

## Problem

The `array_index_subtract` SSCP LIT test fails on the Vulkan backend with
incorrect results or device loss. The same pattern also reproduces with `clvk`
using an OpenCL kernel with `--physical-storage-buffers` / `spir64` addressing.

Root cause: LLVM O3 `instcombine` canonicalises `in_ptr[size - 1]` from a
single typed GEP into a two-level byte-offset chain:

```llvm
; Before instcombine (O0):
%gep = getelementptr inbounds i32, ptr %in_ptr, i64 (size - 1)

; After instcombine (O3):
%5 = getelementptr i32, ptr %in_ptr, i64 %size   ; past-end pointer
%6 = getelementptr i8,  ptr %5,      i64 -4      ; byte GEP, negative offset
```

`clspv --physical-storage-buffers` miscompiles this pattern
(clspv issue [#1292](https://github.com/google/clspv/issues/1292)): seeing `i8`
as the source element type it infers `char` as the pointee type and emits four
separate byte-wide `OpLoad`s instead of a single `i32` `OpLoad`, producing
wrong results or a GPU crash.

## Fix

Add `FoldChainedGEPsPass` — a new LLVM function pass in the clspv backend
pipeline that runs after O3 and before `clspv` invocation. It recognises the
pattern:

```
outer = getelementptr i8, ptr inner, i64 C   (C = constant byte offset)
inner = getelementptr T,  ptr base,  i64 n
```

and folds it into a single typed GEP when `C` is an exact multiple of
`sizeof(T)` (which instcombine always guarantees for this transform):

```
combined = getelementptr T, ptr base, i64 (n + C/sizeof(T))
```

The pass is inserted in `optimizeFlavoredIR()` between `MemsetLoweringPass`
and `AddrSpaceCastRemovalPass`.

The `// UNSUPPORTED: vulkan || vk` guard is removed from the LIT test.

## Testing

- Direct compilation of the new pass with `clang++ -std=c++17` against the
  project headers: passes cleanly.
- The LIT test `tests/compiler/sscp/array_index_subtract.cpp` should now pass
  on Vulkan. A Vulkan-capable build with `clspv` is required to run it
  end-to-end.

## References

- clspv issue: https://github.com/google/clspv/issues/1292

### DIFF
--- a/src/compiler/llvm-to-backend/clspv/LLVMToCLSPV.cpp
+++ b/src/compiler/llvm-to-backend/clspv/LLVMToCLSPV.cpp
@@ -17,6 +17,7 @@
 #include "hipSYCL/compiler/llvm-to-backend/Utils.hpp"
 #include "hipSYCL/compiler/llvm-to-backend/clspv/AddrSpaceCastRemovalPass.hpp"
 #include "hipSYCL/compiler/llvm-to-backend/clspv/ConstantAddrSpacePass.hpp"
+#include "hipSYCL/compiler/llvm-to-backend/clspv/FoldChainedGEPsPass.hpp"
 #include "hipSYCL/compiler/llvm-to-backend/clspv/MemsetLoweringPass.hpp"
 #include "hipSYCL/compiler/llvm-to-backend/clspv/RemoveUnusedIntrinsicsPass.hpp"
 #include "hipSYCL/compiler/sscp/IRConstantReplacer.hpp"
@@ -400,6 +401,8 @@ bool LLVMToCLSPVTranslator::optimizeFlavoredIR(llvm::Module &M,
   MPM.addPass(
       llvm::createModuleToFunctionPassAdaptor(RemoveUnusedIntrinsicsPass()));
   MPM.addPass(llvm::createModuleToFunctionPassAdaptor(MemsetLoweringPass()));
+  MPM.addPass(
+      llvm::createModuleToFunctionPassAdaptor(FoldChainedGEPsPass()));
   MPM.addPass(
       llvm::createModuleToFunctionPassAdaptor(AddrSpaceCastRemovalPass()));
   MPM.addPass(ConstantAddrSpacePass());
--- a/tests/compiler/sscp/array_index_subtract.cpp
+++ b/tests/compiler/sscp/array_index_subtract.cpp
@@ -3,9 +3,6 @@
 // RUN: %acpp %s -o %t --acpp-targets=generic -O3
 // RUN: %t | FileCheck %s
 
-// https://github.com/AdaptiveCpp/AdaptiveCpp/issues/2126
-// UNSUPPORTED: vulkan || vk
-
 #include "common.hpp"
 #include <iostream>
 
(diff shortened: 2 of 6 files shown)


src/compiler/llvm-to-backend/clspv/LLVMToCLSPV.cpp
```diff
@@ -400,6 +401,12 @@ bool LLVMToCLSPVTranslator::optimizeFlavoredIR(llvm::Module &M,
   MPM.addPass(
       llvm::createModuleToFunctionPassAdaptor(RemoveUnusedIntrinsicsPass()));
   MPM.addPass(llvm::createModuleToFunctionPassAdaptor(MemsetLoweringPass()));
+  // Re-fold two-level byte-GEP chains produced by O3 instcombine back into a
+  // single typed GEP.  clspv --physical-storage-buffers miscompiles the i8
+  // byte-offset form (clspv issue #1292), emitting fragmented byte loads that
+  // yield wrong results or a device crash (AdaptiveCpp issue #2126).
```
EwanC: nit: You've documented this pass nicely in other places, so I'd drop the comment here

tests/compiler/sscp/array_index_subtract.cpp
```diff
@@ -4,7 +4,6 @@
 // RUN: %t | FileCheck %s
 
 // https://github.com/AdaptiveCpp/AdaptiveCpp/issues/2126
```
EwanC: Let's remove this comment too, it's out of place by itself without any context and you've already referenced the issue in the pass documentation.

EwanC (commented):
Thanks very much for the contribution, I have some trivial feedback but this is a nice patch. From trying this branch out locally I think it also solves some verification issues I've seen in a couple of benchmarks, but I need to confirm that.

The CI fail in macOS builds I believe is unrelated, I'm seeing it on a PR I have open too, and I'm looking into it.

crystarm:
@EwanC 
Done, removed both comments. Thanks for the feedback!! ദ്ദി(˵ •̀ ᴗ - ˵ ) ✧

EwanC (approved):

EwanC:
Merging this since the CI fail was from the LunarG website temporarily going offline so the SDK couldn't be downloaded, but the other Vulkan jobs passed and I've build + run this branch locally.

PR #2140, opened 2026-07-01
[VK] Support SSCP generated LLVM memmove intrinsic

### LINKED ISSUE #2110: [VK] Support SSCP generated memmove intrinsics
The SSCP compiler can generate `llvm.memmove` intrinsics for SYCL which are not currently consumed by the
clspv compiler used by the Vulkan backend

Example kernel that generate memmove
```cpp
int *device_ptr = sycl::malloc_device<int>(size, q);
q.memcpy(device_ptr, init, sizeof(init)).wait();

 cgh.single_task([=]() {
  for (size_t i = 0; i < size - 1; i++) {
     device_ptr[i] = device_ptr[i + 1];
    }
  });
}).wait(); 
```

This pattern is also seen in the `fdtd3d` HecBench benchmark which generates the intrinsic.

The best fix for this is to add support to clspv so that other users of this tool can benefit from the implementation, see https://github.com/google/clspv/pull/1613

EwanC:
Brings in the clspv change from PR https://github.com/google/clspv/pull/1613 to add support for the `llvm.memmove` intrinsic.

Includes a test that verifies a compile time known size and runtime known size for the intrinsic. As well as testing global and private pointers to the memmove.

Also update benchmark results based on this commit rebased on tip develop.

Closes https://github.com/AdaptiveCpp/AdaptiveCpp/issues/2110

### DIFF
--- a/tests/compiler/sscp/memmove.cpp
+++ b/tests/compiler/sscp/memmove.cpp
@@ -3,6 +3,9 @@
 // RUN: %acpp %s -o %t --acpp-targets=generic -O3
 // RUN: %t | FileCheck %s
 
+// Fails on CI with AMD EPYC 7763 OpenCL driver
+// UNSUPPORTED: opencl || ocl
+
 #include "common.hpp"
 #include <iostream>
 
--- a/tests/compiler/sscp/memmove.cpp
+++ b/tests/compiler/sscp/memmove.cpp
@@ -22,9 +22,7 @@ void dynamic_size_backward(sycl::queue q) {
 
   q.submit([&](sycl::handler &cgh) {
      cgh.single_task([=]() {
-       for (size_t i = 0; i < size - 1; i++) {
-         device_ptr[i] = device_ptr[i + 1];
-       }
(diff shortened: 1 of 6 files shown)


tests/compiler/sscp/memmove.cpp
```diff
@@ -0,0 +1,167 @@
...
+  int init[4] = {0, 1, 2, 3};
+
+  // Test global address space memmove
+  int *device_ptr = sycl::malloc_device<int>(size, q);
+  q.memcpy(device_ptr, init, sizeof(init)).wait();
+
+  q.submit([&](sycl::handler &cgh) {
+     cgh.single_task([=]() {
+       for (size_t i = 0; i < size - 1; i++) {
+         device_ptr[i] = device_ptr[i + 1];
```
illuhad: If you want to test `memmove` specifically, IIRC clang has a `__builtin_memmove()` that should be lowered to the LLVM intrinsic. Maybe that would be a more focused test?

EwanC: Good idea, updates the test to use that.

sbalint98 (approved):

PR #2148, opened 2026-07-12
Minor fix in filepath to enable compilation for llvm22. 

KaruroChori:
The executables can run just fine, not sure if there is any further hidden incompatibility.


KaruroChori:
I just saw https://github.com/AdaptiveCpp/AdaptiveCpp/pull/1986 which covers this.
Withdrawn.

PR #2149, opened 2026-07-12
[VK] Support event profiling

### LINKED ISSUE #2135: [VK] Implement SYCL event profiling
**Bug summary**
SYCL event profiling is enabled by passing `sycl::property::queue::enable_profiling`
to the `sycl::queue` constructor. Execution timestamps are then expected to be
available via `event.get_profiling_info<sycl::info::event_profiling::command_start>()`
and `event.get_profiling_info<sycl::info::event_profiling::command_end>()`.
However, the Vulkan backend throws at runtime when either is queried:
`Operation not profiled: No execution end timestamp available`

**To Reproduce**
```cpp
sycl::queue q{device, {sycl::property::queue::enable_profiling{}}};

auto event = q.submit([&](sycl::handler& h) {
    h.parallel_for(sycl::range<1>(N), [=](sycl::id<1> i) { ... });
});
event.wait();

// Throws: Operation not profiled: No execution end timestamp available
auto start = event.get_profiling_info<sycl::info::event_profiling::command_start>();
auto end   = event.get_profiling_info<sycl::info::event_profiling::command_end>();
```

EwanC:
Implements Vulkan backend support for profiling SYCL events.

Every command creates a single element query pool object per async instrumentation counter. This query pool element then a timestamp written to it by the device as part of command-buffer execution. Either
[VK_KHR_calibrated_timestamps](https://docs.vulkan.org/refpages/latest/refpages/source/VK_KHR_calibrated_timestamps.html) or [VK_EXT_calibrated_timestamps](https://docs.vulkan.org/refpages/latest/refpages/source/VK_EXT_calibrated_timestamps.html) extension is then used to to align device counter with the host side timepoint expected by SYCL.

I decided against a design with a single large query pool because the choice of query pool size would have been arbitrary, it is not possible to dynamically resize the pool if this limit is exceeded, and we don't free the instrumentation counters that often to allow the timestamps to return to the pool because a rt DAG node needs to outlive the user facing `sycl::event` object.

There is a change to the SYCL profiling test to use device USM rather than shared USM, as shared USM isn't supported by the Vulkan backend and it's semantics over device USM aren't required for the test.

Closes https://github.com/AdaptiveCpp/AdaptiveCpp/issues/2135

### DIFF
--- a/include/hipSYCL/runtime/vk/vk_hardware_manager.hpp
+++ b/include/hipSYCL/runtime/vk/vk_hardware_manager.hpp
@@ -12,7 +12,6 @@
 #pragma once
 
 #include "../hardware.hpp"
-#include "hipSYCL/runtime/vk/vk_allocator.hpp"
 #include <vulkan/vulkan_raii.hpp>
 
 namespace hipsycl {
@@ -39,10 +38,21 @@ enum feature_bits {
 };
 }; // namespace vk_device_features
 
+// Bit for each extension we can enable when creating a logical device,
+// improves extension querying compared to C string comparisons against
+// extension names.
+namespace vk_device_extensions {
+enum extension_bits {
+  khr_portability_subset = 1 << 0,
+  khr_calibrated_timestamps = 1 << 1,
+  ext_calibrated_timestamps = 1 << 2,
+};
+} // namespace vk_device_extensions
+
 class vk_hardware_context : public hardware_context {
 public:
   vk_hardware_context(const vk::raii::PhysicalDevice &, int dev_id,
-                      uint16_t features, bool portability_subset);
+                      uint16_t features);
   vk_hardware_context(vk_hardware_context const &) = delete;
   vk_hardware_context(vk_hardware_context &&) = default;
 
@@ -85,6 +95,10 @@ public:
   uint32_t get_subgroup_size() const { return _subgroup_size; }
   uint16_t get_phys_dev_features() const { return _physical_dev_features; }
 
+  bool are_extensions_enabled(uint16_t bits) const {
+    return (bits & _enabled_extensions) == bits;
+  }
+
 private:
   size_t global_mem_size() const;
 
@@ -100,6 +114,7 @@ private:
 
   int _dev_id;
   uint16_t _physical_dev_features;
+  uint16_t _enabled_extensions;
   std::unique_ptr<vk_allocator> _allocator;
 };
 
--- a/include/hipSYCL/runtime/vk/vk_hardware_manager.hpp
+++ b/include/hipSYCL/runtime/vk/vk_hardware_manager.hpp
@@ -42,7 +42,7 @@ enum feature_bits {
 // improves extension querying compared to C string comparisons against
 // extension names.
 namespace vk_device_extensions {
-enum extension_bits {
+enum extension_bits : uint16_t {
   khr_portability_subset = 1 << 0,
   khr_calibrated_timestamps = 1 << 1,
   ext_calibrated_timestamps = 1 << 2,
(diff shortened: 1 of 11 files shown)


tests/sycl/profiler.cpp
```diff
@@ -162,8 +162,8 @@ BOOST_AUTO_TEST_CASE(queue_profiling)
     BOOST_CHECK(t41 <= t43 && t42 <= t43);
 
     // usm
-    auto *src = sycl::malloc_shared<int>(n, queue);
-    auto *dest = sycl::malloc_shared<int>(n, queue);
+    auto *src = sycl::malloc_device<int>(n, queue);
+    auto *dest = sycl::malloc_device<int>(n, queue);
```
EwanC: This is breaking prefetch for cuda in CI, revisit this part of the change

EwanC: Fixed by preferring shared USM and only falling back to device if it's not availabe

src/runtime/vk/vk_profiling.cpp
```diff
@@ -0,0 +1,61 @@
...
+                              : dev.getCalibratedTimestampsEXT(timestamp_infos);
+  auto &timestamps = time_domains.first; // Drop max deviation
+  return calibrated_timestamps{timestamps[0], timestamps[1]};
+}
+
+uint64_t vk_get_query_pool_result(const vk::raii::QueryPool &query_pool) {
+  std::vector<uint64_t> data;
+  vk::Result result;
+  // Get timestamp data as 64-bit int rather than 32-bit default.
+  std::tie(result, data) = query_pool.getResults<uint64_t>(
```
sbalint98: Nit: Maybe we could do structured binding here rather?

include/hipSYCL/runtime/vk/vk_hardware_manager.hpp
```diff
@@ -39,10 +38,21 @@ enum feature_bits {
 };
 }; // namespace vk_device_features
 
+// Bit for each extension we can enable when creating a logical device,
+// improves extension querying compared to C string comparisons against
+// extension names.
+namespace vk_device_extensions {
+enum extension_bits {
```
sbalint98: I think we should type this enum to avoid some pedantic compiler warnings maybe in the future? 
```suggestion
enum extension_bits : uint16_t {
```

src/runtime/vk/vk_queue.cpp
```diff
@@ -143,6 +193,12 @@ result vk_queue::submit_memcpy(memcpy_operation &op, const dag_node_ptr &) {
             print_error(__acpp_here(), error_info{err_msg});
           }
 
+          if (_profiling) {
+            // Since we're dong work before command buffer starts executing,
+            // use an earlier host timestamp
+            _profiling->start_time->take_host_timestamp();
```
sbalint98, line 199: If I understand correctly `start_time`/`finish_time`/`` is only created in case we have the hint for start/finish timestamps. If only one of these would be requested at any point we would be de referencing a unset member here. Since currently these hints are always defined together this is not a current bug, however would lead to unexpected if behaviour if we would change this in the future

sbalint98 (commented):
Sorry for the long delay on this review it took me some time to read myself into this part of the code and the KHR extension. It looks good to me in general, I just had a few comments I wanted to clear up before we merge this. Thank you for this contribution!

src/runtime/vk/vk_queue.cpp
EwanC: Yeah that's a good point, the SYCL frontend sets both when profiling is enabled on a queue. However, if a SYCL extension made profiling more granular in future, or a new AdaptiveCpp frontend language was more granular, then this assumption would break.

I've updated all the control flow guards to be `if (_profiling && _profiling->{start,finish}_time)`

sbalint98 (approved):

PR #2150, opened 2026-07-13
Check for the Clang dev headers

sbalint98:
Currently we only requre the LLVM package to be installed, however the clang development headers are installed as part of a separate package. Since find_package(clang) caused problems in the past, we check if the headers exist and warn the user in case we cannot find them.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -309,7 +309,7 @@ if(BUILD_CLANG_PLUGIN)
     if(NOT CLANG_AST_HEADER_DIR)
       message(WARNING 
         "AdaptiveCpp requires clang development headers, but they were not found.\n"
-        "You might need to install the clang development package.\n"
+        "You might need to install the clang development package (e.g. libclang-dev or similar).\n"
       )
     endif()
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -300,7 +300,18 @@ if(BUILD_CLANG_PLUGIN)
...
-    #find_package(Clang REQUIRED)
+    
+    find_path(CLANG_AST_HEADER_DIR
+      NAMES clang/AST/ASTContext.h
+      PATHS ${LLVM_INCLUDE_DIRS}
+      DOC "Path to the Clang AST headers (usually provided by libclang-dev)"
+    )
+    if(NOT CLANG_AST_HEADER_DIR)
+      message(WARNING 
+        "AdaptiveCpp requires Clang AST development headers, but they were not found.\n"
```
illuhad:
```suggestion
        "AdaptiveCpp requires clang development headers, but they were not found.\n"
```

CMakeLists.txt
```diff
@@ -300,7 +300,18 @@ if(BUILD_CLANG_PLUGIN)
...
+        "You might need to install the clang development package.\n"
```
illuhad:
```suggestion
        "You might need to install the clang development package (e.g. libclang-dev or similar).\n"
```

illuhad (approved):

PR #2153, opened 2026-07-16
[SSCP] Rework HCF registration/unregistration infrastructure

### LINKED ISSUE #2042: nd_range kernel HCF lookup fails when called during static initialisation
Great work on SSCP and the generic JIT 👍  

I am integrating AdaptiveCpp for GPU-accelerated signal processing in [GNU Radio 4.0](https://github.com/fair-acc/gnuradio4), and the proof of concept works well.

## Problem

An `nd_range` kernel with `local_accessor` + `group_barrier` fails HCF lookup at runtime when invoked during static initialisation (global constructor). The same kernel called from `main()` works.

This affects any test framework that runs tests during static init (e.g. Boost.UT's `suite` constructor).

## Reproducer (see attached files below)

**`kernel.cpp`** — nd_range kernel in a separate TU:

```cpp
#include <sycl/sycl.hpp>
#include <cstdio>

extern "C" void run_kernel() {
    constexpr std::size_t N = 256, WG = 64;
    sycl::queue q;
    for (auto& d : sycl::device::get_devices())
        if (d.is_gpu()) { q = sycl::queue(d, sycl::property::queue::in_order{}); break; }

    float* buf = sycl::malloc_shared<float>(N, q);
    float* dst = sycl::malloc_shared<float>(N, q);
    for (std::size_t i = 0; i < N; ++i) buf[i] = float(i);

    q.submit([=](sycl::handler& h) {
        sycl::local_accessor<float, 1> local(sycl::range<1>{WG}, h);
        h.parallel_for(sycl::nd_range<1>{N, WG}, [=](sycl::nd_item<1> item) {
            auto lid = item.get_local_id(0);
            auto gid = item.get_global_id(0);
            local[lid] = buf[gid] * 2.0f;
            sycl::group_barrier(item.get_group());
            dst[gid] = local[lid];
        });
    }).wait();

    std::printf("dst[7] = %.1f (expected 14.0) %s\n", dst[7], dst[7] == 14.0f ? "PASS" : "FAIL");
    sycl::free(buf, q); sycl::free(dst, q);
}
```

**`main_pass.cpp`** — kernel called from main() → PASSES:

```cpp
#include <sycl/sycl.hpp>
extern "C" void run_kernel();
int main() { run_kernel(); }
```

**`main_fail.cpp`** — kernel called from global ctor → FAILS:

```cpp
#include <sycl/sycl.hpp>
extern "C" void run_kernel();
struct Init { Init() { run_kernel(); } };
static Init init;
int main() {}
```

**Build and run:**

```bash
ACPP="acpp -O2 -std=c++23 --acpp-targets=generic"
$ACPP -c kernel.cpp    -o kernel.o
$ACPP -c main_pass.cpp -o main_pass.o
$ACPP -c main_fail.cpp -o main_fail.o

$ACPP main_pass.o kernel.o -o test_pass && ./test_pass  # → PASS
$ACPP main_fail.o kernel.o -o test_fail && ./test_fail  # → "Could not obtain hcf kernel info"
```

## Notes

- `basic parallel_for` (no `local_accessor`) works from global ctors
- Only `nd_range` + `local_accessor` + `group_barrier` fails
- The kernel is in a separate TU — linking as SHARED library works around it
- No dependence on specific linker or flags (mold, bfd, lld all reproduce)

## Environment

- acpp from source (Clang 21.1.8), `--acpp-targets=generic`
- NVIDIA GeForce RTX 3070, CUDA 12.3
- Linux x86_64, openSUSE Tumbleweed

## Workaround

Compile the SYCL-kernel TU as a SHARED library (DSO has its own HCF context, independent of static-init ordering).

[kernel.cpp](https://github.com/user-attachments/files/26638650/kernel.cpp)
[main_fail.cpp](https://github.com/user-attachments/files/26638649/main_fail.cpp)
[main_pass.cpp](https://github.com/user-attachments/files/26638651/main_pass.cpp)
[run_tests.sh](https://github.com/user-attachments/files/26638648/run_tests.sh)

### LINKED ISSUE #2151: C++20 modules don't compile
**Bug summary**

I experimented with C++20 modules some time ago (1-2 months ago). At that time, the module interface compiled successfully, although the program later failed at runtime with `Could not obtain hcf kernel info for kernel`, like in the issue discussed in #2085.

After updating AdaptiveCpp, the modules no longer compile. I am not sure when this regression appeared (maybe after #2088? It stopped compiling just after that fix).

I get the error:

```
modsycl.cppm:1:1: error: 'module;' introducing a global module fragment can appear only at the start of the translation
      unit
    1 | module;
      | ^~~~~~~
modsycl.cppm:5:8: error: module declaration must occur at the start of the translation unit
    5 | export module modsycl;
      |        ^
modsycl.cppm:1:1: note: add 'module;' to the start of the file to introduce a global module fragment
    1 | module;
      | ^
2 errors generated.
```

Could it be because `#2088` injects a line via `-include C:\llvm.acpp.21\bin\..\include\AdaptiveCpp\hipSYCL\glue\llvm-sscp\hcf_registration.hpp`?

**To Reproduce**
Code:

```
module;

#include <sycl/sycl.hpp>

export module modsycl;

export void run_sycl_test()
{
}
```

Compile command:

```
python acpp --acpp-targets=generic -std=c++23 modsycl.cppm --precompile -o modsycl.pcm
```

**Expected behavior**

Module compiles

**Describe your setup**

AdaptiveCpp:

version-major: 25
    version-minor: 10
    version-patch: 0
    version-suffix: +git.98630f7a.20260622.branch.develop

 built as an external LLVM project (both LLVM 20 and 21 tried).

dry-run:

```
C:\llvm.acpp.21\bin\../bin\clang++.EXE -isystem C:\llvm.acpp.21\bin\..\include\AdaptiveCpp -D__OPENSYCL__ -D__HIPSYCL__ -D__ADAPTIVECPP__ -D__ACPP__ -D_USE_MATH_DEFINES -D__ACPP_ENABLE_LLVM_SSCP_TARGET__ -include C:\llvm.acpp.21\bin\..\include\AdaptiveCpp\hipSYCL\glue\llvm-sscp\hcf_registration.hpp -Xclang -disable-O0-optnone -mllvm -acpp-sscp -std=c++23 modsycl.cppm --precompile -o modsycl.pcm -LC:\llvm.acpp.21\bin\..\lib/ -lacpp-rt -lacpp-common
```

illuhad:
Use compiler magic to register/unregister HCF objects by modifying `llvm.global_ctors/global_dtors`, instead of using `-include` to include a stub everywhere.

This should fix the module issues reported in #2151 

EDIT: **Now also** adds another layer where, in case we cannot find a kernel, we do a full ELF scan for HCF data. This means that AdaptiveCpp should be now able to reliably run kernels in static constructors, without static initialization order issues! (only on Linux, it's already unholy enough there. Other people are free to contribute this functionality for Windows/MacOS)
-> potentially fixes #2042

draft because probably needs some additional code paths for early-opaque-ptr LLVM (LLVM 15).


blinkfrog:
I tested this with that reproducer, as well as with more complex code, and it works! The module compiles, the executable links, and it runs successfully, running kernels. Thanks!

The only thing is that during the final compilation/link command, the compiler prints a very large amount of LLVM IR and compilation becomes noticeably slower. Is this some debug output enabled somewhere, or is this something on my side?

illuhad:
@blinkfrog Thanks for testing! Yeah, there's still a debug output enabled in the current PR (which I forgot to remove). If you want to manually disable it, it's line 196 of `HcfRegistrationPass.cpp`.

illuhad:
Now also adds another layer where, in case we cannot find a kernel, we do a full ELF scan for HCF data. This means that AdaptiveCpp should be now able to reliably run kernels in static constructors, without static initialization order issues! (only on Linux, it's already unholy enough there. Other people are free to contribute this functionality for Windows/MacOS)
-> potentially fixes https://github.com/AdaptiveCpp/AdaptiveCpp/issues/2042

PR #2156, opened 2026-07-24
GDB pretty printer for ACPP SYCL data types 

### LINKED ISSUE #2046: GDB pretty printers
It could be of interest to have GDB pretty printer scripts for simplifying the viewing of e.g. sycl::vec classes (not having to unfold until getting to _storage) to see what's inside.

A python script shipped with the acpp repository so that one can easily use it

- Taking https://github.com/root-project/root/blob/master/misc/gdbPrinters/libCore.so-gdb.py as example and adapting it for sycl::vec
- Programming sth specific for QtCreator https://doc.qt.io/qtcreator/creator-debugging-helpers.html

EkremA:
Using a debugger such as GDB to inspect for example SYCL vectors yields information of form `v = {_data = {static effective_size = 4, static alignment = 16, _storage = {1, 2, 3, 4}}}`. 
This PR adds pretty printers for SYCL data types implemented by AdaptiveCpp to simplify the source code debugging process.

### DIFF
--- a/src/tools/debugging/acppvecprinter_alternativeregex.gdb.py
+++ b/src/tools/debugging/acppvecprinter_alternativeregex.gdb.py
@@ -1,0 +1,78 @@
+import gdb
+import re
+
+class AdaptiveCppVecPrinter:
+    """Pretty Printer for AdaptiveCpp sycl::vec<DataT, NumElements>"""
+    def __init__(self, val):
+        self.val = val
+        # Parse out Data Type and Element Count from the full template type string
+        self.sycldatatype = str(val.type)
+        # We understand a SYCL vector as in accordance with following regex  
+        match = re.search(r'(?:\w+::)*sycl::vec<', self.sycldatatype)
+        
+        # Fetch template arguments  
+        if match:
+            template_pattern = re.compile(
+                r"""(?:\w+::)*
+                sycl::vec<
+                    \s*
+                    (?P<T>[^,<>]+)
+                    \s*,\s*
+                    (?P<N>\d+)
+                    \s*,\s*
+                    (?P<VectorStorage>
+                        hipsycl::sycl::detail::vec_storage<
+                            \s*[^,<>]+\s*,\s*\d+\s*
+                        >
+                    )
+                    \s*>
+                """,
+                re.VERBOSE,
+            )
+            template_args = template_pattern.search(self.sycldatatype)
+            if template_args:
+                self.DataT = template_args.group("T")
+                try:
+                    self.NumElements = int(template_args.group("N"))
+                except:
+                    self.NumElements = 0
+        else:
+            self.DataT = None
+            self.NumElements = 0
+
+    def to_string(self):
+        if self.NumElements == 0:
+            return "sycl::vec (could not parse AdaptiveCpp dimensions)"
+        
+        # Resolve the internal storage member for AdaptiveCpp vs hipSYCL legacy
+        m_data = None
+        for field_candidate in ['_data', 'm_Data', 'data']:
+            try:
+                m_data = self.val[field_candidate]
+                break
+            except gdb.error:
+                continue
+        
+        elements = []
+        for i in range(self.NumElements):
+            try:
+                if m_data is not None:
+                    elements.append(str(m_data['_storage'][i]))
+            except gdb.error:
+                elements.append("?")
+
+        return f"sycl::vec<{self.DataT}, {self.NumElements}> = {{{', '.join(elements)}}}"
+ 
+
+    def display_hint(self):
+        return 'SYCL vector'
+
+def lookup_adaptivecpp_printer(val):
+    type_tag = str(val.type.unqualified())
+    if 'sycl::vec<' in type_tag or 'sycl::_V1::vec<' in type_tag:
+        return AdaptiveCppVecPrinter(val)
+    return None
+
+# Register with GDB session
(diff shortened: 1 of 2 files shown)


ferdymercury:
Thanks a lot for this!

I tried it locally, and I see

```
print sv
sycl::vec<float, 4> = {0, 0, 0, 4}
```

which looks great.

However, if I embed the pretty printer into my IDE,  if I look directly at the IDE window, it still shows the full structure.

image

I am used to other GDB pretty printers such as https://github.com/root-project/root/blob/master/misc/gdbPrinters/libCore.so-gdb.py where the script also changes the structure in the IDE window so that one does not need to unfold, etc.
 
Another suggested change:
- Add in the README how to use these printers. It should work easily enough so that the py scripts are installed into the lib folder and the user only needs to write:
`add-auto-load-safe-path /acppinstall/lib` into his `~/.gdbinit` file

src/tools/debugging/acppvecprinter_alternativeregex.gdb.py
```diff
@@ -0,0 +1,78 @@
...
+
+def lookup_adaptivecpp_printer(val):
+    type_tag = str(val.type.unqualified())
+    if 'sycl::vec<' in type_tag or 'sycl::_V1::vec<' in type_tag:
+        return AdaptiveCppVecPrinter(val)
+    return None
+
+# Register with GDB session
+gdb.pretty_printers.append(lookup_adaptivecpp_printer)
+print("AdaptiveCpp sycl::vec pretty printer successfully initialized.")
```
ferdymercury, line 78: GoogleAI suggests to use as a more standard structure by defining the children and display_hint members, see https://sourceware.org/gdb/current/onlinedocs/gdb.html/Pretty-Printing-API.html#Pretty-Printing-API

```suggestion
import gdb
import re

class AdaptiveCppVecPrinter:
    """Pretty Printer for AdaptiveCpp sycl::vec<DataT, NumElements>"""
    def __init__(self, val):
        self.val = val
        # Parse out Data Type and Element Count from the full template type string
        self.sycldatatype = str(val.type)
        # We understand a SYCL vector as in accordance with following regex  
        match = re.search(r'(?:\w+::)*sycl::vec<', self.sycldatatype)
        
        # Fetch template arguments  
        if match:
            template_pattern = re.compile(
                r"""(?:\w+::)*
                sycl::vec<
                    \s*
                    (?P<T>[^,<>]+)
                    \s*,\s*
                    (?P<N>\d+)
                    \s*,\s*
                    (?P<VectorStorage>
                        hipsycl::sycl::detail::vec_storage<
                            \s*[^,<>]+\s*,\s*\d+\s*
                        >
                    )
                    \s*>
                """,
                re.VERBOSE,
            )
            template_args = template_pattern.search(self.sycldatatype)
            if template_args:
                self.DataT = template_args.group("T")
                try:
                    self.NumElements = int(template_args.group("N"))
                except:
                    self.NumElements = 0
        else:
            self.DataT = None
            self.NumElements = 0

    def _get_data_member(self):
        # Resolve the internal storage member for AdaptiveCpp vs hipSYCL legacy
        for field_candidate in ['_data', 'm_Data', 'data']:
            try:
                return self.val[field_candidate]
            except gdb.error:
                continue
        return None

    def to_string(self):
        if self.NumElements == 0:
            return "sycl::vec (could not parse AdaptiveCpp dimensions)"

        m_data = self._get_data_member()
        elements = []
        for i in range(self.NumElements):
            try:
                if m_data is not None:
                    elements.append(str(m_data['_storage'][i]))
            except gdb.error:
                elements.append("?")

        return f"sycl::vec<{self.DataT}, {self.NumElements}>"
 
    def children(self):
        if self.NumElements == 0:
            return
        
        m_data = self._get_data_member()
        if m_data is None:
            return

        for i in range(self.NumElements):
            try:
                # GDB children must yield tuples of (name_string, gdb.Value)
                yield f"[{i}]", m_data['_storage'][i]
            except gdb.error:
                # Suppress or skip broken array indices
                pass

    def display_hint(self):
        # 'array' tells gdb to format the children as an indexed list
        return array

def lookup_adaptivecpp_printer(val):
    type_tag = str(val.type.unqualified())
    if 'sycl::vec<' in type_tag or 'sycl::_V1::vec<' in type_tag:
        return AdaptiveCppVecPrinter(val)
    return None

# Register with GDB session
gdb.pretty_printers.append(lookup_adaptivecpp_printer)
print("AdaptiveCpp sycl::vec pretty printer successfully initialized.")
```

src/tools/debugging/acppvecprinter_alternativeregex.gdb.py
```diff
@@ -0,0 +1,78 @@
...
+        return 'SYCL vector'
+
+def lookup_adaptivecpp_printer(val):
+    type_tag = str(val.type.unqualified())
+    if 'sycl::vec<' in type_tag or 'sycl::_V1::vec<' in type_tag:
+        return AdaptiveCppVecPrinter(val)
+    return None
+
+# Register with GDB session
+gdb.pretty_printers.append(lookup_adaptivecpp_printer)
```
ferdymercury, line 77: A more standard way of registration would be to follow

https://github.com/root-project/root/blob/086fcee4a52790ba84b962edf889b7c67ab4e46d/misc/gdbPrinters/libRooFitCore.so-gdb.py#L103

PR #2158, opened 2026-07-24
Android Cross Compilation Support

### LINKED ISSUE #2116: [VK] Test Vulkan backend on an Android device
One of the benefits of the AdaptiveCpp Vulkan backend is the possibility of running SYCL on Android devices, as they will expose a Vulkan driver. The Vulkan backend hasn't yet been tested on an Android device however. This task is to experiment running SYCL with the Vulkan backend on an Android device through the NDK/adb.

Arm Mali GPUs and Qualcomm Adreno GPUs are likely the most interesting candidates but we should first double check that the drivers have ShaderInt64 and ShaderVariablePointer capabilities.

Documenting the instructions to get this to work as well as the support status. Follow-on issue can be created to investigate issues.

EwanC:
Enable AdaptiveCpp to be cross compiled for Android using the Android NDK to support the host and Vulkan backends running with SSCP compilation.

This is a complicated to perform so has been documented extensively in `doc/installing.md`. However, I was successful in running a SYCL kernel on Qualcomm Adreno 750 GPU on Android!

The two main areas of changes are 1) to allow tooling binaries executed at runtime (opt, llc, lld, clspv) to be pointed to by an environment variable, as the host path used at CMake configure time won't apply. 2) CMake changes to select between the host clang cross compiler used by the NDK and an external cross compiled LLVM build to link with. As well as using the correct cross compile triple for the host libkernel builtin compilation

https://github.com/AdaptiveCpp/AdaptiveCpp/pull/2157 is also required but I opened that up as a standalone fix.

Closes https://github.com/AdaptiveCpp/AdaptiveCpp/issues/2116


illuhad:
Very cool to see!

Cross compilation questions have come up from time to time -- I've always suspected that it should in principle work, but I think you're the first one to actually get it to work :)

What I don't understand yet: Why do we actually need the `ACPP_LLC_PATH` etc environment variables? I understand that you need to point it at the cross-compiled LLVM tools. However, I think that these tools are not actually needed when building AdaptiveCpp. Therefore, would something like `cmake -DACPP_LLC_PATH=...` when building AdaptiveCpp not work? This would have the advantage that
- there are less footguns requiring runtime configuration on behalf of the end user
- deployment feature would "just work", which I believe with the current PR would be broken.
- potentially already works with the existing cmake infrastructure, which in principle already has `ACPP_LLC_PATH` etc.

?

EwanC:
> What I don't understand yet: Why do we actually need the `ACPP_LLC_PATH` etc environment variables? I understand that you need to point it at the cross-compiled LLVM tools. However, I think that these tools are not actually needed when building AdaptiveCpp. Therefore, would something like `cmake -DACPP_LLC_PATH=...` when building AdaptiveCpp not work? This would have the advantage that
> 
>     * there are less footguns requiring runtime configuration on behalf of the end user
> 
>     * deployment feature would "just work", which I believe with the current PR would be broken.
> 
>     * potentially already works with the existing cmake infrastructure, which in principle already has `ACPP_LLC_PATH` etc.

You're right we don't *need* it, I felt this approach added flexibility because it meant that at cross compilation time a user didn't need to know the exact layout on device of the tools, but that's not a use case I feel particularly strongly about. The flip side of that flexibility is indeed potential footguns. I've reverted this part of the change and documented using `-DACPP_FOO_PATH` after checking that approach does indeed works. Introducing the env vars was straightforward, so if there ever is a defined user need for this it should be easy to add back. 

I've also added a CI job in the most recent commit, although it only builds rather than runs a cross compiled SYCL example. I think this does have value though because since I opened this initial PR the [HCF rework](https://github.com/AdaptiveCpp/AdaptiveCpp/commit/f76b2cf89b5b6cc1484ed4c2ec850c76dafbaf58) commit broke the cross compile due to the `__linux__` ifdefs (needed a `__linux__` && !`__ANDROID__`) , so having some CI to check we don't accidentally regress this has some value.

jeongseok-meta:
I ran into the same build-host/target-tool split while enabling conda-forge's Linux AArch64 packages. I opened #2185 as a focused, platform-independent version of that separation: it makes both the Clang and `llvm-link` used for SSCP libkernel generation independently configurable, while preserving the target LLVM paths for the installed runtime.

If that approach looks suitable, this PR's Android-specific `ACPP_LLVMLINK_PATH` branch could use the generic `ACPP_BITCODE_LLVM_LINK` knob instead. The equivalent patch is already exercised by conda-forge's x86_64-to-AArch64 cross-builds across LLVM 16–19, so I wanted to coordinate rather than leave two overlapping interfaces.

.github/workflows/android.yml
```diff
@@ -0,0 +1,235 @@
...
+    name: Cross compile AdaptiveCpp (Android, LLVM ${{ matrix.llvm }})
+    runs-on: ubuntu-24.04
+    strategy:
+      fail-fast: false
+      matrix:
+        llvm: ['20']
+
+    steps:
+    - name: Checkout AdaptiveCpp
+      uses: actions/checkout@v4
```
sbalint98, line 22:
```suggestion
      uses: actions/checkout@v7
```

.github/workflows/android.yml
```diff
@@ -0,0 +1,235 @@
...
+      with:
+        submodules: 'recursive'
+        path: AdaptiveCpp
+
+    # ------------------------------------------------------------------
+    # Android NDK (cached)
+    # ------------------------------------------------------------------
+    - name: Cache Android NDK
+      id: cache-ndk
+      uses: actions/cache@v4
```
sbalint98, line 32:
```suggestion
      uses: actions/cache@v6
```

.github/workflows/android.yml
```diff
@@ -0,0 +1,235 @@
...
+                                libomp-${{ matrix.llvm }}-dev \
+                                llvm-${{ matrix.llvm }}-dev \
+                                clang-${{ matrix.llvm }} \
+                                lld-${{ matrix.llvm }}
+
+    # ------------------------------------------------------------------
+    # ccache shared across LLVM + AdaptiveCpp builds
+    # ------------------------------------------------------------------
+    - name: ccache
+      uses: hendrikmuhs/ccache-action@v1.2
```
sbalint98, line 75:
```suggestion
      uses: hendrikmuhs/ccache-action@v1.2.21
```

.github/workflows/android.yml
```diff
@@ -0,0 +1,235 @@
...
+        key: android-${{ matrix.llvm }}
+        restore-keys: |
+          android-
+
+    # ------------------------------------------------------------------
+    # Cross compiled LLVM for AArch64 / Android
+    # ------------------------------------------------------------------
+    - name: Cache cross compiled LLVM
+      id: cache-llvm
+      uses: actions/cache@v4
```
sbalint98, line 86:
```suggestion
      uses: actions/cache@v6
```

sbalint98 (commented):
Sorry for the delay with the review, The used actions need to be updated, and I think this should be re-based on top of #2185 after it was merged.

PR #2160, opened 2026-07-26
Set proper rpath property for tools

sbalint98:
Previously tools like acpp-hcf-tool or llvm-to-backend tools did not have their rpath set correctly. Setting these corretly will ensure that the user does not need to set LD_LIBRARY_PATH explicitely.

### DIFF
--- a/src/compiler/llvm-to-backend/CMakeLists.txt
+++ b/src/compiler/llvm-to-backend/CMakeLists.txt
@@ -4,7 +4,6 @@ set(LLVM_TO_BACKEND_INCLUDE_DIRS
   ${CMAKE_CURRENT_SOURCE_DIR}
   ${ACPP_BINARY_ROOT}/include)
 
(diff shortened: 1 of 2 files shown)


src/tools/acpp-pcuda-pp/CMakeLists.txt
```diff
@@ -6,4 +6,6 @@ target_include_directories(acpp-pcuda-pp PRIVATE
     ${HIPSYCL_SOURCE_DIR}/include
     ${PROJECT_BINARY_DIR}/include)
 
+set_target_properties(acpp-pcuda-pp PROPERTIES INSTALL_RPATH ${base}/../lib/)
```
illuhad: `acpp-pcuda-pp` has no dependencies apart from libstdc++. Why do we even need that here?

sbalint98: Sorry this does not belong here indeed. removed.

PR #2161, opened 2026-07-26
Use internal linkage for reflection TU registration

### LINKED ISSUE #2159: Dynamic functions fail to resolve reflected functions in multi-TU -O0 builds
Note, I reworked this issue with _Codex_, particularly to provide the reproducable example with acpp's reflection logic rather than a more elaborate example failing at JIT execution.

**Bug summary**

In a multi-translation-unit generic build at `-O0`, host-side function-symbol reflection depends on source/link order. A function enabled for reflection in one translation unit can still resolve to `nullptr` at runtime.

Dynamic functions use this lookup internally, so the same failure can surface from `dynamic_function_config::define()` as:

```text
dynamic_function_config: Could not resolve function symbol
```

The reproducer below exercises only AdaptiveCpp's function-reflection layer. It does not require a SYCL queue or device.

**To Reproduce**

`main.cpp`:

```cpp
#include "hipSYCL/glue/reflection.hpp"

bool other_tu_resolves();

int main() { return other_tu_resolves() ? 0 : 1; }
```

`other.cpp`:

```cpp
#include "hipSYCL/glue/reflection.hpp"

void reflected_from_other_tu() {}

bool other_tu_resolves() {
  hipsycl::glue::reflection::enable_function_symbol_reflection(
      &reflected_from_other_tu);
  return hipsycl::glue::reflection::resolve_function_name(
             &reflected_from_other_tu) != nullptr;
}
```

Compile the empty-reflection TU first:

```bash
acpp main.cpp other.cpp -o repro --acpp-targets=generic -O0
./repro
echo $?
```

When this executable runs, resolve_function_name() returns nullptr, so it returns `1` (expected: `0`).

For comparison, reversing the input order at -O0 or compiling with -O3 produces executables that return 0. 

```bash
acpp other.cpp main.cpp -o repro-reversed --acpp-targets=generic -O0
./repro-reversed
echo $?

acpp main.cpp other.cpp -o repro-o3 --acpp-targets=generic -O3
./repro-o3
echo $?
```

**Expected behavior**

`resolve_function_name()` should find functions enabled for reflection in every translation unit, independently of optimization level and link order. All three executables should return 0.

**Describe your setup**

- AdaptiveCpp `develop` commit `3536de0f7642103cb8bd7dbe4374e5500fc57591`
- Clang 19.1.7
- Linux x86_64
- `--acpp-targets=generic`
- No runtime backend is selected because the reproducer creates no queue and submits no kernel.

The following was observed directly in the affected minimized `-O0` executable using `llvm-nm-19 -C` and `llvm-objdump-19`:

- two local (`t`) functions named `__acpp_reflection_init_registered_function_pointers` are present;
- only one weak (`W`) `__acpp_reflection_tu_init_trigger` constructor is present; and
- the retained constructor calls the empty TU-local initializer. The other initializer contains the missing function registration but is not called.

**Additional context**

I can submit an LLM-assisted patch and regression test as a PR.

illuhad:
Fixes #2159 by using internal linkage for reflection TU registration. The current design is a bit hacky in general; in long term we should perhaps consider doing something similar as in PR #2153 (or merge reflection into that infrastructure somehow?).

Also cleans up some logic in the corresponding LLVM pass that I noticed when going through the code.

@cbyrohl I took the liberty to add your reproducer as a test case. Hope that's okay!

### DIFF
--- a/tests/compiler/reflection/multi-tu-registration/main.cpp
+++ b/tests/compiler/reflection/multi-tu-registration/main.cpp
@@ -1,0 +1,16 @@
+// RUN: %acpp %s %S/other.cpp -o %t --acpp-targets=generic
+// RUN: %t | FileCheck %s
+// RUN: %acpp %s %S/other.cpp -o %t --acpp-targets=generic -O3
+// RUN: %t | FileCheck %s
+// RUN: %acpp %s %S/other.cpp -o %t --acpp-targets=generic -g
+// RUN: %t | FileCheck %s
+
+#include <iostream>
+#include "hipSYCL/glue/reflection.hpp"
+
+bool other_tu_resolves();
+
+int main() {
+  // CHECK: 1
+  std::cout << other_tu_resolves() << std::endl;
+}
--- a/include/hipSYCL/glue/reflection.hpp
+++ b/include/hipSYCL/glue/reflection.hpp
@@ -75,6 +75,7 @@ extern "C" void __acpp_function_annotation_needs_function_ptr_argument_reflectio
 
 
 struct __acpp_reflection_tu_init_trigger {
+  __attribute__((internal_linkage))
   __acpp_reflection_tu_init_trigger() {
     __acpp_reflection_init_registered_function_pointers();
   }
--- a/src/compiler/reflection/FunctionNameExtractionPass.cpp
+++ b/src/compiler/reflection/FunctionNameExtractionPass.cpp
@@ -99,27 +99,23 @@ llvm::PreservedAnalyses FunctionNameExtractionPass::run(llvm::Module &M, llvm::M
     }
   }
 
-  //declare explicitly __acpp_reflection_associate_function_pointer
   static const char* ReflectionAssociateFunctionPointer = "__acpp_reflection_associate_function_pointer";
-  llvm::SmallVector<llvm::Type*> ParamTs;
-  // function pointer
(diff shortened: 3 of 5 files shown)


tests/compiler/reflection/multi-tu-registration/main.cpp
```diff
@@ -0,0 +1,16 @@
+// RUN: %acpp %s -o %t --acpp-targets=generic
+// RUN: %t | FileCheck %s
+// RUN: %acpp %s -o %t --acpp-targets=generic -O3
+// RUN: %t | FileCheck %s
+// RUN: %acpp %s -o %t --acpp-targets=generic -g
+// RUN: %t | FileCheck %s
```
cbyrohl, line 6: isnt `%S/other.cpp` missing in those acpp RUN commands respectively?

cbyrohl (changes requested):
Looks good. I just wonder whether the acpp RUN commands lack the other.cpp includes

tests/compiler/reflection/multi-tu-registration/main.cpp
illuhad: Yep, it is indeed missing. CI hasn't finished yet on this file -- we'll get an error back :)

illuhad: Fixed -- let's see if the test runs correctly now.

PR #2164, opened 2026-07-27
[SSCP] Fix sincos remapping

### LINKED ISSUE #2162: Generic SSCP remaps sincos/sincosf to missing __acpp_sscp_sincos symbols
🤖 _Posted by Codex on behalf of @cbyrohl_

**Bug summary**

With the generic SSCP target, a kernel containing Clang's __builtin_sincosf or __builtin_sincos compiles successfully, but JIT module loading fails because __acpp_sscp_sincos_f32 or __acpp_sscp_sincos_f64 remains unresolved. I reproduced this on both CUDA and the OpenMP host backend using current develop.

**To Reproduce**

Minimal FP32 reproducer:

    #include <sycl/sycl.hpp>

    int main() {
      float* values = sycl::malloc_shared<float>(3, queue);
      values[0] = 0.625f;

      queue.single_task([=]() {
        __builtin_sincosf(values[0], &values[1], &values[2]);
      }).wait_and_throw();

      sycl::free(values, queue);
    }

Compile and run:

    acpp --acpp-targets=generic -O3 repro.cpp -o repro
    ACPP_VISIBILITY_MASK=cuda ./repro

CUDA reports:

    ptxas fatal : Unresolved extern function '__acpp_sscp_sincos_f32'

Changing the values to double and calling __builtin_sincos reports the corresponding __acpp_sscp_sincos_f64 symbol.

The OpenMP host backend reproduces both variants as well when using a default queue and ACPP_VISIBILITY_MASK=omp. Loading the generated JIT shared library then reports an undefined __acpp_sscp_sincos_f32 or __acpp_sscp_sincos_f64 symbol.

**Expected behavior**

The kernel should JIT-compile, load, and write the sine and cosine results for both float and double.

**Describe your setup**

- AdaptiveCpp develop at 3536de0f7642103cb8bd7dbe4374e5500fc57591 (built and installed in an isolated prefix)
- AdaptiveCpp version string: 25.10.0+git.3536de0.20260726.branch.develop
- compiler plugin: Clang/LLVM 21.1.8
- host compiler: GCC 16.1.1
- CUDA toolkit: 12.9.86
- GPU: NVIDIA GeForce RTX 4060 Ti, sm_89
- CUDA driver version reported by AdaptiveCpp: 13030
- configured with the full compiler feature profile and CUDA/OpenMP runtime backends
- application compiled with: acpp --acpp-targets=generic -O3 repro.cpp -o repro

The final optimized CUDA IR dumped with ACPP_S2_DUMP_IR_FINAL contains:

    declare void @__acpp_sscp_sincos_f32(float, ptr, ptr)
    call void @__acpp_sscp_sincos_f32(float %value, ptr %sin_out, ptr %cos_out)

The same behavior was first observed in a downstream THOR CUDA kernel; the downstream reproduction and workaround are documented in https://github.com/cbyrohl/thor/pull/538.

**Additional context**

The exact-symbol issue/PR search found no existing bug report for this module-loading failure. AdaptiveCpp PR #2021 is related to sincos performance/API work, but is not this reproducer.

cbyrohl:
🤖 _Posted by Codex on behalf of @cbyrohl_

Fixes #2162.

## Summary

Generic SSCP remapping currently sends compiler-emitted `sincos`/`sincosf` calls to `__acpp_sscp_sincos_f64`/`__acpp_sscp_sincos_f32`, but those symbols are not part of the SSCP builtin ABI. This leaves unresolved symbols during JIT/module loading. With newer LLVM versions and fast-math, the frontend may instead emit the aggregate-returning `llvm.sincos.f32`/`llvm.sincos.f64` intrinsic.

This change handles both forms with internal always-inline adapters:

- libc `sincos` ABI: write the sine and cosine results through output pointers;
- LLVM intrinsic ABI: return the `{sin, cos}` aggregate;
- lower both adapters through the existing `__acpp_sscp_sin_*` and `__acpp_sscp_cos_*` builtins.

The adapter ABIs are validated before rewriting, and no new SSCP runtime symbol is introduced.

## Validation

- Focused `builtin_remapping.cpp` lit test passed on the OpenMP backend in all four existing modes: default/O0, O3, O3 with `-ffast-math`, and debug.
- The same four modes passed in FP32 and FP64 on CUDA using an RTX 4060 Ti.
- Final optimized CUDA IR contains neither `sincos` calls nor adapter references.
- `clang-tidy` completed without a new diagnostic in the changed source.

### DIFF
--- a/src/compiler/sscp/StdBuiltinRemapperPass.cpp
+++ b/src/compiler/sscp/StdBuiltinRemapperPass.cpp
@@ -11,7 +11,9 @@
 #include "hipSYCL/compiler/sscp/StdBuiltinRemapperPass.hpp"
 #include "hipSYCL/common/debug.hpp"
 
+#include <llvm/IR/Constants.h>
 #include <llvm/IR/Function.h>
+#include <llvm/IR/IRBuilder.h>
 #include <llvm/IR/Module.h>
 
 #include <unordered_set>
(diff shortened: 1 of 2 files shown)


illuhad:
Why can't we just implement a new `__acpp_sscp_sincos_*` for all backends? I suspect that
- this would be cleaner and more maintainable than extra compiler logic to special-case sincos
- and it would allow us to invoke native `sincos()` intrinsics on backends that have them, potentially allowing for better performance.

cbyrohl:
Right! I tried [here](https://github.com/cbyrohl/AdaptiveCpp/tree/fix/sscp-native-sincos). Let me open a new PR.

PR #2165, opened 2026-07-27
[SSCP] Fold basic_parallel_for range guard for non-padded launches

mikaoP:
`basic_parallel_for` wraps every kernel in an `if (id in range)` check to skip the leftover work-items you get when the global range is rounded up to the group size. Problem: that branch also stops the backend from vectorizing the kernel body — even when the range divides evenly and there are no leftover work-items at all.

I ran into this comparing AdaptiveCpp vs DPC++ codegen on pocl: a simple `a[i] = sin(a[i]) + cos(a[i])` came out scalar with AdaptiveCpp but vectorized with DPC++. I traced the difference to this range check — dropping it from the AdaptiveCpp
kernel makes pocl vectorize it too.

So this drops the check when the launch is exact (global range is a multiple of the group size). The runtime detects that and specializes it away via a kernel argument, the JIT folds the branch, and the body vectorizes. Padded launches keep the check, so results are unchanged.

I'm not totally sure I'm not missing something, or whether there's a cleaner way to do this — happy to hear your thoughts.

### DIFF
--- a/include/hipSYCL/glue/llvm-sscp/sscp_kernel_launcher.hpp
+++ b/include/hipSYCL/glue/llvm-sscp/sscp_kernel_launcher.hpp
@@ -139,17 +139,21 @@ class basic_parallel_for {
 public:
   basic_parallel_for(const UserKernel &k,
                      sycl::range<Dimensions> execution_range)
-      : _k{k}, _range{execution_range} {}
+      : _acpp_exact_range{0}, _k{k}, _range{execution_range} {}
 
   [[clang::annotate("hipsycl_kernel_dimension", Dimensions)]]
   void operator()() const {
     auto this_item = sycl::detail::make_item<Dimensions>(
       sycl::detail::get_global_id<Dimensions>(), _range
     );
-    if(item_is_in_range(this_item, _range))
+    // Specialized to 1 for non-padded launches so this range check folds away
+    // and the body vectorizes; 0 (check kept) when the launch is padded.
+    if(_acpp_exact_range || item_is_in_range(this_item, _range))
       _k(this_item);
   }
 private:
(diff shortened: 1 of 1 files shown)


illuhad:
Possible duplicate of #1854 ? Can you check whether this PR also resolves the problem? This PR is a bit lighter in terms of impact compared to what you are doing because it does not require additional kernel specialization on the same device.

include/hipSYCL/glue/llvm-sscp/sscp_kernel_launcher.hpp
```diff
@@ -139,17 +139,21 @@ class basic_parallel_for {
...
     );
-    if(item_is_in_range(this_item, _range))
+    // Specialized to 1 for non-padded launches so this range check folds away
+    // and the body vectorizes; 0 (check kept) when the launch is padded.
+    if(_acpp_exact_range || item_is_in_range(this_item, _range))
       _k(this_item);
   }
 private:
+  // Kernel argument 0 (must stay the first member); set in invoke().
+  int _acpp_exact_range;
```
illuhad, line 156: If you define this as `sycl::detail::__acpp_sscp_emit_param_type_annotation_specialized<int>` or `sycl::specialized<int>`, the compiler will automatically specialize, and you won't have to worry about the argument number.
Maybe it might be even better at the abstraction level of the kernel launcher to funnel that information through the JIT reflection API, but I have not yet thought that through.

mikaoP:
> Possible duplicate of #1854 ? Can you check whether this PR also resolves the problem? This PR is a bit lighter in terms of impact compared to what you are doing because it does not require additional kernel specialization on the same device.

You're right. It solves my use case :)

sbalint98:
Closing since #1854 has been merged.

PR #2172, opened 2026-07-30
[ze] Initialize stype/pNext of Level Zero property structs before API calls

fateshelled:
## Description

Initialize the `stype` and `pNext` fields of Level Zero property structures
before passing them to the respective `ze*Get*Properties` API calls.

Per the Level Zero specification, the `stype` field must be set to the
appropriate structure type enumeration value and `pNext` must be set to
`nullptr` (unless chaining an extension structure) before any API call
that takes the structure. Missing initialization can lead to undefined
behavior or crashes on some runtimes.

### Changes

- `ze_allocator.cpp`: Initialize `stype` for `ze_memory_allocation_properties_t`
  before `zeMemGetAllocProperties`
- `ze_hardware_manager.cpp`: Initialize `stype`/`pNext` for:
  - `ze_device_properties_t` before `zeDeviceGetProperties`
  - `ze_device_compute_properties_t` before `zeDeviceGetComputeProperties`
  - `ze_device_memory_properties_t` (each element) before `zeDeviceGetMemoryProperties`
  - `ze_driver_properties_t` before `zeDriverGetProperties`

### DIFF
--- a/src/runtime/ze/ze_allocator.cpp
+++ b/src/runtime/ze/ze_allocator.cpp
@@ -128,6 +128,7 @@ bool ze_allocator::is_usm_accessible_from(backend_descriptor b) const {
 result ze_allocator::query_pointer(const void* ptr, pointer_info& out) const {
 
   ze_memory_allocation_properties_t props;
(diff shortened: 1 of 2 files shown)


illuhad (approved):
Thanks!

PR #2178, opened 2026-08-05
[OMP] Allow `numBytes == 0` in memset

nilsfriess:
I don't find anything in the SYCL reference that would forbid this. A library that I'm currently trying to make work with AdaptiveCpp uses memsets with size == 0 in several places so I figured fixing this here might be easiest.

Draft because I'm not sure what to do with the instrumentation stuff in `omp_queue::submit_memset`.

### DIFF
--- a/src/runtime/omp/omp_queue.cpp
+++ b/src/runtime/omp/omp_queue.cpp
@@ -668,6 +668,7 @@ result omp_queue::submit_prefetch(prefetch_operation &op, const dag_node_ptr& no
 result omp_queue::submit_memset(memset_operation &op, const dag_node_ptr& node) {
   void *ptr = op.get_pointer();
   std::size_t bytes = op.get_num_bytes();
+  if (bytes == 0) return make_success();
   int pattern = op.get_pattern();
 
   if (!ptr) {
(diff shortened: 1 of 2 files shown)


sbalint98:
I think this should work according to the WG. It did not make the spec yet apparently : https://github.com/KhronosGroup/SYCL-Docs/issues/883

src/runtime/omp/omp_queue.cpp
```diff
@@ -668,6 +668,7 @@ result omp_queue::submit_prefetch(prefetch_operation &op, const dag_node_ptr& no
 result omp_queue::submit_memset(memset_operation &op, const dag_node_ptr& node) {
   void *ptr = op.get_pointer();
   std::size_t bytes = op.get_num_bytes();
+  if (bytes == 0) return make_success();
```
illuhad, line 671: I suspect that other backends might also be affected by this, not just omp.

So, if we want to support this universally, the better approach is probably to implement this at the SYCL layer (`handler.hpp`) by returning a barrier in case the size is 0.
This is also what we do for 0-sized kernels: https://github.com/AdaptiveCpp/AdaptiveCpp/blob/3452d0fa3e764ab603b73612531716bbad5ad947/include/hipSYCL/sycl/handler.hpp#L301

PR #2186, opened 2026-08-12
[VK] Handle kernel arguments set via uniform buffers

EwanC:
This patch updates the Vulkan backend to correctly handle when a SPIR-V kernel expects arguments to be passed via uniform buffers instead of push constants. This requires creating uniform buffers and descriptor sets that can be used on a kernel invocation without conflicting with other invocations of the same kernel.

This path is less efficient than pushing constants to set arguments but is used as workaround for the Qualcomm Adreno GPU Vulkan driver on Android which struggles to correctly execute well formed SPIR-V code where ulong device buffer addresses are passed in the push constant struct.

The clspv commit is also bumped to bring in the clspv change from merged PR https://github.com/google/clspv/pull/1623

### DIFF
--- a/doc/install-vulkan.md
+++ b/doc/install-vulkan.md
@@ -20,7 +20,7 @@ the generic SSCP compilation flow.
   `VK_KHR_buffer_device_address` respectively.
 * The [LunarG Vulkan SDK](https://vulkan.lunarg.com/sdk/home) versions 1.4 or later
   for Vulkan loader, layers, headers, and other tools.
-* A `clspv` executable from commit `88d8ff71000bf995493c8daaed865ec1ac216309`.
+* A `clspv` executable from commit `3d017c6d751cdabed1a5051b20c6bf5f71e8fd32`.
 * Linux, macOS, and Windows operating systems are tested in CI. Ubuntu 22.04 and later are the
   tested distributions for Linux. MacOS 15.7.4 is tested CI with MoltenVK. Windows server 2022
   is used with llvmpipe in CI. However on Windows there are sporadic failures in the `group_functions` suite
@@ -82,6 +82,10 @@ the features used in the kernel:
   instructions to be used in the implementation of SYCL sub-group builtins.
 * `VK_SUBGROUP_FEATURE_SHUFFLE_BIT` - Allows SPIR-V `CapabilityGroupNonUniformShuffle`
   instructions to be used in the implementation of SYCL sub-group builtins.
+* `uniformAndStorageBuffer8BitAccess` - Allows 8-bit arguments to be passed to
+  SPIR-V kernels via uniform buffers.
+* `uniformAndStorageBuffer16BitAccess` - Allows 16-bit arguments to be passed to
+  SPIR-V kernels via uniform buffers.
 
 ## Building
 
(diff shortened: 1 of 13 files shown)


doc/install-vulkan.md
```diff
@@ -310,11 +314,22 @@ each work-group size variant of a kernel. At adaptivity level 0, a specializatio
...
 
-For small numbers of kernel arguments, no buffer descriptors may be required and all
-arguments can be set pushing push constants before the invocation of the Vulkan
-command-buffer containing the kernel. For larger numbers of kernel arguments or struct
-arguments that are decomposed, a single uniform buffer is used with binding 0,
-and each argument is an offset into that uniform buffer.
+#### Kernel Arguments
+
+There are two different mechanisms used for passing arguments to the SPIR-V kernel.
+The preferred was is push constants, which are copied into the command buffer executing
```
sbalint98:
```suggestion
The preferred is push constants, which are copied into the command buffer executing
```

include/hipSYCL/runtime/vk/vk_allocator.hpp
```diff
@@ -59,8 +59,17 @@ class vk_allocator : public backend_allocator {
...
   vk_alloc_info *find_alloc_info(vk::DeviceAddress ptr);
 
+  // Creates a single buffer backed by allocated device memory, and created with
+  // properties which allow its device address returned
   std::pair<vk::raii::Buffer, vk::raii::DeviceMemory>
-  create_buffer(vk::DeviceSize size, vk::BufferUsageFlags usage_flags);
+  create_device_address_buffer(vk::DeviceSize size);
+
+  // Given a list of uniform buffers sets scraped from SPIR-V reflection
+  // allocates enough memory for all of then, then creates buffers into the
```
sbalint98:
```suggestion
  // allocates enough memory for all of them, then creates buffers into the
```

include/hipSYCL/runtime/vk/vk_allocator.hpp
```diff
@@ -59,8 +59,17 @@ class vk_allocator : public backend_allocator {
...
+  // memory which are bound the appropriate offset
```
sbalint98:
```suggestion
  // memory which are bound to the appropriate offset
```

src/compiler/llvm-to-backend/clspv/LLVMToCLSPV.cpp
```diff
@@ -364,6 +374,10 @@ bool LLVMToCLSPVTranslator::applyBuildOption(const std::string &Option,
     this->MaxUniformBufferRange = Value;
   }
 
+  if (Option == "-device-name") {
+    this->DeviceName = Value;
```
sbalint98, line 379: I think we should return true here right? 
```suggestion
    this->DeviceName = Value;
    return true;
```

src/runtime/vk/vk_code_object.cpp
```diff
@@ -516,6 +594,16 @@ static void verify_spv_capabilities(const uint16_t phys_dev_features,
       check_support(vk_device_features::storagePushConstant16,
                     "SPIR-V Capability PushConstant16 not supported by device");
       break;
+    case spv::CapabilityUniformAndStorageBuffer8BitAccess:
+      check_support(vk_device_features::storagePushConstant8,
```
sbalint98: Shouldn't we also check ` uniformAndStorageBuffer8BitAccess` here as well? similarly a few lines below

src/runtime/vk/vk_code_object.cpp
```diff
@@ -164,6 +173,86 @@ size_t vk_kernel_object::get_push_constants_size() const {
...
+    if (desc.is_available(false)) {
+      desc.init(this);
+      return desc;
+    }
+  }
+
+  // If no descriptors are currently free resort to a blocking wait until the
+  // first kernel using them completes
+  _uniform_descriptors[0].is_available(true);
+  return _uniform_descriptors[0];
```
sbalint98, line 189: I think we should init here as well similar to line 179

include/hipSYCL/compiler/llvm-to-backend/clspv/LLVMToCLSPV.hpp
sbalint98:
```suggestion
#pragma once
```

sbalint98 (commented):
Just a few nits and a few questions, otherwise LGTM

src/runtime/vk/vk_code_object.cpp
EwanC: This was a typo, and should be the value you suggested, good catch.

src/runtime/vk/vk_code_object.cpp
EwanC: I've put a code comment here because I don't think we actually need to. The descriptor container needs initalized once, and then any future calls to init() will have an early exit.

If we get to this point in the control flow then the descriptor container has already been initialized because it's been used before, and therefore initalized, which is why were waiting on it. If you feel strongly about this I can add it in though.

sbalint98: I see I think this if fine like this, thanks!

sbalint98 (approved):

Issue #2120, opened 2026-06-24
[VK] Investigate MoltenVK SYCL test suite fails


EwanC:
The `SKIP_IF_MOLTENVK` macro is used in the SYCL test suite to skip certain tests on the Vulkan backend running ontop of MoltenVK. This task is to investigate the skipped tests to work out if this is something we can fix or a fundamental limitation:

- [ ] `vec_tests/`
- [ ] `math_tests/`
- [ ] `relational_tests/`
- [ ] `group_functions_tests/`
- [ ] `atomic_tests/`
- [ ] `accessor_tests/`

PR #2187, opened 2026-08-12
[SSCP] Support >64bit types in shift/broadcast/permute/select

dmitriy-sobolev:
SYCL 2020 spec requires the types to be trivially copyable in group_broadcast, permute_group_by_xor, select_from_group, shift_group_left and shift_group_right. There is no limit to the size of the types, however acpp supports only <=64bit types, at least in SSCP backend.

It does not let to compile a program with by_segment oneDPL algorithms, like exclusive_scan_by_segment and etc.
See https://github.com/uxlfoundation/oneDPL/issues/2790 for more details.

The idea is to split >64bit  elements into 64bit words and do the existing operation on each such word. Sizes that are not a multiple of 64bit are rounded up to whole 64bit words, and only sizeof(T) bytes are copied back. It is very similar to how these operations are implemented for vec and marray types. Note, barriers are used to mirror what vec/marray functions do, but they do not seem to be necessary.

Tested with a 128bit type (multiple of 64bit) and a 96bit type (not multiple, padded), with groups and sub-groups.


dmitriy-sobolev:
I do not quite understand the error in [Self-hosted CI / NVIDIA with clang 18, CUDA 11.0](https://github.com/AdaptiveCpp/AdaptiveCpp/actions/runs/31790639032/job/94736822420?pr=2187): why it happens and how it is connected to the changes.

The error text:
> tests/sycl/profiler.cpp(186): error: in "profiler_tests/queue_profiling": check t71 <= t73 && t72 <= t73 has failed

Could you assist with fixing it?

sbalint98:
I think this is a flaky test case on this GPU and unrelated to this MR

include/hipSYCL/sycl/libkernel/sscp/group_functions.hpp
sbalint98:
```suggestion

```
I think this would be a good occasion to remove this comment

include/hipSYCL/sycl/libkernel/sscp/group_functions.hpp
```diff
@@ -1070,6 +1124,19 @@ HIPSYCL_BUILTIN std::enable_if_t<(sizeof(T) <= 8), T> __acpp_permute_group_by_xo
...
 
+template <class Group, typename T,
+          std::enable_if_t<is_group_v<std::decay_t<Group>> &&
+                                std::is_trivially_copyable_v<T> && (sizeof(T) > 8),
+                            bool> = true>
+HIPSYCL_BUILTIN
+T __acpp_permute_group_by_xor(Group g, T x, typename Group::linear_id_type mask) {
+  return apply_on_words(x, [g, mask](__acpp_int64 word) {
+    __acpp_int64 result = __acpp_permute_group_by_xor(g, word, mask);
+    __acpp_group_barrier(g);
```
sbalint98: Why do we need this group barrier here?

sbalint98 (commented):
Thanks for the contribution. The core changes look good to me, however for the testing, I think we should rather use the existing infrastructure rather than creating new test cases for these types. 

I think we could create the new types in `tests/sycl/group_functions/group_functions.hpp`, and add them to a new type list specific for the shift/broadcast/permute/select tests that allow trivially copyable types.

dmitriy-sobolev:
> I think we could create the new types in tests/sycl/group_functions/group_functions.hpp, and add them to a new type list specific for the shift/broadcast/permute/select tests that allow trivially copyable types.

Good idea, done.

tests/sycl/group_functions/group_functions_misc.cpp
```diff
@@ -175,8 +175,13 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(group_broadcast, T, test_types) {
   }
 }
 
-BOOST_AUTO_TEST_CASE_TEMPLATE(sub_group_broadcast, T, test_types) {
-  if(!sycl::queue{}.get_device().is_host()) {
+BOOST_AUTO_TEST_CASE_TEMPLATE(sub_group_broadcast, T, shuffle_test_types) {
+  sycl::device dev;
+  if constexpr (detail::is_trivially_copyable_test_type_v<T>) {
+    SKIP_IF_MOLTENVK(dev)
+  }
```
dmitriy-sobolev, line 182: https://github.com/dmitriy-sobolev/AdaptiveCpp/actions/runs/32271384993/job/96128418820

```bash
 [mvk-error] SPIR-V to MSL conversion error: Found a buffer packing case which we cannot represent in MSL.
 [mvk-error] VK_ERROR_INITIALIZATION_FAILED: Compute shader function could not be compiled into pipeline. See previous logged error.
 libc++abi: terminating due to uncaught exception of type vk::InitializationFailedError: vk::raii::Device::createComputePipeline: ErrorInitializationFailed
 unknown location:0: fatal error: in "group_functions_tests/sub_group_broadcast<detail__trivially_copyable_12_byte_type>": signal: SIGABRT (application abort requested)
 /Users/runner/work/AdaptiveCpp/AdaptiveCpp/tests/sycl/group_functions/group_functions_misc.cpp:178:
 checkpoint: "sub_group_broadcast" test entry
 Default-selected queue runs on device: Apple Paravirtual device
```


It is difficult for me to tell what is happening here.

Is it because of some underlying limitation in the MoltenVK?
My first assumption is that it is something related to 12-byte type with padding due to this note:
> ... a buffer packing case which we cannot represent in MSL.

I will wait for the CI to finish, then limit the case to 12-byte type only.

dmitriy-sobolev: I got this issue when left only 16-byte type:

```bash
[mvk-error] VK_ERROR_OUT_OF_DEVICE_MEMORY: Lost VkDevice after MTLCommandBuffer "vkQueueSubmit MTLCommandBuffer on Queue 0-0" execution failed (code 9): Invalid Resource (00000009:kIOGPUCommandBufferCallbackErrorInvalidResource)
libc++abi: terminating due to uncaught exception of type vk::DeviceLostError: vk::raii::Queue::submit: ErrorDeviceLost
unknown location:0: fatal error: in "accessor_tests/offset_nested_subscript": signal: SIGABRT (application abort requested)
libc++abi: terminating due to uncaught exception of type vk::DeviceLostError: vk::raii::Queue::submit: ErrorDeviceLost
/Users/runner/work/AdaptiveCpp/AdaptiveCpp/tests/sycl/accessor.cpp:899: last checkpoint: "offset_nested_subscript" test entry
libc++abi: terminating due to uncaught exception of type vk::DeviceLostError: vk::raii::Device::waitSemaphores: ErrorDeviceLost
/Users/runner/work/_temp/41475d12-ab8b-4247-8905-74e2d4b98686.sh: line 2: 17384 Segmentation fault: 11  ACPP_VISIBILITY_MASK="vk" ./sycl_tests
```
Reverted to skipping both large types.

dmitriy-sobolev: If it is not a known limitation in the MoltenVK bakcned, we should likely add a comment to that code section and create an issue. What do you think, do you have a suggestion for such a comment?

EwanC: There's quite a few MoltenVK backend test skips, this is the catch-all ticket https://github.com/AdaptiveCpp/AdaptiveCpp/issues/2120 if you want to add this issue onto that

sbalint98 (approved):

tests/sycl/group_functions/group_functions_misc.cpp
dmitriy-sobolev: There is already a note about the group function tests, so it is not needed. Thanks for confirming.

PR #2190, opened 2026-08-14
Fix division by zero in `memcpy` and `memset` with OMP backend

Stastez:
#2105 introduces a division by zero error for `memcpy` and `memset` when they are called on 0 elements.

In src/runtime/omp/omp_queue.cpp , line 399, `block_size` appears to be 0 when `queue.memcpy(dest, src, 0);` is called, leading to a divison by zero in line 400. I believe that this is not expected behavior.

Reproducer:

```cpp
#include <sycl/sycl.hpp>

int main() {
	sycl::queue queue{sycl::cpu_selector_v};
	const auto* const src = sycl::malloc_device<int>(10, queue);
	auto* const dest = sycl::malloc_device<int>(10, queue);
	queue.memcpy(dest, src, 0);
	queue.memset(dest, {}, 0);
}
```


Stastez:
Sorry, just found the other PR.

This is a duplicate of #2178 .

PR #2191, opened 2026-08-16
[Metal] Fix direct calls with mismatched struct return types

### LINKED ISSUE #2183: Metal reports a direct pair-returning template call as indirect
**Bug summary**

On an Apple M3 Max, the Metal backend rejects a generic SSCP program containing uses of the same pair-returning template in multiple translation units. The diagnostic reports one of those direct calls as indirect:

```text
LLVMToMetal: MetalEmitter failed: Error: Indirect call not supported:
  %2 = call %struct.CallerTwoFloat @_Z17makeAggregatePairIfENSt3__14pairIT_S2_EES2_(float noundef 4.000000e+01)
```

The identical executable produces the expected values with the OpenMP backend. (And no problem with the SSCP backend on an x86 CPU+Nvidia GPU). I observed the Metal failure with both `-O0` and `-O3`.

**To Reproduce**

Create `aggregate_return.hpp`:

```cpp
#pragma once

#include <sycl/sycl.hpp>

#include <utility>

template <class T>
__attribute__((noinline)) std::pair<T, T> makeAggregatePair(T value) {
  return {value + T{1}, value + T{2}};
}

SYCL_EXTERNAL __attribute__((noinline)) float consumeAggregatePair(float value);
```

Create `aggregate_return.cpp`:

```cpp
#include "aggregate_return.hpp"

struct UnrelatedTwoFloat {
  float first;
  float second;
};

SYCL_EXTERNAL __attribute__((noinline)) float consumeAggregatePair(float value) {
  UnrelatedTwoFloat keep_type_alive{3.0f, 4.0f};
  const auto [first, second] = makeAggregatePair(value);
  return first + second + 0.0f * keep_type_alive.first;
}
```

Create `repro.cpp`:

```cpp
#include "aggregate_return.hpp"

#include <iostream>

struct CallerTwoFloat {
  float first;
  float second;
};

int main() {
  sycl::queue queue;
  float *output = sycl::malloc_shared<float>(2, queue);
  output[0] = -11.0f;
  output[1] = -22.0f;

  queue.single_task([=]() {
    CallerTwoFloat keep_type_alive{3.0f, 4.0f};
    const auto [first, second] = makeAggregatePair(10.0f);
    output[0] = first + second + 0.0f * keep_type_alive.first;
    output[1] = consumeAggregatePair(40.0f);
  }).wait();

  std::cout << output[0] << ' ' << output[1] << '\n';
  sycl::free(output, queue);
}
```

Compile and run it with:

```console
$ acpp repro.cpp aggregate_return.cpp -o repro --acpp-targets=generic -O3
$ ACPP_VISIBILITY_MASK=metal ACPP_RT_NO_JIT_CACHE_POPULATION=1 ./repro
```

As a control, the identical executable completes successfully with OpenMP:

```console
$ ACPP_VISIBILITY_MASK=omp ACPP_RT_NO_JIT_CACHE_POPULATION=1 ./repro
23 83
```

`CallerTwoFloat` and `UnrelatedTwoFloat` are dummy types used only to make the failure reproducible. The program does not convert between either type and `std::pair<float, float>`, and their values do not affect the result. The reproducer also calls the same pair-returning template from both translation units; removing these otherwise irrelevant ingredients prevents the failure in this reduced example.

**Observed behavior:** Metal JIT compilation reports the
`Indirect call not supported` error shown above. The program then aborts with exit status 134, and the kernel produces no valid result.

**Expected behavior**

The Metal backend should compile and run the program, which should print:

```text
23 83
```

**Describe your setup**

* AdaptiveCpp was built from source from `develop` at commit `8cb557ea` and installed as version `25.10.0+git.8cb557ea.20260730.branch.develop`.
* Hardware: Apple M3 Max.
* Operating system: macOS 26.5 (build 25F71), arm64.
* AdaptiveCpp LLVM plugin version: 20.
* Clang used by `acpp`: Homebrew Clang 20.1.8 from Homebrew's `llvm@20`
  package (`/opt/homebrew/opt/llvm@20/bin/clang++`).
* Host compiler configured by AdaptiveCpp: Apple Clang 21.0.0
  (`/usr/bin/clang++`).
* Available AdaptiveCpp runtime backends: OpenMP and Metal.
* Targeted backend and hardware: generic SSCP targeting the Metal runtime on
  the Apple M3 Max, selected with `ACPP_VISIBILITY_MASK=metal`.
* Compile command: `acpp repro.cpp aggregate_return.cpp -o repro
  --acpp-targets=generic -O3`.

`acpp --acpp-version` reports:

```text
AdaptiveCpp version: 25.10.0+git.8cb557ea.20260730.branch.develop
Plugin LLVM version: 20
Default clang: /opt/homebrew/opt/llvm@20/bin/clang++
Default targets: generic
Available runtime backends: OpenMP, Metal
```

`acpp-info` reports the Metal device as:

```text
Name: Apple M3 Max
Backend: Metal
Vendor: Apple
Arch: applegpu_g15s
```

**Additional context**

The problem was encountered in a downstream SYCL Monte Carlo application running on an Apple M3 Max. The application uses an ordinary structured binding from a helper returning `std::pair<float, float>`; it does not cast between the pair and the unrelated two-float type. In that application, the Metal emitter reported an analogous `uniform_pair<float>` call as an unsupported indirect call.

A candidate patch and an AdaptiveCpp LIT regression test are available at:

https://github.com/cbyrohl/AdaptiveCpp/tree/fix/metal-direct-call-signatures

This report used Claude Code for working out a reproducable example, collecting the relevant system information, language editing, and the proposed patch above. Since this is outside of my area of expertise, I would like to confirm the issue and the direction of the proposed fix before opening a PR. In that fix, I normalize the Metal direct calls with structurally equivalent but nominally different aggregate return types so the call signature exactly matches the callee.

resetius:
https://github.com/AdaptiveCpp/AdaptiveCpp/issues/2183

### DIFF
--- a/src/compiler/llvm-to-backend/metal/LLVMToMetal.cpp
+++ b/src/compiler/llvm-to-backend/metal/LLVMToMetal.cpp
@@ -70,6 +70,77 @@ namespace compiler {
 
 namespace {
 
+// When linking LLVM IR, structs with identical layout get deduplicated, so the
+// struct return type at a call site can differ from the one of the callee. The
+// Emitter does not support this and assumes in several places that both are the
+// same type, so we just unify them here.
+void unifyCallSiteStructReturnTypes(llvm::Module &M) {
+  llvm::SmallVector<llvm::CallInst *> Calls;
+  for (llvm::Function &F : M)
+    for (llvm::BasicBlock &BB : F)
+      for (llvm::Instruction &I : BB)
+        if (auto *Call = llvm::dyn_cast<llvm::CallInst>(&I))
+          Calls.push_back(Call);
+
+  for (llvm::CallInst *Call : Calls) {
+    // musttail requires an exact signature match, so we cannot repack the result.
+    if (Call->isMustTailCall())
+      continue;
+
+    // Only direct calls, and only those that actually have a mismatch.
+    auto *Callee =
+        llvm::dyn_cast<llvm::Function>(Call->getCalledOperand());
+    if (!Callee || Call->getFunctionType() == Callee->getFunctionType())
+      continue;
+
+    // Everything except the return type must match: anything else is a real
+    // signature conflict that we must not paper over.
+    llvm::FunctionType *CallType = Call->getFunctionType();
+    llvm::FunctionType *CalleeType = Callee->getFunctionType();
+    if (CallType->isVarArg() != CalleeType->isVarArg() ||
+        Call->getCallingConv() != Callee->getCallingConv() ||
+        CallType->params() != CalleeType->params())
+      continue;
+
+    // Both return types must be non-empty structs of identical layout, so that
+    // repacking them field by field is valid.
+    auto *CallReturnType =
+        llvm::dyn_cast<llvm::StructType>(CallType->getReturnType());
+    auto *CalleeReturnType =
+        llvm::dyn_cast<llvm::StructType>(CalleeType->getReturnType());
+    if (!CallReturnType || !CalleeReturnType ||
+        CallReturnType->getNumElements() == 0 ||
+        !CallReturnType->isLayoutIdentical(CalleeReturnType))
+      continue;
+
+    llvm::SmallVector<llvm::Value *> Arguments(Call->args());
+    llvm::SmallVector<llvm::OperandBundleDef> Bundles;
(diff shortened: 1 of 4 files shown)


illuhad (approved):
LGTM

PR #2195, opened 2026-08-21
[CI] Improve the docker infrastructure

sbalint98:
Currently in the CI some LLVM packages are not always installed successfully. The CI only failes after the images are built while installing AdaptiveCpp even though the apt commands for LLVM also failed in the previous step. https://github.com/AdaptiveCpp/AdaptiveCpp/actions/runs/32279703262/job/96261713479

This PR improves build process by the following:

* Fail docker image build jobs in case some components are not installed successfully
* Re try installing LLVM 5 times to account for higher possible transient network issues
* Remove the defaults from the docker image primer workflow, so that images without cuda can be also built manually
* Changes the type of the `FORCE_BUILD` parameter to boolean, I think this was the reason why the base images got re-generated with every run
* ~~Bumps the version of all used actions due to the deprecation of node20~~

### DIFF
--- a/.github/workflows/create-ci-docker/AdaptiveCpp-CI-Linux.dockerfile
+++ b/.github/workflows/create-ci-docker/AdaptiveCpp-CI-Linux.dockerfile
@@ -47,6 +47,7 @@ RUN apt-get -u update \
     && rm -rf /var/lib/apt/lists/*
 
 RUN <<EOF
+    set -e
     # skip if ROCM_VERSION empty (yeah.. Intel not ROCm .. but whatever)
     if [ -z "${ROCM_VERSION}" ]; then exit 0; fi
     wget -q https://github.com/intel/intel-graphics-compiler/releases/download/igc-1.0.16695.4/intel-igc-core_1.0.16695.4_amd64.deb
@@ -63,6 +64,7 @@ RUN <<EOF
 EOF
 
 RUN <<EOF
+    set -e
     # skip if CUDA_MAJOR_VERSION empty
     if [ -z "${CUDA_MAJOR_VERSION}" ]; then exit 0; fi
     mkdir -p /opt/cuda-${CUDA_MAJOR_VERSION}.${CUDA_MINOR_VERSION}
@@ -73,6 +75,7 @@ RUN <<EOF
 EOF
 
 RUN <<EOF
+    set -e
     # skip if ROCM_VERSION empty
     if [ -z "${ROCM_VERSION}" ]; then exit 0; fi
     wget -q -O - https://repo.radeon.com/rocm/rocm.gpg.key | apt-key add -
@@ -84,6 +87,7 @@ RUN <<EOF
 EOF
 
 RUN <<EOF
+    set -e
     wget -q https://apt.llvm.org/llvm.sh
     chmod +x llvm.sh
     ./llvm.sh ${LLVM_VERSION}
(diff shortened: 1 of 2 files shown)


.github/workflows/create-ci-docker/AdaptiveCpp-CI-Linux.dockerfile
```diff
@@ -49,49 +49,49 @@ RUN apt-get -u update \
...
-    wget -q https://github.com/intel/compute-runtime/releases/download/24.17.29377.6/intel-level-zero-gpu-dbgsym_1.3.29377.6_amd64.ddeb
-    wget -q https://github.com/intel/compute-runtime/releases/download/24.17.29377.6/intel-level-zero-gpu_1.3.29377.6_amd64.deb
-    wget -q https://github.com/intel/compute-runtime/releases/download/24.17.29377.6/intel-opencl-icd-dbgsym_24.17.29377.6_amd64.ddeb
-    wget -q https://github.com/intel/compute-runtime/releases/download/24.17.29377.6/intel-opencl-icd_24.17.29377.6_amd64.deb
-    wget -q https://github.com/intel/compute-runtime/releases/download/24.17.29377.6/libigdgmm12_22.3.19_amd64.deb
-    wget -q https://github.com/oneapi-src/level-zero/releases/download/v1.13.5/level-zero-devel_1.13.5+u22.04_amd64.deb
-    wget -q https://github.com/oneapi-src/level-zero/releases/download/v1.13.5/level-zero_1.13.5+u22.04_amd64.deb
-    dpkg -i *.deb
-    rm *.deb
+    wget -q https://github.com/intel/intel-graphics-compiler/releases/download/igc-1.0.16695.4/intel-igc-core_1.0.16695.4_amd64.deb \
```
illuhad: `set -e` ?

sbalint98:
Lets do the update of the actions in a different PR, I think this should go in to fix CI. Could you have a look, @fodinabor ?

.github/workflows/prime-docker.yml
```diff
@@ -12,13 +12,11 @@ on:
         required: true
         default: "ubuntu-22.04"
       cuda_major_version:
-        description: 'CUDA Major Version (default 11)'
```
fodinabor, line 15: The background for these defaults is that for manual priming you need to specify the versions here... and if you dont prime the right version you didnt win anything:)

.github/workflows/create-ci-docker/AdaptiveCpp-CI-Linux.dockerfile
```diff
@@ -84,9 +87,13 @@ RUN <<EOF
 EOF
 
 RUN <<EOF
+    set -e
     wget -q https://apt.llvm.org/llvm.sh
     chmod +x llvm.sh
-    ./llvm.sh ${LLVM_VERSION}
+    for i in 1 2 3 4 5; do 
+        ./llvm.sh ${LLVM_VERSION} && echo "Attempt $i failed, retrying in 15s..."
```
fodinabor, line 96: Feels wrong ^^ but I assume you're trying to counter some flakeyness...

fodinabor (approved):
Mostly looks good, the versions are the only real not.

.github/workflows/create-ci-docker/AdaptiveCpp-CI-Linux.dockerfile
sbalint98: yes, since Friday sometimes the llvm.sh just thinks that there are no repository available because some wget  times out internally. I think this is probably some issue of the github hosted runners. I don't have a better idea how to work around it.

.github/workflows/prime-docker.yml
sbalint98: Dropped this change.

PR #2196, opened 2026-08-21
[Metal] Batch command buffers and deduplicate residency bindings

anudit:
Profiling [GROMACS](https://gitlab.com/gromacs/gromacs) (SYCL/SSCP, GPU-resident) on Apple Silicon showed that the Metal backend committed one command buffer per operation (~11 per MD step vs ~3.5 for a native Metal backend), with each launch additionally making *every* USM allocation in the process resident under the allocator mutex (`metal_queue.cpp`, see the existing `// TODO: switch to MTL4 API for O(1) buffer bindings`). The GPU was idle 25-40% of each step, dominated by per-commit submission cost and inter-buffer bubbles.

This PR changes the Metal backend to encode consecutive operations into a single, lazily-committed command buffer, and reduces the per-launch residency work from O(ops x allocations) to O(allocations) per command buffer.

Measured on M2 Max (GROMACS, grappa benchmark, GPU-resident, 3-rep medians): **+71% / +84% (RF/PME) at 96k atoms, +14-47% at 192-384k, +3-8% at 768k atoms**, with bit-identical potential energies. Command buffers per traced window dropped from 50.6k to 15.9k.

### Command buffer batching (`metal_queue.cpp`)

- Operations append their encoders to a per-queue *open* command buffer which is committed lazily:
  - on `wait()` / `insert_event()` (event must be able to complete),
  - before device-to-host copies and other host-staged memcpys (which need dedicated buffers with CPU-side completion handling),
  - before custom (interop) operations (see below),
  - after a bounded number of ops (`max_ops_per_command_buffer = 32`) to limit memory use and first-work latency.
- D2D memcpys and memsets join the open buffer; host-staging copies keep dedicated committed buffers with unchanged event/staging logic.
- No implicit signal/wait pair is added at commit boundaries: synchronization is expressed exclusively through explicit events and host-visible copies, exactly as before, so the GPU can still execute independent command buffers concurrently. (An earlier version that signal/wait-chained every flush measurably reduced throughput by serializing independent work.)

### Custom operations and native interop ordering

- Host-side custom (interop) operations execute user code synchronously during `invoke()`. That code may create and commit its own native command buffers on the shared `MTLCommandQueue` (e.g. FFT libraries using `get_native_type()`), and Metal executes same-queue command buffers in *commit* order. Batching therefore breaks ordering unless everything encoded so far is committed first.
- `rt::kernel_launcher::is_custom_operation()` exposes whether a launcher describes such an operation; the Metal queue flushes before invoking it, and also in `get_native_type()`.
- Found via a real-world case: GROMACS' VkFFT-on-Metal path produced wrong step-0 energies under batching until this flush was added.

### Residency deduplication (`make_allocations_resident`)

- SSCP kernels may dereference arbitrary USM pointers, so all allocations must be resident - but within one open command buffer each allocation now only gets a single `useResource` call, and the allocator scan is skipped entirely while the allocator state is unchanged.
- `metal_allocator` gains a lock-free monotonic `generation()` counter, bumped on every allocation/free; consumers detect changes without taking the allocator lock. The dedup cache is invalidated whenever the generation changes (a freed buffer's address can be reused by a new allocation).

## Testing

- `sycl_tests` (792 cases), `pcuda_tests`, `rt_tests`: all pass on macOS/arm64 (M2 Max, macOS 26), repeatedly.
- GROMACS end-to-end: step-0 potential energies bit-identical to the unbatched build across RF/PME flavors and system sizes; throughput gains above.

## Throughput results (GROMACS, grappa, GPU-resident, M2 Max, 3-rep medians)

| case | before (ns/day) | after (ns/day) | gain |
|---|---|---|---|
| rf 96k | 52.65 | 89.83 | +70.6% |
| pme 96k | 38.63 | 70.92 | +83.6% |
| rf 192k | 45.35 | 51.66 | +13.9% |
| pme 192k | 35.48 | 41.07 | +15.8% |
| rf 384k | 18.40 | 26.95 | +46.5% |
| pme 384k | 16.29 | 19.15 | +17.6% |
| rf 768k | 11.42 | 11.75 | +3.0% |
| pme 768k | 9.11 | 9.59 | +5.3% |

## xctrace Metal System Trace (192k rf, 18 s window)

| metric | before | after |
|---|---|---|
| GPU command buffers | 50,596 | 15,908 (-69%) |
| compute encoder intervals | 51,590 | 19,675 (merged encoders) |

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,9 +52,3 @@ build*/
 
 # clangd cache
 .cache
-
-# macOS
-.DS_Store
-
-# local install prefix
-/install/
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,9 @@ build*/
 
 # clangd cache
 .cache
+
+# macOS
+.DS_Store
+
+# local install prefix
+/install/
--- a/include/hipSYCL/runtime/metal/metal_queue.hpp
+++ b/include/hipSYCL/runtime/metal/metal_queue.hpp
@@ -195,6 +195,11 @@ private:
 
   void profiling_setup(operation& op, const dag_node_ptr& node);
 
+  // Releases the extra reference held on every entry of _resident_resources
+  // (see comment on that member). Must be called before replacing the cache
+  // with a new snapshot, and on destruction.
+  void release_resident_resources();
+
   MTL::Device* _device = nullptr;
   MTL::CommandQueue* _command_queue = nullptr;
   MTL::SharedEvent* _shared_event = nullptr;
@@ -230,6 +235,12 @@ private:
   // Cached snapshot of the allocations that must be made resident, together
   // with the allocator generation it was taken at. Rebuilt only when the
   // allocator generation changes; independent of command buffer boundaries.
+  // Each entry holds an extra reference taken by metal_allocator::
+  // snapshot_buffers() (retained manually, not via NS::SharedPtr, for the
+  // same reason as _open_buffer above), so a concurrent free() of one of
+  // these buffers on another thread cannot deallocate it while it is cached
+  // here or being passed to useResources(). That reference is released when
+  // the cache is replaced (generation change) or the queue is destroyed.
   std::vector<const MTL::Resource*> _resident_resources;
   uint64_t _residency_generation{~static_cast<uint64_t>(0)};
 
--- a/include/hipSYCL/runtime/metal/metal_queue.hpp
+++ b/include/hipSYCL/runtime/metal/metal_queue.hpp
@@ -24,12 +24,13 @@
 
 #include <mach/mach_time.h>
 
-#include <unordered_set>
+#include <vector>
 
 namespace MTL {
 
 class Device;
 class Buffer;
+class Resource;
 class CommandBuffer;
 class CommandQueue;
 class Fence;
@@ -160,9 +161,10 @@ public:
   virtual ~metal_inorder_queue();
 
   // Makes all USM allocations known to the allocator resident for the given
-  // encoder. Within one open command buffer, each allocation is only passed
-  // to useResource once, and the scan is skipped entirely if the allocator
-  // state has not changed since the last scan.
+  // encoder. Residency established this way is scoped to the encoder, so this
+  // must be called for every encoder; what is cached across encoders is only
+  // the *set* of allocations, which is rebuilt (taking the allocator lock)
+  // exclusively when the allocator generation changes.
   // Public because it is invoked by the free kernel launch helper.
   void make_allocations_resident(MTL::ComputeCommandEncoder* encoder);
 
@@ -225,10 +227,11 @@ private:
 
   static constexpr int max_ops_per_command_buffer = 32;
 
-  // Residency dedup state, valid for the current open command buffer only.
-  std::unordered_set<MTL::Buffer*> _resident_buffers;
+  // Cached snapshot of the allocations that must be made resident, together
+  // with the allocator generation it was taken at. Rebuilt only when the
+  // allocator generation changes; independent of command buffer boundaries.
+  std::vector<const MTL::Resource*> _resident_resources;
   uint64_t _residency_generation{~static_cast<uint64_t>(0)};
-  bool _residency_pass_done{false};
 
   metal_allocator* _allocator = nullptr;
   device_id _device_id;
--- a/include/hipSYCL/runtime/metal/metal_queue.hpp
+++ b/include/hipSYCL/runtime/metal/metal_queue.hpp
@@ -16,13 +16,13 @@
 #include "../kernel_cache.hpp"
 #include "../queue_completion_event.hpp"
 
-#include "hipSYCL/common/spin_lock.hpp"
 #include "hipSYCL/glue/llvm-sscp/jit.hpp"
 
 #include "metal_allocator.hpp"
 #include "metal_event.hpp"
 
 #include <mach/mach_time.h>
+#include <mutex>
 
 namespace MTL {
 
@@ -189,8 +189,7 @@ private:
   // threads concurrently with submit_*() / insert_event().
   std::atomic<uint64_t> _event_counter{0};
 
-  // Only touched by submit_*() and insert_event(), which are serialized
-  // by external mutex, so no atomics needed.
+  // Protected by _mutex.
   uint64_t _pending_cpu_event{0};
   uint64_t _pending_gpu_event{0};
 
@@ -204,7 +203,6 @@ private:
   // because this header must stay free of Foundation headers (see
   // metal_hardware_manager.cpp, which defines the metal-cpp private impls).
   MTL::CommandBuffer* _open_buffer = nullptr;
-  bool _open_buffer_has_trailing_signal{false};
   int _open_op_count{0};
   bool _flush_after_current_op{false};
 
@@ -223,7 +221,7 @@ private:
 
   kernel_configuration _config;
 
-  common::spin_lock _sscp_submission_spin_lock;
+  mutable std::recursive_mutex _mutex;
 
   std::optional<metal_profiling_setup> _profiling_setup;
 };
--- a/include/hipSYCL/runtime/metal/metal_queue.hpp
+++ b/include/hipSYCL/runtime/metal/metal_queue.hpp
@@ -24,17 +24,12 @@
 
 #include <mach/mach_time.h>
 
-#include <vector>
-
 namespace MTL {
 
 class Device;
-class Buffer;
-class Resource;
 class CommandBuffer;
 class CommandQueue;
 class Fence;
-class ComputeCommandEncoder;
 class SharedEvent;
 class SharedEventListener;
 
@@ -160,45 +155,31 @@ public:
 
   virtual ~metal_inorder_queue();
 
-  // Makes all USM allocations known to the allocator resident for the given
-  // encoder. Residency established this way is scoped to the encoder, so this
-  // must be called for every encoder; what is cached across encoders is only
-  // the *set* of allocations, which is rebuilt (taking the allocator lock)
-  // exclusively when the allocator generation changes.
-  // Public because it is invoked by the free kernel launch helper.
-  void make_allocations_resident(MTL::ComputeCommandEncoder* encoder);
-
 private:
-  // Common setup for freshly created command buffers: encodes the wait on the
-  // previous queue work and attaches pending profiling handlers.
+  // Encodes the wait on prior queue work and attaches pending profiling
+  // handlers; shared by get_open_command_buffer() and
+  // new_dedicated_command_buffer().
   MTL::CommandBuffer* prepare_command_buffer(MTL::CommandBuffer* cmd_buf);
 
-  // Returns the currently open (uncommitted) command buffer of this queue,
-  // creating one if none exists. Operations append their encoders to the open
-  // buffer; it is committed lazily by flush().
+  // Returns the open (uncommitted) command buffer, creating one if needed.
+  // Committed lazily by flush().
   MTL::CommandBuffer* get_open_command_buffer();
 
-  // Batching bookkeeping to be invoked after an operation has been fully
-  // encoded into the open buffer: enforces a bound on ops per command buffer
-  // and closes the window for per-op GPU profiling timestamps.
+  // Bounds the number of ops per command buffer and flushes when profiling
+  // needs the buffer to end with exactly the profiled operation.
   void finish_batched_op();
 
-  // Creates a fresh command buffer that is NOT the open buffer, for operations
-  // that must be committed immediately (e.g. device->host copies with CPU-side
-  // completion logic). Serializes against all prior work on this queue.
+  // A command buffer that is committed immediately instead of joining
+  // _open_buffer, for ops needing immediate CPU-side completion handling
+  // (e.g. host-staged copies).
   MTL::CommandBuffer* new_dedicated_command_buffer();
 
-  // Commits the open command buffer, if any. Synchronization with subsequent
-  // work is expressed exclusively through explicit events and host-visible
-  // copies, matching the semantics of the non-batched submission path.
+  // Commits the open command buffer, if any. No implicit signal/wait is
+  // added: that would serialize command buffers the GPU could otherwise run
+  // concurrently, so synchronization stays explicit (events, host copies).
   result flush();
 
   void profiling_setup(operation& op, const dag_node_ptr& node);
-
-  // Releases the extra reference held on every entry of _resident_resources
-  // (see comment on that member). Must be called before replacing the cache
-  // with a new snapshot, and on destruction.
-  void release_resident_resources();
 
   MTL::Device* _device = nullptr;
   MTL::CommandQueue* _command_queue = nullptr;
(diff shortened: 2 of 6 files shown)


resetius:
Thank you,
I ran into that problem too when profiling Llama.cpp.

.gitignore
```diff
@@ -52,3 +52,9 @@ build*/
 
 # clangd cache
 .cache
+
+# macOS
+.DS_Store
```
resetius: This is not relevant to this PR. I would move it to a separate PR.

.gitignore
```diff
@@ -52,3 +52,9 @@ build*/
...
+
+# local install prefix
+/install/
```
resetius: These are environment settings. I would not include them at all 
For example, my install directory is outside the repo working copy

src/runtime/metal/metal_queue.cpp
```diff
@@ -308,11 +289,116 @@ MTL::CommandBuffer* metal_inorder_queue::new_command_buffer() {
...
+    flush();
+  }
+}
+
+void metal_inorder_queue::make_allocations_resident(
+  MTL::ComputeCommandEncoder* encoder) {
+  constexpr auto usage = MTL::ResourceUsageRead | MTL::ResourceUsageWrite;
+
+  const uint64_t gen = _allocator->generation();
+  if (_residency_pass_done && gen == _residency_generation) {
```
resetius: Why is this correct? I thought `useResource` should be called for each encoder, and the encoder is created for each compute pass

You could avoid the `for_each` under the lock by keeping a `vector<> buffers` cache in the queue and replacing it for each new generation. Then you could pass the buffers to the encoder with `useResources`.

resetius:
How exactly is the `sycl::queue` created to make this mode work? Something like this?

```cpp
sycl::queue{
    sycl::property_list{
        sycl::property::queue::in_order{},
        sycl::property::queue::AdaptiveCpp_coarse_grained_events{}
    }
};
```

src/runtime/metal/metal_queue.cpp
anudit: Ah! updated it with your approach.

anudit:
> How exactly is the `sycl::queue` created to make this mode work? Something like this?
> 
> ```c++
> sycl::queue{
>     sycl::property_list{
>         sycl::property::queue::in_order{},
>         sycl::property::queue::AdaptiveCpp_coarse_grained_events{}
>     }
> };
> ```

Yup, 

Without AdaptiveCpp_coarse_grained_events the executor inserts an event per node (inorder_executor.cpp:181), and insert_event() has to commit the open command buffer so the event can complete, so you get one command buffer per op again and the batching is a no-op. There's also a per-command-group variant (property::command_group::AdaptiveCpp_coarse_grained_events) if you only want it for some submissions. This is the same combination stdpar uses in sycl_glue.hpp.

anudit:
Looks like some of the linux issues on the CI are the Git's safe.directory check tripping inside the Docker container (the repo is checked out as one UID but built as another), which then breaks CMake's compiler detection

On the macOS side, 
`metal_inorder_queue::make_allocations_resident()` cached raw `MTL::Buffer*` pointers across the allocator lock's scope, so a concurrent raw_free() on another thread could release()/deallocate a buffer between the snapshot and the useResources() call (or while it sat in the per-queue cache) - a use-after-free explaining the CI Bus error: 10.

Fix (metal_allocator.hpp/.cpp, metal_queue.hpp/.cpp):
- metal_allocator::snapshot_buffers() now retain()s every buffer it hands back, under the same lock as the snapshot, so the extra reference exists before the lock is ever released. Had to make it a non-template member (implemented in metal_allocator.cpp) since retain() needs MTL::Buffer complete, which the header deliberately doesn't include.
- metal_inorder_queue releases the previous snapshot's references (release_resident_resources()) right before replacing the cache on a generation change, and again in the destructor to avoid leaking the last snapshot.

Verification:
- rt-backend-metal and sycl_tests/pcuda_tests build clean against a fresh local install.
- Ran sycl_tests (793 cases) 4 times and pcuda_tests (31 cases) 2 times — all pass with *** No errors detected, no crash, matching exactly what the CI Metal backend [macOS, LLVM 20] job runs.

resetius:
Thanks, looks good. Could you move the residency bindings deduplication into a separate PR? It looks like independent work now. I'd like to review the batch command buffers separately and more carefully, because when I tried something similar myself, I ran into many bugs.

Also, could you clean up the comments? They should be short and explain why. If it is clear from the code what a function does, the comment is not needed.

anudit:
- Split out the residency bindings dedup into #2215, so this PR is now just the command buffer batching.
- Trimmed the comments.

resetius:
@anudit could you rebase it?

resetius:
```cpp
#include <sycl/sycl.hpp>

#include <chrono>
#include <cstdio>
#include <cstdlib>

using clk = std::chrono::steady_clock;
static double ms_since(clk::time_point t) {
  return std::chrono::duration<double, std::milli>(clk::now() - t).count();
}

int main(int argc, char **argv) {
  const int n = argc > 1 ? std::atoi(argv[1]) : 2000;
  const int reps = argc > 2 ? std::atoi(argv[2]) : 3;

  auto run = [&](const char *name, const sycl::property_list &props) {
    sycl::queue q{props};
    int *data = sycl::malloc_device<int>(64, q);
    q.memset(data, 0, 64 * sizeof(int)).wait();

    for (int i = 0; i < 64; ++i)
      q.single_task([=]() { data[i & 63] = i; });
    q.wait();

    double best_submit = 1e18, best_total = 1e18;
    for (int r = 0; r < reps; ++r) {
      auto t0 = clk::now();
      for (int i = 0; i < n; ++i)
        q.single_task([=]() { data[i & 63] = i; });
      double submit = ms_since(t0);
      q.wait();
      double total = ms_since(t0);
      if (total < best_total) {
        best_total = total;
        best_submit = submit;
      }
    }

    std::printf("%-32s submit %8.2f ms   total %8.2f ms   %7.1f us/op\n", name,
                best_submit, best_total, best_total * 1000.0 / n);
    std::fflush(stdout);
    sycl::free(data, q);
  };

  std::printf("kernels per rep: %d, best of %d\n\n", n, reps);
  std::fflush(stdout);
  run("in-order", {sycl::property::queue::in_order{}});
  run("in-order + coarse events",
      {sycl::property::queue::in_order{},
       sycl::property::queue::AdaptiveCpp_coarse_grained_events{}});
  run("out-of-order", {});
  run("out-of-order + coarse events",
      {sycl::property::queue::AdaptiveCpp_coarse_grained_events{}});
  return 0;
}
```

Apple M5 Pro, macOS 26

## Before the PR (develop, no batching)

2000 kernels per run:

| queue mode | submit (ms) | total (ms) | per kernel (us) |
|---|---|---|---|
| in-order | 22.68 | 23.05 | 11.5 |
| in-order + coarse events | 15.81 | 16.35 | 8.2 |
| out-of-order | 29.82 | 34.06 | 17.0 |
| out-of-order + coarse events | 28.33 | 49.55 | 24.8 |

## With the PR (batching)

2000 kernels per run:

| queue mode | submit (ms) | total (ms) | per kernel (us) |
|---|---|---|---|
| in-order | 281.30 | 290.30 | 145.1 |
| in-order + coarse events | 8.09 | 9.32 | 4.7 |
| out-of-order | 32.11 | 35.80 | 17.9 |
| out-of-order + coarse events | — | — | crash |

## Comparison

| queue mode | before (us) | after (us) | change |
|---|---|---|---|
| in-order | 11.5 | 145.1 | 12.6x slower |
| in-order + coarse events | 8.2 | 4.7 | 1.7x faster |
| out-of-order | 17.0 | 17.9 | about the same |
| out-of-order + coarse events | 24.8 | crash | — |

src/runtime/metal/metal_queue.cpp
```diff
@@ -333,11 +311,79 @@ MTL::CommandBuffer* metal_inorder_queue::new_command_buffer() {
...
+  if (!_open_buffer) {
+    return make_success();
+  }
+
+  MTL::CommandBuffer* cb = _open_buffer;
+
+  // No implicit signal/wait is added on commit: that would serialize command
+  // buffers the GPU could otherwise run concurrently. If this buffer already
+  // has an explicit signal (from insert_event()), later buffers must wait on it.
+  if (_open_buffer_has_trailing_signal) {
```
resetius: perf degradation: 11.5 -> 145.1 us/op
Command buffers on one queue are already ordered by hazard tracking and the fence
I think this block could be dropped

include/hipSYCL/runtime/metal/metal_queue.hpp
```diff
@@ -177,6 +199,17 @@ class metal_inorder_queue : public inorder_queue
   MTL::Fence* _fence = nullptr;
   bool _fence_chain_active = false;
 
+  // Currently open command buffer that ops append encoders to; committed
+  // lazily by flush(). Retained manually rather than via NS::SharedPtr
+  // because this header must stay free of Foundation headers (see
+  // metal_hardware_manager.cpp, which defines the metal-cpp private impls).
+  MTL::CommandBuffer* _open_buffer = nullptr;
```
resetius, line 205: _open_buffer it no protected by a mutex
I think this is the reason for the crash

src/runtime/metal/metal_queue.cpp
anudit: Fixed. I dropped the trailing-signal bookkeeping and the redundant _pending_gpu_event wait. Command ordering continues to rely on Metal’s hazard tracking and the existing fence. Ordinary in-order execution is back to 14.7 µs/op instead of 145.1 µs/op on M2 Max.

include/hipSYCL/runtime/metal/metal_queue.hpp
anudit: Fixed. The mutable queue state, including _open_buffer, event-chain fields, profiling state, and fence state, is now protected by a queue-level recursive mutex. Recursive locking is needed because kernel submission re-enters through the SSCP invoker and wait() calls insert_event().
The out-of-order + coarse-events reproducer now completes without crashing. The existing coarse-event and cross-queue tests also pass.

src/runtime/metal/metal_queue.cpp
resetius: this site is broken now
```
#include <sycl/sycl.hpp>
#include <cstdio>

int main() {
  sycl::property_list props{
      sycl::property::queue::in_order{},
      sycl::property::queue::AdaptiveCpp_coarse_grained_events{}};
  sycl::queue qa{props}, qb{props};

  constexpr std::size_t n = 1 << 16;
  int *data = sycl::malloc_shared<int>(n, qa);
  int *out = sycl::malloc_shared<int>(n, qa);

  int bad_total = 0;
  for (int rep = 0; rep < 20; ++rep) {
    for (std::size_t i = 0; i < n; ++i) { data[i] = 0; out[i] = -1; }

    sycl::event e = qa.parallel_for(sycl::range<1>(n), [=](sycl::id<1> idx) {
      int acc = 0;
      for (int k = 0; k < 20000; ++k) acc += k & 1;  
      data[idx[0]] = 42 + (acc & 0);
    });

    e.get_info<sycl::info::event::command_execution_status>();

    qb.submit([&](sycl::handler &cgh) {
      cgh.depends_on(e);
      cgh.parallel_for(sycl::range<1>(n), [=](sycl::id<1> idx) {
        out[idx[0]] = data[idx[0]];
      });
    });
    qb.wait();
    qa.wait();

    int bad = 0;
    for (std::size_t i = 0; i < n; ++i) if (out[i] != 42) ++bad;
    if (bad) { std::printf("rep %2d: %d of %zu elements stale\n", rep, bad, n); bad_total += bad; }
  }
  std::printf(bad_total ? "DEPENDENCY DROPPED\n" : "all reps correct\n");
  sycl::free(data, qa); sycl::free(out, qa);
  return bad_total != 0;
}
```

the problem existed before but it was not reachable because the operation was committed immediately

query_status() should flush and compare against a signal recorded by flush(), not _event_counter

src/runtime/metal/metal_queue.cpp
```diff
@@ -709,6 +794,7 @@ result metal_inorder_queue::submit_external_wait_for(const dag_node_ptr& node) {
 }
 
 result metal_inorder_queue::wait() {
+  std::lock_guard<std::recursive_mutex> lock{_mutex};
```
resetius: I think this lock should be taken after wait

src/runtime/metal/metal_queue.cpp
```diff
@@ -691,9 +774,11 @@ result metal_inorder_queue::submit_queue_wait_for(const dag_node_ptr& node) {
   inorder_queue_event<metal_event_handle>* metal_evt = cast<inorder_queue_event<metal_event_handle>>(evt.get());
   auto handle = metal_evt->request_backend_event();
   NS::SharedPtr<NS::AutoreleasePool> pool = NS::TransferPtr(NS::AutoreleasePool::alloc()->init());
-  auto* cmd_buf = new_command_buffer();
+  // Appending to the open buffer preserves order: encoders within a command
+  // buffer run in encoding order, so prior batched work precedes the wait.
```
resetius: Metal does not promise that encoders in a command buffer run in encoding order
the code is correct for a different reason: an MTLEvent wait applies to the commands encoded after it

src/runtime/metal/metal_queue.cpp
```diff
@@ -333,11 +311,70 @@ MTL::CommandBuffer* metal_inorder_queue::new_command_buffer() {
...
+
   return cmd_buf;
 }
 
+MTL::CommandBuffer* metal_inorder_queue::new_dedicated_command_buffer() {
+  // Relies on the caller's autorelease pool, like the rest of Metal-cpp.
+  return prepare_command_buffer(_command_queue->commandBuffer());
+}
+
+result metal_inorder_queue::flush() {
```
resetius, line 341: commits without an autorelease pool
but it is reached from get_native_type() and from the destructor with no pool on the stack

src/runtime/metal/metal_queue.cpp
anudit: Fixed. query_status() now flushes under the queue mutex and checks the recorded completion signal. Staged copies also update that completion target.
Your reproducer passes, and I added regression coverage for partial batches and the automatic flush boundary.

src/runtime/metal/metal_queue.cpp
anudit: Fixed. The GPU wait runs without holding the queue mutex. Afterward, the lock protects fence cleanup, which is skipped if another thread submitted newer work. Verified queue access remains available during a blocked GPU wait.

src/runtime/metal/metal_queue.cpp
anudit: Corrected the comment: the event wait applies to commands encoded after it. Removed the claim that encoders execute in encoding order.

src/runtime/metal/metal_queue.cpp
anudit: Added a local autorelease pool inside flush(), covering signal encoding, commit, and release. Exercised native access and queue destruction without a caller-provided pool.

resetius:
The patch revealed another issue 
We should wait for the command buffer to complete not for the `MTLEvent`, since the command buffer may contain callbacks

Smth like this:
```cpp
  // metal_queue.cpp, insert_event()
  NS::SharedPtr<MTL::CommandBuffer> cb = NS::RetainPtr(get_open_command_buffer());
  flush();
  return std::make_shared<metal_node_event>(metal_event_handle{...}, cb.get());

  // metal_event.cpp
  void metal_node_event::wait() {
    _command_buffer->waitUntilCompleted();   // was: waitUntilSignaledValue(...)
  }
```

anudit:
> The patch revealed another issue We should wait for the command buffer to complete not for the `MTLEvent`, since the command buffer may contain callbacks
> 
> Smth like this:
> 
> ```c++
>   // metal_queue.cpp, insert_event()
>   NS::SharedPtr<MTL::CommandBuffer> cb = NS::RetainPtr(get_open_command_buffer());
>   flush();
>   return std::make_shared<metal_node_event>(metal_event_handle{...}, cb.get());
> 
>   // metal_event.cpp
>   void metal_node_event::wait() {
>     _command_buffer->waitUntilCompleted();   // was: waitUntilSignaledValue(...)
>   }
> ```

Addressed. insert_event() now retains the command buffer before flushing and passes it to metal_node_event. metal_node_event retains/releases the command buffer, and wait() uses waitUntilCompleted(), ensuring command-buffer completion handlers have finished before returning. The full Metal suite passes locally.

resetius:
> The full Metal suite passes locally

Confirmed. By the way, I could only reproduce it in a VM, not locally 
It looks like locally the queue was cleared before the destructor had a chance to run

resetius:
the build is broken after merging `import events`
need to call
```cpp
return std::make_shared<metal_node_event>(*handle, nullptr); // metal_backend.cpp
```

```cpp
metal_node_event::metal_node_event(metal_event_handle handle, MTL::CommandBuffer* command_buffer)
    : _handle{handle}, _command_buffer{command_buffer} {
    _handle.event->retain();
    if (_command_buffer) {
      _command_buffer->retain();
    }                                                                                                                                                                                                    
  }

  metal_node_event::~metal_node_event() {
    if (_command_buffer) {
      _command_buffer->release();
    }
    _handle.event->release();
  }                                                                                                                                                                                                      

  void metal_node_event::wait() {
    // The shared event may signal before completion handlers have finished.
    if (_command_buffer) {
      _command_buffer->waitUntilCompleted();
      return;
    }
    while (!_handle.event->waitUntilSignaledValue(_handle.value, 1000)) {
      std::this_thread::yield();
    }                                                                                                                                                                                                    
  }
```

resetius (approved):

resetius:
LGTM

resetius:
The new test revealed a bug in the OpenMP backend. Obviously, this should be fixed in a separate PR. The fix may be simple

If this PR needs to be merged urgently, we can limit the new test to the Metal backend for now. But I would wait

resetius:
@anudit rebase please

PR #2198, opened 2026-08-23
[Metal] Fix structurization pass on LLVM-22

resetius:
prep step for llvm22

### DIFF
--- a/src/compiler/llvm-to-backend/metal/LLVMToMetal.cpp
+++ b/src/compiler/llvm-to-backend/metal/LLVMToMetal.cpp
@@ -612,7 +612,9 @@ bool LLVMToMetalTranslator::translateToBackendFormat(llvm::Module& FlavoredModul
     FPM.addPass(llvm::DCEPass());
     FPM.addPass(llvm::ADCEPass());
     FPM.addPass(llvm::StructurizeCFGPass());
-    FPM.addPass(llvm::SimplifyCFGPass());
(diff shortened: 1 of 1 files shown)


src/compiler/llvm-to-backend/metal/LLVMToMetal.cpp
```diff
@@ -612,7 +612,9 @@ bool LLVMToMetalTranslator::translateToBackendFormat(llvm::Module& FlavoredModul
     FPM.addPass(llvm::DCEPass());
     FPM.addPass(llvm::ADCEPass());
     FPM.addPass(llvm::StructurizeCFGPass());
-    FPM.addPass(llvm::SimplifyCFGPass());
+    // Since LLVM 22 this pass removes the Flow blocks that StructurizeCFG creates,
+    // so a loop gets more than one exit again. HLExtractorPass needs exactly one
+    FPM.addPass(llvm::SimplifyCFGPass(llvm::SimplifyCFGOptions().setSimplifyCondBranch(false)));
```
sbalint98, line 617: I wonder shouldn`t be this gated for later versions than LLVM 22?

resetius: I don't think we need to gate this. It is better to move SimplifyCFG before StructurizeCFG and remove the flag. Otherwise, a future LLVM version may break the structured CFG again

sbalint98 (approved):

PR #2199, opened 2026-08-23
[Metal] Fix code generation for 128 bit integers

resetius:
prep step for llvm22

### DIFF
--- a/src/compiler/llvm-to-backend/metal/Emitter.cpp
+++ b/src/compiler/llvm-to-backend/metal/Emitter.cpp
@@ -75,6 +75,22 @@ std::optional<uint64_t> getConstU64(llvm::Value* V) {
   return std::nullopt;
 }
 
+bool isWideInteger(const Type* T) {
+  return T->isIntegerTy() && T->getIntegerBitWidth() > 64;
+}
+
+constexpr unsigned supportedWideIntegerOps[] = {
+  Instruction::Shl, Instruction::LShr, Instruction::AShr,
+  Instruction::And, Instruction::Or,   Instruction::Xor,
+};
+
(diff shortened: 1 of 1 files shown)


al42and:
Thanks! Can confirm that this fixes tests for me too :+1: 

But there is yet another case which fails with a similar error even with this PR: `as_type<uint4>((__as_signed(t_sroa_01158_0_copyload_i)) >> (__as_signed(uint4(0x60u, 0x0u, 0x0u, 0x0u))));`

```cpp
#include <sycl/sycl.hpp>

sycl::vec<int, 4> load4(const int* ptr)
{
    sycl::vec<int, 4> v;
    for (int i = 0; i < 4; ++i)
        v[i] = ptr[i];
    return v;
}

int main()
{
    sycl::queue q;
    int         in[4] = { 1, 2, 3, 4 };
    int64_t     out   = 0;
    {
        sycl::buffer<int>     b_in{ in, 4 };
        sycl::buffer<int64_t> b_out{ &out, 1 };

        q.submit(
                [&](sycl::handler& cgh)
                {
                    auto in_acc  = b_in.get_access<sycl::access_mode::read>(cgh);
                    auto out_acc = b_out.get_access<sycl::access_mode::write>(cgh);

                    cgh.single_task(
                            [=]()
                            {
                                sycl::vec<int, 4> v = load4(&in_acc[0]);
                                out_acc[0] = static_cast<int64_t>(v.s3());
                            });
                });
    }
    return 0;
}
```

resetius:
> But there is yet another case which fails with a similar error even with this PR

I hope I've covered all possible cases. I intentionally don't support `__int128` arithmetic.

al42and (approved):
Works for me!

PR #2200, opened 2026-08-24
[SSCP][prototype] Implement sycl_khr_free_function_kernels

koparasy:
Proof-of-concept implementation of the KHR free function kernels extension ([KhronosGroup/SYCL-Docs#1033](https://github.com/KhronosGroup/SYCL-Docs/pull/1033)) on the generic SSCP flow.

A namespace-scope function decorated with SYCL_KHR_KERNEL() becomes a device kernel entry point whose parameters are the kernel arguments. The kernel is identified by its address via kernel_function<Func> and launched with launch_task / launch_grouped, routed through the normal handler/DAG path (events, dependencies, graphs work as usual).

The macro maps onto the existing SSCP kernel-entry-point annotations (hipsycl_sscp_kernel + hipsycl_sscp_outlining), reusing the same outlining/HCF machinery as PCUDA free kernels, but WITHOUT the acpp_free_kernel host-call rewrite: an ordinary host call to the function is not a launch; only the explicit launch_* functions launch.

Host launch resolves the function address to its mangled HCF kernel name (templated __acpp_sscp_extract_kernel_name variant), packs each argument as a separate kernel parameter, and submits by name. A name->object index in hcf_cache (get_kernel_info_by_name) enables cross-translation-unit launches: define the kernel in one TU, launch from another. EntrypointPreparationPass now only emits a kernel for a function with a body, so a decorated declaration in the launching TU does not register a body-less kernel.

This should be used for illustration purposes only. 

Assisted by Claude@4.8

### DIFF
--- a/include/hipSYCL/glue/kernel_launcher_data.hpp
+++ b/include/hipSYCL/glue/kernel_launcher_data.hpp
@@ -67,6 +67,13 @@ struct kernel_launcher_data {
   const char* sscp_kernel_id = nullptr;
   const rt::hcf_kernel_info* kernel_info = nullptr;
   invoker_function_t sscp_invoker;
+
+  // Used by the free function kernel launcher (sycl_khr_free_function_kernels):
+  // Unlike lambda/functor kernels - which pass a single closure struct as one
+  // kernel argument - a free function kernel passes each function parameter as
+  // a separate kernel argument. This holds the per-argument sizes; the argument
+  // values themselves are packed contiguously into kernel_args.
+  std::vector<std::size_t> sscp_argument_sizes;
 };
 
 
--- a/include/hipSYCL/glue/llvm-sscp/free_function_kernel_launcher.hpp
+++ b/include/hipSYCL/glue/llvm-sscp/free_function_kernel_launcher.hpp
@@ -1,0 +1,202 @@
+/*
+ * This file is part of AdaptiveCpp, an implementation of SYCL and C++ standard
+ * parallelism for CPUs and GPUs.
+ *
+ * Copyright The AdaptiveCpp Contributors
+ *
+ * AdaptiveCpp is released under the BSD 2-Clause "Simplified" License.
+ * See file LICENSE in the project root for full license details.
+ */
+// SPDX-License-Identifier: BSD-2-Clause
+#ifndef HIPSYCL_FREE_FUNCTION_KERNEL_LAUNCHER_HPP
+#define HIPSYCL_FREE_FUNCTION_KERNEL_LAUNCHER_HPP
+
+#include "hipSYCL/glue/kernel_launcher_data.hpp"
+#include "hipSYCL/glue/llvm-sscp/s1_ir_constants.hpp"
+#include "hipSYCL/runtime/kernel_cache.hpp"
+#include "hipSYCL/runtime/code_object_invoker.hpp"
+#include "hipSYCL/runtime/operations.hpp"
+#include "hipSYCL/runtime/dag_node.hpp"
+#include "hipSYCL/runtime/error.hpp"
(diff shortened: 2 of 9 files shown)


illuhad:
I haven't looked through it in detail, because large parts of it seem to just be duplicated functionality and mechanisms that we already have elsewhere.

But I don't really understand how it solves cross-TU kernel launches.

> EntrypointPreparationPass now only emits a kernel for a function with a body, so a decorated declaration in the launching TU does not register a body-less kernel.

That's not the challenge.

AdaptiveCpp decomposes every non-scalar aggregate argument into scalars for the kernels that it generates. There are layers on top that assume that this works reliably. (because of this, the assumption of the KHR that we can have 1:1 mapping of actual kernel arguments to function arguments is also a bit bogus)

When you have a sufficiently complex aggregate argument and have opaque pointers, then, depending on your ABI, the decomposition can fail because the kernel compiler might not be able to reconstruct the actual argument type.
This is not a problem in the same TU because the kernel launch usually provides sufficient information for the device compiler (e.g. `alloca` instructions for the parameters) to reconstruct the type.

As far as I can see the PR does not solve this.

koparasy:
> I haven't looked through it in detail, because large parts of it seem to just be duplicated functionality and mechanisms that we already have elsewhere.
> 
> But I don't really understand how it solves cross-TU kernel launches.
> 
> > EntrypointPreparationPass now only emits a kernel for a function with a body, so a decorated declaration in the launching TU does not register a body-less kernel.
> 
> That's not the challenge.
> 
> AdaptiveCpp decomposes every non-scalar aggregate argument into scalars for the kernels that it generates. There are layers on top that assume that this works reliably. (because of this, the assumption of the KHR that we can have 1:1 mapping of actual kernel arguments to function arguments is also a bit bogus)
> 
> When you have a sufficiently complex aggregate argument and have opaque pointers, then, depending on your ABI, the decomposition can fail because the kernel compiler might not be able to reconstruct the actual argument type. This is not a problem in the same TU because the kernel launch usually provides sufficient information for the device compiler (e.g. `alloca` instructions for the parameters) to reconstruct the type.
> 
> As far as I can see the PR does not solve this.

It is likely/expected for this to have duplicated code, cause I don't have internal knowledge of the code base and I take everything that the model spits out at face value. 

This is a PR that should show the intend of the KHR, not adopt the implementation as is. 

> When you have a sufficiently complex aggregate argument and have opaque pointers, then, depending on your ABI, the decomposition can fail because the kernel compiler might not be able to reconstruct the actual argument type.
This is not a problem in the same TU because the kernel launch usually provides sufficient information for the device compiler (e.g. alloca instructions for the parameters) to reconstruct the type.

I believe I understand what you are pointing at, but I would expect this to become an issue when arguments are coerced. You should be able  to get the type of the argument before ABI lowering by looking the `alloca`'s type (IIRC `getAllocatedType`). 

In any case, I now understand what your expected challenge is, and you should also understand what the KHR is about.

illuhad:
> This is a PR that should show the intend of the KHR, not adopt the implementation as is.

Yes, I understand, and that was also my impression -- and also why I did not look through it in detail. I just don't want review a pile of AI-generated code which is clearly not in shape to be reviewed :)

> I believe I understand what you are pointing at, but I would expect this to become an issue when arguments are coerced. You should be able to get the type of the argument before ABI lowering by looking the alloca's type (IIRC getAllocatedType).

We already use that API. It's really only an issue if the call is not seen by the device compiler as I said.

> In any case, I now understand what your expected challenge is, and you should also understand what the KHR is about.

And I think it's become clear that there's a mismatch between what the KHR assumes, and what existing implementations assume.

koparasy:
> And I think it's become clear that there's a mismatch between what the KHR assumes, and what existing implementations assume.

We can let the workgroup decide this.

PR #2205, opened 2026-08-26
[VK] Improve device driver version reporting

EwanC:
[VkPhysicalDeviceDriverProperties](https://docs.vulkan.org/refpages/latest/refpages/source/VkPhysicalDeviceDriverProperties.html) is available from Vulkan version 1.2, which is our minimum backend version requirement. It provides a `driverInfo` field that contains a version string, this PR updates the code to use that as opposed to the current approach of assembling a driver string from an integer value that may not be correct for all vendors.

For example with `acpp-info` on my Intel MTL Vulkan device:
* Before -  `Driver version: 26.0.3`
* After - `Driver version: Intel open-source Mesa driver, Mesa 26.0.3-1ubuntu1`

This PR also includes a couple of other small fixes:
* Allow valdation layer checking of  Vulkan instance creation
* Document the NVIDIA Ada device workaround reported in https://github.com/AdaptiveCpp/AdaptiveCpp/pull/2009#discussion_r3847010970

### DIFF
--- a/doc/install-vulkan.md
+++ b/doc/install-vulkan.md
@@ -175,18 +175,20 @@ own bugs that need addressed (which may turn out to be not device specific once 
 Devices known to be well supported by the backend include:
 
 | Driver Name  | Driver Version    | Device Type          | Status
-| ------------ | ----------------- | -------------------- | ------------------------------------------------------------------------------------- |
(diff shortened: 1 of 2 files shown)


sbalint98 (approved):
LGTM

PR #2206, opened 2026-08-26
[CI] Fix current CI failures

sbalint98:
The logic merged before in #2195 was broken, this PR ~~reverts~~ fixes that. 

Additionally this PR pins the lit version to 18.1.18 to temporally fix the issue reported in #2207

### DIFF
--- a/.github/workflows/create-ci-docker/AdaptiveCpp-CI-Linux.dockerfile
+++ b/.github/workflows/create-ci-docker/AdaptiveCpp-CI-Linux.dockerfile
@@ -90,7 +90,19 @@ RUN <<EOF
     set -e
     wget -q https://apt.llvm.org/llvm.sh
     chmod +x llvm.sh
-    ./llvm.sh ${LLVM_VERSION}
+    success=0
+    for i in 1 2 3 4 5; do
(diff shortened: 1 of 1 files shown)


tdavidcl:
I'm not sure if you should remove that, did it help making it more stable ? 
Currently i'm experiencing a **lot** of failures on Shamrock repo at the ./llvm.sh step. Only yesterday i got 9 auto retries of that job alone because of that -> https://dashboard.mergify.com/orgs/Shamrock-code/repos/Shamrock/ci-insights/auto-retry
The use of llvm.sh is always unstable in runner ...

sbalint98:
I think it is fine to revert this, since these images are only re-built in case there are changes to the image building workflow or the dockerfile. So after a successful build, this script is most likely not going to be running until the next time we want to support a new LLVM version. If manually re-starting these rare jobs gets too annoying we can always add back the logic for retrying.

sbalint98:
@tdavidcl Actually I think you are right to keep that in. I tested this change on my fork, and it took 5 retries until I managed to get all 5 docker images building... https://github.com/sbalint98/AdaptiveCpp/actions/runs/32964901097

However it seems that there is some issue with the intel repo as well...

illuhad (approved):

Issue #2208, opened 2026-08-27
allocation_map lifetime issues with ACPP_PERSISTENT_RUNTIME


al42and:
**Bug summary**

When `ACPP_PERSISTENT_RUNTIME=1` is set, the application sporadically segfaults at shutdown when a backend with allocation tracking is used (verified with OpenCL and Vulkan).

**To Reproduce**

The example below most reliably triggers the failure. I've seen it even with a single `foo` invocation, but is takes longer to manifest.

```console
$ acpp sycl-app-reinit.cpp --acpp-targets=generic -O3 -g
$ i=0; while true; do i=$(($i+1)); ACPP_PERSISTENT_RUNTIME=1 ACPP_VISIBILITY_MASK=opencl:A770 ./a.out &>/dev/null || break; done; echo "Crashed after $i repeats"
Aborted (core dumped)
Crashed after 17 repeats
$ i=0; while true; do i=$(($i+1)); ACPP_PERSISTENT_RUNTIME=1 ACPP_VISIBILITY_MASK=vk:A770 ./a.out &>/dev/null || break; done; echo "Crashed after $i repeats"
Segmentation fault (core dumped)
Crashed after 6 repeats
$ i=0; while true; do i=$(($i+1)); ACPP_PERSISTENT_RUNTIME=1 ACPP_VISIBILITY_MASK=vk:NVIDIA ./a.out &>/dev/null || break; done; echo "Crashed after $i repeats"
Segmentation fault (core dumped)
Crashed after 4 repeats
```

<details><summary>sycl-app-reinit.cpp</summary>
<p>

```cpp
#include <sycl/sycl.hpp>

#include <iostream>

class Kernel;

void foo() {
  // Creating buffer of 4 elements to be used inside the kernel code
  sycl::buffer<size_t, 1> Buffer(4);

  for (const auto &Device : sycl::device::get_devices()) {
    // Creating SYCL queue
    //sycl::queue queue(Device, {sycl::property::queue::in_order()});
    sycl::queue queue{Device, {sycl::property::queue::AdaptiveCpp_priority{1}, sycl::property::queue::in_order()}};

    std::cout << "Running on "
              << queue.get_device().get_info<sycl::info::device::name>()
              << "\n";

    static const int numThreads = 8;

    sycl::buffer<int, 1> buffer(numThreads);
    queue
        .submit([&](sycl::handler &cgh) {
          auto d_buffer =
[...]

PR #2209, opened 2026-08-27
allocator: Unregister the pointer before freeing it

al42and:
If one worker thread was deallocating a data region, it
was first freeing the pointer and only then unregistering it.

If the application was allocating something at the same time,
it could just so happen that the just-freed pointer is returned,
inserted into the map over the existing one, and immediately
unregistered by the worker thread, leaving `allocation_map`
in a broken state.

Note: applying this patch masks #2208 a bit since `raw_free` was
delaying the `on_deallocation` call.


illuhad (approved):
LGTM.

Also, couldn't help but read your text as a poem.
Love the enjambement before "in a broken state" -- the form mirrors the breakage. Perfect craftsmanship! :D

PR #2210, opened 2026-08-27
[VK] Fix lifetime issue in device::has(execution_timestamps)

al42and:


EwanC (approved):
Ah woops, `getQueueFamilyProperties()` is returning a `std::vector` and we're taking a reference into it even though it's immediately being destroyed,

Thanks for the fix.

PR #2212, opened 2026-08-28
Use internal lit shell

sbalint98:
see https://github.com/AdaptiveCpp/AdaptiveCpp/pull/2207

cc @EwanC


EwanC (approved):
LGTM, thanks for finishing this off

PR #2215, opened 2026-08-30
[Metal] Deduplicate residency bindings

anudit:
Split out of #2196 per [review](https://github.com/AdaptiveCpp/AdaptiveCpp/pull/2196#issuecomment-5462113940): this part is independent of the command buffer batching work there, so it's reviewable on its own.

## Background

SSCP kernels may dereference arbitrary USM pointers, so every allocation must be made resident for each dispatch's compute encoder. Previously this was done with `metal_allocator::for_each_buffer`, which took the allocator lock and called `useResource()` once per allocation on every single kernel launch.

## Changes

- `metal_allocator` gains a lock-free monotonic `generation()` counter, bumped on every allocation/free, and a `snapshot_buffers()` method that returns a retained snapshot of all backing buffers together with the generation it was taken at.
- `metal_inorder_queue::make_allocations_resident()` caches that snapshot and only rebuilds it (taking the allocator lock) when the generation has changed since the last check; otherwise it just passes the cached set to `useResources()`. This turns the per-launch residency cost from O(allocations) under the allocator mutex into a single lock-free generation check plus one `useResources()` call in the common case.

## Testing

`sycl_tests`, `pcuda_tests`, `rt_tests` pass on macOS/arm64 (M2 Max, macOS 26).

### DIFF
--- a/src/runtime/metal/metal_queue.cpp
+++ b/src/runtime/metal/metal_queue.cpp
@@ -106,6 +106,7 @@ result launch_kernel_from_library(
   MTL::Library* library,
   MTL::Device* device,
   MTL::CommandBuffer* command_buffer,
+  metal_inorder_queue* queue,
   metal_allocator* allocator,
   std::string_view kernel_name,
   const rt::range<3>& num_groups,
@@ -196,10 +197,10 @@ result launch_kernel_from_library(
       encoder, device, allocator, function.get(), args, arg_sizes, num_args, is_pointer_arg, buffers_out, buf_offset);
   }
 
+  // SSCP kernels may dereference arbitrary USM pointers, so every allocation
+  // must be resident for this encoder.
   // TODO: switch to MTL4 API for O(1) buffer bindings
-  allocator->for_each_buffer([&](MTL::Buffer* buf) {
-    encoder->useResource(buf, MTL::ResourceUsageRead | MTL::ResourceUsageWrite);
-  });
+  queue->make_allocations_resident(encoder);
 
   MTL::Size num_groups_size = MTL::Size::Make(
     num_groups[0],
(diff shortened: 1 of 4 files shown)


src/runtime/metal/metal_queue.cpp
```diff
@@ -855,6 +881,7 @@ metal_inorder_queue::~metal_inorder_queue() {
   // Drain pending work before releasing queue-owned Metal objects that may be
   // referenced by completion handlers.
   wait();
+  release_resident_resources();
```
resetius, line 884: I'm a bit concerned about release_resident_resources() 
Retaining the snapshot entries effectively makes the queue an owner of allocator resources, so free() may not release memory until the cache is refreshed or the queue is destroyed

Perhaps an RCU-style approach would fit better: keep ownership in the allocator, cache a non-owning generation-tagged snapshot in the queue, and defer reclamation only while a snapshot is being passed to useResources(). Please try this direction 
if useful, I can post a concrete sketch later.

resetius: I'll try to provide some concrete code snippets on Saturday to show how this could work without breaking memory ownership

I've just hit the same lock in my application :) I removed it for now locally. Luckily, I don't have any indirect structures

resetius: I started to use `MTL::ResidencySet` here: https://github.com/AdaptiveCpp/AdaptiveCpp/pull/2221

anudit: Tested #2215 @ 1be7c970, #2221 @ 42d8261

Launch-only benchmark (2nd comment, 32 threads, N allocations, 500 launches), µs/launch:

Allocations | #2215 | #2221
-- | -- | --
10 | 35.1 | 26.9
100 | 38.5 | 25.6
1000 | 158.1 | 25.8
4000 | 504.0 | 25.9

#2221  is definitely better

sbalint98:
@anudit do I understand correctly that this change is superseded by #2221 ?

anudit:
> @anudit do I understand correctly that this change is superseded by #2221 ?

It should I think.

PR #2216, opened 2026-09-01
add default contructor for nd_range

TApplencourt:
Implemented https://github.com/KhronosGroup/SYCL-Docs/pull/1046

And added tests for extra-fancyness, but CTS tests it too, so as you want. I can remove it .


illuhad (approved):
Thanks!

@sbalint98 Please delay merging until SYCL spec change https://github.com/KhronosGroup/SYCL-Docs/pull/1046 is actually merged.

PR #2219, opened 2026-09-03
SSCP: simplify functions before force-inlining them into the kernel

mikaoP:
Run GVN + SimplifyCFG on every function before AlwaysInlinerPass, as the -O3 pipeline does before its own always-inliner. Inlining drops the callee's parameter attributes, and `dereferenceable` on a `const T&` parameter is what SimplifyCFG needs to fold a short-circuited condition into selects; once inlined, the proof is gone. Adds a lit test for it.

### DIFF
--- a/src/compiler/llvm-to-backend/LLVMToBackend.cpp
+++ b/src/compiler/llvm-to-backend/LLVMToBackend.cpp
@@ -29,6 +29,9 @@
 #include <cstdint>
 
 #include <llvm/Transforms/IPO/AlwaysInliner.h>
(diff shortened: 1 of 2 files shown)


illuhad (commented):
Conceptually I think the change is good; the added LIT test however seems to fail because for some reason the IR file that it looks for does not exist. Can you check what's going on there?

mikaoP:
@illuhad hopefully now passes the ci

illuhad (approved):
Thanks!

PR #2223, opened 2026-09-05
[Metal] Import native events into SYCL

resetius:
The Metal backend already exports a `sycl::event` as a native `MTL::SharedEvent` via `sycl::get_native<sycl::backend::metal>()` 
The PR adds the opposite direction: importing a native `MTL::SharedEvent` as a `sycl::event`

This makes it possible to mix SYCL kernels with command buffers submitted directly through the Metal API: on a `MTL::CommandQueue` that AdaptiveCpp does not own 
Without it the only way to order such work against a SYCL kernel is a host round-trip: `waitUntilCompleted()` on the producing command buffer before submitting to the SYCL queue 

With an imported event `handler::depends_on()` lowers to `MTL::CommandBuffer::encodeWait()` on the SYCL queue, so the dependency is resolved on the GPU timeline and the host is not blocked

Snippet:
```
  #include <Metal/Metal.hpp>
  #include <sycl/sycl.hpp>

  MTL::CommandBuffer *cb = my_queue->commandBuffer();
  // ... encode work producing `data` ...
  cb->encodeSignalEvent(my_event, signal_value);
  cb->commit();               // no waitUntilCompleted(): the host does not block

  sycl::queue q{sycl::property::queue::in_order{}};
  sycl::event produced = sycl::make_event<sycl::backend::metal>(
      {my_event, signal_value}, q.get_context());

    cgh.depends_on(produced);
    cgh.parallel_for(N, [=](sycl::id<1> i) { data[i] *= 2.f; });
  });
```

Concrete use-case example: https://github.com/resetius/fdm/blob/master/test/ns_cyl_sycl_demo.cpp#L443

### DIFF
--- a/include/hipSYCL/glue/metal/metal_interop.hpp
+++ b/include/hipSYCL/glue/metal/metal_interop.hpp
@@ -21,6 +21,7 @@
 #include "hipSYCL/runtime/error.hpp"
 #include "hipSYCL/runtime/inorder_queue_event.hpp"
 #include "hipSYCL/runtime/metal/metal_event.hpp"
+#include "hipSYCL/runtime/operations.hpp"
 #if defined(__APPLE__)
 #include "hipSYCL/runtime/metal/metal_hardware_manager.hpp"
 #endif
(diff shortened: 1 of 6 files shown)


include/hipSYCL/glue/metal/metal_interop.hpp
```diff
@@ -173,6 +174,75 @@ template <> struct backend_interop<sycl::backend::metal> {
...
+    rt::runtime *runtime = ctx.AdaptiveCpp_runtime();
+    rt::backend *backend =
+        runtime->backends().get(rt::backend_id::metal);
+    if (!backend) {
+      rt::register_error(
+          __acpp_here(),
+          rt::error_info{"make_event: Metal backend not available",
+                         rt::error_type::runtime_error});
+      return {};
+    }
```
sbalint98, line 193: I think here we should also check if the passed context is a Metal context, and if not throw a `errc::backend_mismatch`

sepc [4.5.1.3](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:backend-interoperability-make) says:


> Overloads of the make_{sycl_class} function which take a SYCL [context](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#context) object as an argument must throw an exception with the errc::backend_mismatch error code if the backend of the provided SYCL context doesn’t match the target backend.

sbalint98 (commented):
One issue with regard to conformance to the spec, otherwise LGTM!

include/hipSYCL/glue/metal/metal_interop.hpp
resetius: Added expections

sbalint98 (approved):

Issue #2217, opened 2026-09-03
[VK] Add pipeline barrier synchronization


EwanC:
In the Vulkan backend we are current using timeline semaphores to synchronize command execution. However, in vulkan memory synchronization is also required through the use of pipeline barriers that express memory dependencies.  As the backend only ever uses a single command in a command-buffer we have got away without using these to order commands within a command-buffer. However, on some devices they may be required for correct execution.

I think we can use buffer specific `VkBufferMemoryBarrier` objects rather than the more generic `VkMemoryBarrier` because we'll know what buffers are used in a command.

crystarm:
Hi!! (˶ᵔᗜᵔ˶)ﾉﾞ

One thing worth checking before adding pipeline barriers is that semaphore signal/wait operations already establish a memory dependency between submissions. Since the backend currently puts one command into each command buffer and chains submissions through a timeline semaphore, additional buffer barriers may not be necessary.

However, `submit_command_buffer()` currently uses `eComputeShader` as the wait stage for every command. This does not cover `copyBuffer()` operations running at the transfer stage, so those may execute before the semaphore wait is satisfied. Using `eAllCommands` or passing the appropriate stage for each command may address the actual synchronization issue.

I'll likely dig deeper into this and work on a fix.

EwanC:
> One thing worth checking before adding pipeline barriers is that semaphore signal/wait operations already establish a memory dependency between submissions. Since the backend currently puts one command into each command buffer and chains submissions through a timeline semaphore, additional buffer barriers may not be necessary.

Thanks for this comment! I taken a closer look at the semaphore [signalling](https://docs.vulkan.org/spec/latest/chapters/synchronization.html#synchronization-semaphores-signaling) and [waiting](https://docs.vulkan.org/spec/latest/chapters/synchronization.html#synchronization-semaphores-waiting) sections of the specification and see the languages about memory ordering. So I would agree that finer grain per-buffer barriers wouldn't add anything when there's already a "all memory accesses" dependency access scope.

> However, submit_command_buffer() currently uses eComputeShader as the wait stage for every command. This does not cover copyBuffer() operations running at the transfer stage, so those may execute before the semaphore wait is satisfied. Using eAllCommands or passing the appropriate stage for each command may address the actual synchronization issue.

> I'll likely dig deeper into this and work on a fix.

This is a better diagnosis of a potential bug and improvement we could make, thank you.

EwanC:
[...]

PR #2225, opened 2026-09-07
Fix Vulkan semaphore wait stage mask

crystarm:
Related to #2217.

I spent quite some time double-checking whether `eAllCommands` is the right fix here, and I believe it is. The semaphore already provides the required memory dependency, but `eComputeShader` does not cover transfer commands such as `copyBuffer()`. Since submitted command buffers may contain different types of commands, `eAllCommands` seems to be the correct and safest choice.

## Summary

Change the Vulkan semaphore wait stage mask in `vk_queue::submit_command_buffer()` from `eComputeShader` to `eAllCommands`.

This ensures that queue dependencies are respected by every command in the submitted command buffer, including transfer operations such as `copyBuffer()`.

## References

[Vulkan semaphore signaling](https://docs.vulkan.org/spec/latest/chapters/synchronization.html#synchronization-semaphores-signaling)
[Vulkan semaphore waiting](https://docs.vulkan.org/spec/latest/chapters/synchronization.html#synchronization-semaphores-waiting)

### DIFF
--- a/src/runtime/vk/vk_queue.cpp
+++ b/src/runtime/vk/vk_queue.cpp
@@ -519,7 +519,7 @@ void vk_queue::submit_command_buffer(vk::CommandBuffer &cmd_buf) {
   vk::TimelineSemaphoreSubmitInfo timeline_info{
       num_wait_semaphores, wait_values.data(), 1u, &signal_value};
   std::vector<vk::PipelineStageFlags> wait_stages(
-      num_wait_semaphores, vk::PipelineStageFlagBits::eComputeShader);
+      num_wait_semaphores, vk::PipelineStageFlagBits::eAllCommands);
   vk::SubmitInfo submit_info(num_wait_semaphores, semaphores.data(),
                              wait_stages.data(), 1, &cmd_buf, 1,
                              semaphores.data());


crystarm:
Ping @EwanC

src/runtime/vk/vk_queue.cpp
```diff
@@ -519,7 +519,7 @@ void vk_queue::submit_command_buffer(vk::CommandBuffer &cmd_buf) {
   vk::TimelineSemaphoreSubmitInfo timeline_info{
       num_wait_semaphores, wait_values.data(), 1u, &signal_value};
   std::vector<vk::PipelineStageFlags> wait_stages(
-      num_wait_semaphores, vk::PipelineStageFlagBits::eComputeShader);
+      num_wait_semaphores, vk::PipelineStageFlagBits::eAllCommands);
```
EwanC, line 522: Could this be ` vk::PipelineStageFlagBits::eComputeShader | vk::PipelineStageFlagBits::eTransfer`, since those are the only two types of command we could have in our command-buffers right now? or do you think being explicit does buy us much and could be a potential source of bugs in the future if that changes

crystarm: Well, that would work for the current operations, but `eAllCommands` better matches the full queue/event dependency semantics imao and avoids having to update the mask when new command types are added. If you don't mind, I'd keep it.

EwanC: Thanks for the explanation, wanted to make sure I understood. Yes let's keep it as is.

EwanC (approved):
LGTM

Merging this since the Vulkan Windows CI fail is due to an unrelated issue I'm addressing in https://github.com/AdaptiveCpp/AdaptiveCpp/pull/2226 and the other CI fail is for hip multipath which doesn't stress the Vulkan backend.

PR #2226, opened 2026-09-07
[CI] Reduce rate limiting on Vulkan-SDK downloads

EwanC:
In Vulkan jobs we sometimes see errors in the stage for downloading the Vulkan-SDK via the https://github.com/jakoch/install-vulkan-sdk-action action. Example https://github.com/AdaptiveCpp/AdaptiveCpp/actions/runs/34134711168/job/101804537331?pr=2225

```
To avoid hitting GitHub API rate limits, please set github_token as action input or a GITHUB_TOKEN in your environment.
Error: HttpClientError: API rate limit exceeded for 4.155.237.0. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
    at mr.<anonymous> (file:///D:/a/_actions/jakoch/install-vulkan-sdk-action/v1/dist/index.js:63:31542)
    at Generator.next (<anonymous>)
    at a (file:///D:/a/_actions/jakoch/install-vulkan-sdk-action/v1/dist/index.js:63:20903)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
```

This can be mitigate by providing an Github Token API key. This PR does this for the 2 jobs where we use the third party action, but explicitly locks down the job permissions to read-only so that if the third party action is compromised the token can't be abused in the CI job.

### DIFF
--- a/.github/workflows/linux-vulkan.yml
+++ b/.github/workflows/linux-vulkan.yml
@@ -2,6 +2,8 @@ name: Linux Vulkan
 
 on: [push, pull_request]
 
+permissions: read-all
+
 jobs:
   prep-container:
     name: Prepare CI container clang ${{ matrix.clang }}, ${{ matrix.os }}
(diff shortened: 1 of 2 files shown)


sbalint98 (approved):
Thank you!

PR #2221, opened 2026-09-05
[Metal] Make USM allocations resident via MTLResidencySet

resetius:
benchmark:

```
#include <sycl/sycl.hpp>

#include <atomic>
#include <chrono>
#include <cstdio>
#include <cstdlib>
#include <random>
#include <thread>
#include <vector>

namespace {

std::atomic<long> g_errors{0};
std::atomic<bool> g_stop{false};

void run_worker(int worker_id, int iterations, int num_blocks, int elements,
                bool use_shared_queue, sycl::queue &shared_queue) {
  sycl::queue own_queue;
  sycl::queue &q = use_shared_queue ? shared_queue : own_queue;

  std::vector<int *> blocks;
  for (int b = 0; b < num_blocks; ++b) {
    int *blk = sycl::malloc_shared<int>(elements, q);
    for (int i = 0; i < elements; ++i)
      blk[i] = worker_id * 100000 + b * 1000 + i;
    blocks.push_back(blk);
  }

  int **ptrs = sycl::malloc_shared<int *>(num_blocks, q);
  for (int b = 0; b < num_blocks; ++b)
    ptrs[b] = blocks[b];

  int *out = sycl::malloc_shared<int>(num_blocks * elements, q);

  std::mt19937 rng(worker_id * 7919 + 13);

  for (int it = 0; it < iterations; ++it) {
    int *pre = sycl::malloc_shared<int>(elements, q);
    for (int i = 0; i < elements; ++i)
      pre[i] = i;

    const int bias = it;
    auto ev = q.parallel_for(sycl::range<2>(num_blocks, elements),
                             [=](sycl::id<2> idx) {
                               int b = idx[0], i = idx[1];
                               int *blk = ptrs[b]; // pointer chasing
                               out[b * elements + i] = blk[i] * 2 + bias;
                             });

    sycl::free(pre, q);
    int *mid = sycl::malloc_device<int>(elements, q);
    int *mid2 = sycl::malloc_shared<int>(elements, q);
    sycl::free(mid, q);

    ev.wait();

    for (int b = 0; b < num_blocks; ++b) {
      for (int i = 0; i < elements; ++i) {
        int want = (worker_id * 100000 + b * 1000 + i) * 2 + bias;
        if (out[b * elements + i] != want) {
          if (g_errors.fetch_add(1) < 5)
            std::printf("worker %d iter %d: out[%d][%d] = %d, want %d\n",
                        worker_id, it, b, i, out[b * elements + i], want);
        }
      }
    }

[...]

PR #2227, opened 2026-09-08
[SSCP] Fix kernel stats being collected after dead argument elimination

resetius:
`IsFreeOfIndirectAccess` still doesn't work because the DAE optimization replaces the function 
I moved the statistics collection loop earlier

https://github.com/AdaptiveCpp/AdaptiveCpp/pull/2222

### DIFF
--- a/src/compiler/llvm-to-backend/LLVMToBackend.cpp
+++ b/src/compiler/llvm-to-backend/LLVMToBackend.cpp
@@ -30,6 +30,7 @@
 
 #include <llvm/Transforms/IPO/AlwaysInliner.h>
 #include <llvm/ADT/APFloat.h>
+#include <llvm/ADT/StringMap.h>
 #include <llvm/IR/Attributes.h>
 #include <llvm/IR/DerivedTypes.h>
 #include <llvm/IR/DiagnosticInfo.h>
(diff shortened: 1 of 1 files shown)


resetius:
Needed for better performance of https://github.com/AdaptiveCpp/AdaptiveCpp/pull/2221

src/compiler/llvm-to-backend/LLVMToBackend.cpp
```diff
@@ -396,10 +396,18 @@ bool LLVMToBackendTranslator::prepareIR(llvm::Module &M) {
...
     for(const auto& KN : Kernels) {
+      KernelStats KS;
+      KS.Name = KN;
+      KS.IsFreeOfIndirectAccess = false;
       if(auto* F = M.getFunction(KN)) {
-        KernelIsFreeOfIndirectAccess[F] = utils::IsFunctionFreeOfIndirectAccess(F);
+        KS.IsFreeOfIndirectAccess = utils::IsFunctionFreeOfIndirectAccess(F);
+        KernelIsFreeOfIndirectAccess[F] = KS.IsFreeOfIndirectAccess;
       }
+      KernelCompilationStats.push_back(KS);
```
illuhad: I wonder whether it might be a better solution to change `KernelIsFreeOfIndirectAccess` to index on the kernel name instead of the function pointer. That way, the site of collection of the kernel compilation stats can remain mostly independent of other things.

We currently only use it for the indirect access, but in principle, if we were to extend it in the future, I suspect that there might be a natural incentive to have statistics generation late in the pipeline when we can somewhat reason about the end result.

sbalint98 (approved):
I think this looks good now, lets wait for @illuhad to merge

resetius:
@illuhad can this be merged?

illuhad (approved):
Thanks!

Issue #2228, opened 2026-09-09
OpenCL backend on Arm Mali-G610 (libmali, coarse-grained SVM): Testing report after #2048


aleph-ra:
This is a follow up to a discord conversation from September 2025 about running acpp on the Mali-G610 usually found on compute boards like RK3588 (with proprietary libmali OpenCL driver, only coarse-grained SVM available). The conclusion then was that the device needs at least device resident allocations to be usable. #2048 has now provided exactly that. This would enable ACPP usage on many compute boards with integrated GPUs widely used in robotics. So I tested the develop branch (1166873) on a Firefly RK3588 board with libmali g24p0 (OpenCL 3.0, SPIR-V 1.0 IL) and our open source [Kompass](https://github.com/automatika-robotics/kompass-core) robotic navigation library, which uses `malloc_device`, `memcpy`, `fill`, `nd_range`, `atomic_ref` on int and float, and reductions, in its kernels.

As @illuhad suggested back then, the vendor should support `cl_khr_unified_svm`. And that should indeed be the long term approach but its not going to happen any time soon. The extension is still unratified (KhronosGroup/OpenCL-Docs#1282) and no driver I know of ships it. The open source alternative (rusticl on panfrost, enabled by default since Mesa 26.2 and OpenCL 3.0 conformant on the Mali-G610 since July 2026) does not expose SVM at all yet. So for now the proprietary driver with coarse-grained SVM is the only way to run acpp on this GPU, which is what this issue is about.

## Things that needed to be patched

The following is a list of problems that occur on this class of integrated GPUs that I had to patch during testing. I would venture to guess that a similar class of bugs would appear when testing with other vendors (like Qualcomm Adreno). After patching everything worked quite nicely so the proposed fixes are for the maintainers consideration.

### 1. Every kernel is rejected at program creation

```
ocl_code_object: Construction of CL program failed (error code = CL:-30)
```

[...]

Issue #2154, opened 2026-07-16
[OpenCL] AdaptiveCpp assumes capability for alias domains in SPIR-V consumers


linehill:
**Bug summary**
Kernel compilation fails on Rusticl/LLVMpipe because of an unsupported SPIR-V capability/extension involving `OpAliasDomainDeclINTEL`.

**To Reproduce**
Download reproducer from [here](https://github.com/ORNL/HeCBench/blob/a7b20ff8d66182fdf9e80abbcac7e3e8df44fd5d/src/adam-sycl/main.cpp). The needed `reference.h` header is in [here](https://github.com/ORNL/HeCBench/blob/a7b20ff8d66182fdf9e80abbcac7e3e8df44fd5d/src/adam-cuda/reference.h). Compile and run:
```
acpp -O3 main.cpp -o main
./main 100 100 1
```

**Expected behavior**
The reproducer should complete without failures. I believe the `OpAliasDomainDeclINTEL` is part of something non-essential feature.

**Describe your setup**

* Built AdaptiveCpp from sources (330677a58d68242a3b675084eee3a8b01073bf0a), depending on LLVM-21. Only OpenCL backend is enabled.
* Built Mesa/Rusticl/LLVMpipe from sources on [this revision](https://gitlab.freedesktop.org/mesa/mesa/-/commit/05cd2b9509ea1b341d9f07932c030c243211f6d4), depending on LLVM-22.

**Optional additional diagnostic information**
Diagnostics that get emitted due to the issue:
```
$ ./main 100 100 1
pci id for fd 5: 10de:2216, driver (null)
pci id for fd 6: 10de:2216, driver (null)
[AdaptiveCpp Warning] OpenCL device llvmpipe (LLVM 22.1.5, 256 bits) does not support Intel USM extensions; falling back to fine-grained system SVM. USM pointer info queries have limited support.
[AdaptiveCpp Warning] kernel_cache: This application run has resulted in new binaries being JIT-compiled. This indicates that the runtime optimization process has not yet reached peak performance. You may want to run the application again until this warning no longer appears to achieve optimal performance.
[AdaptiveCpp Error] from /mnt/disk0/linehill/ws-adaptivecpp/AdaptiveCpp/src/runtime/ocl/ocl_code_object.cpp:101 @ ocl_executable_object(): ocl_code_object: Building CL program failed. Build log: (file=input,line=0,column=0,index=43): AliasDomainDeclINTEL must appear in a block
[...]

PR #2231, opened 2026-09-10
[SSCP][OpenCL] Compare-exchange fallback for float atomics on devices without cl_ext_float_atomics

aleph-ra:
Fixes item 3 of #2228.

## Problem

The SSCP SPIR-V atomic builtins lower float `fetch_add`, `fetch_sub`, `fetch_min` and `fetch_max` to `__spirv_AtomicFAddEXT`, `__spirv_AtomicFMinEXT` and `__spirv_AtomicFMaxEXT`. The resulting module is rejected by OpenCL devices without `cl_ext_float_atomics` at program creation:

```
ocl_code_object: Construction of CL program failed (error code = CL:-30)
```

## Fix

This PR adds new JIT reflection query `spirv_has_native_float_atomics`. The OpenCL backend sets it from `cl_ext_float_atomics` in `CL_DEVICE_EXTENSIONS`. The float atomic builtins branch on it. The native `__spirv_Atomic*EXT` call when the query is unknown or true, and otherwise a compare and exchange loop on the integer representation of the value. When the optimisation pipeline runs, the unused branch folds away at JIT time.

Details:

- `fetch_min` / `fetch_max` use `fmin` / `fmax` semantics in the fallback, matching `atomic_fetch_min` / `atomic_fetch_max` of `cl_ext_float_atomics` (a NaN operand yields the other operand).
- The failure memory order of the compare-exchange is derived from the requested order (`release` -> `relaxed`, `acq_rel` -> `acquire`), as required by SPIR-V.
- The query is set only by OpenCL backend and I have exposed it in the public API, so kernels can branch on it as well. If you think it should stay internal, we can be drop it.
- The capability bits of `cl_ext_float_atomics` are not consulted, only the presence of the extension is checked. The current devices advertising it would behave similar to how they behaved before.

## Testing

On a Mali-G610 (libmali g24p0, no `cl_ext_float_atomics`):

- A test of `atomic_ref<float>` `fetch_add`, `fetch_sub`, `fetch_min` and `fetch_max` on global and local memory with all five memory orders passes, including the uniqueness of the returned old values under contention.
- The [Kompass GPU test suites](https://github.com/automatika-robotics/kompass-core) pass, whose kernels use float `fetch_add`, `fetch_min` and `fetch_max`.

I don't have a device with `cl_ext_float_atomics` available to test the native path, which is unchanged apart from the folded branch.

### DIFF
--- a/doc/extensions.md
+++ b/doc/extensions.md
@@ -101,6 +101,12 @@ struct target_is_cpu;
 /// Return type: bool
 struct target_is_gpu;
 
+/// Returns whether the SPIR-V consumer accepts native floating point atomic fetch_add, 
+/// fetch_min and fetch_max instructions. Only known on the SPIR-V compiler backend when
+/// the runtime backend provides it (currently only OpenCL provides it from cl_ext_float_atomics).
+/// Return type: bool
+struct spirv_has_native_float_atomics;
+
 /// Returns the AdaptiveCpp runtime backend that is managing the execution of this kernel.
 /// Return type: int (sycl::backend cast to int)
 struct runtime_backend;
--- a/include/hipSYCL/glue/llvm-sscp/jit-reflection/queries.hpp
+++ b/include/hipSYCL/glue/llvm-sscp/jit-reflection/queries.hpp
@@ -46,12 +46,14 @@ extern "C" bool __acpp_sscp_jit_reflect_knows_runtime_backend();
 extern "C" bool __acpp_sscp_jit_reflect_knows_compiler_backend();
 extern "C" bool __acpp_sscp_jit_reflect_knows_target_is_cpu();
 extern "C" bool __acpp_sscp_jit_reflect_knows_target_is_gpu();
+extern "C" bool __acpp_sscp_jit_reflect_knows_spirv_has_native_float_atomics();
 
 extern "C" int __acpp_sscp_jit_reflect_target_vendor_id();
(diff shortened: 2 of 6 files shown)


illuhad (commented):
I think we should also take this into account in the hash of the binary in `ocl_queue::submit_sscp_kernel_from_code_object` (`binary_configuration_id`). With this change, the generated SPIR-V can vary between different devices, therefore, we should ensure that the device is taken into account in the binary hash for the kernel cache so that we don't get incorrect lookups.

As far as I can see, the device currently only enters the code object hash, which identifies the runtime object (i.e. `cl_program`), but not the SPIR-V blob that sits in the persistent on-disk cache.

I would suggest that we make `ocl_hardware_context` store a unique id for each device, which must be invariant over program restarts and changes of system configuration. Maybe PCI id if available, and a hash of vendor name+device name otherwise?

Then we don't have to recompute the unique id for every kernel launch, and can just include it in the hash.

aleph-ra:
Agreed that is the best approach (and as a bonus this would strengthen the design case for item 1 in #2228 and #2154). I have added computing device id at discovery from the PCI bus info (when exposed) and the vendor and device name, and included in the `binary_configuration_id`. 

Tested it on Mali (no PCI info, so it takes the vendor/name path). And a second process reuses cached binaries without reJITting.

PR #2232, opened 2026-09-11
Let the runtime hold a shared_ptr to the allocation map

nilsfriess:
The allocation_map is constructed lazily at the first USM allocation and destroyed during static destruction at program exit. Other objects with static storage duration (some other singleton e.g.) that is constructed *before* the first USM allocation (e.g. another singleton that ends up holding a sycl::queue and USM memory) will be destroyed *after* the allocation_map. It might then happen that sycl::free is called after the allocation_map is already destroyed which results in a double free or similar invalid memory accesses. 

Here's a minimal example:
```cpp
#include <sycl/sycl.hpp>
#include <memory>

struct Buf {
  sycl::queue q;
  double* p = sycl::malloc_device<double>(10, q);
  ~Buf() { sycl::free(p, q); }
};

std::unique_ptr<Buf>& keep() { static std::unique_ptr<Buf> k; return k; }

int main()
{
  keep();                            // the static is constructed here, before SYCL is touched
  keep() = std::make_unique<Buf>();  // the first queue and USM allocation happen after it
}
```

By deliberately leaking the allocation_map, this issue goes away. I guess this might be ok here, because if all USM allocations are properly freed then the allocation map shouldn't really have to free anything in its destructor anyway (?)

Draft because I am not sure if this is a good fix. Ultimately I think this shows that the allocation_map should be tied to the lifetime of the runtime in some way but I wasn't sure how to do that properly.

### DIFF
--- a/src/runtime/allocation_tracker.cpp
+++ b/src/runtime/allocation_tracker.cpp
@@ -19,8 +19,8 @@ namespace {
 using amap_t = common::allocation_map<allocation_info>;
 
 amap_t& get_allocation_map() {
-  static amap_t amap;
-  return amap;
+  static amap_t *amap = new amap_t;
+  return *amap;
 }
 
 }


nilsfriess:
Just saw #2208 and #2209 which seems to be related to this

illuhad:
As mentioned in #2208, can we turn it into a `shared_ptr`?

nilsfriess:
> As mentioned in #2208, can we turn it into a `shared_ptr`?

Done. It's now a shared_ptr and the runtime holds a copy of that shared_ptr

PR #2234, opened 2026-09-11
Fixes a compile-time failure in std::copy/std::move support under --acpp-stdpar for iterator types whose operator*() returns a value rather than a reference (e.g. counting/generator iterators with no backing storage).

jabruniessner:
# Summary

  Fixes a compile-time failure in std::copy/std::move support under --acpp-stdpar for iterator types whose operator*() returns a value rather than a reference (e.g. counting/generator iterators with no backing storage).

# Problem

  copy()/move() in algorithm.hpp gate their memcpy fast-path with a plain runtime if:

```cpp
  if (std::is_trivially_copyable_v<value_type1> && ... &&
      util::is_contiguous<ForwardIt1>() && util::is_contiguous<ForwardIt2>() &&
      detail::should_use_memcpy(q.get_device())) {
    return q.memcpy(&(*d_first), &(*first), ...);
  } else { ... }
```
 This mixes two conditions of different nature: `util::is_contiguous<T>()` is a pure compile-time type trait, while `detail::should_use_memcpy(q.get_device())` 

genuinely needs a runtime device. Because the whole thing is a plain if rather than if constexpr, the compiler must still typecheck the memcpy branch for every instantiation — including iterator types where is_contiguous<T>() is statically false and that branch could
  never be taken. For iterators whose operator*() returns by value, &(*first) then fails with cannot take the address of an rvalue, even though such iterators are otherwise perfectly valid ForwardIterators.

# Fix

  Hoist the compile-time-only conditions into if constexpr, keeping only the genuinely-runtime should_use_memcpy check as a nested plain if. This mirrors the pattern fill() already uses correctly for its own should_use_memset fast-path in the same file. The memcpy expression is now only ever instantiated for iterator types already known at compile time to be contiguous; non-contiguous/non-addressable iterators take the if
  constexpr's else branch and get the existing parallel_for-based fallback.

  No behavior change for previously-working cases: the runtime decision (should_use_memcpy) and both code paths are unchanged, only the compile-time gating is corrected.

# Testing

  - Standalone test covering both the fast path (std::copy/std::move with std::execution::par_unseq over std::vector<float>) and the previously-failing slow path (a custom random_access_iterator_tag counting iterator returning value_type by value) — both compile and produce correct results.
  - Found and fixed while building a real-world downstream project (OpenFOAM's GPU-offload branch) using labelRange/IndexIterator, a counting-iterator pattern that triggered this exact failure in std::copy.

### DIFF
--- a/.github/workflows/linux-vulkan.yml
+++ b/.github/workflows/linux-vulkan.yml
@@ -53,7 +53,7 @@ jobs:
       matrix:
         docker: ${{fromJson(needs.prep-container.outputs.docker_matrix)}}
         clang_version: ['19', '21']
-        clspv_commit: ['88d8ff71000bf995493c8daaed865ec1ac216309']
+        clspv_commit: ['3d017c6d751cdabed1a5051b20c6bf5f71e8fd32']
       fail-fast: false
     steps:
     - uses: actions/checkout@v7
--- a/.github/workflows/macos-vulkan.yml
+++ b/.github/workflows/macos-vulkan.yml
@@ -2,7 +2,7 @@ name: macOS Vulkan backend
 
 on: [push, pull_request]
 env:
-  clspv_commit: '88d8ff71000bf995493c8daaed865ec1ac216309'
+  clspv_commit: '3d017c6d751cdabed1a5051b20c6bf5f71e8fd32'
 jobs:
   build_vulkan:
     name: Vulkan backend macOS
--- a/.github/workflows/windows-vulkan.yml
+++ b/.github/workflows/windows-vulkan.yml
@@ -13,7 +13,7 @@ jobs:
         clang_version: ['19', '20']
         base_compiler: ['19']
         os: [windows-2022]
-        clspv_commit: ['88d8ff71000bf995493c8daaed865ec1ac216309']
+        clspv_commit: ['3d017c6d751cdabed1a5051b20c6bf5f71e8fd32']
     steps:
     - uses: actions/checkout@v7
       with:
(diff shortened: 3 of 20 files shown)


resetius:
It would be good to add a test for this

illuhad:
@jabruniessner can you remove Claude from the commit history? @sbalint98 and I discussed some time ago that we'd prefer AI tools to not enter the commit history, to emphasize that contributors must take full responsibility for their contributions independently of which tools were involved.

jabruniessner:
> @jabruniessner can you remove Claude from the commit history? @sbalint98 and I discussed some time ago that we'd prefer AI tools to not enter the commit history, to emphasize that contributors must take full responsibility for their contributions independently of which tools were involved.

Couldn't agree more. This was just an accident. Fixed now

jabruniessner:
> It would be good to add a test for this

What do you think about the tests I added?

resetius:
> @jabruniessner can you remove Claude from the commit history? @sbalint98 and I discussed some time ago that we'd prefer AI tools to not enter the commit history, to emphasize that contributors must take full responsibility for their contributions independently of which tools were involved.

Maybe it makes sense to add a CI check for the Co-authored-by signature

jabruniessner:
> > @jabruniessner can you remove Claude from the commit history? @sbalint98 and I discussed some time ago that we'd prefer AI tools to not enter the commit history, to emphasize that contributors must take full responsibility for their contributions independently of which tools were involved.
> 
> Maybe it makes sense to add a CI check for the Co-authored-by signature

At least, it would be good to have it written somewhere (If it is not somewhere already). 
Perhaps a CI test that prints the policy and its reasoning upon failure.

resetius (approved):

resetius:
LGTM 
I would shorten the commit title and description when merging

PR #2235, opened 2026-09-15
[VK] Staging buffer refactor

### LINKED ISSUE #2218: [VK] Use memory pooling in Vulkan backend
In the current backend implementation every time a user allocates memory, or the runtime internally does an allocation, this is a new vkMemory object allocation that is then bound to a newly created vkBuffer. Allocating vulkan memory is an expensive call however, so pooling memory by making a single large allocation and chunking it out to individual buffers would be a more performant approach.

The header only library https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator already exists for this as an established solution, so trying to integrate it could be a good path forward.

EwanC:
### Motivation

Don't uniformly apply the same memort properties to all types of buffers being created and allow the runtime to select preferred and required memory properties for uniform/staging/device buffers.

Closes https://github.com/AdaptiveCpp/AdaptiveCpp/issues/2218

### Details

* Move `vk_queue::submit_memcpy` code for doing staging host side buffer copies before and after the command-buffer copy into their own `vk_queue` helper functions.
* Create separate `vk_allocator` helper functions for allocating each of device, staging, and uniform buffers. Allowing finer grained memory property selection and the ability to only set device address bit for user allocated pointers (device USM), not uniform or staging buffers that don't need this.
* Improve memset implementation so we try to use the command-buffer command if the size is 4 byte aligned, and only fall back to host impl if not. Also add an assert in the unlikely case the host fallback isn't possible.

### Experiment

I used the following code to stress the changes across 3 devices. CPU driver, iGPU driver, and discrete GPU driver. There was no performance change in the CPU and iGPU drivers because they are still selecting the same memory type. However, the discrete GPU now picks a cached host memory type for the staging buffer, resulting in a performance bump when copying back to host.

```cpp

int main() {
  sycl::queue q{sycl::property_list{sycl::property::queue::in_order{}}};
  std::string device = q.get_device().get_info<sycl::info::device::name>();
  std::cout << "Default-selected queue runs on device: " << device << std::endl;

  // 256 MB allocation
  size_t count = 64 * 1024 * 1024;
  size_t bytes = count * sizeof(int);

  auto start = std::chrono::high_resolution_clock::now();
  int *ptrA = sycl::malloc_device<int>(count, q);
  int *ptrB = sycl::malloc_device<int>(count, q);
  auto end = std::chrono::high_resolution_clock::now();
  auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start);
  std::cout << "malloc_device time: " << duration.count() << "ms" << std::endl;

  int* initial_data = (int*)malloc(bytes);
  int* out_data = (int*)malloc(bytes);

  for (std::size_t i = 0; i < count; ++i)
    initial_data[i] = i;

  start = std::chrono::high_resolution_clock::now();
  q.memcpy(ptrA, initial_data, sizeof(int) * count).wait();
  end = std::chrono::high_resolution_clock::now();
  duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start);
  std::cout << "memcpy host->device: " << duration.count() << "ms" << std::endl;

  start = std::chrono::high_resolution_clock::now();
  q.memcpy(ptrB, ptrA, sizeof(int) * count).wait();
  end = std::chrono::high_resolution_clock::now();
  duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start);
  std::cout << "memcpy device->device: " << duration.count() << "ms" << std::endl;

  start = std::chrono::high_resolution_clock::now();
  q.memcpy(out_data, ptrB, sizeof(int) * count).wait();
  end = std::chrono::high_resolution_clock::now();
  duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start);
  std::cout << "memcpy device->host: " << duration.count() << "ms" << std::endl;

  for (std::size_t i = 0; i < count; ++i) {
    if (out_data[i] != i) {
       std::cout << "Error at index " << i << " was " << out_data[i] << std::endl;
       return -1;
    }
  }

  sycl::free(ptrA, q);
  sycl::free(ptrB, q);
  std::cout << "SUCCESS\n";
  return 0;
}
```

#### Develop
```
Default-selected queue runs on device: NVIDIA RTX 500 Ada Generation Laptop GPU (NVK AD107)
malloc_device time: 4ms
memcpy host->device: 99ms
memcpy device->device: 10ms
memcpy device->host: 15720ms
SUCCESS
```

#### Patch
See order of magnitude improvement in copy back to host.

```
Default-selected queue runs on device: NVIDIA RTX 500 Ada Generation Laptop GPU (NVK AD107)
malloc_device time: 4ms
memcpy host->device: 149ms
memcpy device->device: 7ms
memcpy device->host: 265ms
SUCCESS
```


include/hipSYCL/runtime/vk/vk_allocator.hpp
```diff
@@ -57,11 +66,13 @@ class vk_allocator : public backend_allocator {
...
-  vk_alloc_info *find_alloc_info(vk::DeviceAddress ptr);
+  vk_alloc_info *find_user_alloc(vk::DeviceAddress ptr);
+
+  vk_alloc_info *staging_allocate(size_t size_bytes);
 
   // Creates a single buffer backed by allocated device memory, and created with
   // properties which allow its device address returned
-  std::pair<vk::raii::Buffer, vk::raii::DeviceMemory>
+  std::tuple<vk::raii::Buffer, vk::raii::DeviceMemory, vk::MemoryPropertyFlags>
   create_device_address_buffer(vk::DeviceSize size);
```
sbalint98: why not return a vk_allocation_info here instead of the tuple?

include/hipSYCL/runtime/vk/vk_queue.hpp
```diff
@@ -28,6 +28,22 @@ class vk_hardware_manager;
 class vk_hardware_context;
 struct vk_alloc_info;
 
+// Maps command signal value to any temporary memory allocations
+// that need freed asynchronously when it completes
+struct protected_map {
```
sbalint98: Maybe `async_alloaction_map` would better describe the purpose of this struct?

include/hipSYCL/runtime/vk/vk_queue.hpp
```diff
@@ -70,11 +86,20 @@ class vk_queue : public inorder_queue {
...
   void submit_command_buffer(vk::CommandBuffer &cmd_buf);
 
-  std::pair<vk_alloc_info *, bool>
-  find_or_create_allocation(vk::DeviceAddress ptr, unsigned size);
+  vk_alloc_info *find_or_create_allocation(vk::DeviceAddress ptr,
+                                           unsigned size);
 
   void profile_if_enabled(operation &op, const dag_node_ptr &node);
 
+  std::pair<vk_alloc_info *, vk_alloc_info *>
```
sbalint98: Maybe instead of passing around a pair we could introduce a small struct with src, and dest members to better track which one is which?

src/runtime/vk/vk_queue.cpp
```diff
@@ -25,6 +25,28 @@
...
+  return _alloc_map.insert({key, val});
+}
+
+protected_map::ValueType protected_map::get(uint64_t wait_value) {
+  std::lock_guard<std::mutex> lock{_mutex};
+  return _alloc_map[wait_value];
+}
+
+void protected_map::erase(uint64_t wait_value) {
+  assert(_alloc_map.count(wait_value));
```
sbalint98: I think we are missing a mutex here

src/runtime/vk/vk_allocator.cpp
```diff
@@ -23,26 +23,36 @@ vk_allocator::vk_allocator(vk_hardware_context *hw_ctx, rt::device_id dev)
...
             properties) {
-      return i;
+      return std::make_pair(i, _mem_properties.memoryTypes[i].propertyFlags);
     }
   }
 
-  print_error(__acpp_here(),
-              error_info{"failed to find suitable memory type!"});
-  return UINT32_MAX;
+  return std::make_pair(UINT32_MAX, vk::MemoryPropertyFlags{});
```
sbalint98: Why remove the error message here?

include/hipSYCL/runtime/vk/vk_allocator.hpp
```diff
@@ -21,13 +21,22 @@ namespace rt {
...
 class vk_hardware_context;
 
+enum class vk_alloc_type {
+  USER = 0,    // User application allocated memory
+  STAGING = 1, // Staging buffer to use for memcpy to/from host pointer
+  UNIFORM = 2, // For passing uniform arguments to the kernel
+};
+
 struct vk_alloc_info {
   vk_alloc_info() = delete;
```
sbalint98: Nit: Although this has not been added in this PR, declaring a constructor as deleted, will disable the aggreage intitiallization starting with cpp20. this is fine for now since we use cpp17 but I think we should redisign this class anyway.

include/hipSYCL/runtime/vk/vk_allocator.hpp
```diff
@@ -21,13 +21,22 @@ namespace rt {
...
 
-  vk::DeviceAddress _base_ptr;
+  vk_alloc_type _type;
+
+  vk::DeviceAddress _base_ptr; // For USER allocations only
```
sbalint98: If this classes intended to be initialized as an aggregate, we will have to provide a value for _base_ptr even if the type is not USER. I wonder if it wouldn't make more sense to have this as an optional, and pass std::nullopt, instead of passing 0.

src/runtime/vk/vk_queue.cpp
```diff
@@ -207,84 +195,32 @@ result vk_queue::submit_memcpy(memcpy_operation &op, const dag_node_ptr &node) {
...
 
+  // Track temporary buffers created in map
+  if (src_alloc_info->_type == vk_alloc_type::STAGING) {
     temp_allocs.first = src_alloc_info;
   }
 
-  if (dst_alloc.second) {
-    temp_allocs.second = dst_alloc.first;
+  if (dst_alloc_info->_type == vk_alloc_type::STAGING) {
+    temp_allocs.second = dst_alloc_info;
```
sbalint98: Can it happen that source and dest are not both staging?

sbalint98 (commented):
Thank you! I left some comments mostly related to nits :) Mostly looks good

include/hipSYCL/runtime/vk/vk_queue.hpp
EwanC: I agree, updated the name.

This struct is used to define a member variable of the `vk_queue` object in this header, so we can't forward declare it and have the definition in the cpp file. Putting it in an anonymous namespace in the header I think is an anitpattern because each TU gets it's own copy. So I'd probably leave it as-is unless you can direct me more on the change.

include/hipSYCL/runtime/vk/vk_queue.hpp
EwanC: Done that now, agree it's more readable.

src/runtime/vk/vk_queue.cpp
EwanC: good spot, updated

include/hipSYCL/runtime/vk/vk_allocator.hpp
EwanC: Deleted this, think it was a development leftover. Thanks for the tip about c++20 changes though.

include/hipSYCL/runtime/vk/vk_allocator.hpp
EwanC: Made this a std::optional for clarity

src/runtime/vk/vk_queue.cpp
EwanC: Yes that can happen, if you do a memcpy with a src and dst pointer which come from a `sycl::malloc_device` operation then we don't need any staging buffers because the USM malloc operation created the underlying buffers already.

Happy to add a code comment if there's a way that could be described better

PR #2239, opened 2026-09-18
[stdpar] Fix memory_pool::release() size rounding mismatch

jabruniessner:
claim() rounds allocation size up to _page_size before computing the buddy-allocator level, but release() used the original unrounded size, filing the freed block under the wrong level and losing it from the pool permanently.


include/hipSYCL/std/stdpar/detail/sycl_glue.hpp
```diff
@@ -435,6 +435,10 @@ class memory_pool {
 
   void release(void* ptr, std::size_t size) {
     if(_pool && is_from_pool(ptr)) {
+      // Must match the rounding claim() applies, or release() computes the
+      // wrong buddy-allocator level and the block is lost from the pool.
+      if(size < _page_size)
```
resetius: There is some code duplication here. I would move this into a helper. Smth like
```
effective_size(size) -> return size < _page_size ? _page_size : size
```

`effective_size` should be called by `release/claim`

resetius:
Could you add unit tests for this in `tests/pstl/memory.cpp`?

include/hipSYCL/std/stdpar/detail/sycl_glue.hpp
jabruniessner: Done

jabruniessner:
> Could you add unit tests for this in `tests/pstl/memory.cpp`?

Done appended in the latest commit

tests/pstl/memory.cpp
```diff
@@ -65,4 +67,21 @@ BOOST_AUTO_TEST_CASE(pstl_aligned_alloc) {
   std::free(p1);
 }
 
+BOOST_AUTO_TEST_CASE(pstl_memory_pool_release_reclaims_space) {
+  const std::size_t page_size = static_cast<std::size_t>(sysconf(_SC_PAGESIZE));
+  const std::size_t pool_size = page_size * 4;
+  hipsycl::stdpar::memory_pool pool{pool_size};
```
resetius: We only need to check that the invariant holds here. There is no need to run this with a real `page_cache`. We can use a constant and remove `unistd.h`. e.g.
```cpp
hipsycl::stdpar::memory_pool pool{1 << 20};
```

jabruniessner: I reverted to 1<<14 as this is exactly what the previous pool_size evaluated to. Your suggested number is much larger, which (correct me if I am wrong) might also require us to use more iterations in the for loop

resetius (approved):

PR #2240, opened 2026-09-18
[CUDA] Avoid tripping caller FPE traps during PTX JIT compilation

jabruniessner:
The NVIDIA PTX JIT compiler can raise benign FP exceptions internally. Disable hardware FPE trapping around cuModuleLoadDataEx so callers with FPE traps enabled (e.g. OpenFOAM's trapFpe) don't crash on unrelated JIT-internal exceptions.


resetius:
have you tried `std::feholdexcept` ? https://en.cppreference.com/cpp/numeric/fenv/feholdexcept

jabruniessner:
> have you tried `std::feholdexcept` ? https://en.cppreference.com/cpp/numeric/fenv/feholdexcept

Thank you very much I really appreciate your feedback. 

Done.

src/runtime/cuda/cuda_code_object.cpp
```diff
@@ -85,10 +86,16 @@ result build_cuda_module_from_ptx(CUmod_st *&module, int device,
   std::string error_log_buffer(error_log_buffer_size, '\0');
   option_vals[1] = error_log_buffer.data();
 
+  // Suppress caller FPE traps during JIT; fesetenv (not feupdateenv) avoids replaying trapped exceptions.
```
resetius:
```suggestion
  // Suppress caller FPE traps during JIT, as some apps and libs may enable them during execution and break the JIT
```

resetius: The second part of the comment is not needed. It does not relate directly to this line and gives extra details that can be found on cppreference. At the same time, the "why" is missing, so I suggest this change

resetius (approved):

resetius:
I think it is better to use squash and merge. This matches the style used in this repository.
Next time, please select `Squash and merge`

jabruniessner:
Ooops, I thought it was going to squash and merge by default. Sorry! Next time!

PR #2241, opened 2026-09-18
[VK] Document KosmicKrisp driver support

EwanC:
I've managed to get my hands on a Apple device to run the sycl tests locally on the Vulkan backend with both MoltenVK and KosmicKrisp, both of which layer on top of Metal.

The first thing I noticed is that the device name for MoltenVK was different than we are expecting in the current test skipping code, so this patch updates that to use the driver version string for the check.

Secondly, I was hoping that Kosmic Krisp would be a more reliable driver to use because it is Vulkan conformant unlike MoltenVK. However, it exhibited similar failures plus a few new ones which unfortunately manifested by tests hanging. This patch documents this behavior.


sbalint98 (approved):
Thanks! LGTM

PR #2242, opened 2026-09-18
[CI] Do not run workflows twice for in-repo pull requests

resetius:
When you click `Update` in a PR, it triggers a `push` event for the merge commit and also updates the PR. With the current configuration, the tests run twice

If the PR is from a fork, the `push` workflow runs in the fork. Otherwise, it runs directly in the PR, as here: https://github.com/AdaptiveCpp/AdaptiveCpp/pull/2239

Also note that this does not happen for the `macos-metal` workflow because its configuration is already correct


sbalint98:
Thanks! I think this change is an improvement for the CI. However this way, the self-hosted workflows are only run in case one opens a PR. i could see this as an issue when working on long-lived feature branches. Although one could open a draft PR, I think in that case all the watchers on the repository would get notified nonetheless for every activity on the work in progress development, which might result in a lot of spam.

sbalint98:
I tried to look into this, but there does not seem to be an easy solution to this problem. I would suggest that we trigger the push workflow for  `feature/**` branches as well. After a PR is opened, these would still run the workflows twice, but at least we save some CI time on smaller fix PRs.

PR #2243, opened 2026-09-19
[Metal] Follow-up fixes for command buffer batching

resetius:
This PR replaces the recursive mutex with a regular mutex to make the code cleaner 
Because this changes where we lock and unlock, profiling setup is moved closer to command buffer submission

Before
```
    submit_kernel()
      LOCK recursive_mutex
      profiling_setup()
      normal kernel:
        invoke()
          LOCK recursive_mutex again
          encode kernel
          UNLOCK recursive_mutex
      custom operation:
        flush()
        invoke callback()
      UNLOCK recursive_mutex
```

After
```
    submit_kernel()
      normal kernel:
        LOCK mutex
        profiling_setup()
        invoke()
          submit_sscp_kernel_unlocked()
            encode kernel
        UNLOCK mutex
      custom operation:
        LOCK mutex
        host_profiling_setup()
        flush()
        UNLOCK mutex
        invoke callback()
```


sbalint98 (approved):
thanks LGTM

src/runtime/metal/metal_queue.cpp
```diff
@@ -706,49 +730,39 @@ result metal_inorder_queue::submit_memcpy(memcpy_operation& op, const dag_node_p
...
-    if (result r = flush(); !r.is_success()) {
-      return r;
+  {
+    // invoke() can re-enter the queue and must run without the mutex
+    std::lock_guard<std::mutex> lock{_mutex};
+
+    // Custom operations may submit directly to the native queue. Commit pending
+    // work first and use host timestamps for work that we do not encode
+    if (op.get_launcher().is_custom_operation()) {
+      host_profiling_setup(op, node);
```
sbalint98, line 749: Maybe we should document somehow that custom_operations return phony profiling results?

resetius: Added a comment

PR #2247, opened 2026-09-21
[VK] Fix device_id mismatch

al42and:
`vk_hardware_manager` was recording Vulkan device index as its id, but `vk_allocator` was treating this id as its index in the list of active devices.

Seen in `usm_tests/device_allocation_functions` failures when several Vulkan devices are present and `ACPP_VISIBILITY_MASK=vk:1` is used.


EwanC (approved):
LGTM, thanks 😃

PR #2248, opened 2026-09-21
[Metal] Add fences for non-relaxed atomic memory orders

resetius:
map non-relaxed memory orders to fences, since MSL atomics only support memory_order_relaxed 
needed for upcoming stdpar support


sbalint98 (approved):
LGTM

